PR #1, opened 2019-07-16
Set up Azure Pipelines.

tobiasraabe:
This PR adds Azure Pipelines which is a continuous integration service. Tests ensure that the project runs on all major operating systems.

### Todo

- [x] Add azure badge to README.


tobiasraabe:
@hmgaudecker The build step stops early since a file is missing, but it is enough for ``bld/project_paths.py`` to be created. I don't know if the build step takes too long, but do you know a way to make a separate waf command which creates only ``project_paths.py`` and then stops?

hmgaudecker:
That was easy :-) We will probably kick out Waf altogether for this project eventually. Still from the days of a joint behavioural / taxes&transfers model.

tobiasraabe:
@hmgaudecker After your review, I will merge the changes.

.azure-pipelines.yaml
```diff
@@ -0,0 +1,69 @@
+pr: none
+
+jobs:
+- job:
+  displayName: Linux
+  pool:
+    vmImage: "ubuntu-16.04"
```
hmgaudecker: Does 18.04 exist already?

.azure-pipelines.yaml
```diff
@@ -0,0 +1,69 @@
...
+    displayName: Add conda to PATH.
+
+  - bash: conda update conda --yes --quiet
+    displayName: Update conda.
+
+  - bash: conda env create
+    displayName: Create Anaconda environment.
+
+  - bash: |
+      source activate gettsim
```
hmgaudecker: Why does that still work? With modern conda it should be `conda activate gettsim`, right?

.azure-pipelines.yaml
```diff
@@ -0,0 +1,69 @@
...
+    vmImage: "ubuntu-16.04"
+
+  steps:
+  - bash: echo "##vso[task.prependpath]$CONDA/bin"
+    displayName: Add conda to PATH.
+
+  - bash: conda update conda --yes --quiet
+    displayName: Update conda.
+
+  - bash: conda env create
```
hmgaudecker: Make explict by adding ` -f environment.yml` ?

.azure-pipelines.yaml
tobiasraabe: No, here are all available images: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent.

.azure-pipelines.yaml
tobiasraabe: I think it is still supported, but in the future one should move to the other commands. The problem here is that ``conda init`` alters the configuration files which have to be resourced. That did not work when I tried it the last time.

.azure-pipelines.yaml
tobiasraabe: Sure, I will also set the name.

.azure-pipelines.yaml
hmgaudecker: Cool, thanks. Looks to me from a very quick glance like that MS is well aware of this and `source activate` is their own workaround rather than old conda style. Can you add a comment to the yaml file pointing to: https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/anaconda

raholler:
Why don't you update conda for windows in the azure pipeline?

tobiasraabe:
Because it breaks the testing. I think the pre-installed conda is something like 3.5 and newer versions changed a lot concerning the activation of an environment.

PR #3, opened 2019-07-19
Update README.

hmgaudecker:


tobiasraabe:
I think the first point is only possible with ``.rst``, right? Because markdown has no option to refer to a different target than the image url. Are you ok with conversion to ``rst``?

hmgaudecker:
> I think the first point is only possible with `.rst`, right? Because markdown has no option to refer to a different target than the image url. Are you ok with conversion to `rst`?

The move alrready happened :-)

hmgaudecker:
@tobiasraabe : Cool, thanks! I can take it from here!

PR #7, opened 2019-07-19
Integrating changes from dynamod 

MaxBlesch:
There were some little changes on my dynamod branch, which didn't get transferred. I will add them and complete the features, this includes:

- Tests for altersfreibetrag (no person above 64 in the test data)
- Tests for the calculation of Sonderkosten.
- Restructuring modules
- Restructure code, such that functions are self containing.(not just called for the persons relevant and instead for all)


src/analysis/tax_transfer_funcs/zve.py
```diff
@@ -332,10 +352,16 @@ def vorsorge2010(df, tb):
 def calc_hhfreib_until2014(df, tb):
     """Calculates tax reduction for single parents. Used to be called
     'Haushaltsfreibetrag'"""
-    return tb["hhfreib"]
+    hhfreib = pd.Series(index=df.index, data=0)
+    hhfreib[df["alleinerz"]] = tb["hhfreib"]
```
hmgaudecker: Can we have multiple single parents in one household? Else we should make the check more readable.

Also we should not introduce function names like this... Fine for now, but this is exactly where we will need to be more modular eventually (i.e., do not only allow to replace parameters for different years/scenarios, but entire (sets of) functions).

hmgaudecker:
@MaxBlesch: Can you just update those few things so I can merge this? Thanks!

src/analysis/tax_transfer_funcs/zve.py
MaxBlesch: I think so far. We account for that possibility. This is because we differ between household and tax unit.

PR #25, opened 2019-08-15
Add codecov.io.

tobiasraabe:


tobiasraabe:
I need to add a configuration file for when the report errors and so on. I am a little bit confused as there exist two formats, ``coveragerc`` and ``codecov.yml``, but I will figure that out.

PR #32, opened 2019-10-04
Wohngeld can now be calculated by 'Mietstufe'

### LINKED ISSUE #16: Add features for wohngeld(wg function)
- [ ] Wohngeld: Collect more maximum rents for all *Mietstufen*. This affects the parameters `wgmax?p_m` (? = [1, 2, 3, 4, 5, plus5]).

Eric-Sommer:
Closes #16.


src/analysis/tax_transfer_funcs/benefits/wohngeld.py
```diff
@@ -63,31 +73,31 @@ def calc_wg_rent(df, tb, hhsize):
...
     cnstyr_dict = {1: "a", 2: "m", 3: "n"}
     key = cnstyr_dict[cnstyr]
     # fixed amounts for the households with size 1 to 5
     # afterwards, fix amount for every additional hh member
     if hhsize <= 5:
-        max_rent = tb["wgmax" + str(hhsize) + "p_" + cnstyr_dict[cnstyr]]
+        max_rent = tb["wgmax{}p_{}_st{}".format(hhsize, key, mietstufe)]
     else:
-        max_rent = tb["wgmax5p_" + key] + tb["wgmaxplus5_" + key] * (hhsize - 5)
+        max_rent = tb["wgmax5p_{}_st{}".format(key, mietstufe)] + tb["wgmaxplus5_".format(key, mietstufe)] * (hhsize - 5)
```
MaxBlesch: tb["wgmaxplus5_".format(key, mietstufe)] is missing any {} for formatting.

src/analysis/tax_transfer_funcs/benefits/wohngeld.py
```diff
@@ -42,13 +44,21 @@ def calc_wg_rent(df, tb, hhsize):
...
     # There are also min and max values for this.
+    # If 'Mietstufe' is not given, choose '3'.
+    if 'mietstufe' in df.columns:
+        mietstufe = int(df['mietstufe'].max())
+    else:
+        mietstufe = 3
+
+    assert(mietstufe in range(1, 7))
     # First max rent
     if tb["yr"] >= 2009:
```
MaxBlesch: I know this is a leftover from my code. But didn't we agree to implement unified function interfaces and then hand over the functions via a dictionary for now? If we want each different method of calculating things as function on his own, this would be the way to go. Otherwise we could also just check the year in the function and if before 2009 we could set the key to "m".

MaxBlesch:
@janosg @hmgaudecker 
Some general questions:

- Should we agree on some string formatting style to go forward. Either f-strings or "...".format()?
- I think we always need tests if we include new features?

src/analysis/tax_transfer_funcs/benefits/wohngeld.py
Eric-Sommer: Correct. Will be addressed

src/analysis/tax_transfer_funcs/benefits/wohngeld.py
Eric-Sommer: where is `any()` in the old code? 88 is new to me, I recall agreeing on 99, but fair enough...

hmgaudecker: At some point, we let black's default be. Please make use of the pre-commit hooks, i.e., type `pre-commit install` once and then this should work automatically. A bit more context is [here](https://econ-project-templates.readthedocs.io/en/stable/pre-commit.html).

hmgaudecker:
> Should we agree on some string formatting style to go forward. Either f-strings or "...".format()?

The pyupgrade hook should take care of that automatically. Will end up as f-strings. 

> I think we always need tests if we include new features?

Yes, please!

MaxBlesch:
This looks very good now. But I think you should incorporate the new design of gettsim before you merge it.

Eric-Sommer:
> This looks very good now. But I think you should incorporate the new design of gettsim before you merge it.

done.

tobiasraabe:
I forgot to add pre-commit to the CI checks so that ``.format`` is currently not converted to f-strings (will be fixed in #38).

The checks failed because coverage decreased. Currently, the new code must have the same coverage as the base branch which is ~89.55%. We can allow for a different target ratio and we can be lenient if coverage drops by only a small amount. We can also differentiate between the PR and the whole project and different folders. Here is the piece of documentation: https://docs.codecov.io/docs/commit-status. Pandas targets 82% in the project and 50% in a PR 8https://github.com/pandas-dev/pandas/blob/master/codecov.yml).

Do you have any suggestions for the target ratios? We could go with something like the pandas defaults and if the quality deteriorates quickly, we change it.

Ping @hmgaudecker.

tobiasraabe:
And I had to change the title to use Github keywords: https://help.github.com/en/articles/closing-issues-using-keywords.

hmgaudecker:
So what is the issue with leaving things as they are? Is merging blocked then? 

If not, I think it is actually useful to leave as is because it makes you look for things. IIUC, line 51 here is not covered? https://codecov.io/gh/iza-institute-of-labor-economics/gettsim/pull/32/diff?src=pr&el=tree#D4-51  This would actually be super-important to have in the tests, because it defines implicit behaviour. I am not sure we want that in the end, but we should never silently change something like that and break users' code in subtle ways.

Back to the general question by @tobiasraabe: Measuring helps, but one has to be careful not to take adherence to these measures for the ultimate goal.

hmgaudecker:
> And I had to change the title to use Github keywords: https://help.github.com/en/articles/closing-issues-using-keywords. It has to be in the title of the commit or in the longer commit description.

I am personally no big fan of putting the issues in the title as they do not become links ever, AFAICT. Better just keep these references in the text ("work on #x"  and write "closes #x" only upon merging. Thoughts?

hmgaudecker:
Addendum: Another issue is that the two numbers show up directly next to each other in the title **... (loses #Issue) #PR** and people like me get confused easily...

tobiasraabe:
In the commit history (https://github.com/iza-institute-of-labor-economics/gettsim/commits/master) both hashes are represented as links. The other way is to write ``Closes #16`` in the first PR message which is also very accessible.

hmgaudecker:
> In the commit history (https://github.com/iza-institute-of-labor-economics/gettsim/commits/master) both hashes are represented as links. The other way is to write `Closes #16` in the first PR message which is also very accessible.

Fair  enough. But the most important views of PRs seem to be during the time they are being worked on. :-)

hmgaudecker:
Thanks! Please never use names like 'fun_x", "fun_x2", etc. but names that carry meaning. Like in this case "test_wohngeld_no_col_mietstufe". 

Also, is there a reason to put them in a separate file?

Eric-Sommer:
can't complete it right now. pre-commit fails because I'm behind a proxy server.

PR #37, opened 2019-10-07
Draft on tax_transfer.py

MaxBlesch:
I will fix the function calls, align the call of the functions with the layer(pid, tu_id, hid) we already use in the tests.


Eric-Sommer:
- The calculation of `dpi_ind`, `dpi` and `gross` also deserves a function on its own, along with a new test.

- I'd add a test for `tax_transfer()` as a whole. No need to check for correct calculations, but to make sure a proper data set is returned (same number of rows, key column names present).

gettsim/tax_transfer.py
```diff
@@ -125,16 +132,20 @@ def tax_transfer(df, datayear, taxyear, tb, tb_pens=None, mw=None, hyporun=False
 
     # 6. SOCIAL TRANSFERS / BENEFITS
     # 6.0.1 Alimony Advance (Unterhaltsvorschuss)
-    df["uhv"] = uhv(df, tb, taxyear)
+    # tu_id
+    df["uhv"] = uhv(df, tb)
 
     # 6.1. Wohngeld, Housing Benefit
     # TODO: rename wohngeld ('wohngeld_basis') until final check.
```
Eric-Sommer: this can be removed

MaxBlesch:
This is an early draft on the tax_transfer.py function. The test data doesn't provide any results and therefore the correctness can't be guranteed. What I think should be discussed or even be transferred separate issues:

1) I created the file func_out_columns.py, where the output columns for each tax transfer function is stored. I think the same would be really handy for the necessary inputs. We could store the input and output columns in the pyfile of each function, but I am not sure if that is then to messy. For me there are several reasons for that:

- To have this information all in one place, would make it really easy for a user to add a particular variable as input or output and not to forget it somewhere in the layout. 

- It would be an easy way to require tests for all outputs. 

- If we have the input columns stored, we could also call each function with just the information needed.

2) We need to align variable names. I was really confused to not find these variables https://github.com/iza-institute-of-labor-economics/gettsim/blob/427cce3ad00b39b9f06bfe3a6c8a6030fe28193d/gettsim/incomes.py#L29 . Also why is bp not used at all?

gettsim/incomes.py
```diff
@@ -0,0 +1,60 @@
...
+
+def disposible_income(df):
+    disp_inc = pd.DataFrame(index=df.index, columns=DPI)
+    disp_inc["dpi_ind"] = df[
+        [
+            "m_wage",
+            "m_kapinc",
+            "m_self",
+            "m_vermiet",
+            # "m_imputedrent", Do we need this?
```
Eric-Sommer: Whether one "needs" imputed rents is a matter of conventions. Note though that if you want to align to "official" inequaliy aggregates, imputed rents need to be added

gettsim/tax_transfer.py
```diff
@@ -36,151 +50,58 @@ def tax_transfer(df, datayear, taxyear, tb, tb_pens=None, mw=None, hyporun=False
...
+
+    # We start with the top layer, which is household id. We treat this as the
+    # "Bedarfsgemeinschaft" in the german tax law.
+    for hid in df["hid"].unique():
+        hh_index = df[df["hid"] == hid].index
+        for tu_id in df.loc[hh_index, "tu_id"].unique():
+            tu_index = df[df["tu_id"] == tu_id].index
+            for i in tu_index:
+
+                df.loc[i, SOC_SEC] = soc_ins_contrib(df.loc[i], tb)
```
Eric-Sommer: `soc_ins_contrib` runs on personal level

gettsim/incomes.py
```diff
@@ -0,0 +1,60 @@
...
+            "uhv",
+        ]
+    ].sum(axis=1) - df[
+        ["incometax", "soli", "abgst", "gkvbeit", "rvbeit", "pvbeit", "avbeit"]
+    ].sum(
+        axis=1
+    )
+
+    # Disposible income on hh level
+    # TODO: Why we do not have m_alg2, wohngeld and kiz
```
Eric-Sommer: because you need to run through `benefit_checks()` first. Before, this was done within the `kiz` function. We haven't touched `tax_transfer()` for quite a while, so this was not updated.

gettsim/tests/test_data/test_dfs_tax_transfer.csv
```diff
@@ -0,0 +1,15 @@
+hid,tu_id,pid,head_tu,hhtyp,m_wage,east,age,selfemployed,haskids,m_self,m_pensions,pkv,m_wage_l1,months_ue,months_ue_l1,months_ue_l2,alg_soep,m_pensions,w_hours,child_num_tu,adult_num_tu,byear,exper,avg_ep,EP,child,m_childcare,m_imputedrent,m_kapinc,m_vermiet,renteneintritt,zveranl,m_pensions,w_hours,ineducation,zveranl,handcap_degree,alleinerz,miete,wohnfl,eigentum,heizkost,child11_num_tu,cnstyr,mietstufe,m_transfers,hh_korr,year
```
Eric-Sommer: While the values don't matter much here, they don't make sense for `exper`, `EP` (some are negative)

gettsim/incomes.py
MaxBlesch: But it is nowhere else used?

gettsim/tax_transfer.py
```diff
@@ -36,151 +50,58 @@ def tax_transfer(df, datayear, taxyear, tb, tb_pens=None, mw=None, hyporun=False
...
+                # Unemployment benefits
+                df.loc[i, UI] = ui(df.loc[i], tb)
+                # Pension benefits
+                df.loc[i, PENS] = pensions(df.loc[i], tb, tb_pens)
+
```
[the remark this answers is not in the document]
MaxBlesch: Yes. This calls the functions on individual levels, but transforms the row to Series. I wrote the functions on Series level.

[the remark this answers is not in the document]
Eric-Sommer: you loop through values of `tu_id`, not `pid`.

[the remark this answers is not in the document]
MaxBlesch: In line 64 I loop over the tu_ids. In 66 I loop over the index of one particular tu_id.

gettsim/tests/test_data/test_dfs_tax_transfer.csv
MaxBlesch: Shouldn't we calculate some of these variables inside anyway? They are in the excel sheet, but is this just for convenience or just general? For example in the sheets 'EP' = 'exper' * 'avg_ep'

gettsim/tax_transfer.py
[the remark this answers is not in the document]
Eric-Sommer: you are right sorry. It does mot become obvious from the variable names though. Maybe rename `tu_index` to `persons_in_tu`?

gettsim/incomes.py
Eric-Sommer: We haven't decided yet on what's mandatory input and what's not.

gettsim/tests/test_data/test_dfs_tax_transfer.csv
Eric-Sommer: I don't know, you created that sheet, didn't you? ;)

gettsim/incomes.py
MaxBlesch: Then we should decide at some point and in the same move also which variables we compute inside the functions. I'll put that in a separate issue!

gettsim/tests/test_data/test_dfs_tax_transfer.csv
MaxBlesch: I don't know! :) See above, I'll move that!

gettsim/tax_transfer.py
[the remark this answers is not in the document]
MaxBlesch: But then I should also change `hh_index` into `persons_in_hh`. As it should be clear, that each index in the DataFrame is a person, I would instead go with `hh_indicies` and `tu_indicies`, which makes clear that this variable collects all indicies for one hh and one tu!

gettsim/incomes.py
MaxBlesch: I included it, but commented out the assert checks. Can we agree on having a general checking section, if the inputs are valid? Then we would get valid outcomes.

Eric-Sommer: we can outsource the validation checks. The first one (`assert bp[ben].notna().all()`) could be dropped completely. It was mainly important when writing the function.

Issue #31, opened 2019-10-02
Datatype of params


MaxBlesch:
We need to discuss how to move away from excel sheets.

Eric-Sommer:
what exactly bothers you?

MaxBlesch:
The thing with excel sheets is the conflict of bool variables if you open it in other editors. If we want to have community code, which is editable by a large community we need something like csv. We haven't yet agreed on one format or structure yet. Will keep this updated, as soon we have.

Eric-Sommer:
Right, I remember. The benefit of spreadsheets is their ability to work with formulas. Our tests are non-trivial. One could stick to 0/1 variables altogether and possibly reformat them within the test routine.

janosg:
Another benefit of not having spreadsheets is version control. I think it was important to have the formulae while debugging the test-cases, but now it's just important to have the correct numbers. If it's about documentation, then the formulae should go in docstrings or doctests.

Eric-Sommer:
What bothers me more right now is that my LibreOffice crashes everytime I want to save changes to the param file. What about switching to `ods` format all the way and using [pandas-ods-reader](https://pypi.org/project/pandas-ods-reader/)?

Eric-Sommer:
It turns out that versioning of [xlsx files](https://github.com/willu47/git_diff_xlsx) and [OpenOffice](https://git.wiki.kernel.org/index.php/GitTips#How_to_use_git_to_track_OpenDocument_.28OpenOffice.2C_Koffice.29_files.3F) files is possible without too much trouble. I might have closer look on these.

mjbloemer:
A few notes on this matter from my perspective as the maintainer of the Stata based microsimulation model (forked from the ZEW fork of izamod two years ago) at ifo.

1. At least in the context of the Stata version of the taxtransfer model I fully agree that moving away from binary xls files is/was beneficial.
2. It might be good to move a step further and get rid of the year paradigm.

On 1: At ifo we got rid of the binary param.xls a while ago replacing it with csv files for every year. While this change seemed subtile at first, it made a huge difference for our workflow especially with version control and multiple developers. The binary xls file was too error-prone. We also briefly tested one big csv file (like it was in the early days of izamod btw) as well as text based xlsx and LibreOffice files and thought about xml or json based databases fo the parameters.

[...]

PR #40, opened 2019-10-11
convert test data to open office format

Eric-Sommer:

Existing test cases are stored in Excel files, which creates problems when edited in LibreOffice/OpenOffice. To this end, all test data are now stored in as `ods` files.

addresses #31.

One could extend this PR by automatically convert them into `csv` in order to track changes.


.azure-pipelines.yaml
```diff
@@ -21,7 +21,10 @@ jobs:
   - bash: echo "##vso[task.prependpath]$CONDA/bin"
     displayName: Add conda to PATH.
 
-  - bash: conda update conda --yes --quiet
```
MaxBlesch: Why is that necessary? @tobiasraabe!

hmgaudecker: @MaxBlesch : https://github.com/conda/conda/issues/9337

MaxBlesch: Ok thanks!

Issue #41, opened 2019-10-15
Decide on variables


MaxBlesch:
### Current and desired situation

We need to decide which variables to be mandatory and which can either be assumed or if they appear only in a nonrelevant calculation( `m_imputedrent` in gross_income) can be left out. Furthermore there are variables right now, which can be computed from basic household data(`hhsize`, `child_num_tu` etc.)

### Proposed implementation

Catch in the beginning missing variables and either alert user or assume fill values, but still alert user.

### Considered alternatives

- Make all variables mandatory

hmgaudecker:
I strongly favor the alternative. Let the user be explicit about this. 

We should, however, document eventually how to prepare various datasets and what sensible default values would be if a variable cannot be reasonably prepared. Andreas Peichl mentioned that something like this exists for TAXSIM (? or Euromod?). @mjbloemer, would you know a reference from the top of your head? 

As for internal function calls, the goal is make everything usable from [CuDF](https://github.com/rapidsai/cudf) eventually. In order to use something like apply_grouped on [this](https://docs.rapids.ai/api/cudf/stable/) page. We will need to think through exactly how to make it work, but for now we should definitely strive to call functions only with the minimal necessary subset of columns of df.

Eric-Sommer:
potentially stupid, slightly off-topic question: If we are using CuDF, can "standard" users with a CPU make use of the package at all?

CruelChris:
For EUROMOD, there is a document with modelling conventions: https://www.euromod.ac.uk/publications/euromod-modelling-conventions
It has info on variable names, values for missings or N/A etc. 

There is more detailed information for every input dataset in its "DRD file" (Data  Requirement  Document). The default data for EUROMOD are derived from EU-SILC, and the DRD files are Excel tables that specify how each variable has been derived from its SILC source.

Eric-Sommer:
Hi @CruelChris  ;)

I don't know TAXSIM, but I'm pretty sure Euromod does not fill columns if they are not there. The error message "row too short" is what you see as a beginner pretty ofen. Their case is however somewhat different than ours as their input data are standardized and shipped along with the model.

CruelChris:
Indeed, the EUROMOD case is a bit different as the datasets basically come with the model, although there are also attempts to use it with different datasets (HFCS based, if I'm not mistaken).

EUROMOD has default values in case a variable is not there or missing, but it is not very explicit about it, I think.

hmgaudecker:
[...]

PR #45, opened 2019-10-18
Rewrite of Tax & Transfer module using apply

hmgaudecker:

Get rid of loops.

### Todo

- [x] Check that example is working
- [x] Apply (sic!) to other function calls in tax_transfer.py


Eric-Sommer:
I like. But was `inout` supposed to be `input`?

hmgaudecker:
We are all learning... That is basically preparation for CuDF support, where the output columns have to be pre-allocated and are modified in-place. I am sure I had other suggestions at a different point  in time.

Btw, this was the output of a meeting with @MaxBlesch on Friday, he will take this PR from there.

gettsim/benefits/alg2.py
```diff
@@ -12,61 +11,44 @@ def alg2(df, tb):
...
-        ]
-    ]
+    return df
 
 
 def regelsatz_alg2(df, tb):
     """Creating the variables need for the calculation of the alg2 regelsatz. Then
     according to the year the appropriate function is called"""
-    regel_df = pd.DataFrame(index=df.index)
+    child_dict = {}
```
Eric-Sommer: This dict contains the number of kids in various ages, but also adults. Therefore, I find `child_dict` slightly misleading. Maybe something like `hhage_count`?

MaxBlesch: It only contains the number of adults which directly follows from the number of children. So it should be clear that it contains information on the children mainly. I renamed it to a more self explainable variable.

gettsim/benefits/arbeitslosengeld.py
```diff
@@ -15,22 +15,20 @@ def ui(df_row, tb):
 
     eligible = check_eligibility_alg(df_row)
 
-    if eligible:
-        if df_row["child_num_tu"] == 0:
-            return alg_entgelt * tb["agsatz0"]
+    if eligible.all():
```
Eric-Sommer: Why `.all()`? ALG1 is an individual benefit and only individuals should run through the function.

gettsim/pensions.py
```diff
@@ -85,7 +82,7 @@ def _zugangsfaktor(df_row):
 def _regelaltersgrenze(df_row):
     """Calculates the age, at which a worker is eligible to claim his full pension."""
     # If born after 1947, each birth year raises the age threshold by one month.
-    if df_row["byear"] > 1947:
+    if (df_row["byear"] > 1947).all():
         return np.minimum(67, ((df_row["byear"] - 1947) / 12) + 65)
```
Eric-Sommer: again. this should be one row.

gettsim/benefits/arbeitslosengeld.py
MaxBlesch: The function is called on the individual level. Therefore the Series contains only one element!

gettsim/social_insurance.py
```diff
@@ -20,160 +21,148 @@ def soc_ins_contrib(df, tb):
...
-            "grvbs_mini"
-        ] * np.maximum(175, df["m_wage"])
-    """
+    # """
+    # if yr > 2012:
+    #     ssc.loc[inout["m_wage"].between(1, tb["mini_grenzew"]), "rvbeit"] = tb[
+    #         "grvbs_mini"
+    #     ] * np.maximum(175, inout["m_wage"])
+    # """
     # Self-employed may insure via the public health and care insurance.
```
Eric-Sommer: Should we delete this bit altoghether? I added it following a discussion in the topics course, but it does not seem to be the default.

gettsim/tax_transfer.py
```diff
@@ -56,47 +46,200 @@ def tax_transfer(df, tb, tb_pens=None):
...
+        "gross_e5_tu",
+        "gross_e6_tu",
+        "gross_e7_tu",
+        "ertragsanteil",
+        "sonder",
+        "hhfreib",
+        "altfreib",
+        "vorsorge",
+    ]
+    df = df.groupby(tax_unit)[in_cols + out_cols].apply(zve, tb=tb)
```
Eric-Sommer: Don't you need a `df[out_cols] = np.nan`?

gettsim/tax_transfer.py
```diff
@@ -56,47 +46,200 @@ def tax_transfer(df, tb, tb_pens=None):
...
+    out_cols = [
+        "incometax_tu",
+        "incometax",
+        "kindergeld",
+        "kindergeld_hh",
+        "kindergeld_tu",
+    ]
+    for col in out_cols:
+        df[col] = np.nan
+    df = df.groupby(tax_unit)[in_cols + out_cols].apply(favorability_check, tb=tb)
```
Eric-Sommer: This whole sequence is highly repetitive. Can be possibly put in a loop / function as well.

gettsim/taxes/favorability_check.py
```diff
@@ -12,51 +11,46 @@ def favorability_check(df, tb):
...
-        fc.loc[:, "abgst_tu"] = 0
-    # Why do we calculate that? Can we not just set it as tu?
+    if (not ("nokfb" in max_inc)) | (tb["yr"] <= 1996):
+        df.loc[:, "kindergeld"] = 0
+        df.loc[:, "kindergeld_tu"] = 0
+    if "abg" in max_inc:
+        df.loc[:, "abgst"] = 0
+        df.loc[:, "abgst_tu"] = 0
+    # TODO: Why do we calculate that? Can we not just set it as tu?
     # Aggregate Child benefit on the household.
```
Eric-Sommer: Because you may have several tax units with different kids in one household => Patch-work families, two parents with one child each, non-married.

gettsim/benefits/arbeitslosengeld.py
Eric-Sommer: so there is no need for `.all()`?

hmgaudecker: There should not be, indeed. 

Also, as you go through these things, can we please directly call them:
```
person
tax_unit
household
```
instead of `df_row` and `df`, respectively? If you insist, it can also be `p`, `tu`, `hh`. But this way it is clear immediately which level a functions touches.

gettsim/benefits/benefit_checks.py
```diff
@@ -94,47 +72,44 @@ def wealth_test(df, tb):
...
     """
-
-    wt = pd.DataFrame(index=df.index)
-    for v in ["regelbedarf", "wohngeld_basis_hh", "kiz_temp"]:
-        wt[v] = df[v]
-    # Initiate birth year series
-    wt["byear"] = tb["yr"] - df["age"]
+    if "byear" not in df.columns.values:
+        # Initiate birth year series
+        df["byear"] = tb["yr"] - df["age"]
```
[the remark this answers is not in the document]
MaxBlesch: I think this is something for #41 . If it is given in DataFrame, but if not we can just calculate it.

gettsim/social_insurance.py
MaxBlesch: Kicked this out

gettsim/benefits/benefit_checks.py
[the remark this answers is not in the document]
hmgaudecker: That will have to be in the data. Anything else may lead to inconsistencies.

gettsim/benefits/arbeitslosengeld.py
MaxBlesch: Will adapt it and use `iloc[0]` instead of `.all()`.

hmgaudecker: But why is it not a scalar?

MaxBlesch: When i use apply and on a groupby object which has single row groups, the functions are called with a one row dataframe. If I call by label, I get a one elemented Series.

hmgaudecker: Then we should rewrite `check_eligibility_alg` so that it returns a scalar.

MaxBlesch: Okay. So I shifted the iloc call to `check_eligibility_alg`.

gettsim/tax_transfer.py
MaxBlesch: I wrote a function, which does that for me.

gettsim/benefits/benefit_checks.py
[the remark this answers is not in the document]
MaxBlesch: Included it in test data!

gettsim/taxes/favorability_check.py
MaxBlesch: Deleted TODO.

gettsim/tax_transfer.py
MaxBlesch: Yes. Fixed it
.

gettsim/taxes/zve.py
```diff
@@ -13,100 +12,76 @@ def zve(df, tb):
...
+    df = tb["calc_hhfreib"](df, tb)
 
     # Taxable income (zve)
     # For married couples, household income is split between the two.
     # Without child allowance / Ohne Kinderfreibetrag (nokfb):
-    zve.loc[~df["child"], "zve_nokfb"] = zve_nokfb(zve, tb)
+    df.loc[~df["child"], "zve_nokfb"] = zve_nokfb(df, tb)
     # Tax base including capital income
-    zve["zve_abg_nokfb"] = zve_abg_nokfb(df, zve, tb)
+    df = zve_abg_nokfb(df, tb)
```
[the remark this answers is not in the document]
MaxBlesch: You are probably right. But I would now adapt all functions that way. We have varaibles like earnings points (`EP`) which get updated and therefore overwritten.

hmgaudecker:
After defining `person` in `tax_transfer`, please check the resulting groups are unique (for now, we do not need to support a panel).

Then please add a `squeeze` to the interface of `_apply_tax_transfer_func`, which is passed to the `groupby` call inside of it. It should be `True` if the grouping variable is `person` and False otherwise. 

Then we do not need the `iloc` and the code is much clearer and less error-prone.

MaxBlesch:
I added the check. But the issue with squeeze is more complicated. The squeeze argument of groupby applies the squeeze function to the output of the tax_transfer function. We want to have it applied to the input. I could wirte a wrapper which is called from `_apply_tax_transfer_func`?

MaxBlesch:
I wrote a draft on the wrapper https://github.com/iza-institute-of-labor-economics/gettsim/blob/0ee5b25e96723de19dc7714055ec346a4f15ae2c/gettsim/tax_transfer.py#L438

hmgaudecker:
Sorry, this is what haltend of you look too  briefly. Wrapper sounds good, but please do not call the keyword "squeeze" then.

MaxBlesch:
I adjusted all individual functions.

hmgaudecker:
Cannot check before Monday anymore, but that should make it much clearer. Thanks!

MaxBlesch:
Can we still close the PR? It changed a lot throughout all functions and if somebody else wants to work over the weekend. It would be good if the new design is in master.

hmgaudecker:
Fine with me. With a bit of luck I will manage in the next half hour. If somebody else wants to go ahead, please do so!

hmgaudecker:
Ah, just go ahead and merge it, @MaxBlesch. I just realise that I cannot because I actually opened it...

hmgaudecker:
@Eric-Sommer, @MaxBlesch: When merging, please edit the automatically-generated list of commit messages so that only the relevant end result is described. There is 30 lines or so in this commit including stuff like "    * Deleted todo. Wrote answer as comment.", which seems irrelevant for somebody trying to understand what happened there. 

But this really looks great!

Issue #12, opened 2019-07-22
Check whether some collaborative editing / communication tools are useful for us


hmgaudecker:
E.g., 

- https://stackedit.io/
- https://hackmd.io/

look nice for drawing not-so-heavy coders into the community. 

Any experiences?

hmgaudecker:
Via numpy mailing list: Slack alternative 

FWIW we use Zulip for scikit-image and we are very happy with it. See https://skimage.zulipchat.com

Advantages over Slack:

- free for FOSS, including unlimited archives (https://zulipchat.com/for/open-source/)
- login with GitHub
- Zulipchat is itself FOSS (I have successfully set up my own instance before)
- (upcoming) public archive (read without login)

mjbloemer:
On communication tools: I have never used slack or zulip. They seem as full featured communication solutions to me. Plus mainly web browser or electron based client software.

As slim chat alternatives I would suggest:
- IRC
- XMPP
- matrix

In these cases I imagine we could have a public community room and a nonpublic room.

I am also fine with mailing lists.

Eric-Sommer:
Let me express strong distaste for mailing lists 😃. Slack seems the way to go to me, with the downside of not being FOSS.

janosg:
I have a strong preference for using one tool for everything. It looks like Zulip has some advantages over Slack, but I am only in favor of using it if it also replaces Slack for internal communication. Ideally, we would also align the tools between GETTSIM and OpenSourceEconomics ...

hmgaudecker:
Browsing a little bit, I would favor Zulip: 

- Free for FOSS (=us). IIRC, Slack ends up being quite expensive for many users, so it would not scale well. The 10k-searchable messages limit of the free version is reached quite early. 
- Searchable archives. This is at least as much for users as it is for developers. You do not want to open an issue just because as a Python newbie, some error is puzzling you (re @tobiasraabe's comment [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/47#issuecomment-544968923)). If some problem crops up repeatedly, google will pick that up (well, eventually [Coming soon: Allow users to read and search public stream history in Zulip's UI without first creating an account.](https://github.com/zulip/zulip/issues/13172)).

hmgaudecker:
[...]

PR #47, opened 2019-10-22
Suggestion for GEP process.

hmgaudecker:

We started out by making design decisions face-to-face in Bonn; with more geographic diversity this has shifted towards discussions in PRs (e.g., #31 or #41). Both forms are not ideal for various reasons. We will soon have to make a number of decisions with long-term impact that will require careful thought.

I thus want to propose adopting the same approach Python has had for a long time and which is also used by NumPy: GETTSIM Enhancement Protocols, or GEPs. To quote from the N(umpy)EP descriptions: *NumPy Enhancement Proposals (NEPs) describe proposed changes to NumPy. NEPs are modeled on Python Enhancement Proposals (PEPs), and are typically written up when large changes to NumPy are proposed.*. As in Numpy, they would live in subfolder of the docs folder. We could just render them in a "developer" section of the docs, a separate repo would be a bit too much IMHO.

My suggestion for a list of initial GEPs would be:

1. Aims and scope of GETTSIM
2. High-level API (essentially what we want to support in ``tax_transfer()`` — e.g., do we force people to convert their datasets to true long format with one row per household member, or do we also support variables nkids, age_kid_1, age_kid_2, ..., nkids, age_kid_10?)
3. Format for input data (see #31)
4. Generic API for internal functions
5. Mechanism for overriding specific functions and specification of custom parameters (crucial for reforms)

Of course, neither are titles set in stone nor is the order. I am sure I have forgotten something that is important. So please add any suggestions you may have. Similarly if you think this is complete overkill etc..

Looking forward to some discussion!

### Todo

- [x] We do not have a mailing list or other forum for discussion outside of Github, also see #12. We need to establish **one** place for general discussion / questions / etc. (i.e., not related to issues / code in PR's).


hmgaudecker:
@janosg, @mjbloemer, @MaxBlesch, @tobiasraabe, @Eric-Sommer  - thoughts?

Eric-Sommer:
As my experience in participating in large scale projects à la numpy is limited to say the least, I'm not sure if I understand the fundamental difference between GEP and PR. Apparently, the idea is to source out larger PRs.

Just one thought on this: In my view, numpy and co. are inherently much broader in their scope. Gettsim will, one day, cover the German tax-benefit system in a comprehensive manner. In other words, I feel there is a natural ceiling for us in terms of scope. Once this is reached, there will be few major enhancements. You may have other views, so let's discuss them upfront (or in GEP no. 1, if you like ;) ). I totally agree that the five points you listed above are among the ones we need to decide now, as they are quite fundamental. Whether we do this via PR or GEP, I'm rather agnostic. Personally, I have a preference for tackling the actual issues rather than thinking too much about another 'institutional layer' we might not need (again, I might be missing something).

The ToDo ("place for general discussion") could be added as GEP no.6.

hmgaudecker:
Thanks for your thoughts, @Eric-Sommer! Very valid points. 

> Gettsim will, one day, cover the German tax-benefit system in a comprehensive manner. 

That is, hopefully, true. But I think it is good to be more precise than that. What do we mean by "comprehensive"? How do we think about the speed-accuracy trade-off? How do we allow users to specify reforms?

> In other words, I feel there is a natural ceiling for us in terms of scope. Once this is reached, there will be few major enhancements. 

True, and I have been debating with myself whether I should call the document "nep-0" or "nep-00" :-)

> You may have other views, so let's discuss them upfront (or in GEP no. 1, if you like ;) ). I totally agree that the five points you listed above are among the ones we need to decide now, as they are quite fundamental. Whether we do this via PR or GEP, I'm rather agnostic.

The problem with PRs is that they are hard to dig out once merged and closed. Plus, they reflect the entire discussion, much of which will be misleading and irrelevant because the solution ended up being something else than what was initially discussed.

I believe that there is considerable value for future users and developers in being able to concisely see why design decisions have been made. But of course, the discussion on GEPs would happen in PRs  :-)

MaxBlesch:
I just read some NEPs and in my opinion something similar will become very handy. I understand them as a detailed summary of discussions in Issues and PRs and therefore these protocols can provide a detailed roadmap of the project. I think the 5 mentioned are a good start and we can extend or shrink at any time. 

For the communication I would suggest a slack workspace with a link in the gettsim readme to join.

tobiasraabe:
I also like the idea of GEPs as a summary of discussions. If a discussion gets to long, the participants write a GEP and move forward. Thus, anyone entering the discussion can read the highly condensed GEP instead of the long back-and-forth in comments.

I am not sure whether one needs a separate channel for discussions like Slack or Google Groups. What is the additional benefit instead of using issues? 

I assume 5) includes the discussion on the parameter format in #31.

janosg:
I think GEPs can provide two advantages:

1) They have a higher level of abstraction. Issues should have a size that can be tackled in one PR. PRs are even tied to a concrete implementation. GEPs could help the developers to discuss problems or implementations that are much bigger than one PR.

2) We could have econ GEPs that are written by users who are completely unfamiliar with the implementation of GETTSIM or have never used GitHub. They could just describe what they would like to be able to do and how the interface for that should look like. For this we would need an extremely easy way of editing the GEPs in a browser, but It seems that those exist (see #12).

hmgaudecker:
@tobiasraabe: On my machine, the TOC now renders as


    GETTSIM Enhancement Protocols
    Indices and tables
    Code of Conduct (CoC)

I have no idea why it puts "Indices and Tables" in the middle. If you spot why, that would be great to fix, but nothing to waste time on.

hmgaudecker:
Also, we do not want the template to be rendered, but rather have a link to the rst file. Could you please change that? I have to move on for today.

Eric-Sommer:
> @tobiasraabe: On my machine, the TOC now renders as
> 
> ```
> GETTSIM Enhancement Protocols
> Indices and tables
> Code of Conduct (CoC)
> ```
> 
> I have no idea why it puts "Indices and Tables" in the middle. If you spot why, that would be great to fix, but nothing to waste time on.

Because `Indices and Tables` are on the same hierarchy in `index.rst` as the headline

hmgaudecker:
https://github.com/numpy/numpy/pull/14734

I did not have a chance to read it, but the idea sounds right. Could somebody check, please?

hmgaudecker:
Based on the discussion at https://gettsim.zulipchat.com, I think we can consider this as accepted. 

@tobiasraabe, could you have a look into the Numpy link and update the template if it seems useful? If somebody could merge after that, it would be great!

tobiasraabe:
Their changes sound reasonable to me. The changes move the discussions on the mailing list to discussions with end-users which is what we also intended.

- clearer focus on usability
- focus on backward compatibility

tobiasraabe:
Hi @Eric-Sommer, chose you because you participated most frequently in the discussion. Feel free to pass the review to someone else if you are busy.

docs/geps/gep-00.rst
```diff
@@ -0,0 +1,278 @@
...
+The proposal should be submitted as a draft GEP via a `GitHub pull request`_ to the
+``doc/geps`` directory with the name ``gep-<n>.rst`` where ``<n>`` is an appropriately
+assigned two-digit number (e.g., it is ``gep-00.rst`` for this document). The draft must
+use the :ref:`gep-template` file.
+
+Once the PR is in place, the GEP should be announced on the in the #general/geps topic
+on `Zulip`_ for discussion. Discussion about implementation details will take place on
+the pull request, but once editorial issues are solved, the PR should be merged, even if
+with draft status. The mailing list e-mail will contain the NEP upto the section titled
+"Backward compatibility", so as to make it digestible to a wide audience. The mailing
```
Eric-Sommer: NEP => GEP

docs/geps/gep-00.rst
```diff
@@ -0,0 +1,278 @@
...
+be impractical when subjected to the test of implementation.  Often it makes sense for
+the prototype implementation to be made available as PR to the GETTSIM repository
+(making sure to appropriately mark the PR as a WIP).
+
+
+Review and Resolution
+^^^^^^^^^^^^^^^^^^^^^
+
+GEPs are discussed on the mailing list.  The possible paths of the status of GEPs are as
+follows:
```
Eric-Sommer: ...are discussed in the #gep topic (?) on Zulip

tobiasraabe:
I can add the link, but everyone without an account is asked for registration which is then denied because it is invite-only. Is it possible to make the channel publicly visible, but accounts are invite-only or need to be confirmed?

ChristianZimpelmann:
> I can add the link, but everyone without an account is asked for registration which is then denied because it is invite-only. Is it possible to make the channel publicly visible, but accounts are invite-only or need to be confirmed?

This doesn't seem to be possible, yet: https://github.com/zulip/zulip/issues/13172

hmgaudecker:
> Their changes sound reasonable to me. The changes move the discussions on the mailing list to discussions with end-users which is what we also intended.
> 
>     * clearer focus on usability
> 
>     * focus on backward compatibility

Great, can you add that version, then?

I think, however, that we can tone down the backward-compatibility bit of NumPy by a lot. I do not expect people to embed this as deeply as people embed NumPy in other code. A reasonably stable API should be a primary goal, of course, but NumPy is a completely different animal in that respect.

hmgaudecker:
@tobiasraabe: Can you please replace the template by the updated version? If one of the others could review and merge over the weekend, that would be great!

hmgaudecker:
More or less. We should leave a sentence on it, but it is just not a primary concern to me.

PR #48, opened 2019-10-22
Add how to contribute section to documentation.

### LINKED ISSUE #36: Docs: How to contribute sections
Before we do anything like API docs, we should outline the workflow, this will be more stable... 

Both for contributors (remember we will have many newbies, so maybe even call a section "What to do in case of errors" and then explain in detail how to open an issue) and for developers (always open PR, request review, rebase on current master and squash merge).

tobiasraabe:

Closes #36. Thanks to Janos for the template!


hmgaudecker:
Shall we link to the style guide at https://estimagic.readthedocs.io/en/latest/contributing/styleguide.html for the moment? Not sure we want to do exactly the same and quite a bit of stuff depends on the architecture we will end up having, so I do not think that writing something down now is helpful. 

One thing we should add, however, is something on language. Like here: https://openfisca.org/doc/contribute/language.html#language. I would suggest having all variable names in German (just for precision) and make a translation in a central table mandatory. Also, we will need to add readability requirements. IMHO, we are still too Stata-biased (read: championing brevity over readability. I always have to take a second look at zve...)

Thoughts, @mjbloemer, @Eric-Sommer, @stichnoth ? How do the Stata-codebases handle that and how would that fare in your view in a setting where you do not only have maintainers looking at the code?

stichnoth:
My old ZEW model uses a wild mix of English and German. In the current Python model I went for German names in order to be more consistent and precise. I try to use readable names, but I'm not always consistent here; for instance, I use wohngeld and not wog, but kiz instead of kinderzuschlag. I also use zve and the perhaps equally cryptic est for einkommensteuer.

You are right that these acronyms and other abbreviations are not ideal for non-maintainers looking at the code. Then again, even with tab completion longer names can be a bit of a pain when writing the code. And names such as calc_arbeitslosengeld_II_erwerbstaetigenfreibetraege improve readability when seen in isolation, but may reduce readability of the overall code by leading to longer lines and/or more frequent line splits, and by generally increasing the character-to-programming-logic ratio.

Eric-Sommer:
Variable naming conventions is a very important issue on which I have a couple of ideas. I'd argue they deserve their own GEP which is going to be referenced in the contribution guidelines. From my user experience with Euromod, naming conventions play an equivally large role for users when it comes to

1. alter/specify the model input 
2. understand the model output.

hmgaudecker:
Yes, that is very valid. I just wanted to bring up the language point here.

hmgaudecker:
@tobiasraabe: From my perspective this looks good for now since @Eric-Sommer just wrote over on Zulip that he is in the process of drafting a GEP on variable names. So hopefully we will have enough material to link soon enough, no need to go into any detail there.

tobiasraabe:
Great. Review?

PR #49, opened 2019-10-25
Elterngeld

### LINKED ISSUE #14: Add Elterngeld
2/3 of previous income for up to 1 year.

boryana-ilieva:
With this PR we want to include Elterngeld as part of the benefits individuals in Germany are eligible for.

When done Closes #14


hmgaudecker:
Great, thanks!

@stichnoth: would you have a chance to help look at this?

stichnoth:
I'm new to GETTSIM, so I hope I understand the code and the review process correctly. 

Am I correct that the code computes elterngeld as 2/3 of previous net earnings? This would be fine as a first approximation, but there are a few special cases. The standard rate is indeed 67%, but it is higher for lower previous earnings and lower for higher ones. There is also a cap at 1800 euros. And the elterngeld only replaces the earnings that are lost, so if a parent works part-time the amount would not equal 67% or previous earnings, but 67% of previous minus current earnings.

There are also a few extra rules for siblings and twins etc., but I guess it's fine to leave them aside for now.

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+        elt_geld += list(
+            eltg_df.loc[
+                (eltg_df["pid"] == pid)
+                & (eltg_df["year"] == referece_year.to_numpy()[0]),
+                "m_wage",
+            ]
+            * 2
+            / 3
+        )
```
Eric-Sommer: No hard coded numbers please. Parameters should be taken from param.xls (which is yet an excel file) and written into the `tb` dictionary. If I'm not mistaken, the parameters are already in the file.

hmgaudecker:
Hey guys, let us take it easy, please. This was explicitly a first pass and not supposed to be a completed patch. In general, we are  a collaborative effort and not everybody has spent the last ten years coding up tax rules 😉. All valid points, but let us please try to be a bit more encouraging.

Eric-Sommer:
Sure thing. Sorry @boryana-ilieva for being too laconic after a long week. This is very valuable and a good start!

stichnoth:
Apologies if I came across as too critical. I think it makes perfect sense to start with the base case and to extend it later if the need arises.

Eric-Sommer:
fyi: the parameters you need are called `elgfaktor` (0.67), `elgmax` (1800) and `elgzwst` (1201).

boryana-ilieva:
Thank you for your input. As clarified above, this was only an initial commit without a claim for completeness, which is why I did not request a review.

MaxBlesch:
I wrote a draft on elterngeld. I also implemented the reduction/increase of the payment share according to § 2 (2) BEEG. We don't really have a variable `net_inc_last_worked`so I used `m_wage_l1`. The description you can find in https://gettsim.readthedocs.io/en/latest/getting-started.html . Has some of you a better idea or some other other comments? Especially @stichnoth, @Eric-Sommer and @mjbloemer ?

hmgaudecker:
I have not checked in detail, but it definitely would have to be net income. Ideally before making any deductions for Werbungskosten. We probably need to require Steuerklassen (I do understand this is not possible to get from the SOEP, but nevertheless...).

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,36 @@
...
+        person["elt_geld"] = max(min(elt_geld_calc, params["elgmax"]), params["elgmin"])
+
+    else:
+        person["elt_geld"] = 0
+
+    return person
+
+
+def calc_elterngeld_percentage(considered_wage, params):
+    if considered_wage < params["elg_st_1"]:
```
Eric-Sommer: can you explain in a docstring briefly what's happening in this function?

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,36 @@
+def elt_geld(person, params):
+    """This function calculates the monthly benefits for having
+    a child that is up to one year old (Elterngeld)"""
+
+    if person["elt_zeit"]:
+        considered_wage = person["m_wage_l1"] - person["m_wage"]
+
```
Eric-Sommer: you definitely need net wage. you can just feed this into the tax tariff, similar to how it is done with the unemployment benefit. social sec. contributions also need to subtracted.

gettsim/tests/test_data/test_dfs_eltg.csv
```diff
@@ -0,0 +1,16 @@
+hid,tu_id,pid,m_wage,elt_zeit,year,,,elt_geld
+1,1,23,1500,FALSE,2017,,,0
```
Eric-Sommer: why is there no `m_wage_l1`?

Eric-Sommer:
> Ideally before making any deductions for Werbungskosten. We probably need to require Steuerklassen (I do understand this is not possible to get from the SOEP, but nevertheless...).

Werbungskosten also need to deducted. Vorsorgeaufwendungen as well. Soli gets deducted, social security contributions as well (there are however different rates for Elterngeld, see §2f BEEG).

hmgaudecker:
How would that work in practice? I can see it if you pre-registered them. But if you only claim them in the tax declaration? Speaking based on n=2 here, of course 😉

gettsim/benefits/elterngeld.py
MaxBlesch: Included docstring!

gettsim/benefits/elterngeld.py
MaxBlesch: Yeah about that. I was really wondering if we want to do this every time for some functions. I mean we have a detailed ssc function and then just use and approximation for alg and then maybe also here...Shouldn't we rely on our code instead of approximations?

Eric-Sommer:
In practice you hand in your wage bills on which basis deductions are calculated. You get only lump sum werbungskoten which might be higher in your income tax. For vorsorgraufwendungen its not a big deal either

gettsim/benefits/elterngeld.py
Eric-Sommer: We could use the values from the alg call of course. It's the same lagged wage.

hmgaudecker:
So you are saying that 1k / year are automatically deducted without me doing anything? Is there a reference for that?

Eric-Sommer:
> So you are saying that 1k / year are automatically deducted without me doing anything? Is there a reference for that?

§2e (2) BEEG:

> (2) Bemessungsgrundlage für die Ermittlung der Abzüge für Steuern ist die monatlich durchschnittlich zu berücksichtigende Summe der Einnahmen nach § 2c, soweit sie von der berechtigten Person zu versteuern sind, und der Gewinneinkünfte nach § 2d. Bei der Ermittlung der Abzüge für Steuern nach Absatz 1 werden folgende Pauschalen berücksichtigt:
> - der Arbeitnehmer-Pauschbetrag nach § 9a Satz 1 Nummer 1 Buchstabe a des Einkommensteuergesetzes, wenn die berechtigte Person von ihr zu versteuernde Einnahmen hat, die unter § 2c fallen

§9a EStG:

> Für Werbungskosten sind bei der Ermittlung der Einkünfte die folgenden Pauschbeträge abzuziehen, wenn nicht höhere Werbungskosten nachgewiesen werden: von den Einnahmen aus nichtselbständiger Arbeit [...] ein Arbeitnehmer-Pauschbetrag von 1.000 Euro;

gettsim/benefits/elterngeld.py
MaxBlesch: I renamed and reused the function.

gettsim/tests/test_data/test_dfs_eltg.csv
MaxBlesch: I adapted the test.

hmgaudecker:
Looking at this: We ignore Elterngeld Plus completely? Is that on purpose? What are the empirical fractions these days?

MaxBlesch:
Actually I ignored this on purpose and just implemented the basic rule. This is what I interfered from the previous discussion i this PR.

gettsim/benefits/arbeitslosengeld.py
```diff
@@ -14,7 +14,7 @@ def ui(
 
     """
 
-    alg_entgelt = _alg_entgelt(
+    alg_entgelt = proxy_net_wage_last_year(
         person,
```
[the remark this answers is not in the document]
MaxBlesch: I changed the interface of the function to give the social security contribution separate. Also I added a parameter to elterngeld. We do not check for private insurance in here yet?! Also not for arbeitslosengeld. Should we change that?

gettsim/data/elterngeld.yaml
```diff
@@ -34,3 +45,89 @@ elgzwst:
...
+    2012-09-18:
+      value: 180
+      note: Artikel 1 G. v. 10.09.2012 BGBl. I S. 1878
+soz_vers_pausch:
+  name:
+    de: Sozialversicherungspauschale zur Berechnung des Nettoeinkommens
+    en:
+  description:
+    de: § 153 (1) Nr. 1 SGB III
+    en:
```
Eric-Sommer: It's §2f BEEG and it's 0.21. Could you at least mention the individual components (9 + 10 + 2) in the note?

gettsim/benefits/arbeitslosengeld.py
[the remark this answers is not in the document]
Eric-Sommer: Good point. I don't know how they are treated. Do self-employed get Elterngeld? These lump-sum values are supposed to get a fictive net wage, which is what Elterngeld is supposed to replace to some extent.

Eric-Sommer:
We'd also need the legislation from 2008 to 2011 without sibling bonus and a different calculation of net income.

gettsim/benefits/arbeitslosengeld.py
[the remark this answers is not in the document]
hmgaudecker: Yes, but for them the last calendar year matters, not the 12 months before birth. I only know because having some solar panels on your roof makes turns you into a business owner by that rule... 

So we can just take last year's tax statement, will need to figure out how income tax classes are handled in couples.

MaxBlesch:
> We'd also need the legislation from 2008 to 2011 without sibling bonus and a different calculation of net income.

There is already 0 for sibling bonus from 2007 to 2011. In this time there was "Mehrlingsbonus". Do we have that information on the kids? 

There were no extra parameters before for the calculation right? I just saw that the relevant wage was the one from the employers payment sheets.

Also I only found that the degression of the percentage has not always been there. Til 2009 there was no degression. Since 2011 there is degression and it was caped later.

I haven't touched eligiblity yet. Should we that include?

Eric-Sommer:
> There is already 0 for sibling bonus from 2007 to 2011. In this time there was "Mehrlingsbonus". Do we have that information on the kids?

Usually not, and I'd vote for not including it.
 
> There were no extra parameters before for the calculation right? I just saw that the relevant wage was the one from the employers payment sheets.

...which means we need the actual social contribution parameters.
> 
> Also I only found that the degression of the percentage has not always been there. Til 2009 there was no degression. Since 2011 there is degression and it was caped later.

Right.
 
> I haven't touched eligiblity yet. Should we that include?

It should be definitely included here. In principle, very tricky in principle since you can also take Elterngeld if the child is 5. It's probably OK to base eligibility on the fact that a child is 0 or 1 years.

gettsim/tests/test_data/test_dfs_eltg.csv
```diff
@@ -0,0 +1,13 @@
+hid,tu_id,pid,m_wage,m_wage_l1,east,incometax,soli,svbeit,elt_zeit,geschw_bonus,year,,,elt_geld,considered_wage,payed_percentage
+1,1,23,0,0,FALSE,0,0,0,FALSE,FALSE,2018,,,0,,
```
Eric-Sommer: do we need `incometax`, `soli` `svbeit` here or aren't they calculated by the function? In addition, `geschw_bonus` needs to be made endogenous.

MaxBlesch: Yes. For the net income of the current year! Or is this all calculated by these proxy parameters? `geschw_bonus` is defnetly endogenous as well as `elt_zeit`. Just was not sure if we touch eligibility here. As I am now, I will include.

Eric-Sommer: Right. As soon as you're done, I'll add my own elterngeld calculation as a test ;)

hmgaudecker:
> > There is already 0 for sibling bonus from 2007 to 2011. In this time there was "Mehrlingsbonus". Do we have that information on the kids?
> 
> Usually not, and I'd vote for not including it.

How would we not see if kids are twins? Just to get an idea of whether it is worth it: How large is the difference? 

> > I haven't touched eligiblity yet. Should we include that?
> 
> It should be definitely included here. In principle, very tricky in principle since you can also take Elterngeld if the child is 5. It's probably OK to base eligibility on the fact that a child is 0 or 1 years.

We would need a state variable that we can store with the child: `elt_geld_mon_mut`, `elt_geld_mon_vat` or so. Then check whether at any point each is <= 12 individually and the sum <= 14 (IIRC)?

Eric-Sommer:
> We would need a state variable that we can store with the child: `elt_geld_mon_mut`, `elt_geld_mon_vat` or so. Then check whether at any point each is <= 12 individually and the sum <= 14 (IIRC)?
Actually, we'd need that for every child.

hmgaudecker:
> > We would need a state variable that we can store with the child: `elt_geld_mon_mut`, `elt_geld_mon_vat` or so. Then check whether at any point each is <= 12 individually and the sum <= 14 (IIRC)?
> 
> Actually, we'd need that for every child.

Of course.

MaxBlesch:
This is a firrst draft including eligibility. I did not include Elterngeld Plus, but as I understand it, it extends the time at which you are eligible to claim elterngeld? Do we have the necessary information on that? Also I am really not sure, where to include it in the `tax_transfer.py` @Eric-Sommer.

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,215 @@
...
+    current_net_wage = calc_net_wage(person)
+
+    return net_wage_last_year - current_net_wage
+
+
+def calc_elterngeld_percentage(considered_wage, params):
+    """ This function calculates the percentage share of net income, which is
+    reinbursed in the time of recieving elterngeld.
+
```
Eric-Sommer: "... which is reimbursed when receiving Elterngeld"

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,215 @@
...
+    According to § 2 (2) BEEG the percentage increases below elg_st_1 and decreases
+    above elg_st_2.
```
Eric-Sommer: "According to ..., the percentage increases for low incomes and decreases for high incomes until its minimum of 65%.

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,215 @@
...
+            (children["byear"] == np.max(children["byear"]))
+            & (children["bmonth"] == np.max(children["bmonth"]))
+        ]
+        # Get the birthdate of the youngest child(If "Mehrlinge", then just take one)
+        birth_date = datetime.date(
+            day=youngest_child["bday"].iloc[0].astype(int),
+            month=youngest_child["bmonth"].iloc[0].astype(int),
+            year=youngest_child["byear"].iloc[0].astype(int),
+        )
+        age_youngest_child = relativedelta.relativedelta(params["date"], birth_date)
```
Eric-Sommer: what if this is `< 0`? Set it to zero or raise Exception? I vote for the latter.

gettsim/benefits/elterngeld.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if eligible_age & eligible_consumed:
+            # Each parent can't claim more than 12 month
+            eligible = ~household["child"] & (
+                household["elterngeld_mon"] <= params["max_months"]
+            )
+
+            household.loc[eligible, "elternzeit_anspruch"] = True
+
+            if (len(children[(params["year"] - children["byear"]) < 3]) == 2) | (
```
Eric-Sommer: please add one comment on what this if-condition contains

gettsim/tests/test_data/test_dfs_eltg.csv
```diff
@@ -0,0 +1,20 @@
...
+3,3,56,3400,3600,FALSE,FALSE,708.27,38.96,680,1976,7,7,41,0,0,0,,2018,,,,FALSE,FALSE,0,0,,
+3,3,56,3600,3400,FALSE,FALSE,788.34,43.36,756,1976,8,8,42,0,0,0,,2019,,,,TRUE,FALSE,0,0,,
+4,4,59,0,0,FALSE,TRUE,0,0,0,2017,12,6,0,1,0,0,,2018,,,,FALSE,FALSE,0,0,,
+4,4,57,3600,3400,FALSE,FALSE,779.33,42.86,756,1976,9,9,41,0,0,0,,2018,,,,TRUE,FALSE,0,0,,
+5,5,58,0,3600,FALSE,FALSE,0,0,0,1978,10,10,40,0,0,0,,2019,,,,TRUE,FALSE,0,1314.1765,2021.81,0.65
+5,5,60,0,0,FALSE,TRUE,0,0,0,2018,12,4,0,1,0,0,,2019,,,,FALSE,FALSE,0,0,,
+6,19,19,900,900,FALSE,FALSE,7.68,0,180,1980,11,11,37,0,0,0,,2018,,,,FALSE,FALSE,0,0,,
+6,19,20,0,0,FALSE,TRUE,0,0,0,2017,12,2,1,0,0,0,,2019,,,,FALSE,FALSE,0,0,,
+6,19,21,0,0,FALSE,TRUE,0,0,0,2018,12,10,0,2,2,0,,2019,,,,FALSE,FALSE,0,0,,
+6,19,19,0,900,FALSE,FALSE,0,0,0,1982,13,2,36,0,0,2,,2019,,,,TRUE,TRUE,0,650.554889,703.32,0.81834
```
Eric-Sommer: can we have at leat a few test cases from some (semi-) official calculator (https://familienportal.de/familienportal/rechner-antraege/elterngeldrechner)

Eric-Sommer:
> Also I am really not sure, where to include it in the `tax_transfer.py` @Eric-Sommer.

Needs to be right before income tax (#126), maybe right after `ui`.

gettsim/benefits/elterngeld.py
MaxBlesch: This would mean the birthdate of the child is after the date for the desired policy. Will we allow to calculate taxes for individuals, which have not lived at the policy date? If no, then this should be a general check in the beginning of the gettsim code.

gettsim/benefits/elterngeld.py
MaxBlesch: Added comment

gettsim/benefits/elterngeld.py
MaxBlesch: Changed comment, but didn't include specific values and instead the variable names.

gettsim/benefits/elterngeld.py
MaxBlesch: Corrected spelling.

gettsim/tests/test_data/test_dfs_eltg.csv
MaxBlesch: This calculator requires Steuerklassen. We assume Steuerklasse 4 for everybody? Are there maybe test cases from an elterngeld implementation at ZEW @stichnoth or IFO @mjbloemer .

gettsim/benefits/elterngeld.py
Eric-Sommer: If someone runs own households with this, there is most likely a mishap concerning the correct dates. At the very least, we should issue a warning.

gettsim/tests/test_data/test_dfs_eltg.csv
Eric-Sommer: It should work with Steuerklasse 4 in the absence of other deductions.

gettsim/benefits/elterngeld.py
MaxBlesch: Raised Error!

gettsim/tests/test_data/test_dfs_eltg.csv
MaxBlesch: As we do not have Lohnsteuer yet and just use a proxy for net income, we would pass a external test yet. However I included an example and we were close up to 10 Euro. I documented this the test file.

hmgaudecker: Well, it is hard to tell with that calculator where the difference might be coming from, isn't it? Rather unclear where they might make different assumptions on Social Insurance contributions, ... ? So within 10 € is fine, I think.

Please double-check the other entries with the calculator, though, to the extent that this is possible

MaxBlesch: I checked and we are quite close. I think we should do that again, once we have Lohnsteuer modeled.

MaxBlesch:
> can you at least introduce one test case where elterngeld is > 0 for wohngeld and alg2? Is it relevant for Kiz too?

I included tests with elterngeld >0 for alg2 and wohngeld calculation. Kiz uses directly the income calculated in the alg2 function. Therefore we can't write a separate test for kiz!

Eric-Sommer:
> I included tests with elterngeld >0 for alg2 and wohngeld calculation.

I'd just altered the `elterngeld` component in one of the existing tests, but why not :)

hmgaudecker:
@MaxBlesch: Do not forget to log your changes :-) Maybe easiest to quickly hijack #50 and approve again?

hmgaudecker:
And: Thanks, everyone!

PR #50, opened 2019-10-29
Modelling of old 'Vorsorgeaufwendungen' rules

Eric-Sommer:
Legislation on 'Vorsorgeaufwendungen' (the major tax deduction for employees) changed three times since 2004. To make things more complicated, there are checks which legislation is more profitable. This PR extends the model by incorporating the old rules. 

- [x] model pre-2004 legislation
- [x] add tests to 2004 legislation
- [x] replicate a test case post 2004 where 2004 is better
- [x] incorporate parameters in param file
- [x] Created a separate test file and fill it with semi-offiicial test cases.
- [x] require private pension contributions in order to a) match tests and b) account for their importance.
- [x] make sure 2004 "vorwegab" is not used after 2019 (see #135 for discussion).


Eric-Sommer:
closes #50. 

@mjbloemer, could you check whether the 2005 modelling is appropriate?

gettsim/taxes/zve.py
```diff
@@ -318,25 +319,64 @@ def vorsorge2010(tax_unit, tb):
...
+def vorsorge2005(tax_unit, tb):
+    """ Vorsorgeaufwendungen pre 2010
+    Pension contributions are accounted for up to €20k.
+    From this, a certain share can actually be deducted,
+    starting with 60% in 2005.
+    Other deductions are just added, up to a ceiling of 2400 p.a. for standard employees.
 
+    Background: https://bit.ly/32oqCQq
+    """
+    tb["max_altersvors_2005"] = 20000
```
mjbloemer: This parameter is not fixed from 2015 on: Artikel 5 G. v. 22.12.2014 BGBl. I S. 2417, see https://www.buzer.de/gesetz/4499/al46488-0.htm

gettsim/taxes/zve.py
```diff
@@ -318,25 +319,64 @@ def vorsorge2010(tax_unit, tb):
...
+    rvbeit_vors_max = np.minimum(
+        tb["max_altersvors_2005"] * vorsorge_year_faktor(tb["yr"]),
+        12 * tax_unit["rvbeit"] * 2,
+    )
+    # intermediate step
+    altersvors2005_int = ~tax_unit["child"] * (
+        vorsorge_year_faktor(tb["yr"]) * (12 * 2 * tax_unit["rvbeit"])
+        - (12 * tax_unit["rvbeit"])
+    ).astype(int)
+    altersvors2005 = np.minimum(rvbeit_vors_max, altersvors2005_int)
```
mjbloemer: My first draft given the explanations on the website you cited would be:

```python
altersvors2005 =
    (0.6 + 0.02 * (min(tb["yr"], 2025) - 2005)) 
    * np.minimum(tb["max_altersvors_2005"], 12 * tax_unit["rvbeit"] * 2)
    - 12 * tax_unit["rvbeit"]
```

where the factor is only once multiplied with the min(20000,2*rvbeit).

I think it leads to different results when 2*rvbeit (+ possibly private retirement provisions) gets close to the 20000. 

I might be wrong.

Eric-Sommer: you are right, the maximum threshold is fix.

mjbloemer: Even with the added commit 3910c23 it looks wrong to me.

Now you have altersvors2005 = `min(A,B)` with `A = min(20k, 2*RV)` and `B = f*(2*RV)-RV`.

From the linked website it should simply be altersvors2005 = `f*min(20k,2*RV)-RV`.

gettsim/taxes/zve.py
Eric-Sommer: You are right, but isn't this essentially irrelevant after 2010 since the new ruling is always better than the old one?

mjbloemer: I don't know if it is always better. Still, it might become again relevant in a reform or counterfactual. Probably depends on Riester stuff too.

gettsim/taxes/zve.py
Eric-Sommer: I'm not sure that source is 100% correct. I just found this one: https://www.smartsteuer.de/online/lexikon/g/guenstigerpruefung/ . They even provide the test cases ;)

mjbloemer: Which one of the examples in your new link is for the 2005 rule? 

The table under
> Abziehbar nach § 10 Abs. 3 EStG

shows the rule I sketched.

Eric-Sommer: OK the factor needs definitely to be applied to the threshold as well. the 2005 and 2010 rules actually only differ in their treatment of other contributions

tobiasraabe:
Sorry, updated branch to check CI feature.

docs/getting-started.rst
```diff
@@ -93,6 +93,8 @@ Required columns in input data
 +--------------+---------------------------------------------------------+-------------+
 |pkv           |Whether individual is (only) privately health insured    | Bool        |
 +--------------+---------------------------------------------------------+-------------+
+|priv_pens_exp |Monthly contributions to private pension insurance       | Float       |
```
hmgaudecker: `priv_pens_contr`

gettsim/taxes/zve.py
```diff
@@ -313,57 +318,145 @@ def calc_gross_e7(tax_unit, params):
...
+    # calculate x% of relevant employer and employee contributions and private contributions
     # then subtract employer contributions
+    altersvors2010 = calc_altersvors_aufwend(tax_unit, params)
+
     # also subtract health + care + unemployment insurance contributions
-    altersvors2010 = ~tax_unit["child"] * (
-        (0.6 + 0.02 * (np.minimum(params["year"], 2025) - 2005)) * (12 * rvbeit_vors)
-        - (12 * 0.5 * rvbeit_vors)
+    # 'Basisvorsorge': Health and old-age care contributiosn are deducted anyway.
+    sonstigevors2010 = 12 * (
```
hmgaudecker: Remove the year from the variable name. Typo in contributions.

gettsim/taxes/zve.py
```diff
@@ -313,57 +318,145 @@ def calc_gross_e7(tax_unit, params):
...
         § 10 (3) EStG
         The share of deductable pension contributions increases each year by 2 pp.
         ('nachgelagerte Besteuerung'). In 2018, it's 86%. Add other contributions;
         4% from health contributions are not deductable.
         only deduct pension contributions up to the ceiling. multiply by 2
         because it's both employee and employer contributions.
+
+        Contributions to other security branches can also be deducted.
+
+        This ruling differs to vorsorge2005() only in the treatment of other contributions.
```
hmgaudecker: So are they not close enough to each other to have one function with a case distinction?

gettsim/taxes/zve.py
```diff
@@ -313,57 +318,145 @@ def calc_gross_e7(tax_unit, params):
...
+        item_2 = 0.5 * np.minimum(params["grundbet"], vorsorg_rest)
+        item_3 = 0.5 * np.minimum((vorsorg_rest - item_2), 2 * params["grundbet"])
+
+    # Finally, add up all three amounts and assign in to the adults.
+    vorsorge2004 = ~tax_unit["child"] * (item_1 + item_2 + item_3).astype(int)
+
+    return vorsorge2004
+
+
+def vorsorge04_05(tax_unit, params, soz_vers_beitr_params):
```
hmgaudecker: The names are trying to give too much information. Without looking at the functions, it is unclear what this means.

I would suggest the following names:
```
vorsorge_pre_2005 (=current 2004)
vorsorge_since_2005 (=current 04_05, 04_10)
_vorsorge_since_2005 (current 2005 / 2010)
```

Btw, does the max(04, x)-thing apply also to insurance contracts that were made after 2005? Or only to old ones?

gettsim/policy_for_date.py
```diff
@@ -55,9 +56,11 @@ def get_policies_for_date(year, group, month=1, day=1, raw_group_data=None):
...
             tax_data["calc_hhfreib"] = calc_hhfreib_from2015
         if year >= 2010:
-            tax_data["vorsorge"] = vorsorge2010
-        else:
-            tax_data["vorsorge"] = vorsorge_dummy
+            tax_data["vorsorge"] = vorsorge04_10
+        elif year >= 2005:
+            tax_data["vorsorge"] = vorsorge04_05
+        elif year <= 2004:
+            tax_data["vorsorge"] = vorsorge2004
```
hmgaudecker: No test case for this, if I interpret the coverage statistics correctly?

docs/getting-started.rst
Eric-Sommer: I went for `pr_pens_contr`...I don't feel like extending the column width

gettsim/taxes/zve.py
Eric-Sommer: They are pretty close and hence both use `calc_altersvors_aufwand()`. The case distinction is 'one level above'

gettsim/taxes/zve.py
Eric-Sommer: OK with refactoring. It does not matter when insurance contracts started. Only the year in which contributions are paid matter.

docs/getting-started.rst
hmgaudecker: Done in 24ca3034 (<1min with Atom + sublime-style-column-selection :-)). Please add the `r` in the code.

gettsim/policy_for_date.py
Eric-Sommer: Test is added. I am struggling with a correct implementation though. The test wants to compare the column "vorsorge", so I need to return a named item. This however leads to problems if `_vorsorge_pre_2005()` returns only an `int`. Any ideas, @hmgaudecker , @tobiasraabe ?

hmgaudecker: I do not quite understand. In my world, the column "vorsorge" should already exist in the DataFrame (`tax_data` in the test) and it should get modified in-place by the functions? And all of them should share a common interface.

Eric-Sommer: working now...

gettsim/tests/test_zve.py
```diff
@@ -80,7 +81,7 @@ def test_zve(
     e_st_abzuege_raw_data,
 ):
     year_data = input_data[input_data["year"] == year]
-    df = year_data[IN_COLS].copy()
+    df = year_data[IN_COLS].copy()
```
MaxBlesch: Why trailing whitespaces?

gettsim/taxes/zve.py
```diff
@@ -147,6 +150,7 @@ def kinderfreibetrag(tax_unit, params, kindergeld_params):
 
 def zve_nokfb(tax_unit, params):
```
MaxBlesch: Maybe we should strart giving `nokfb` a german descriptive name. Like `kein_kind_freib

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
-    # calculate x% of relevant employer and employee contributions
+    # calculate x% of relevant employer and employee contributions and private contributions
     # then subtract employer contributions
+    altersvors = calc_altersvors_aufwend(tax_unit, params)
     # also subtract health + care + unemployment insurance contributions
-    altersvors2010 = ~tax_unit["child"] * (
-        (0.6 + 0.02 * (np.minimum(params["year"], 2025) - 2005)) * (12 * rvbeit_vors)
-        - (12 * 0.5 * rvbeit_vors)
+    # 'Basisvorsorge': Health and old-age care contributions are deducted anyway.
+    sonstigevors = 12 * (
```
MaxBlesch: Use lowerscore between words. `sonstige_vors`

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
+    # The test requires a named item. If there is only one value, create a list.
+    return vorsorge
+
+
+def vorsorge_since_2005(tax_unit, params, soz_vers_beitr_params):
+    """ With the 2005 reform, no taxpayer was supposed to be affected negatively.
+        Therefore, one needs to compute amounts
+        (2004 and 2005 regime) and take the higher one.
+    """
+    vors2004 = vorsorge_pre_2005(tax_unit, params, soz_vers_beitr_params)
```
MaxBlesch: Use lowerscore

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
+
+def vorsorge_since_2010(tax_unit, params, soz_vers_beitr_params):
+    """ After a supreme court ruling, the 2005 rule had to be changed in 2010.
+        Therefore, one needs to compute amounts
+        (2004 and 2010 regime) and take the higher one. (§10 (3a) EStG).
+        Sidenote: The 2010 ruling is by construction
+        *always* more or equally beneficial than the 2005 one,
+        so no need for a separate check there.
+    """
+    vors2004 = vorsorge_pre_2005(tax_unit, params, soz_vers_beitr_params)
```
MaxBlesch: Lowerscore

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
+    vors2010 = _vorsorge_since_2010(tax_unit, params, soz_vers_beitr_params)
+
+    return np.maximum(vors2004, vors2010)
+
+
+def calc_altersvors_aufwend(tax_unit, params):
```
MaxBlesch: Something more general: As we use german words for functions, shouldn't we replace calc everywhere?

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
+
+def vorsorge_year_faktor(year):
+    """ in the calculation of *Vorsorgeaufwendungen*,
+    there is year-dependent factor to be calculated.
+
+    year: int
+
+    returns the factor
+    """
+    return 0.6 + 0.02 * (min(year, 2025) - 2005)
```
MaxBlesch: Is there a better solution than hardcoding?

gettsim/taxes/zve.py
```diff
@@ -312,58 +316,151 @@ def calc_gross_e7(tax_unit, params):
...
+        # From what is left from vorsorg_rest, you may deduct 50%.
+        # (up until 50% of 'Grundhöchstbetrag')
+        item_3 = np.minimum(0.5 * (vorsorg_rest - item_2), 0.5 * params["grundbet"])
+    # For the married couple, the same stuff, but with tu totals.
+    if tax_unit["zveranl"].max():
+        for var in ["m_wage", "rvbeit", "gkvbeit"]:
+            tax_unit[f"{var}_tu"] = tax_unit.loc[~tax_unit["child"], "m_wage"].sum()
+        if params["year"] <= 2019:
+            item_1 = 0.5 * np.maximum(
+                2 * params["vorwegab"]
```
MaxBlesch: `vorwegab`doesn't sound very self-explanatory.

gettsim/taxes/zve.py
```diff
@@ -118,7 +121,6 @@ def kinderfreibetrag(tax_unit, params, kindergeld_params):
         kifreib_total = params["kifreib_s_exm"]
 
     diff_kifreib = nokfb_lower - (kifreib_total * child_num_kg)
```
MaxBlesch: Lowerscore in `kifreib` and german word for `child_num_kg`

gettsim/tests/test_zve.py
hmgaudecker: This should be automatically caught by the pre-commit hooks. @Eric-Sommer, can you please check that you have them turned on and they run on your local machine?

gettsim/taxes/zve.py
Eric-Sommer: OK, but I did not introduce `nokfb/kfb` here so might be better suited in a large rename PR.

gettsim/taxes/zve.py
Eric-Sommer: maybe worth a discussion point in #60?

gettsim/taxes/zve.py
Eric-Sommer: I don't know. The alternative would be to specify the values year-wise in the param file.

gettsim/taxes/zve.py
Eric-Sommer: changed to `vorwegabzug`

gettsim/taxes/zve.py
Eric-Sommer: > Lowerscore in `kifreib`

what exactly?

gettsim/taxes/zve.py
hmgaudecker: I would just get rid of `calc_`. The only case where this might matter is where we have assignments like:

```python
altersvors_aufwend = calc_altersvors_aufw(...)
```
but this does not seem worth adding a nonsense string to every function.

Thinking of it, another option would be to only ever import functions as follows: 

```
import gettsim.taxes as tx
import gettsim.transfers as tr
```

and get rid of the intermediate module level by `__init__.py` files in `gettsim.taxes` and `gettsim.transfers` (these files would only include things like `from .zve import altersvors_aufwend`). So the above example would become:

```python
import gettsim.taxes as tx

altersvors_aufwend = tx.altersvors_aufw(...)
```

I kind of like that idea (the Zen... **Namespaces are one honking great idea -- let's do more of those!**)

gettsim/taxes/zve.py
MaxBlesch: Either `kindfreibetrag`, but if we abbreviate a word, then use underscore. So this would be `ki_freib`, which looks a bit ugly. I would just go with the whole word!

gettsim/taxes/zve.py
MaxBlesch: Same as before. If we abbreviate a word, then mark that with an underscore. I think it is not a good idea to glue together abreviations...

hmgaudecker: But `vorwegabzug` is the whole word, no?

MaxBlesch: Oh sorry. Just googled it. Didn't even know this word. German language is really suprising....
Thought this is some kind of abbreviation.

MaxBlesch: But we didn't include that rule in #60 ?! When I program, I always follow: If I abbreviate something, I mark it with an underscore.

gettsim/taxes/zve.py
hmgaudecker: In any case we should stay consistent across variations of the identifier. No need to change the entire old code right now, but I just added this to GEP 1:

  Across variations that include the same identifier, this identifier should not be
  changed, even if it leads to long variable names (e.g., `kinderfreibetrag`,
  `zu_verst_e_ohne_kinderfreibetrag`). This makes searching for that identifier easier
  and less error-prone.

gettsim/taxes/zve.py
hmgaudecker: "Names follow standard Python conventions (`lowercase_with_underscores`).", but explicit is better than implicit, hence 4feda43.

The whole thing is indeed not fully obvious in German where you just create long words where you would have separate ones in English. 

Anyhow: Good catch!

gettsim/taxes/zve.py
Eric-Sommer: While it looks unusual, this is not "harder coded" than any other tax tariff of benefit amount calculation.

hmgaudecker: This seems indeed special enough so as not to move it to the parameters files.

A whole function seems too much to me, though. I would just do:

```python
einführungsfaktor = 0.6 + 0.02 * (min(year, 2025) - 2005)
```

directly in   `calc_altersvors_aufwend`. And we should add to the docstring of that function that it includes the phasing in, explictly cautioning against using it for years prior to 2005 (who knows whether somebody wants to to a policy change of the type 'what if the current system had been around in 1990'?)

Eric-Sommer:
> Thinking of it, maybe it is clearer the opposite way around? `1 - 0.02 * ... ` ?

This way is probably closer to how it is stated in the law. "Der Prozentsatz in Satz 4 erhöht sich in den folgenden Kalenderjahren bis zum Kalenderjahr 2025 um je 2 Prozentpunkte je Kalenderjahr"

hmgaudecker:
Yeah, but how often do you have to compare law and code? If so, it is easy to do that calculation. It also made sense 15-10 years ago to think about it that way. But now we are much closer to 2025.

The typical situation by now is that you want to see that the factor will be 1 eventually and I would find that much easier when seeing 1 - (2025 - min(2025, year)) / 50. 

But no need to make a big fuzz about that.

hmgaudecker:
Sorry, should have looked at 70b3c5a first -- that makes it more than obvious enough! 

Will add a slight formatting change in a sec (remember that these docstrings might be rendered as ReStructuredText and in the current way it would all go into one paragraph).

PR #51, opened 2019-10-29
Add maintenance manual.

### LINKED ISSUE #24: Set up development workflow
Things like required reviews etc. over at OpenSourceEconomics / estimagic etc. look very useful to me. Now that we have ported everything from the old Gitlab repository, we can also be a bit more defensive on merging things. 

@tobiasraabe, @janosg: Do you have any workflow written up somewhere? We should discuss this eventually and I'll set up the Github Project accordingly.

tobiasraabe:

We have no document on how to maintain GETTSIM.  Fixes #24.

The document covers the following sections.

- Versioning
- Branching model
- How to release
- How to backport

### Todo

- [x] Add note on how to restart test builds failed by https error and other stuff.
- [x] Add config for bump2version.


docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+^^^^^
+
+As backports can introduce new regressions the scope is limited to critical bug fixes
+and documentation changes. Performance enhancements and new features are not backported.
+
+Procedure
+^^^^^^^^^
+
+In the following we will consider an example where GETTSIM stable version is 0.2.0.
+Version 0.2.1 is currently developed on the master branch but not released. And a bug
```
hmgaudecker: As I understood the model, if `0.2.0` is released, we have a branch `0.2` which takes backports and we develop 0.3 on master?

If correct, we should take out one level here and backport to 0.2.1, forgetting about 0.1.8 but noting that we may even want to backport to multiple minor versions.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+fix was found which should be backported to minor version 0.1 which latest patch version
```
hmgaudecker: `a critical bug was found. The fix should go both into master and into 0.1.8.`

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+is 8.
+
+1. Create a PR which targets the master branch and includes the bug fix which should be
+   backported.
+2. Add a note to the release notes for version 0.1.9.
+3. Add a label ``backport-to-0.1.9`` to the PR.
+4. Squash merge the PR into master and note down the commit sha.
+5. Create a new PR against branch 0.1.x. Either the branch exists or create the branch
```
hmgaudecker: Branch 0.1.x or 0.1 -- whatever we call it -- will exist.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+we can use bump2version to create a tagged commit. Otherwise, we stick to PRs and create
+the tag via Github. I would go with the latter.**
+
+3. Create a PR to finalize the new release.
+
+   1. Update ``CHANGES.rst`` with all necessary information regarding the new release.
+   2. Use ``bump2version [major|minor|patch]`` to increment all version strings.
+   3. Merge it.
+
+4. Go to the `page for releases <https://github.com/iza-institute-of-labor-economics/
```
hmgaudecker: Create a new branch with the major version number called 0.2 (0.2-maint?). This will be used for maintenance of the 0.2.x release series.

hmgaudecker:
@janosg: What are your thoughts?

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+   gettsim/releases>`_ and draft a new release. The tag and title become ``v0.2.1``.
+   Make sure to target the master branch. A description is not necessary as the most
+   important information lies in ``CHANGES.rst``.
+
+   Release the new version by clicking "Publish release".
```
janosg: What will publish release do? Does it build the package on a server?

docs/how-to-maintain.rst
janosg: I agree that we would develop `0.3` and not `0.2.1` on master, so we would have to backport to `0.2.1`, `0.1.9` and potentially some `0.0.x`. 

I hadn't really thought about backporting before, but have a few general questions:

- In the described workflow we never just delete a release. This means that there will be versions with known bug on our conda channel. Is this what we want? I'm thinking about a bug that leads to wrong results without warnings or exceptions. 
- Maybe we should suggest that people never specify the third digit in their environment files. That way they will have higher chances of getting the latest fixes.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+* ``minor`` is incremented when you add functionality which is backwards compatible.
+* ``patch`` is incremented when you make backwards compatible bug fixes.
+
+You can add a suffix to signal more information on the current version.
+
+* Use ``0.0.1dev0`` to signal that the release is a development version which is
+  currently tested and to be released soon. Increment the last digit for multiple
+  development versions.
+
+.. note::
```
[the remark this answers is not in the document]
tobiasraabe: Done.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+2. There can exist branches for minor versions called, for example, ``0.1.``. The
+   branches contain changes, e.g. backports, which are not released yet. To find out how
+   to release the next patch version, see the following section. If there are no
+   additional changes waiting to be released, the branches should be deleted for
+   clarity.
+
+
+.. _releases:
+
+How To Release
```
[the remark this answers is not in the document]
tobiasraabe: I think in the case you want to release a new patch version, but somebody wants to merge additional functionality which requires a minor update, you actually want to call for a halt. Or what do you have in mind?

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,113 @@
...
+To release a new version of GETTSIM, do the following.
+
+1. To start a new release for version ``0.2.1``, create a new milestone with the version
+   as its name on Github to collect issues and PRs which should be included in the new
+   version.
+
+2. Call a halt on changes to the master branch. Make sure all changes to the master
+   branch since the last version are included in the milestone.
+
+**Here, we have to decide whether we allows pushing to master for releases or not. Then,
```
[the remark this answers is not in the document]
tobiasraabe: It does not make sense to me anymore either. It was related to the fact that the master branch is in my projects protected so that only PR merges are allowed, but it does not make sense anymore.

docs/how-to-maintain.rst
tobiasraabe: I do not agree. I think, depending on what PRs are in the pipeline, we are working on 0.2.1 or 0.3.0. It could be the case that we wanted a preliminary release and did not want to wait for more unrelated patches. If we plan to release 0.2.1, we develop the version on master and merge it to branch 0.2. once the release is finished. Does it make sense?

I agree that depending on the changes and the range of versions we want to support, you have to backport to multiple minor versions.

Regarding the questions:
1) Yes, I think it is normal to have releases with flaws. Depending on the severity of the issues and whether the version is still used, we decide for or against backporting.
2) That is a good advice!

hmgaudecker:
> In the described workflow we never just delete a release. This means that there will be versions with known bug on our conda channel. Is this what we want? 

Yes. If only to reproduce wrong results and show the difference. 

> I'm thinking about a bug that leads to wrong results without warnings or exceptions.

That is the nature of things. We need to make clear that people should always update to the latest minor version number. 

> Maybe we should suggest that people never specify the third digit in their environment files. That way they will have higher chances of getting the latest fixes.

Yes, that is one way to help ensuring this.

hmgaudecker:
Would it be useful to run bump2version from release.py? I.e., `python release.py 0.2.1` would bump the version number and upload the release?

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Branching Model
+---------------
+
+The branching model for GETTSIM is very simple.
+
+1. The master branch is the target of ongoing development. Every PR should be merged
+   into master.
+
+   (The only deviation can happen with respect to backports. See :ref:`backports` for
```
hmgaudecker: This sentence can go. Or we need to say that this may happen additionally. This sounds too much like an either / or thing.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+   more.)
+
+2. There can exist branches for minor versions called, for example, ``0.1.``. The
```
hmgaudecker: Minor versions have their own branches. Bugfixes are included in these branches and patch releases are made from these branches.

Once a minor version is not supported anymore (=will not receive any more bugfixes), the branch should be deleted for clarity.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+
+
+.. _releases:
+
+How To Release
+--------------
+
+To release a new version of GETTSIM, do the following.
+
+1. To start a new release for version ``0.2.1``, create a new milestone with the version
```
hmgaudecker: Change example to `0.2`.

docs/how-to-maintain.rst
hmgaudecker: See the previous comment - conda upload needs to be included now.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+In the following we will consider an example where GETTSIM stable version is 0.2.0.
+Version 0.2.1 is currently developed on the master branch but not released. And a
+critical bug was found which should go both into master and into the minor predecessor
+0.1.8.
+
+1. Create a PR which targets the master branch and includes the bug fix which should be
+   backported.
+2. Add a note to the release notes for version 0.1.9.
+3. Add a label ``backport-to-0.1.9`` to the PR.
+4. Squash merge the PR into master and note down the commit sha.
```
hmgaudecker: Somebody else will have to do that, right?

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+5. Create a new PR against branch ``0.1.``. Call the branch for the PR
+   ``backport-<#pr>`` where #pr is the PR number.
+6. Use ``git cherrypick -x <commit-sha>`` with the aforementioned commit sha to apply
```
hmgaudecker: Do we need to distinguish between 5. and 6.  or can we just do the cherry-pick? 

I would say if somebody approved merging into master and backporting was explicitly mentioned, it is fine to do so directly.

tobiasraabe: Three points.

1. Maintenance branches could be protected against changes without PRs just like master.
2. With PRs backporting is more explicit and documented than a simple cherry-pick merge.
3. Changes definitely pass the test suite whereas maintainers might forget a step while merging via the command line.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+^^^^^^^^^
+
+In the following we will consider an example where GETTSIM stable version is 0.2.0.
+Version 0.2.1 is currently developed on the master branch but not released. And a
+critical bug was found which should go both into master and into the minor predecessor
+0.1.8.
+
+1. Create a PR which targets the master branch and includes the bug fix which should be
+   backported.
+2. Add a note to the release notes for version 0.1.9.
```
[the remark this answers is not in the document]
tobiasraabe: Done!

docs/how-to-maintain.rst
tobiasraabe: Done.

docs/how-to-maintain.rst
tobiasraabe: done

docs/how-to-maintain.rst
tobiasraabe: I thought about your comments again and I agree. I changed the example to cover versions 0.2.0, 0.2.1 and 0.3.0.

docs/how-to-maintain.rst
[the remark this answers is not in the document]
tobiasraabe: I added a note that consensus among developers determines the scope of a release.

docs/how-to-maintain.rst
tobiasraabe: Included release on anaconda and clarified what the release does.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,97 @@
...
+3. Add a label ``backport-to-0.1.9`` to the PR.
```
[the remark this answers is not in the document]
tobiasraabe: Labels? It is a proposal from the pandas team and I thought it might be helpful. But it is not necessary.

tobiasraabe:
What happens if we release a new patch version for some older version which includes changes which could not be merged into master first (e.g. bug fix to functionality which does not exist anymore). Then, do we have to update ``CHANGES.rst`` on master with the release notes of the new patch version in a separate PR?

docs/how-to-maintain.rst
hmgaudecker: My preference would be to leave it as is. It is good to keep this in mind but for the moment, I think we have enough restrictions for a project this size.

hmgaudecker:
> What happens if we release a new patch version for some older version which includes changes which could not be merged into master first (e.g. bug fix to functionality which does not exist anymore). Then, do we have to update `CHANGES.rst` on master with the release notes of the new patch version in a separate PR?

Yes. But I would not expect much of this except we decide to support super-old versions. With enough manpower, we might one day. For now, no chance for that to happen in the first place.

docs/how-to-maintain.rst
hmgaudecker: Except for this issue, this looks good to go to me. @janosg and @mjbloemer, what do you think?

janosg: I agree with Tobias and would protect the maintenance branches. Probably backporting does not happen very often, so the maintainers will not have too much practice with cherry picking. Therefore, it is good to have all tests triggered before the backport can be merged. But there is no need for an additional review on those branches. 

All the rest is really good.

docs/how-to-maintain.rst
```diff
@@ -0,0 +1,109 @@
...
+
+1. To start the release for any new version, e.g., ``0.2.0``, `create a new milestone
+   <https://github.com/iza-institute-of-labor-economics/gettsim/milestones/new>`_ with
+   the version as its name on Github to collect issues and PRs. A consensus among
+   developers determines the scope of the new release.
+
+2. To finalize a release
+
+   1. Update :ref:`changes` with all necessary information regarding the new release.
+   2. Use ``bumpversion [major|minor|patch]`` to increment all version strings.
```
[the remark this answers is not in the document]
tobiasraabe: You can use both. The command is ``bumpversion minor``.

Issue #44, opened 2019-10-17
Automated parameter documentation


mjbloemer:
### Current and desired situation

Currently the param.xls gives a relatively good overview of all the parameters over time. With aspects and issues of xls files raised in #31 and the move to csv files (see https://github.com/iza-institute-of-labor-economics/gettsim/issues/31#issuecomment-540616594) for the parameters there is scope for a different approach on automated parameter documentation.

### Proposed implementation

Add a script that replicates a param.xls file as well as a PDF documentation with nice graphs (example see https://github.com/iza-institute-of-labor-economics/gettsim/issues/31#issuecomment-543072292) or a nice website/html with all the parameters.

### Considered alternatives

- Ignore this, because the csv file as discussed in  is already easy to read

mjbloemer:
The new project https://offenegesetze.de might be a good source for parameters. Also they promise stable links.

mjbloemer:
A few notes:
- We should add a web based parameter documentation/browser with plotly graphs (for parameters change over t).
- ideally this would have a button "found a bug?" with a very low effort functionality that files a bug report
- It could also directly link and highlight the parameter over at offenegesetze e.g. https://offenegesetze.de/veroeffentlichung/bgbl1/2015/41?q=404#page=8 (not so easy in all cases)

Eric-Sommer:
I'd prefer bokeh over plotly, for reasons of experience in the former ;)

ChristianZimpelmann:
Linking this: https://gist.github.com/mjbloemer/834409a80758e5354a8298b67ba52968

(Stata code to create .tex files from parameter .csv-files)

Eric-Sommer:
Do we still need it as PDF, given the bikes dashboard?

ChristianZimpelmann:
Probably, you are right.

PR #54, opened 2019-10-30
Migration of the parameter database from xls to yaml.

### LINKED ISSUE #31: Datatype of params

mjbloemer:
Work in Progress - do not merge!

When merged, his PR migrates the binary param.xls to a param.yaml. Closes #31.

This PR includes the following changes:
- [X] A full conversion of parameters and annotations from the original param.xls
- [X] Remove the old param.xls
- [x] The necessary changes related to database read-in

This PR explicitly does not include the following changes:
- general cleanup of strings
- correction of values

No information is added, removed or modified, except:
- Add group names based on the original annotations. E.g. `[ALG2]` which was part of the xls column "comment_DE" are now stored under the key `group: alg2`. 
- Removed usernames and hostnames from comment/annotation strings
- dates are stored as `yyyy-01-01` irrespective of the actual starting date
- dropped six duplicate definitions: `wgmax*_m_st3`

After merge, I will open separate issues on cleanup and fixing of the param.yaml, these will be:
- fix values
- fix start dates to precise day
- cleanup of name, description and values
- translation task for these strings
- check group names
- sorting by groups
- assign unit key (euro, dm, one, percent, thousand..)
- assign period key (month, year..)

Please comment in this PR on aspects specific to this implementation. When it comes to alternative implementations leave the discussion in #31.

The automated parameter documentation is addressed in #44.


MaxBlesch:
I added nans for not existing values. Right now the code relies for years a policy would actually not exist on nans. A example for this is https://github.com/iza-institute-of-labor-economics/gettsim/blob/1633e1e08f7ac895cb18b1397633e37a5a3e4551/gettsim/taxes/zve.py#L288 .

I think we should in the end have code, which doesn't calls variables for years in which they don't exist. Right now we can't gurantee that the code spits out the right values for every possible year.

@mjbloemer any feedback on the current implementation?

MaxBlesch:
@mjbloemer gave me some private feedback. Do you have anything @hmgaudecker ? Otherwise I'll merge!

hmgaudecker:
This is great and an obvious improvement without changing functionality or pinning down long term decisions.

So do not worry about me. Not like I am the dictator here or anything, though strong opinions might sometimes give that impression... 😉

MaxBlesch:
Ok, thanks :)

PR #57, opened 2019-11-03
implement 07/2019 reform on children transfers

Eric-Sommer:

No Problem. Just some child benefit and kiz changes which became effective 07/2019. This affects both parameters, but also the calculation of the kiz.

in addition some minor edits to `param.yaml`, mostly notes on the actual law, but also one correction on the kiz withdrawal rate from 2005 to 2008.

# To Dos
- [x] update `param.yaml`
- [x] update `kiz` calculation rule: no maximum income threshold anymore and other treatment of child and parental income.


gettsim/data/param.yaml
```diff
@@ -8782,7 +8797,7 @@ ust_high:
     1993-01-01:
       value: .15
       note: .
-    1999-01-01:
+    1999-04-01:
```
mjbloemer:
```suggestion
    1998-04-01:
```

There was indeed a big amendment of the UStG in 1999, see https://offenegesetze.de/veroeffentlichung/bgbl1/1999/30?q=umsatzsatzsteuer%20sechzehn#page=2 with the new law in force from 1999-04-01.

But the change from .15 to .16 already happened a year before on 1998-04-01, see https://offenegesetze.de/veroeffentlichung/bgbl1/1997/86?q=Umsatzsteuer#page=30 or  https://www.welt.de/print-welt/article645301/Mehrwertsteuer-steigt-zum-1-April-auf-16-Prozent.html.

gettsim/data/param.yaml
```diff
@@ -8782,7 +8797,7 @@ ust_high:
...
       value: .16
       note:
```
mjbloemer:
```suggestion
       note: Artikel 5 G v 19.12.1997 BGBl. I 3122.
```

gettsim/data/param.yaml
Eric-Sommer: right that was a typo. 

> I would appreciate if the changes of the StaFamG are addressed in a separate PR as they are relatively complex.

I'm actually doing it the other way round now and do only StaFamG stuff here :)

hmgaudecker: > I'm actually doing it the other way round now and do only StaFamG stuff here :)

Pls update the title then.

Eric-Sommer:
This is really weird. If I merge the master locally, I don't see any conflicts.

gettsim/benefits/kiz.py
```diff
@@ -71,28 +71,60 @@ def kiz(household, tb):
...
-    household["kiz_temp"] = household["kiz"].max()
-    # Transfer some variables for eligibility check
-    # kiz["ar_base_alg2_ek"] = household["ar_base_alg2_ek"]
-    # kiz["n_pens"] = household["pensioner"].sum()
+
+    return household
+
+
+def calc_kiz_amount_2020(household, tb):
+    """ Kinderzuschlag since 07/2019
```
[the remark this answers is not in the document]
Eric-Sommer: The latter one, as indicated in param. I could rename the function of course.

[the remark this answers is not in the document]
mjbloemer: It might be helpful then to rename the function then or alternatively add the date switch inside the function.

gettsim/data/param.yaml
```diff
@@ -8791,9 +8821,6 @@ ust_high:
     1993-01-01:
       value: .15
       note: .
-    1999-01-01:
```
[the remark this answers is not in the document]
Eric-Sommer: Sorry, this should of course not be taken out.

gettsim/data/param.yaml
```diff
@@ -871,6 +874,21 @@ a2kiz_minek_sin:
...
+  name:
+    de: Rate, mit der relevantes Einkommen auf den Kinderzuschlag angerechnet wird.
+    en: transfer withdrawal of additional child benefit
+  description:
+    de: § 6a (3) BKGG
+    en:
+  values:
+    2005-01-01:
+      value: 0.5
+      note:
```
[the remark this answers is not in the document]
Eric-Sommer: thanks for the remark, I just found out it was 0.7 initially. Since 07/2019, income is deducted continuously.

Eric-Sommer:
you are right. infact, parents income is still deducted with 50%, although without rounding now. will be corrected

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+      note: G. v. 24.09.2008 BGBl. I S. 1854
+a2kiz_withdrawal_rate_child:
+  group: kiz
+  name:
+    de: Rate, mit der Einkommen des Kindes auf den Kinderzuschlag angerechnet wird. Parameter ist vor 07/2019 unwirksam, da er in der Funktion nicht auftaucht.
+    en:
+  values:
+    2005-01-01:
+      value: 1
+      note: Art. 46 v. 24.12.2003, BGBl. I. S. 2954
```
mjbloemer:
```suggestion
      note: Art. 46 G. v. 24.12.2003 BGBl. I. S. 2954.
```
in #69 I decided not to give the exact "Artikel" of the law as it would have been a lot more effort. I think it is nice to add as a service. We could do that consistently everywhere (more effort), not at all (less service for those who double check the law) or just if we want to (less consistency).

Drop the comma behind the date as nothing will show up in the search over at offenegesetze.de if there is a comma (yes it is only a few keystrokes more to remove that, but when you look up a lot of laws it starts to matter).

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+  name:
+    de: Rate, mit der relevantes Einkommen auf den Kinderzuschlag angerechnet wird. Bis 07/2019 wurde das Einkommen nur in 10€-Schritten berücksichtigt.
+    en: transfer withdrawal of additional child benefit
+  description:
+    de: § 6a (6) BKGG, bis 01.07.2019 § 6a (4) BKGG
+    en:
+  values:
+    2005-01-01:
+      value: 0.7
+      note: Art. 46 v. 24.12.2003, BGBl. I. S. 2954
```
mjbloemer:
```suggestion
      note: Art. 46 G. v. 24.12.2003 BGBl. I. S. 2954.
```

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+    2008-10-01:
+      value: 0.5
+      note: G. v. 24.09.2008 BGBl. I S. 1854
```
mjbloemer:
```suggestion
      note: G. v. 24.09.2008 BGBl. I S. 1854.
```

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
       value: 600
-      note:
+      note: G. v. 24.09.2008 BGBl. I S. 1854
+a2kiz_withdrawal_rate:
+  group: kiz
+  name:
+    de: Rate, mit der relevantes Einkommen auf den Kinderzuschlag angerechnet wird. Bis 07/2019 wurde das Einkommen nur in 10€-Schritten berücksichtigt.
+    en: transfer withdrawal of additional child benefit
+  description:
+    de: § 6a (6) BKGG, bis 01.07.2019 § 6a (4) BKGG
```
mjbloemer:
```suggestion
    de: Rate, mit der der Gesamtkinderzuschlag nach Erreichen der Bemessungsgrenze durch Einkommen der Eltern gemindert wird. Bis 07/2019 wurde das Einkommen nur in 10€-Schritten berücksichtigt. § 6a (6) BKGG, bis 01.07.2019 § 6a (4) BKGG.
```

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
       note:
-    2009-01-01:
+    2008-10-01:
       value: 600
-      note:
+      note: G. v. 24.09.2008 BGBl. I S. 1854
+a2kiz_withdrawal_rate:
+  group: kiz
+  name:
+    de: Rate, mit der relevantes Einkommen auf den Kinderzuschlag angerechnet wird. Bis 07/2019 wurde das Einkommen nur in 10€-Schritten berücksichtigt.
```
mjbloemer:
```suggestion
    de: Minderungsrate durch Einkommen der Eltern über BMG
```
moved the long description to the description de key and added a suggestion for a short name here.

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+    en: transfer withdrawal of additional child benefit
```
mjbloemer:
```suggestion
    en: transfer withdrawal rate, parent's income above BMG
```

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+    2005-01-01:
+      value: 0.7
+      note: Art. 46 v. 24.12.2003, BGBl. I. S. 2954
+    2008-10-01:
+      value: 0.5
+      note: G. v. 24.09.2008 BGBl. I S. 1854
+a2kiz_withdrawal_rate_child:
+  group: kiz
+  name:
+    de: Rate, mit der Einkommen des Kindes auf den Kinderzuschlag angerechnet wird. Parameter ist vor 07/2019 unwirksam, da er in der Funktion nicht auftaucht.
```
mjbloemer:
```suggestion
    de: Anrechnungsrate durch Kindereinkommen
```
could you add the long description to the description key?
```
Rate, mit der Einkommen des Kindes auf den maximal möglichen Kinderzuschlag angerechnet wird. Parameter ist vor 07/2019 unwirksam, da er in der Funktion nicht auftaucht.
```

gettsim/data/param.yaml
```diff
@@ -868,9 +871,36 @@ a2kiz_minek_sin:
...
+    en:
```
mjbloemer:
```suggestion
    en: rate at that child income is deducted
```
or "considered"?

gettsim/data/param.yaml
Eric-Sommer: OK with the comma. We need these guidelines on how to update / add params somewhere. Usually, I don't mention the "Art.". The reason is that most interesting changes happen in Art. 1. Here, this was not the case.

hmgaudecker:
Looks good, thanks! A couple of points: 
* Please always merge with master before requesting a review. In this case, actually, it will be useful to wait for #111. I will take a look at that in a sec.
* Re @mjbloemer's comment on the previous errors in the KIZ calculation: Can you please move those into a separate PR? Really has nothing to do with this one and seems orthogonal enough to merit a separate commit for later checking. Should not be too hard to do with git cherry-pick. Again, please work off current master / #111 for that. 
Thanks !

hmgaudecker:
While reviewing that, I noticed that we are re-doing the calculation of most ALG-2 stuff in KIZ. We should rely on what we did before. If that means adding a temporary variable to the household dataframe, so be it.

Eric-Sommer:
>     * Re @mjbloemer's comment on the previous errors in the KIZ calculation: Can you please move those into a separate PR? 

Done now

gettsim/data/kinderzuschlag.yaml
```diff
@@ -54,3 +57,14 @@ a2kiz_withdrawal_rate:
     2008-10-01:
       value: 0.5
       note: G. v. 24.09.2008 BGBl. I S. 1854.
+a2kiz_withdrawal_rate_child:
+  name:
+    de: Anrechnungsrate durch Kindereinkommen
+    en: rate at that child income is deducted
```
mjbloemer: I think the description keys are missing here.

suggestion:
> Rate, mit der individuelles zu berücksichtigendes Kindereinkommen beim Kinderzuschlag angerechnet wird. § 6a (3) BKGG. Bis 30.06.2019 wird Kindereinkommen zu 100% angerechnet.

Eric-Sommer:
> I think a self-explanatory name would make sense here. This is stuff that happens many times (I guess?) and it might actually be worth keeping it around. Something like `kiz_no_xchecks` ? Ideas welcome.
> 
> At that point, why do we have both `kiz` and `kiz_temp` and why the max operator? Does not seem intuitive.

maybe `kiz_prelim`?

hmgaudecker:
> maybe `kiz_prelim`?

Not quite, preliminary could be anything again. It should be clear that this is before cross-checks / interactions with other programmes.

Eric-Sommer:
> > maybe `kiz_prelim`?
> 
> Not quite, preliminary could be anything again. It should be clear that this is before cross-checks / interactions with other programmes.

OK, seems to be not that easy right now. Probably also better suited in a separate PR.

Eric-Sommer:
> At that point, why do we have both `kiz` and `kiz_temp` and why the max operator? Does not seem intuitive.

renaming makes definitely sense. the `.max()` operator secures that every row is filled. But there's also a more intuitive way to achieve this.

hmgaudecker:
But do we **want** every row filled? I really need to work on the data structures GEP...

Eric-Sommer:
> But do we **want** every row filled? I really need to work on the data structures GEP...

It's way more robust in my opinion than to assign it to the household head only. And it's the default for all benefits paid on the household level.

hmgaudecker:
Good to know, thanks!

Issue #58, opened 2019-11-04
Remove some cruft and set up coverage statistics


hmgaudecker:
A look at the coverage statistics reveals that:

* Not surprisingly, `release.py` and `setup.py` have zero coverage: https://codecov.io/gh/iza-institute-of-labor-economics/gettsim. @tobiasraabe, is it possible to tell codecove.io to focus on the gettsim subfolder? Is there anything outside of that folder that we would want to know about?
* In the gettsim folder, only auxiliary.py has some missing stuff: https://codecov.io/gh/iza-institute-of-labor-economics/gettsim/src/master/gettsim/auxiliary.py. @MaxBlesch @Eric-Sommer @mjbloemer:
  * Can we just delete the uprate function?
  * Can we delete the "adult_married" *unit* ?

MaxBlesch:
If there is not a real reason to keep the uprate functio, I will open a PR and submit a suggestion to remove auxiliary.py.

Eric-Sommer:
For now, we can put the uprating aside. In practice, it is pretty important if data year and policy year differ. But this can be easily re-implemented.

`adult_married` is now defined within the functions where it is needed and can also be discarded.

mjbloemer:
I have the impression that every uprating of monetary variables of the input data set should be done on the data preparation side.

MaxBlesch:
Removed auxiliary.py with #59

MaxBlesch:
This can be closed?

hmgaudecker:
Closed via #59 and #64

PR #59, opened 2019-11-04
Remove auxiliary.py

MaxBlesch:

Right now we kept some functions in auxiliary.py, which are either not need in our new design or a one-liner. Therefore I will delete this module. Adressing partly #58 .

### Todo

- [x] Implement household and tax_unit aggregation directly in the functions.
- [x] Delte get_params and uprate.


hmgaudecker:
Good idea to delete the entire module. Rather have five modules with names that tell us something than one container module with five unrelated functions.

gettsim/benefits/alg2.py
```diff
@@ -33,7 +31,7 @@ def alg2(household, tb):
     )
     # Aggregate on HH
     for var in ["ar_alg2_ek", "alg2_grossek", "uhv"]:
-        household[var + "_hh"] = aggr(household, var, "all_hh")
+        household[var + "_hh"] = household[var].sum()
```
tobiasraabe: Go for ``f"{var}_hh"`` which is more readable.

gettsim/benefits/kiz.py
```diff
@@ -17,7 +15,7 @@ def kiz(household, tb):
         This is done by some fixed share which is updated on annual basis
         ('jährlicher Existenzminimumsbericht')
     """
-    household["uhv_tu"] = aggr(household, "uhv", "all_tu")
+    household["uhv_tu"] = household.groupby(["tu_id"])["uhv"].transform("sum")
```
tobiasraabe: no need for squared brackets in groupby.

gettsim/tests/auxiliary_test_tax.py
```diff
@@ -28,20 +28,19 @@ def load_test_data(year, file_name, columns, *pd_args, **pd_kwargs):
...
         for col in bool_cols:
             if col in df.columns:
                 df[col] = df[col].astype(bool)
-
-    df = df.loc[df["year"].eq(year), columns].copy()
-
-    return df
+    else:
+        raise ValueError("File does not exist.")
+    return df[df["year"] == year][columns]
```
tobiasraabe: Do not assign on a return statement. ``df.loc[df.year.eq(year), columns].copy()`` avoids chained selection and probably prevents warnings like "do not alter view".

MaxBlesch: Assigned before returning

gettsim/benefits/kiz.py
MaxBlesch: Fixed

gettsim/benefits/alg2.py
MaxBlesch: Converted to f-string

Issue #30, opened 2019-10-02
API for children


Eric-Sommer:
There are a couple of issues which would make the setup of a custom data set as input to `tax_transfer()` cumbersome at the moment. Let's collect them here.

- [ ] There are many variables counting the number of children in different age brackets: `[child_num, child3_6_num, child7_13_num, ...]`. This is due to the scattered system of benefits in Germany. To get rid of this, each function could compute the necessary values itself, based on age of each household member. This would also reduce the amount of data carried from one function to the next.

hmgaudecker:
Thanks. Let us please go for one issue per topic rather collecting many different things in one.

hmgaudecker:
Content-wise: Sounds good. We also need to add links between parents and children (relevant for larger households,  re-married people, kids > 16,  etc.)

E.g., not at all obvious if you have members age 55, 52, 22 and the youngest being a student. Probably parents receiving child benefits, but better make the link explicit.

I would think that all taxes / transfers are going to the parents (any exceptions? BAFöG?). So we would loop over the connected kids and calculate tax breaks, child benefits, Unterhaltsvorschuss, who knows what.

Eric-Sommer:
> Content-wise: Sounds good. We also need to add links between parents and children (relevant for larger households, re-married people, kids > 16, etc.)
> E.g., not at all obvious if you have members age 55, 52, 22 and the youngest being a student. Probably parents receiving child benefits, but better make the link explicit.

This should actually be done in the data preparation already. If the 22-year old son is a student, he is in the same tax unit. If he is employed, he is a tax unit on his own. 

> I would think that all taxes / transfers are going to the parents (any exceptions? BAFöG?). 

It is usually easier to assign benefits (e.g. kindergeld) to the children and compare it with their income / hours worked (there used to be thresholds) first. Child alimony is explicitly an income of the child which e.g. matters for the means test of the Kinderzuschlag. We don't care on which bank account the benefit is transferred to.

> So we would loop over the connected kids and calculate tax breaks, child benefits, Unterhaltsvorschuss, who knows what.

I rather thought along the lines of inserting the household / tax unit as a whole and then count the children below 7 years (e.g. for ALGII).

hmgaudecker:
> > Content-wise: Sounds good. We also need to add links between parents and children (relevant for larger households, re-married people, kids > 16, etc.)
[...]

PR #60, opened 2019-11-04
GEP 1 - Naming Conventions

Eric-Sommer:

Agree on a convention on how to name variables.


docs/geps/gep-01.rst
```diff
@@ -0,0 +1,148 @@
...
+| b            | benefits (i.e. non-market income) | balim, boldagepen, bsocass                               |
++--------------+-----------------------------------+----------------------------------------------------------+
+| h            | housing                           | hsize, hrent, hheat                                      |
++--------------+-----------------------------------+----------------------------------------------------------+
+| x            | expenses                          | xchildcare                                               |
++--------------+-----------------------------------+----------------------------------------------------------+
+| a            | assets/wealth                     |  afinancial, aproperty                                   |
++--------------+-----------------------------------+----------------------------------------------------------+
+
+The table above demonstrates the virtue of the first letter identification. `hsize` can easily be misunderstood as the household size, which is a demographic variable. The first letter `h` avoids this misinterpretation.
```
hmgaudecker: This.

In my world, the variable would be called `house_square_meters` or `house_n_rooms`, whatever is relevant.

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,148 @@
...
+- in principal, use lower case letters only. Exceptions are specified below.
+- longer names might be interrupted with underscores, e.g. `alg2_anzurechnendes_ek`
+- use of booelan type for dummy variables rather than integer. Boolean variables can be negated easily (mind the difference between `~` and `not`) and make the code easier to read.
+
+Input Variables
+-----------------
+
+First letter specifies type of variable
+----------------------------------------
+In the spirit of the Euromod_ framework, I suggest that the type of a variable is indicated by its first letter
```
hmgaudecker: Thanks for all your work on this. 

Personally, I am not a big fan of the EUROMOD approach, which I would summarise as "Make the variable names incomprehensibly short, but easy to look up." I can see how this will grow on you if you work with it for sufficiently long time, but our goal is precisely to draw in people who do not.

My suggestion would be closer to what @janosg describes [here](https://estimagic.readthedocs.io/en/latest/contributing/styleguide.html). We are a Python package, after all and should try to be pythonic. Though we have [this](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds), of course ...

docs/geps/gep-01.rst
Eric-Sommer: Or, `h_size_sqm` ;)

docs/geps/gep-01.rst
Eric-Sommer: Very valid points. The reason I started with (not copied) the EUROMOD approach is that, while having worked with a few survey datasets, the Euromod approach is the one where, at one point, you stop consulting the codebook every time. This never happened with SOEP to me. I'm not at all against longer, self-speaking names, while I see a lot of use in a categorical first letter and the last letter indicating the time reference.

MaxBlesch:
I strongly vote for two. If somebody looks at our code, he/she would need to have a table with all the definitions beside. Self explanatory code is the best way to go for me!

Eric-Sommer:
I am OK with (2) but like to insist on the time reference letter. I don't know how often I subtracted annual taxes from monthly income.

hmgaudecker:
That is a valid point, but only partly related to variable names. Ideally, I would like to keep it out and only ever work at either monthly or yearly level. Just spent lots of time this morning explaining my students that a big portion of good programming is to minimise the amount of state variables you have to keep in mind! 

To be a bit more concrete, I think two things would be helpful for me

- A list of relevant column names. Maybe those that that ever leave a single function in the current code? Would that be hard to do?
- Related is the question of what we do with household / tax-unit level functions and variables. I think that came up somewhere the other day but I cannot find it right now. That is, to what extent can we work with aggregates (e.g., total income at the tax unit level is enough to calculate anything on taxes), when do we actually need individual-level data? @mjbloemer @Eric-Sommer: Do you have a feeling for that?

Eric-Sommer:
- We could agree on working always with monthly data. The only exception would need to be weekly hours of work and taxable incomes.
- A list of necessary columns could be set up. As a start, take those from the `test_tax_transfer` input. Some of those will be replaced when #30 is tackled. Instead of counting the number of children, you'd rather have parents and kids pointers to the `pid`s.
- I assume we are talking about model output. The decision on where to assign a variable needs to be driven by common sense. For example, children don't pay taxes. In a couple, you could assign the `tu` sum of taxes to the head. But the "non-head" adult is also a taxpayer. The decision on where to assign monetary variables should be driven by 
  - what bears the most straightforward interpretation for a user when looking at an individual household?
  - what makes aggregation on the dataset as a whole (and with which sample weights, i.e. personal or individual) the most straightforward?
  
  Based on my suggestion, summing up those personal weights with positive tax payments yields the number of taxpayers. Summing up household weights over those with ALG2 benefits yields the number of dependent households. Actually, this could also be discussed in a separate GEP ;)

hmgaudecker:
>     We could agree on working always with monthly data. The only exception would need to be weekly hours of work and taxable incomes.

Sounds good. I would like to see the empirics, though (=the column names + currently used time period)

>     A list of necessary columns could be set up. As a start, take those from the `test_tax_transfer` input. Some of those will be replaced when #30 is tackled. Instead of counting the number of children, you'd rather have parents and kids pointers to the `pid`s.

Fine. @MaxBlesch, could you try to pull those together?

>     I assume we are talking about model output. The decision on where to assign a variable needs to be driven by common sense. For example, children don't pay taxes. In a couple, you could assign the `tu` sum of taxes to the head. But the "non-head" adult is also a taxpayer. The decision on where to assign monetary variables should be driven by

I think the concept of a "household head" is on its way out :-)
  
>       what bears the most straightforward interpretation for a user when looking at an individual household?
>       what makes aggregation on the dataset as a whole (and with which sample weights, i.e. personal or individual) the most straightforward?

Yes, all true. What I was trying to get at was actually how often we want things that never vary across individuals within a household (you could just work with one row per household then) and how often we need things that do vary.

Eric-Sommer:
> What I was trying to get at was actually how often we want things that never vary across individuals within a household (you could just work with one row per household then) and how often we need things that do vary.

What about two separate outputs? One on the personal level in the spirit of this draft, one with one row per household and household sums?

mjbloemer:
I think this is an important question. And it is a bit more than only thinking on how to name the inputs and outputs. 

It also has a lot to do with the readability of the code. While some parts of the model are fix and rarely looked at, the implementation of tax and transfer rules are frequently looked at by users. Users that want to apply the model and simulate counter-factual policies will work with the code and want to use intuitive variable names there.

I see two problems with long variable names:

- frequent line breakes make the code hard to read, especially since we wrap variables in `household[""]` and break lines at x chars.

- users will add suffixes like `_rs2019ls` or prefixes like `diff_sq_` or `eq_oecd_` on their side after applying GETTSIM, so leave them some space. Is the Stata macro name length also 32? (users might want to create macros based on variable names like `mean_...`).

Things like `household["hh_size"]` will not happen, right? (the household attribution is redundant)

MaxBlesch:
> > A list of necessary columns could be set up. As a start, take those from the `test_tax_transfer` input. Some of those will be replaced when #30 is tackled. Instead of counting the number of children, you'd rather have parents and kids pointers to the `pid`s.
> 
> Fine. @MaxBlesch, could you try to pull those together?

The list in documentation is right the one needed. There are variables like hhsize, adult_num_tu, child_num_tu, which we can really create internally and therefore reduce this list. In `test_tax_transfer` is no list right now!

Eric-Sommer:
> The list in documentation is the one needed. There are variables like hhsize, adult_num_tu, child_num_tu, which we can really create internally and therefore reduce this list. In `test_tax_transfer` is no list right now!

Could you amend #30 by listing all those variables which need to be internally calculated (and rename the issue)?

hmgaudecker:
Cool, so no internal variables anywhere that we might also need to worry about?

hmgaudecker:
> I think this is an important question. And it is a bit more than only thinking on how to name the inputs and outputs.

Oh yes, that part is trivial to switch out. As I said, I would even try allowing DE/EN and potentially things like EUROMOD/CNEF as options

> It also has a lot to do with the readability of the code. While some parts of the model are fix and rarely looked at, the implementation of tax and transfer rules are frequently looked at by users. Users that want to apply the model and simulate counter-factual policies will work with the code and want to use intuitive variable names there.

Precisely.
 
> I see two problems with long variable names:
> 
> frequent line breaks make the code hard to read, especially since we wrap variables in `household[""]` and break lines at x chars.

I think we will eventually be able to switch out much of that extra wrapping inside the core functions.

In general, I do not mean to imply that "longer is better". `hh` for `household` is clear enough in my world, at the same time I would prefer `zu_verst_eink` over `zve` over `zu_versteuerndes_einkommen`.

> users will add suffixes like `_rs2019ls` or prefixes like `diff_sq_` or `eq_oecd_` on their side after applying GETTSIM, so leave them some space. 

Fair enough. We should set soft and hard limits well below 32 for user-facing variables and try to adhere to these internally, too. Maybe 15 (soft) and 20 (hard)?

> Is the Stata macro name length also 32? (users might want to create macros based on variable names like `mean_...`).

I do not think so. More like 256, IIRC. Though there were weird differences depending on whether you define them using `=` or not...

> Things like `household["hh_size"]` will not happen, right? (the household attribution is redundant)

Probably not, see above.

mjbloemer:
> I do not think so. More like 256, IIRC. Though there were weird differences depending on whether you define them using `=` or not...

````
. local a2345678901234567890123456789012 = 1
_a2345678901234567890123456789012 invalid name
r(198);

. local a2345678901234567890123456789012 1
_a2345678901234567890123456789012 invalid name
r(198);
````
it seems to be 31 characters for locals in Stata 14.2.

In internal functions and graph routines Stata also sometimes add a few `_` here and there (to variables and maybe locals) which is also true for a lot of user written commands (I think povdeco is a candidate).

**Comment by hmgaudecker** [comment]
I thought the relevant measure would be what they can hold, not what they can be called?

**Comment by mjbloemer** [comment]
> not what they can be called?

yes.

A user gets `inc_net` from GETTSIM, 
- calls it `inc_net_rs2019ls` afterwards, 
- then `eq_inc_net_rs2019ls`,
- later `diff_sq_eq_inc_net_rs2019ls`,
- and later creates a local `mean_diffsq_eq_inc_net_rs2019ls` or calls a Stata function that creates this local or uses this macro internally.

And these are even short pre/suffixes.

(Not that I like that, but this is what people do in Stata).

**Comment by hmgaudecker** [comment]
I agree we are not in the business of educating Stata users. And as I said, I would be fine with a reasonable hard limit on user-facing variables.

But if people really want to do what you describe above, I am fine with asking them to rename their variables to whatever they can handle rather than having our optimal choice be constrained by the possibility of such patterns.

**Comment by Eric-Sommer** [comment]
for the sake of completeness:
From the Stata Help (`help limit`):

    length of a variable name          32
    length of ado-command name         32
    length of a global macro name      32
    length of a local macro name       31

**Comment by tobiasraabe** [comment]
Sorry @Eric-Sommer for also updating this PR. But, now look at the CI checks and see the new rtd feature which checks whether the docs can be built. Clicking on the check redirects you to the actual build for inspection.

Pinging also @mjbloemer, @MaxBlesch, @hmgaudecker  so you know it is there.

File: docs/geps/gep-01.rst [hunk 3]
```diff
@@ -0,0 +1,139 @@
...
+parameters, Python identifiers (functions, variables), etc. should be called. In a
+nutshell and without explanations, these conventions are:
+
+* Names follow standard Python conventions (``lowercase_with_underscores``)
+* Names should be long enough to be readable, but
+
+  - For column names in the user-facing API, there is a hard limit of 15 characters
+  - For others, there is a soft limit of 15 and a hard limit of 20 characters
+
+* The language should generally be German
```
Eric-Sommer: I agree that gettsim-internal (tax-transfer-related) variables should be named in German. But should this also hold for input and output data? Do we want `alter` instead of `age` and `familienstand` instead of `marital_status`?

hmgaudecker: For the reference implementation, I think so, yes. There is plenty of German already and some of it would be hard to translate. Also, it would be useful to have stuff like `gkv` line up with `gkv_beitrag` rather than trying to figure out from the English name that something is related to the same concept.

It is trivial to provide an English translation, so this mostly affects the internal workings of the code.

ChristianZimpelmann: I also think it would feel strange to work with only German variables and functions. Additionally, it would make it almost impossible for non-German speakers to participate in the project.

We have this nice differentiation in three basic building blocks of the code:
>1.     The input and output data, including  any values stored intermediately. The scope of this GEP are the column identifiers. The index and the variable contents are defined elsewhere could link if we have something).
> 2.     The parameters of the tax transfer system, as detailed in the YAML files.
 >3.     Python identifiers, that is, variables and functions.

So I am more in favor of having 1. and 3. in English and only 2. in German. 

> There is plenty of German already and some of it would be hard to translate.
That's a good point. What do you those guys, who have worked more with the code think? In any case we should decide for one language and change all variables, functions (and comments) accordingly.

hmgaudecker: > I also think it would feel very strange to work with only German variables and functions. @hmgaudecker, would you then have also docstrings and comments in German?

No, only identifiers. 

> Additionally, it would make it almost impossible for non-German speakers to participate in the project.
> 
> We have this nice differentiation in three basic building blocks of the code:
> 
> > 1. The input and output data, including  any values stored intermediately. The scope of this GEP are the column identifiers. The index and the variable contents are defined elsewhere could link if we have something).

I would say no with respect to the scope of this GEP. I do not think it makes sense to separate these three things at the level of pinning down naming conventions.

> > 2. The parameters of the tax transfer system, as detailed in the YAML files.

> > 3. Python identifiers, that is, variables and functions.

> So I am more in favor of having 1. and 3. in English and only 2. in German.
> 
> > There is plenty of German already and some of it would be hard to translate.
> 
> That's a good point. In any case we should decide for one language and change all variables, functions (and comments) accordingly. What do the guys, who have worked more with the code, think is easier?

I was not trying to suggest that and I think it would be a bad idea indeed. Sorry if anything gave that impression.

> > Also, it would be useful to have stuff like gkv line up with gkv_beitrag rather than trying to figure out from the English name that something is related to the same concept.
> 
> I think it would be an option to have all names that are directly related to an institutional feature (parameter in tax-transfer-system) in German. If we adopt this proposal, we could have a function name like `calculate_elterngeld()`. I prefer this over `berechne_elterngeld()`

I am totally in favour of that. Maybe something like "verbs should always be in English"? 

We will continue to have a mix of languages in 3. (Python identifiers) for sure. 

But I personally would prefer to get rid of the language mixing in the column identifiers and tax / transfer parameters. There are too many grey areas what constitutes an institutional feature and what does not. Some components of income / deductions might have reasonable translations, but "werbungskosten" or "zu_versteuerndes_einkommen" ? I would then prefer having only German names in the formula. Same for other stuff. "Eheähnliche Gemeinschaft" ?

ChristianZimpelmann: Ok, I misunderstood something. We agree on 2. (German) and 3. (English - except referring directly to something institutional). The open question is block 1. 

I tend to agree with your proposed solution then (German).

hmgaudecker: Upon re-reading, I see where the confusion is coming from. Sorry about that! It is a draft, but still...

hmgaudecker: Upon re-reading, I see where the confusion is coming from. Sorry about that! It is a draft, but still...

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,139 @@
...
+
+The length of a variable name should be proportional to its scope. In a list
+comprehension or short loop, it might be an acceptable name for the running variable,
+but variables that are used at many different places should have descriptive names.
+
+The name of variables should reflect the content or meaning of the variable and not the
+type. As for column names and parameters, in some casees it might be useful to append an
+underscore plus either of {``y``, ``m``, ``w``, ``d``} to indicate the timeframe.
+
+Function names should contain a verb. Moreover, the length of a function name is
```
Eric-Sommer: Which verbs will we use other than `calculate`?

hmgaudecker: Check? Eligibility for social assistance, for example. I have not tried to think of a complete list. Should we try?

docs/geps/gep-01.rst
Eric-Sommer: > I do not think it makes sense to separate these three things at the level of pinning down naming conventions.

I disagree. Once the household API is set up, we will require relatively few input columns (30-40) which are generic enough to abstract from German tax-transfer specifics. `marital_status` might be one exception but we'd need to provide some additional guidelines (data type, scope, ...) anyway. As @ChristianZimpelmann, demanding `alter`, `weiblich` and `am_status` would raise additional burden for non-German users and possibly for German ones as well.

hmgaudecker: Again, I was not precise enough. I tried to say "this should be in one GEP rather than in three", not that rules for all three broad concepts need to be the same.

And again: the question is not about the API (trivial to provide in various formats), but about the internal code. And there we will end up with all those cases (health insurance, ...) where there is a 1:1 link between the parameters and the data columns. 

As for `weiblich`: To be future-proof we should go for `geschlecht` anyhow to allow for more than two values ...

In the end of the day, I think it will be hard or unnatural to avoid German altogether there and I do not see a good reason to do so in the forest place. I tend to prefer going all the way in terms of German column names; but if we decide in the end of the day that we should leave stuff like `age` in English, this would be fine, too, of course.

docs/geps/gep-01.rst
Eric-Sommer: after skimming over the code I found:

- `calculate`
- `check`
- `update`
- `set`
- `apply`
- `get`

never mind ;)

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,253 @@
...
+* The parameters are stored by group (e.g., ``arbeitsl_geld``, ``kinderzuschlag``).
+  These groups or their abbreviations should not re-appear in the name.
+* Use Python containers like tuples and namedtuples (the mutable versions of these,
+  i.e., lists and dictionaries, might be better known) where relevant.
+
+E.g., instead of ``kgeld1: 204``, ``kgeld2: 204``, ``kgeld3: 210``, and ``kgeld4: 235`` in the file ``kindergeld.yaml``, use
+
+    .. code-block:: yaml
+
+        beträge:
```
mjbloemer: Could we avoid umlauts? Difficult to enter on most non-German keyboards.

hmgaudecker: I have a pretty strong opinion here and it is pro-Umlaut. 

One of the founding principles of Python is "Code is read much more often than it is written." This is particularly true here. We would deviate from much better readability (and uniqueness; in English it is never that obvious whether ö will become oe or o) for sake of avoiding a minor annoyance for very few people in very few circumstances -- this code is just not written from scratch all that often.

In all likelihood, the vast majority of GETTSIM developers will have access to a German keyboard. And anybody else can use the replacement and ask for somebody else to change it (as a case in point, see [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/50#discussion_r364433143).

mjbloemer: Hmmm... `bezugsgröße_ost`? I see no big advantage in readability. I guess I am used to `ae` replacements.

So we specifically allow umlauts (and sharp s?) but require all lowercase words and a lower bar `_` in place for spaces?

And I think being open for non German speakers and users is a very valid point (sonst können wir das hier alles auch auf Deutsch machen. Dann wären einige Diskussionen um das deutsche Steuer- und Transfersystem weniger umständlich). I think writing code or addressing/discussing functions parameter names in mails or github issues for instance is part of that. 

When browsing through OpenFisca I have to look up French words pretty often (they don't use accents btw, `._impot`, `._bareme`,...). I'm not sure what is the right way but it could be good to use English language for clear cases such as vat or income tax.

Plus aren't there some issues with Umlaut Handling in other programming languages? Let's say someone wants to analyze the param database in a different language or re-use/share the param database with the Stata model?

hmgaudecker: > Hmmm... `bezugsgröße_ost`? I see no big advantage in readability. I guess I am used to `ae` replacements.

Yes, I find "bezugsgröße" much more readable than "bezugsgroesse". The "ost" part will be solved differently, anyhow.

> So we specifically allow umlauts (and sharp s?) but require all lowercase words and a lower bar `_` in place for spaces?

That depends on whether we are sure we never need these guys as Python identifiers. E.g. `arbeitsl_geld_2.beträge` vs. `arbeitsl_geld["Beträge"]`. As long as we want to keep the former an option (which I think we should), Python conventions it is. 

> And I think being open for non German speakers and users is a very valid point (sonst können wir das hier alles auch auf Deutsch machen. Dann wären einige Diskussionen um das deutsche Steuer- und Transfersystem weniger umständlich). I think writing code or addressing/discussing functions parameter names in mails or github issues for instance is part of that.

Who would mind someone being a bit sloppy in github issues /  Zulip discussions? 

> When browsing through OpenFisca I have to look up French words pretty often (they don't use accents btw, `._impot`, `._bareme`,...). I'm not sure what is the right way but it could be good to use English language for clear cases such as vat or income tax.

The bigger problem is that English translations refer to concept rather than identifiers for laws. E.g., the Solidaritätssteuer would have to be part of income tax even though it is not part of Einkommensteuer. How to distinguish between Elterngeld, Elterngeld +, and Erziehungsgeld? Etc. I would choose one language to avoid discussing each time we add a new parameter whether it is in German or English. But this is somewhat orthogonal to the ASCII / non-ASCII discussion.

> Plus aren't there some issues with Umlaut Handling in other programming languages? Let's say someone wants to analyze the param database in a different language or re-use/share the param database with the Stata model?

There are good converters. This is a very distant possibility; I really would not want to make design decisions based on it.

mjbloemer: I'm not convinced by the pro-Umlaut argument. Readability is the only argument if I understand correctly.

I don't think we need to improve readability. Again: And if so, why only Umlauts and sharp-s? I think it would be ridiculous to make use of anything that slightly improves readability. E.g. you could also improve readability of the variable name by using [UTF-8 whitespace characters](http://jkorpela.fi/chars/spaces.html) instead of `_` (assuming not all are reserved by python) or use full stop characters when abbreviating names.

What do you say on lowercase vs. uppercase ([I think you prefer lowercase](https://github.com/iza-institute-of-labor-economics/gettsim/pull/60#pullrequestreview-313397771)) in the light of this readability argument? I have the impression that the readability argument would definitely require uppercase German nouns (please not).



> Code is read much more often than it is written.

That refers to writing clean source code in general, and not specifically to the naming of variables.

Also I think that underestimates how often variable names will be *typed*:
* in github issues, mails and chat, as mentioned before;
* when browsing source code, searching and grepping occurrences of a variable name in and across files

There are also people in Germany that work with US keyboards (some say US keyboards are more comfy for coding).

I think the disadvantages still overweight.
   

> There are good converters.

OK, good to know. Could you point me to one for the mentioned tasks?


> This is a very distant possibility; I really would not want to make design decisions based on it.

I already try to keep the param names of our Stata based model in sync with the new gettsim names to improve maintainability.

hmgaudecker: > I'm not convinced by the pro-Umlaut argument. Readability is the only argument if I understand correctly.

And uniqueness, to some extent. We could of course agree always to use the output of something like https://pypi.org/project/Unidecode/ (re your question on converters) throughout, but that would be more rules to keep in mind... 

> I don't think we need to improve readability. Again: And if so, why only Umlauts and sharp-s? I think it would be ridiculous to make use of anything that slightly improves readability. 

Can we please try to tone things down a little? I think there are two very valid opinions around here and I do not find anything that would be ridiculous. 

> E.g. you could also improve readability of the variable name by using [UTF-8 whitespace characters](http://jkorpela.fi/chars/spaces.html) instead of `_` (assuming not all are reserved by python) or use full stop characters when abbreviating names.

> What do you say on lowercase vs. uppercase ([I think you prefer lowercase](https://github.com/iza-institute-of-labor-economics/gettsim/pull/60#pullrequestreview-313397771)) in the light of this readability argument? I have the impression that the readability argument would definitely require uppercase German nouns (please not).

Well, there are clear rules for Python variable/function names, which are `lowercase_with_underscores`. The moment you see an uppercase identifier, you think it is a class.

Whether we are talking about identifiers or strings in general, it all depends on what is "normal". Using a character that looks like another but is actually different would indeed be a no-go.

There are trade-offs here, I am not saying that there is a completely obvious solution one way or another.

> > Code is read much more often than it is written.
> 
> That refers to writing clean source code in general, and not specifically to the naming of variables.

Yes, of course. And naming is one of them.

> Also I think that underestimates how often variable names will be _typed_:
> 
>     * in github issues, mails and chat, as mentioned before;
> 
>     * when browsing source code, searching and grepping occurrences of a variable name in and across files

The latter is a fair enough point. Usually you will be able to copy & paste, though. 

> There are also people in Germany that work with US keyboards (some say US keyboards are more comfy for coding).

I do. But switching them is a keystroke which I am more than happy to do for readability. 

> > This is a very distant possibility; I really would not want to make design decisions based on it.
> 
> I already try to keep the param names of our Stata based model in sync with the new gettsim names to improve maintainability.

So where are the limits of UTF-8 support in Stata? Assuming that supporting <= 13 is not an issue?

mjbloemer: > So where are the limits of UTF-8 support in Stata? Assuming that supporting <= 13 is not an issue

Stata seems to use/allow unicode UTF-8 encoding since version 14: https://www.stata.com/stata14/unicode/. you can `generate einkommen_tsd€` or even `gen zufriedenheit_😀️ = 1`.

No other problems with non-ascii-characters? Filesystems? URLs? path specifications?

hmgaudecker: > > So where are the limits of UTF-8 support in Stata? Assuming that supporting <= 13 is not an issue
> 
> Stata seems to use/allow unicode UTF-8 encoding since version 14: https://www.stata.com/stata14/unicode/. you can `generate einkommen_tsd€` or even `gen zufriedenheit_😀️ = 1`.

They did make it to the 21st century, after all :-)

> No other problems with non-ascii-characters? Filesystems? URLs? path specifications?

I am sure there are, but I do not think this is relevant for our naming conventions here? A URL is a URL, we should use the correct one regardless of what is in there. Paths should work in general, but I am not sure I would want to probe that.

Eric-Sommer: As long as people are saving their stuff in `utf-8`, this should not be major problem. Might be worth mentioning in the contributor guidelines, even though it should be self-evident nowadays.

docs/geps/gep-01.rst
```diff
@@ -121,7 +120,7 @@ or value change has a reference to a law or a source.
             de: Regelsatz
             en: Standard rate
 
-2. The `description` key has two sub-keys `de` and `en`, which
```
hmgaudecker: I just got stuck on this example; I think it is a great case in point to illustrate why an approach with dicts etc. will be much clearer.
- In `arbeitsl_geld_2.yaml`, the parameters `a2an1` ... `a2eg3` essentially repeat the description six times. This is very easy to get wrong at some point / forget a copy & paste.
- I find it very hard to see the schedule for a given point in time. You essentially have to check six (seven with `a2eg3ki`... who comes up with this kind of policies?) over 90 lines.

I suggest we walk through this example to see what we might want to do in general. My proposal (certainly not complete), explanations below:

```yaml
e_anr_frei:
    name:
      de: Anrechnungsfreie Einkommensanteile
      en: Income shares not subject to transfer withdrawal
  description:
    de: Einkommensanteile, die anrechnungsfrei bleiben. § 30 SGB II. Seit 01.10.2005 zudem definiert durch Freibetrag in § 11 SGB II, siehe auch § 67 SGB II. Seit 01.04.2011 § 11b (2) SGB II (neugefasst durch B. v. 13.05.2011 BGBl. I S. 850. Artikel 2 G. v. 24.03.2011 BGBl. I S. 453).
    en: ??? Do we repeat everything or just translation of the first bit?
  type: piecewise_linear
  value:
    2005-01-01:
      value: 
        400: 0.15
        900: 0.3
        1500: 0.15
      note: G. v. 24.12.2003 BGBl. I S. 2954.
    2005-10-01:
      value: 
        100: 1
        800: 0.2
        1200: 0.1
      note: G. v. 14.08.2005 BGBl. I S. 2407.
    2011-04-01:
      value: 
        100: 1
        1000: 0.2
        1200: 0.1
      note: SGB II. B. v. 13.5.2011 BGBl. I S. 850. Artikel 5 G. v. 20.12.2011 BGBl. I 2854. G. v. 14.08.2005 BGBl. I S. 2407.
```
Notes:
* `e_...` to be consistent with `e_st`
* I do not like the value(s).value nested construct that we have, but could not come up with anything better for now
* Stuff like `piecewise_linear` will need a clear definition somewhere
* Not sure how we want to handle the cases where only one parameter of six changes.
* Neither for the current `a2eg3ki` -- I would tend to repeat the whole thing and make clear the former is for households without, the second with kids (I would do that in the code for sure, not so obvious in the YAML).

docs/geps/gep-01.rst
ChristianZimpelmann: For me `bezugsgroesse` isn't much less readable than `bezugsgroesse`. 
I am therefore in favour of replacing all German-specific characters.

hmgaudecker: Now that @mjbloemer has started using Unicode all over the place in his Stata code? ;-)

docs/geps/gep-01.rst
hmgaudecker: Maybe we could use `upper_thresholds_rates` instead of (the inner) `value`? This would mean that the keys of `value[date]` would be differ across parameters, but it would become very descriptive.

docs/geps/gep-01.rst
ChristianZimpelmann: Did he? If this is already decided, that's also fine for me!

hmgaudecker: That is how I interpreted his last message :-) But I was mostly joking, shouldn't do that here. Sorry.

Nothing is decided yet. My interpretation is, however, that the mood has become more favourable towards special characters. 

In any case, this will not affect a huge amount of code and it is not the question deciding the future of GETTSIM. I would of course go with the majority.

Somewhat related: Do we all agree that names / descriptions / notes should keep correct spelling? These may end up directly in reports etc., I would not want to rename them before publishing something in German.

mjbloemer: > Did he?

no of course not. Why would you think so?

hmgaudecker: > > Did he?
> 
> no of course not. Why would you think so?

https://github.com/iza-institute-of-labor-economics/gettsim/pull/60#discussion_r369011217

Yes, I am sorry. I should not joke around.

docs/geps/gep-01.rst
Eric-Sommer: I see the benefits of this approach when the param file shall be easier to read. An even stronger case can be made for the Wohngeld parameters by household size (see the changes in #134). One has to be careful though to limit this to policies which are, structurally, stable over time. This won't hence work for the income tax tariff.

Is this how the param in the code would look like: `params["e_anr_frei"][100]`? I find this borderline acceptable. We certainly need to balance readability of the param file and readability of the code. After all, the param file is primarily a data source and only secondarily a documentation. A more user-friendly documentation could be compiled from this fairly easy, e.g. extending #122, while not touching the way these params are used in the code.

> * Stuff like `piecewise_linear` will need a clear definition somewhere

These are, after all, only the numeric parameters. In my view, the docstring of the high-level function (the one called from `tax_transfer()` is the natural place to describe what a function is actually doing.

hmgaudecker: > I see the benefits of this approach for the purpose of making the param file easier to read. An even stronger case can be made for the Wohngeld parameters by household size (see the changes in #134). One has to be careful though to limit this to policies which are, structurally, stable over time. This won't hence work for the income tax tariff.

If this latter case is prevalent, it would be a good reason to put the "type" parameter inside the value.date objects (along with `value` [`upper_thresholds_rates`] / `note`). We have a misunderstanding about what I have in mind there, see below.

> Is this how the param in the code would look like: `params["e_anr_frei"][100]`? I find this borderline acceptable. 

No. See below.

> We certainly need to balance readability of the param file and readability of the code. After all, the param file is primarily a data source and only secondarily a documentation. A more user-friendly documentation could be compiled from this fairly easy, e.g. extending #122, while not touching the way these params are used in the code.

> > Stuff like `piecewise_linear` will need a clear definition somewhere
> 
> These are, after all, only the numeric parameters. In my view, the docstring of the high-level function (the one called from `tax_transfer()` is the natural place to describe what a function is actually doing.

We will have a set of generic functions doing this, `piecewise_linear` will be one of them. The `piecewise_polynomial` we [discussed about before](https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/GEPs/near/185481518) will be another one. `step_function` another one. Names of course not pinned down. We could have some with jumps etc.. The point is there is nothing special about the precise application in the context of social assistance / wohngeld / tax schedules / ... and we do not need to re-write the same things over and over.

docs/geps/gep-01.rst
ChristianZimpelmann: > Somewhat related: Do we all agree that names / descriptions / notes should keep correct spelling? These may end up directly in reports etc., I would not want to rename them before publishing something in German.

Is this still about the special characters or what should be the reason to 
misspell?

hmgaudecker: Yes, but in another context than what we were discussing before — i.e., fields that will never end up in code. 

Checking the stuff I would consider less controversial first 😉

ChristianZimpelmann: Ok, then I am not sure what 

> names / descriptions / notes

refers to.

hmgaudecker: This is all about the contents of the yaml files. See [below](https://github.com/iza-institute-of-labor-economics/gettsim/pull/60#discussion_r369256241), for example.

Our original discussion was about the identifiers of parameters (e.g., `e_anr_frei`) and it is unresolved.

I am now probing the waters first if people have an aversion against special characters per se -- also in settings that have nothing to do with code --- or whether this is more about a reluctance to see them as Python identifiers.

E.g., if I was building a table with parameters, my identifying column would typically be what we include in "name". I would not want to (pay people to) manually fix all occurrences of "Bezugsgroesse" to "Bezugsgröße" etc. and I hope we agree that in published reports, ASCII-normalised spelling is not an option.

ChristianZimpelmann: Ok, thanks for the clarification! 

> Do we all agree that names / descriptions / notes should keep correct spelling?
I agree!

docs/geps/gep-01.rst
Eric-Sommer: > If this latter case is prevalent, it would be a good reason to put the "type" parameter inside the value.date objects (along with `value` [`upper_thresholds_rates`] / `note`). We have a misunderstanding about what I have in mind there, see below.

Now I see where this is going. I don't think the latter is the prevalent case, but there will be a bunch of generic cases which can hardly be generalized. Vorsorgeaufwendungen and income test for Wohngeld come to my mind.

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,279 @@
...
+
+
+Abstract
+--------
+
+This GEP pins down naming conventions for GETTSIM — general rules for what data columns,
+parameters, Python identifiers (functions, variables), etc. should be called. In a
+nutshell and without explanations, these conventions are:
+
+* Names follow standard Python conventions (``lowercase_with_underscores``).
```
Eric-Sommer: If I'm not mistaken, Python conventions require constant values in `UPPERCASE`. In the context of a tax-ben system the parameters are constants. Should we write the param dictionary(ies) in upper case?

hmgaudecker: This only applies to module-level constants.

E.g., after the import statements, define 
```python
PI = 3.14....
```
and never change it.

It does not apply to identifiers you would pass strings to functions etc.

docs/geps/gep-01.rst
hmgaudecker: My understanding is that the functions in question are usually rather generic. Some of the super-old tax schedules were actually so ingenious that they had jumps at the cutoffs, so we want to keep that case in mind. Other than that, all these schedules---whether they are taxes or transfer-reducing---seem pretty straightforward.

The hard part, and indeed the one you cannot generalise much, is coming up with the quantities. E.g., once you have taxable income the schedule is some lines of generic code. Coming up with that number requires several hundred or thousand lines of code, though.

Is that any different with the examples you were citing?

Eric-Sommer: OK, I guess one could count vorsorgeaufwendungen as quantities. The wohngeld income test is basically a share of income that is subtracted. But this share is endogenous to whether or not one pays income taxes, health care contributions or old-age pension contributions. So this might be done with a general linear function, with the "slope" not directly coming from the parameters.

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,279 @@
...
+              value: 16956
+              note: Artikel 1 G. v. 10.12.2019 BGBl. I S. 2115.
+
+
+   the `note` key:
+
+   * cites the law (Gesetz "G. v."), decree (Verordnung "V. v.") or proclamation (Bekanntmachung "B. v.") that changes the law
+   * in German
+   * follows the style ``Artikel [n] [G./V./B.] v. [DD.MM.YYYY] BGBl. I S. [SSSS]. [optional note].``
+   * do not add information "geändert durch" (it is always a change) or the date the law comes into force (this is exactly the date key in the previous line)
```
Eric-Sommer: the last parantheses is not correct. The date in the note is the date of publishing. The date of coming into force is below the `values` key.

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,279 @@
...
+   .. code-block:: yaml
+
+        description:
+          de: Einkommensanteil, der anrechnungsfrei bleibt, Intervall 2 [a2eg1, a2eg2]. § 30 SGB II. Seit 01.04.2011 § 11b SGB II.
+          en: Income share not subject to transfer withdrawal, interval 2 [a2eg1, a2eg2]. § 30 SGB II. Since 01.04.2011 § 11b SGB II.
+
+
+3. The ``value`` key
+
+   * contains the value as defined in the law
```
Eric-Sommer: Before:
* specifies the date at which the value below comes into force.

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,279 @@
...
+   * values in percentages can alternatively be expressed to the base of one
+   * add a leading zero for values smaller than 1 and greater than -1
+   * DM values have to be converted to Euro using the excange rate 1:1.95583.
+
+   Example 1:
+
+   .. code-block:: yaml
+
+        value:
```
Eric-Sommer: values:

docs/geps/gep-01.rst
hmgaudecker: I think that is precisely what @mjbloemer wrote. Am I missing something? Do you have a better suggestion less prone to misunderstandings?

Feel free to edit directly, also regarding the two comments below, it is getting kind of hard to follow here.

hmgaudecker:
With most of the contentious bits move to GEP-3, I would kind of like to get this GEP to "provisionally accepted" status soon. The only thing left that included hot debates are special characters and the only way to find out whether they cause problems is to throw them out there. This will be easy to revert and that is what the provisionally accepted status is for.

@amageh: Could you please try to summarize the discussion that led to the name `arbeitsl_geld` from https://github.com/iza-institute-of-labor-economics/gettsim/issues/79 as the example, tackling the [todo](https://gettsim--60.org.readthedocs.build/en/60/geps/gep-01.html#examples)?

The functions bit will require a bit of cleanup, too. Repeating the discussion [that just came up](https://github.com/iza-institute-of-labor-economics/gettsim/pull/50#discussion_r384444214), I think we want to get rid of the suggestion that function names should contain a verb. At least in the context of specific taxes and transfers, that does not make much sense (I also remember someone mentioning that, so no copyright on my behalf). I would actually suggest to only import functions as follows: 

```
import gettsim.taxes as tx
import gettsim.transfers as tr
```

and get rid of the intermediate module level by `__init__.py` files in `gettsim.taxes` and `gettsim.transfers` (these files would only include things like `from .zve import altersvors_aufwend`). So the above example would become:

```python
import gettsim.taxes as tx

altersvors_aufwend = tx.altersvors_aufw(...)
```

I kind of like that idea (the Zen... **Namespaces are one honking great idea -- let's do more of those!**)

hmgaudecker:
Any additional thoughts / suggestions / worries, @ChristianZimpelmann, @Eric-Sommer, @janosg, @MaxBlesch, @mjbloemer, @stichnoth, @tobiasraabe ?

MaxBlesch:
Thanks for clarifying the use of underscores. I think this GEP will give us a good guideline for further development. I have nothing else to add and vote for acceptance!

docs/geps/gep-01.rst
```diff
@@ -0,0 +1,178 @@
...
+might be useful to append an underscore plus one of {``j``, ``m``, ``w``, ``t``} to
+indicate the time unit.
+
+Across variations that include the same identifier, this identifier should not be
+changed, even if it leads to long variable names (e.g., ``kinderfreibetrag``,
+``zu_verst_e_ohne_kinderfreibetrag``). This makes searching for that identifier easier
+and less error-prone.
+
+Function names should contain a verb. Moreover, the length of a function name is
+typically inversely proportional to its scope. The public functions like maximize and
```
janosg: I feel flattered that you copied from estimagic but `maximize` and `minimize` are not part of gettsim, so maybe not a good example ;-)

hmgaudecker: Maybe I liked it because parts of that sounded very familiar :-). 

Yip, that part still needs change.

Eric-Sommer:
All fine with me as well. Only the issue of "verbs in function names" came up recently.

Eric-Sommer:
Fine with me as well. Maybe clarifying the use of verbs in function names? E.g. everything other than *calculate*?

hmgaudecker:
> Fine with me as well. Maybe clarifying the use of verbs in function names? E.g. everything other than _calculate_?

Yip, that's why I copied it up there in the discussion to this PR. 

What do you guys think of the namespaces (`tx.soli` or `st.soli` / `tr.arb_los_geld_2` /  / `ins.ges_rente` or `vers.ges_rente`) thing then? I think that will be the cleanest way out, otherwise it will always be awkward to find names at some point because you'd want to assign to the same variable that a function is called

hmgaudecker:
@tobiasraabe I just wanted to finally accept this once, but CI fails on Windows -- can you please check?

tobiasraabe:
@hmgaudecker The update of the environment timed out. Should be resolved with a restart.

hmgaudecker:
Thanks! Was in a rush, sorry.

Never debug while standing up...

PR #61, opened 2019-11-04
Upload GETTSIM to anaconda.org.

tobiasraabe:

The package should be accessible via conda.

### Todo

- [x] Blocked by #62.
- [x] Add a getting started page?
- [ ] Upload a last release once this pr is merged.


hmgaudecker:
Anything missing here? 

Once ready for merging, please include a sentence on how to install it in the README, modifying this one: *"The current version is usable, but probably only by the initiated. Please get in touch (ideally by joining our Zulip Chat at https://gettsim.zulipchat.com) if you want to use it already, else keep an eye out for what there is to come!"*

Also, would it be too hard to go for a very simple "Getting started" page in the docs? Something like Install via `conda install -c ... gettsim` and then five lines of example code including the list of required variables and the ones you get back? @MaxBlesch ?

tobiasraabe:
Good point. Just copied it from respy 😬

hmgaudecker:
Looks great. I'd like to wait for a few sentences on getting started as outlined above. 

Maybe you can add a page to the docs already, @tobiasraabe, with the installation instructions? @MaxBlesch and myself can fill it with a bit of example code and prose tomorrow.

docs/index.rst
```diff
@@ -2,13 +2,16 @@ Welcome to GETTSIM's documentation!
...
-   :end-line: 35
+   :start-line: 32
+   :end-line: 41
+
 
 .. toctree::
    :caption: Table of Contents
    :maxdepth: 1
 
+   installation
```
hmgaudecker: Given that installation is short and trivial, I think one page "getting started" with subsections "installation" and "usage" might be more sensible.

MaxBlesch:
I'll take over!

MaxBlesch:
I uploaded a preliminary code block. This includes a minimal example. This could would not work though, as we would need to assign the different reform functions. See https://github.com/iza-institute-of-labor-economics/gettsim/blob/976923323f2368ea52395c7b2a6b2ed040270d22/gettsim/tests/test_tax_transfer.py#L62 . 

There are several options:
- We leave this out.
- We include the same code as in the example of tax_transfer.
- As we now have a function, which selects the state of the policy by year, we could write the if conditions in this function and then they will also be selected by year. This also would save some code.

I prefer the last option!

docs/getting-started.rst
```diff
@@ -22,6 +22,15 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
-    gettsim.test()
+    import pandas as pd
+    from gettsim.config import ROOT_DIR
+    from gettsim.tax_transfer import tax_transfer
+    from gettsim.policy_for_date import get_policies_for_date
+
+    tax_data = yaml.safe_load((ROOT_DIR / "data" / "param.yaml").read_text())
+    year_of_policy = 2015  # We cover 1985 to 2019
+    tax_policy = get_policies_for_date(tax_policy_data, year=year)
+    tax_policy_pensions = pd.read_excel(ROOT_DIR / "data" / "pensions.xlsx").set_index(
```
hmgaudecker: What is this? Why separate from the rest?

docs/getting-started.rst
```diff
@@ -22,6 +22,15 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
 
-    import gettsim
-
-    gettsim.test()
+    import pandas as pd
+    from gettsim.config import ROOT_DIR
+    from gettsim.tax_transfer import tax_transfer
+    from gettsim.policy_for_date import get_policies_for_date
+
+    tax_data = yaml.safe_load((ROOT_DIR / "data" / "param.yaml").read_text())
```
hmgaudecker: YAML should be hidden from the user. Please include that in the `get_policies_for_date` function.

docs/getting-started.rst
```diff
@@ -22,6 +22,15 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
+    year_of_policy = 2015  # We cover 1985 to 2019
+    tax_policy = get_policies_for_date(tax_policy_data, year=year)
```
hmgaudecker: Should be `tax_data`

docs/getting-started.rst
```diff
@@ -22,6 +22,15 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
+    tax_policy_pensions = pd.read_excel(ROOT_DIR / "data" / "pensions.xlsx").set_index(
+        "var"
+    )
+    tax_transfer(df, tax_policy, tax_policy_pensions)
```
hmgaudecker: What is `df` supposed to be ?

Please call by name, not by location.

tobiasraabe:
We could use a notebook for the getting started page if the examples become too complex and we would like to inspect the output.

docs/getting-started.rst
MaxBlesch: This is the data for calculating the pensions(meanwages etc.) @Eric-Sommer always kept this separate because we are note sure yet, if this will be exogenously given or maybe estimated data. Maybe @Eric-Sommer you could elaborate on this!

docs/getting-started.rst
MaxBlesch: > > I uploaded a preliminary code block. This includes a minimal example. This
> > would not work though, as we would need to assign the different reform functions. See
> > https://github.com/iza-institute-of-labor-economics/gettsim/blob/976923323f2368ea52395c7b2a6b2ed040270d22/gettsim/tests/test_tax_transfer.py#L62
> > 
> > .
> > There are several options:
> > ```
> > * We leave this out.
> > 
> > * We include the same code as in the test of tax_transfer.
> > 
> > * As we now have a function, which selects the state of the policy by year, we could write the if conditions in this function and then they will also be selected by year. This also would save some code.
> > ```
> > 
> > 
> > I prefer the last option!
> 
> I do not understand much of this. What is the problem?

We adapted the code, such that reforms have to be explicitly assigned and not selected internal by year. I will just implement the third option, because then we have one function, which can be called if the user decides to select on the year paradigma.

hmgaudecker: Sorry, my bad. Let us forget about reforms at this point. Until we have a good API for that, some time will pass. 

The use case is simply: I have some dataset and want to get the taxes and transfers for a particular year.

docs/getting-started.rst
hmgaudecker: Ah, I see. See below. For now, we should target the use case that I want to have taxes and transfers for existing data. That means using the actual past values. In any case, we want to hide all of this from the user by default.

docs/getting-started.rst
MaxBlesch: Included loading the standard tax data as default in `get_policies_for_date`

MaxBlesch:
I simplified now a bit more. This should be okay. If everyone agees, I would add a table with mandatory variables and explanation!

docs/getting-started.rst
MaxBlesch: Changed to dataset

docs/getting-started.rst
MaxBlesch: Renamed!

hmgaudecker:
Ideally, of course, we would just pick the year from the dataset.

MaxBlesch:
I write a wrapper in tax_transfer!

MaxBlesch:
Implemented wrapper and a test. Also started the table. Is there any way to automate the writing of the table. This takes forever otherwise! @mjbloemer @Eric-Sommer

docs/getting-started.rst
```diff
@@ -30,44 +30,44 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
 |child0_18_num|Number of children between 0 and 18 in household          | Int         |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |hh_wealth    |Wealth of household                                       | Float       |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |m_wage       |Monthly wage of each individual                           | Float       |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |age          |Age if Individual                                         | Int         |
```
hmgaudecker: of

docs/getting-started.rst
```diff
@@ -30,44 +30,44 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |selfemployed |Self-employment of Individual                             | Bool        |
```
hmgaudecker: Whether self-employed

docs/getting-started.rst
```diff
@@ -30,44 +30,44 @@ To validate the installation, start a Python shell or a Jupyter notebook and typ
...
 |haskids      |Individual has kids                                       | Bool        |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |m_self       |Monthly wage of selfemployment of each individual         | Float       |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |m_pensions   |Monthly pension payments of each individual               | Float       |
-+-------------+------------------------------------------------------------------------+
++-------------+----------------------------------------------------------+-------------+
 |pkv          |Individual is private health insured                      | Bool        |
```
hmgaudecker: Whether (only) privately insured

MaxBlesch: Changed

docs/getting-started.rst
MaxBlesch: Changed

docs/getting-started.rst
MaxBlesch: Fixed

MaxBlesch:
I think this should be done from my side!

hmgaudecker:
Please also add a table with outputs, or did I miss that?

MaxBlesch:
Added it!

MaxBlesch:
Oh, yes. Fixed it!

hmgaudecker:
> /azp run

?

tobiasraabe:
The badge in the README signaled that the last build failed due to an https error. I tried to rerun the tests with this command (https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml). Did not work. I found another approach. Now, the badge is green again :)

hmgaudecker:
Thanks. Would the alternative be worth an entry in the maintenance guide?

tobiasraabe:
I propose something in #51.

hmgaudecker:
By "the alternative" I meant the "other approach" you used to re-run the Azure pipeline. At least I would have no idea of how to do that.

PR #64, opened 2019-11-04
Ignore files while testing coverage.

tobiasraabe:

Partially adresses #58.


tobiasraabe:
See here: https://codecov.io/gh/iza-institute-of-labor-economics/gettsim/tree/5805a32416b1e937cf448256f813729f3dfe4ffd. It works. `setup.py` and `release.py` are removed from coverage calculations.

Issue #62, opened 2019-11-04
Convert test data to yaml.


tobiasraabe:
### Current and desired situation

- The tests require ods-files which can only be read with a pip package which blocks making gettsim conda installable. #61 
- We have five data formats (csv, json, yaml, ods, xlsx). json can be replaced with yaml.
- Consider to tackle #63 along the way.

### Proposed implementation

*Discussing*

### Considered alternatives

*Discussing*

tobiasraabe:
@MaxBlesch @mjbloemer Hi Maximilans, can we simply convert ods to csv? Do we need the ods, but keep duplicated csvs for the tests? You definitely have a better overview. What are your thoughts?

mjbloemer:
I have never looked at the test data and I have actually no idea how that works. So I have no opinion on that.

My gut feeling is that it should be csv. yaml was for the params especially because of the structure.

tobiasraabe:
@mjbloemer I am sorry for pinging you. I thought you have already worked on that, but @MaxBlesch just told me that I was mistaken.

hmgaudecker:
That was my suggestion actually because I thought you had some code for converting Excel to YAML. But csv sounds reasonable for a table structure.

Eric-Sommer:
As I mentioned earlier (#31), I have a strong preference for keeping `ods` along with `csv`. Our tests are pretty complicated. Infact, most tests I set up fail initially because the test is wrong. I just created a new test by copying one from existing rows and building on top of that. Without the formulas, one would need to start from scratch and possibly repeat the same mistakes.

MaxBlesch:
I agree for different reasons with Eric and would like for now to keep the sheets as well. I suggest a subfolder in test_data called formula sheets, where we keep the ods files. In the test_data folder we keep just csv files. If all agree I would set this up and adapt the code.

hmgaudecker:
Fine. We just do not want to have an ods dependency because of the way we develop our tests. 

Sounds like doing that in Python/Pandas would be the better way to go, though, if the formulas are important.

MaxBlesch:
I would open a separate issue for that and just do the discussed way first. Then we can upload to Anaconda!

Eric-Sommer:
> Sounds like doing that in Python/Pandas would be the better way to go, though, if the formulas are important.

True, but then you test python code with python code.

PR #65, opened 2019-11-04
Move2csv

MaxBlesch:

Adressing #62 and converting the test data to csv. We will keep the ods sheets in a subfolder.

### Todo

- [x] Create csv with test data.
- [x] Delete ods dependency in code.


hmgaudecker:
Just keep the files side by side, please.

And remove the ods reader from the environment, just to be sure.

gettsim/tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import yaml
+
+from gettsim.config import ROOT_DIR
+
+
+@pytest.fixture(scope="session")
+def tax_policy_data():
+    return yaml.safe_load(open(f"{ROOT_DIR}/data/param.yaml", "rb"))
```
tobiasraabe: ``yaml.safe_load((ROOT_DIR / "data" / "param.yaml").read_text())``

tobiasraabe: See https://docs.python.org/3/library/pathlib.html for information on how pathlib works. It is amazing :)

gettsim/tests/policy_for_date.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+
+import numpy as np
+
+
+def get_policies_for_date(tb_pre, year, month=1, day=1):
+    tb = {}
+    this_year = datetime.date(year=year, month=month, day=day)
+    for key in tb_pre.keys():
```
tobiasraabe: No need for ``.keys()``.

gettsim/tests/policy_for_date.py
```diff
@@ -0,0 +1,19 @@
...
+        if tb_pre[key]["values"] is not None:
+            policy_dates = list(tb_pre[key]["values"].keys())
```
tobiasraabe: list and keys are not necessary.

gettsim/tests/test_alg2.py
```diff
@@ -47,14 +47,20 @@
 
 
 years = [2006, 2009, 2011, 2013, 2016, 2019]
-tax_policy_data = load_tax_benefit_data()
+
+
+@pytest.fixture
+def input_data():
+    file_name = "test_dfs_alg2.csv"
+    out = pd.read_csv(f"{ROOT_DIR}/tests/test_data/{file_name}")
```
tobiasraabe: Same issue as above. pathlib lets you concat paths with ``/``.

gettsim/tests/conftest.py
MaxBlesch: Changed accordingly. We should right down somewhere that we want paths that way. I changed some in the beginning from this format to f-strings.

gettsim/tests/policy_for_date.py
MaxBlesch: Done

gettsim/tests/policy_for_date.py
MaxBlesch: Done

gettsim/tests/test_alg2.py
MaxBlesch: See above

gettsim/tests/test_arbeitslosengeld_ui.py
```diff
@@ -27,17 +28,22 @@
...
+    out = pd.read_csv(ROOT_DIR / "tests" / "test_data" / file_name)
+    return out
 
 
 @pytest.mark.parametrize("year", years)
-def test_ui(year):
-    file_name = "test_dfs_ui.ods"
-    df = load_test_data(year, file_name, input_cols, pd_kwargs={"true_values": "TRUE"})
+def test_ui(input_data, tax_policy_data, year):
+    year_data = input_data[input_data["year"] == year]
```
tobiasraabe: Chaining commands is more readable. ``df = input_data.loc[input_data.year.eq(year), input_cols].copy()``

gettsim/tests/test_kindergeld.py
```diff
@@ -1,26 +1,32 @@
...
 from gettsim.taxes.kindergeld import kg_eligibility_hours
 from gettsim.taxes.kindergeld import kg_eligibility_wage
 from gettsim.taxes.kindergeld import kindergeld
-from gettsim.tests.auxiliary_test_tax import get_policies_for_date
-from gettsim.tests.auxiliary_test_tax import load_tax_benefit_data
-from gettsim.tests.auxiliary_test_tax import load_test_data
+from gettsim.tests.policy_for_date import get_policies_for_date
 
 
 input_cols = ["hid", "tu_id", "pid", "age", "w_hours", "ineducation", "m_wage"]
```
tobiasraabe: The convention is to capitalize module variables.

gettsim/tests/test_arbeitslosengeld_ui.py
MaxBlesch: But is `input_data.year` really nicer than `input_data["year"]`. Philipp told me the other way around.

gettsim/tests/test_wg.py
```diff
@@ -62,16 +66,20 @@ def test_wg(year):
...
         df[col] = np.nan
     df = df.groupby("hid").apply(wg, tb=tb)
-    expected = load_test_data(year, file_name, test_column)
-    assert_frame_equal(
-        df[test_column], expected, check_exact=False, check_less_precise=2
-    )
+    assert_frame_equal(df[test_column], year_data[test_column])
+
+
+@pytest.fixture
```
tobiasraabe: This fixture is not necessary. Just reuse the other one.

gettsim/tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import yaml
+
+from gettsim.config import ROOT_DIR
+
+
+@pytest.fixture(scope="session")
+def tax_policy_data():
+    return yaml.safe_load(open(ROOT_DIR / "data" / "param.yaml", "rb"))
```
tobiasraabe: No need for open(). Just use read_text() as a path method.

gettsim/tests/test_kindergeld.py
MaxBlesch: Done

gettsim/tests/test_wg.py
MaxBlesch: True. But there is a second csv for test_wg. Will ask Eric why its there

gettsim/tests/test_arbeitslosengeld_ui.py
tobiasraabe: My argument was mainly about the chaining. Feel free to adopt whatever style you like to pick a column. I am in favor of the dot, because it seems cleaner to me. I think Janos is in favor of your approach because it better distinguishes between columns and methods.

gettsim/tests/test_wg.py
MaxBlesch: Sorry. The mistake was to not use the second csv. Now its right!

gettsim/tests/conftest.py
MaxBlesch: Did it!

gettsim/tests/test_arbeitslosengeld_ui.py
hmgaudecker: indexes, no dots please. Pandas explicitly discourages using dot syntax (basically they just added it so you get tab completion in interactive mode). Makes it difficult to change things, too, although that might be less of a concern for 'year'

MaxBlesch: Quote Philipp: The dot is an remaining from the stata users. i have no idea if thats true :)

MaxBlesch: I didn't chain because I use the year data again later to get the expected results. Then I would need to condition a second time on year.

hmgaudecker: No idea where you may have dot syntax in Stata...

MaxBlesch:
I would go with `scope="module"`. You are right, that we could just write a `pd.read_csv`, but this would be repeated the same way, the fixtures are now. Also I think the structure with the fixture above seems very nice readable to me!

Going forward I think we will not get more csvs, because adding a feature can use the existing test cases plus some extra variables.

gettsim/tests/test_arbeitslosengeld_ui.py
tobiasraabe: I will shed this bad habit from now on :).

Eric-Sommer: > because it better distinguishes between columns and methods

I fully agree

PR #68, opened 2019-11-05
Fix display of logos on the landing page of the documentation.

tobiasraabe:

*Still investigating*


hmgaudecker:
Shamelessly included in #61, changing the paths did the trick.

PR #69, opened 2019-11-07
Cleanup of ALG II parameters and documentation

mjbloemer:
In this PR I clean up the ALG II parameters and documentation in param.yaml.

- Date keys were corrected.
- Missing parameters changes where added.
- All parameters have meaningful German and English names which are relatively short.
- All parameters have a meaningful description and cite the relevant law and history of the law if applicable.
- Every introduction or change of a parameter has a law (Gesetz "G. v."), decree (Verordnung "V. v.") or proclamation (Bekanntmachung "B. v.") in the Bundesgesetzblatt which is the source for that parameter. The date the introduction/change is coming into effect can also be found in the respective law (look for "Inkrafttreten" at the end of the law in the BGBl). Sources can be checked using the value of the note key at https://offenegesetze.de/.
- If a parameter is calculated the note will also give information on the calculation. 
- In some rare cases I add a date/value for a parameter that did not change the parameter. This is usually the case when a whole block of parameters is changed but one of the parameters did not change (by coincidence or not). In these cases there the same value is still listed in the BGBl. I add a "Betrag unverändert." to the note. This is also to signal that the BGBl. has been taken into account.

This PR does explicitly not (should be done in separate PRs):
- add initial ALG II withdrawal rules (done after merge, see #66)
- add new parameters
- add recent values


Eric-Sommer:
While we are cleaning params here, do we want to add the `wohnbedarf` array into `param.yaml` here? Has also to do with alg2.

mjbloemer:
> While we are cleaning params here, do we want to add the `wohnbedarf` array into `param.yaml` here? Has also to do with alg2.

[This one?](https://github.com/iza-institute-of-labor-economics/gettsim/blob/master/gettsim/benefits/kiz.py#L140)

Yes. But in a separate PR. 

I think these are the parents share for the housing cost (only used for KiZ, right?). 

There are probably multiple approaches to add these as parameters:

At ifo we do not use these numbers from the KiZ Merkblätter. They are functions from other parameters defined by a law and published in the Existenzminimumbericht.

Or is there a specific source for this array?

Eric-Sommer:
They are taken from the "Merkblatt Kinderzuschlag"

stmlitsche:
I reviewed the parameters in param.yaml and approve them.

Issue #71, opened 2019-11-07
Missing pytest dependency in conda package


hmgaudecker:
### Bug description

GETTSIM's conda package is missing the dependency on pytest

### To Reproduce

```
$ conda create -n tmp
$ conda activate tmp
$ conda install -c gettsim gettsim
$ python
Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gettsim
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/xxx/miniconda3/envs/gst/lib/python3.7/site-packages/gettsim/__init__.py", line 1, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

hmgaudecker:
Fixed in #70.

PR #72, opened 2019-11-07
Add pytest to runtime dependencies.

tobiasraabe:
 
Solves #71.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -10,7 +10,6 @@ source:
 build:
   number: 0
 
-
 requirements:
   build:
     - python {{ python }}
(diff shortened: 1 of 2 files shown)


meta.yaml
```diff
@@ -19,6 +18,7 @@ requirements:
   run:
     - python {{ python }}
     - pandas
+    - pytest
     - pyyaml
     - xlrd
```
hmgaudecker: While we are at it -- do we still need the Excel reader?

tobiasraabe: Yes. There is some test which needs excel files.

tobiasraabe: It is because of `pensions.xlsx` use in `tax_transfer.py` and tests.

hmgaudecker: Thanks. #73

PR #74, opened 2019-11-08
proper modelling of Unterhaltsvorschuss

### LINKED ISSUE #55: unterhaltsvorschuss ignores interaction with other parameters
The current modeling of Unterhaltsvorschuss ist not precise and interactions with child benefits are ignored.

Currently there are fixed values in the param file for the age groups while in reality the calculation of these values follows a specified function of existing parameters.

E.g. currently in 2019 this results for a child in the second age group in 212 Euro (which is the parameter uhv11).

Actually the calculation is: `Kindliches Existenzminimum / 12 - Kindergeld`.

E.g. for 2019-01-01 `4872/12 - 194 = 212.`

Now, if there is a change in Kindergeld in status quo or in a counterfactual, we would want to keep the interaction alive.

Actually from 2019-07-01 there was an increase in Kindergeld by 10 Euro and the Unterhaltsvorschuss automatically decreased by this, so now it is for the child in the middle age group `4872/12 - 204 = 202`. See e.g. https://www.berit-sander.de/2019/unterhaltsvorschuss-ab-01-juli-2019/

For the calculation of the amount § 7 Abs. 1 Satz 1 UhVorschG links to § 1612a BGB (Mindestunterhalt). There are factors 0.87 and 1.17 for the first and the third age group which could be considered as parameters.

Kindergeld is already a parameter and the Kindliches Existenzminimum can be found in Existenzminimumbericht or calculated from Kinderfreibetrag if Betreuungs- und Ausbildungsbedarfe can be added as parameters .

Eric-Sommer:

closes #55.

Unterhaltsvorschuss is now explicitly defined depending on 'Mindestunterhalt' (Mindestunterhaltsverordnung) and child benefit.


gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -57,6 +62,13 @@ def uhv_since_2017(tax_unit, params):
...
         & (uhv_inc_tu > 600),
         "uhv",
-    ] = params["uhv17"]
+    ] = (
+        params["uhv17"] * e_st_abz_params["kifreib_s"] / 12
+        - kindergeld_params["kgeld1"]
+    )
+
+    # round up
+    tax_unit["uhv"] = np.ceil(tax_unit["uhv"]).astype(int)
```
MaxBlesch: Didn't we for now agree to not include rounding. Could you maybe just comment it out?

MaxBlesch:
Very nice coded. Could maybe @mjbloemer look for the correctness of the law?

gettsim/benefits/unterhaltsvorschuss.py
Eric-Sommer: Still not decided if I'm not mistaken (#88).

MaxBlesch: Okay, mea culpa!

gettsim/data/e_st_abzuege.yaml
```diff
@@ -51,6 +51,39 @@ kifreib:
     2019-01-01:
       value: 7620
       note:
+kifreib_s:
+  group: kinderfreibetrag
```
mjbloemer:
```suggestion
```

Eric-Sommer:
> Maybe only Mindestunterhaltsverordnung is relevant from 2016 on?

Yes.

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,107 @@
...
+      note: G. v. 03.12.2015, BGBl. I S. 2188
+    2018-01-01:
+      value: 348
+      note: Artikel 1 G. v. 28.09.2017, BGBl. I S. 3525
     2019-01-01:
-      value: 160
-      note:
-uhv11:
+      value: 354
+      note: Artikel 1 G. v. 28.09.2017, BGBl. I S. 3525
```
mjbloemer:
```suggestion
      note: Artikel 1 V. v. 28.09.2017 BGBl. I S. 3525.
```
it is a Verordnung, not Gesetz. Also remove the comma to make it slightly easier to search for on buzer.de and on offenegesetze.de.

This probably applies to more entries.

mjbloemer: Is that even the correct decree for the 2019-01-01 policy date? The decree it is from 2017.

Eric-Sommer: OK. not sure whether the full stop makes things easier though.

mjbloemer: The full stop makes no difference for the search bar over on offenegesetze.de.

In some cases there is additional info (maybe an explation) on the note key (e.g. [here](https://github.com/iza-institute-of-labor-economics/gettsim/blob/79ca7cedd79566c453f41d4bd6ddddfb81731219/gettsim/data/arbeitsl_geld_2.yaml#L273) where I added a note). I think that is why I added a full stop everywhere. I have no strong opinion on omitting it when there is nothing more in the note.

Eric-Sommer: Should we leave it open for now and at some point define guidelines for the param files?

mjbloemer: Looking at [Artikel 1 V. v. 28.09.2017 BGBl. I S. 3525](https://offenegesetze.de/veroeffentlichung/bgbl1/2017/66#page=13) again: You are right. It defines both 2018-01-01 and 2019-01-01.

I think we should not leave a "G. v." there as it is a "V. v." ;) Simply is wrong. Nothing to be defined in note style guidelines. 

I have not much to say about note style guidelines. This really is a detail. We can have a separate discussion on that, sure (about the comma, the full stop and things like "Artikel.."). [Here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/69) I basically just follow the principle to make it is as easy as possible to look up the parameter in the law over on offenegesetze.de and buzer.de plus just have a consistent citing style (it is the citing style  buzer.de uses btw. So it is even very easy to copy and paste the citation).

Eric-Sommer: Well, that sounds like a guideline, but it would need to be made consistent in all files. Actually, in order to make full-text search work, one would need to replace the month to words. ('September' instead of '.09.'). Using this bit: `2017 BGBl. I S. 3525` should however lead to the correct bit.

mjbloemer: OK, let's discuss that at https://github.com/iza-institute-of-labor-economics/gettsim/issues/127

gettsim/data/e_st_abzuege.yaml
```diff
@@ -51,6 +51,38 @@ kifreib:
...
+  name:
+    de: Kinderfreibetrag (nur sächliches Existenzminimum).
+    en:
+  description:
+    de: § 32(6) EStG
+    en:
+  values:
+    2002-01-01:
+      value: 1824
+      note:
```
mjbloemer: Could you add a note here? [Quick search](https://offenegesetze.de/veroeffentlichung/bgbl1/2001/42?q=%221%20824%22#page=2) reveals 
```suggestion
      note: Artikel 1 G. v. 16.08.2001 BGBl. I S. 2074.
```

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,125 @@
-uhv5:
+uhv6_share:
   name:
-    de: Unterhaltsvorschuss für Kinder unter 5 Jahren. Der Prozentsatz in §2 UVG wird angewendet auf (2 x Kinderfrebeitrag (32 (6) EStG) / 12) - Kindergeld für das 1. Kind
+    de: Unterhaltsvorschuss für Kinder bis 6 Jahre. Anteil am Kinderfreibetrag.
```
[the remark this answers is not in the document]
Eric-Sommer: Not since 2016. Before it was used and you need this share.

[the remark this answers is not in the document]
mjbloemer: Please state the parameter you mean with Kinderfreibetrag here.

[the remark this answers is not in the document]
hmgaudecker: Maybe this would be a case where an "end date" or so for a parameter would be helpful? "obsolete_after"?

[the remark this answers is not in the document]
mjbloemer: I don't think that adding end dates is helpful (in general and here). 

You will always have to look at the source code to see which law is in force at a specific date.

Anyway, this share is a parameter of a law that in still in force.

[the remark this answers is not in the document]
hmgaudecker: Sure, but this share is not.

I agree we should not duplicate things and have the same dates in two places. I also hope that eventually, we can define a bit more in the parameters files than we do currently.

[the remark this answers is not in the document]
mjbloemer: > Sure, but this share is not.

Not part of §1612a anymore? Not in force?

The description key of this parameter (`uhv6_share`) says: "§1612a (1) Nr .1 BGB". Looking at https://www.buzer.de/gesetz/6597/a93318.htm it seems to me the law still has this parameter and is in force. With §1612a (4) maybe irrelevant for gettsim. But it is still there.

Btw how does that even work: having §1612a (1) and §1612a (4) at the same time? Seems they lead to different results. Do we just decide to only look at (4) and ignore (1)?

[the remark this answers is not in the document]
hmgaudecker: > > Sure, but this share is not.
> 
> Not part of §1612a anymore? Not in force?
> 
> The description key of this parameter (`uhv6_share`) says: "§1612a (1) Nr .1 BGB". Looking at https://www.buzer.de/gesetz/6597/a93318.htm it seems to me the law still has this parameter and is in force. With §1612a (4) maybe irrelevant for gettsim. But it is still there.

I meant the reference to Kinderfreibetrag; did not check the source until now. 

But this brings us back to the old "depends on" discussion we had at some point. I guess this would be a case where that would have changed over time? 

> Btw how does that even work: having §1612a (1) and §1612a (4) at the same time? Seems they lead to different results. Do we just decide to only look at (4) and ignore (1)?

Probably yes. https://www.buzer.de/gesetz/11821/al64917-0.htm 

Are the results different?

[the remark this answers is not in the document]
mjbloemer: > Are the results different?

yes I think that is e.g. 221 vs 212 euro in the example I mention here: 
https://github.com/iza-institute-of-labor-economics/gettsim/pull/74#pullrequestreview-330520136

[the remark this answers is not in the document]
hmgaudecker: I am lost now. The link I just sent for (4) defines precise numbers and none is below 335 € ?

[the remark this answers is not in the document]
mjbloemer: that is correct. subtract a `kgeld1` to get the Unterhaltsvorschuss.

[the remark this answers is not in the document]
Eric-Sommer: Because you have to subtract Kindergeld (204€ right now) in order to get the proper UHV. 

I don't know if the Kinderfreibetrag Referenz in §1612a BGB does have any purpose now. It used to have until 2016, which is why I did not delete it.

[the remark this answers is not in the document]
hmgaudecker: Great, you just have to be patient with an old man.

So until end of 2015, `uhv*_share` is relevant and `uhv*_amount` thereafter. Whatever we decide in the end, we should have some cross-reference eventually so one can infer that the parameter is superseded by another one.

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -1,41 +1,46 @@
...
+    The amount is specified in §1612a BGB and, ultimately, in Mindesunterhaltsverordnung.
 
     returns:
     tax_unit: Updated DataFrame including uhv
     """
+
+    # Before 2017, UHV was only paid up to 6 years, which is why we model it only since then.
     if params["year"] >= 2017:
-        return uhv_since_2017(tax_unit, params)
+        return uhv_since_2017(tax_unit, params, kindergeld_params)
```
mjbloemer: The `uhv_since_2017` function returns positive values for children `between(13, 17)`. Looking at https://www.buzer.de/gesetz/2978/al63880-0.htm children in that age range only get UV from 2017 on.

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -53,10 +58,14 @@ def uhv_since_2017(tax_unit, params):
         .sum()
     )
     tax_unit.loc[
-        (tax_unit["age"].between(12, 17))
+        (tax_unit["age"].between(13, 17))
```
mjbloemer: > bis zur Vollendung des 18. Lebensjahres.

(https://www.buzer.de/gesetz/2978/a42168.htm)

Are all the between() correct and return the correct result for non-integer age variables?

hmgaudecker: Good catch. Please drop this `.between()` construct everywhere. It is confusing, too lazy to look up the other discussion we had about it.

gettsim/data/unterhalt.yaml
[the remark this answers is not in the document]
mjbloemer: I would sketch it as follows:
until 2016-12-31 `uhv = kifrei_s * _share - kgeld1` and thereafter `uhv = _amout - kgeld1`. 

Plus, I thin Eric only implemented uhv starting 2017 because before UV was only a temporary benefit?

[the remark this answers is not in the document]
Eric-Sommer: Exactly. But if someone wants to model it pre 2017, the params are there.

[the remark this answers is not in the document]
hmgaudecker: So what is missing? The time limits?

gettsim/benefits/unterhaltsvorschuss.py
Eric-Sommer: could it be that the whole extension was done on 07-2017?

gettsim/data/unterhalt.yaml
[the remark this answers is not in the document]
Eric-Sommer: There is no function right now at all pre 2017 doing `kifreb_s * share - kgeld1`.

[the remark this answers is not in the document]
mjbloemer: Nothing is really missing. I basically just wanted:
> Please state the parameter you mean with Kinderfreibetrag here.

as we now have `kifrei` and `kifrei_s`.

I did not question the fact that the shares are added but not (yet) used. I'm absolutely fine with that.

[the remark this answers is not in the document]
hmgaudecker: Can somebody pls open an issue for the correction of that stuff mentioned above and also include renaming these guys? Maybe `ki_freib_gesamt` and `ki_freib_ex_min` ? Not married to those terms, but it should become clear from the name what is meant.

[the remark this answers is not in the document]
Eric-Sommer: See #130.

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -53,10 +53,23 @@ def uhv_since_2017(tax_unit, params):
         .sum()
     )
     tax_unit.loc[
-        (tax_unit["age"].between(12, 17))
+        (tax_unit["age"] >= 13)
+        & (tax_unit["age"] <= 17)
```
mjbloemer: This is for children 
> bis zum vollendeten 18. Lebensjahr

should this condition be `<18`?

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -1,41 +1,41 @@
...
-    # Calculate them on the level of the tax unit
-
+    # Benefit amount depends on the tax allowance for children ("sächliches Existenzminimum")
+    # and the child benefit for the first child.
     tax_unit["uhv"] = 0
     # Amounts depend on age
-    tax_unit.loc[tax_unit["age"].between(0, 5) & tax_unit["alleinerz"], "uhv"] = params[
-        "uhv5"
-    ]
+    tax_unit.loc[(tax_unit["age"] <= 6) & tax_unit["alleinerz"], "uhv"] = (
```
mjbloemer: This is
> bis zur Vollendung des sechsten Lebensjahrs

I think usually the age variable will be the number of completed years. So for someone 6 years and 6 month old this variable will be either 6 or 6.5. I think in both cases they will be in the second age group.
 
is it better to write it as <6?

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -1,41 +1,41 @@
...
+        params["uhv6_amount"] - kindergeld_params["kgeld1"]
+    )
     tax_unit.loc[
-        tax_unit["age"].between(6, 11) & tax_unit["alleinerz"], "uhv"
-    ] = params["uhv11"]
+        (tax_unit["age"] >= 7) & (tax_unit["age"] <= 12) & tax_unit["alleinerz"], "uhv"
```
mjbloemer: currently there is a gap between `tax_unit["age"] <= 6` and `tax_unit["age"] >= 7` if we allow for non integer age variables.

Eric-Sommer: the current GEP #60 requires integers. But we can make it robust to this.

hmgaudecker: Oh, I thought that was on *names* only, not on data types... I really need to look at that.

Yes, please use actual cutoffs where possible, not assuming integers. Thanks for implementign it already!

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,125 @@
...
+      value: 378
+      note: Artikel 1 V. v. 12.09.2019 BGBl. I S. 1393
+uhv12_amount:
   name:
-    de: Unterhaltsvorschuss für Kinder von 6-11 Jahren
+    de: Mindestunterhalt für Kinder von 7-12 Jahren
     en:
   description:
-    de: §2 UVG, §1612a BGB
+    de: §1 Mindesunterhaltsverordnung (MUV)
```
mjbloemer:
```suggestion
    de: Mindestunterhalt für Kinder von in der zweiten Altersstufe (vom siebten bis zur Vollendung des zwölften Lebensjahrs, nach § 1612a BGB). § 1 Mindesunterhaltsverordnung.
```

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,125 @@
...
+  values:
+    2008-01-01:
+      value: 1.17
+      note: Artikel 1 G. v. 21.12.2007 BGBl. I S. 2631
+uhv6_amount:
+  name:
+    de: Mindestunterhalt für Kinder bis 6 Jahre
+    en:
+  description:
+    de: §1 Mindesunterhaltsverordnung (MUV)
```
mjbloemer:
```suggestion
    de: Mindestunterhalt für Kinder von in der ersten Altersstufe (bis zur Vollendung des sechsten Lebensjahrs, nach § 1612a BGB). § 1 Mindesunterhaltsverordnung.
```

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,125 @@
...
+      value: 434
+      note: Artikel 1 V. v. 12.09.2019 BGBl. I S. 1393
+uhv17_amount:
   name:
-    de: Unterhaltsvorschuss für Kinder von 7 bis 12 Jahren
+    de: Mindestunterhalt für Kinder von 13-17 Jahren
     en:
   description:
-    de: §2 UVG, §1612a BGB
+    de: §1 Mindesunterhaltsverordnung (MUV)
```
mjbloemer:
```suggestion
    de: Mindestunterhalt für Kinder von in der dritten Altersstufe (vom 13. Lebensjahr an, solange midnerjährig, nach § 1612a BGB). § 1 Mindesunterhaltsverordnung.
```

gettsim/data/unterhalt.yaml
```diff
@@ -1,51 +1,125 @@
...
+      note: Artikel 1 V. v. 12.09.2019 BGBl. I S. 1393
+    2021-01-01:
+      value: 508
+      note: Artikel 1 V. v. 12.09.2019 BGBl. I S. 1393
+uhv17_inc:
+  name:
+    de: Monatliches Mindesteinkommen, um UHV für Kinder ab 13 Jahren zu erhalten
+    en:
+  description:
+    de: §1 (1a) Nr. 2
```
mjbloemer:
```suggestion
    de: § 1 (1a) Nr. 2 Unterhaltsvorschussgesetz
```

Issue #76, opened 2019-11-09
Move zve_list to `policy_for_date` and make it year-dependent


Eric-Sommer:
### Bug description

In #61, the definition of `zve_list` was not moved into `policy_for_date` and is still in `test_favorability_check`. It also must not be in `calculate_tax_transfers`. Moreover, it needs to be year-sensitive.

### Solution

    if year >= 2009:
        tb["zve_list"] = ["nokfb", "kfb", "abg_nokfb", "abg_kfb"]
    else:
        tb["zve_list"] = ["nokfb", "kfb"]

hmgaudecker:
Seems like  `policy_for_date` is the right place for this?

Eric-Sommer:
Absolutely

MaxBlesch:
Will do it these days!

Issue #81, opened 2019-11-11
Favorability Check with capital income tax


Eric-Sommer:
### Current and desired situation

Since [2009](https://www.buzer.de/gesetz/4499/al8065-0.htm), capital income is taxed with a flat rate of 25% (plus soli) (*Abgeltungssteuer*). In principle, tax authorities are supposed to tax capital income with a lower rate if the personal tax rate of all incomes would be lower than 25%. In order to do this, taxpayers have to claim their capital income [on a separate sheet](https://www.vlh.de/wissen-service/steuer-abc/wie-funktioniert-die-abgeltungssteuer.html). Whether this happens on a regular basis is unclear. 

### Proposed implementation

If we want to implement this, we need to extend `tb["zve_list"]` with the taxable income definitions including capital income, i.e. `zve_abg_nokfb` and `zve_abg_kfb`. Renaming `abg` to e.g. `cap` might also make sense here.

### Considered alternatives

leave as is because it's not relevant.

hmgaudecker:
It is good to keep this in mind, but I would indeed think this is very rare so it would end up being meaningful in quantitative terms. I would think that we have more pressing things to work on at present.

PR #82, opened 2019-11-11
Moving assignment zve_list to policy_for_date. 

MaxBlesch:

In #61, the definition of zve_list was not moved into policy_for_date and is still assigned in each test_case. As of 2009 the Abgeltungssteuer was introduced we also need to include this in the favorability check of the different ways to calculate the income tax. For more details see #76 and #81.

### Todo

- [x] We move the assignment of zve_list to policy_for_date
- [ ] We adapt favorability_check.py as discussed in #81 .

### DIFF
--- a/gettsim/policy_for_date.py
+++ b/gettsim/policy_for_date.py
@@ -60,10 +60,14 @@ def get_policies_for_date(year, tax_data_raw=None, month=1, day=1):
         tax_data["vorsorge"] = vorsorge2010
     else:
         tax_data["vorsorge"] = vorsorge_dummy
-    if year >= 2009:
-        tax_data["zve_list"] = ["nokfb", "kfb", "abg_nokfb", "abg_kfb"]
-    else:
-        tax_data["zve_list"] = ["nokfb", "kfb"]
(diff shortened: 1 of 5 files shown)


hmgaudecker:
Are issues 1. and 2. intimately related? If it is a trivial change, we may want to include the changed favourability check; else I would tend to ignore that for now.

gettsim/policy_for_date.py
```diff
@@ -60,6 +60,10 @@ def get_policies_for_date(year, tax_data_raw=None, month=1, day=1):
         tax_data["vorsorge"] = vorsorge2010
     else:
         tax_data["vorsorge"] = vorsorge_dummy
+    if year >= 2009:
+        tax_data["zve_list"] = ["nokfb", "kfb", "abg_nokfb", "abg_kfb"]
+    else:
```
Eric-Sommer: If we want to postpone #81, take out the `abg` items again here.

MaxBlesch: Removed year dependency.

MaxBlesch:
As we not tackling year dependency. This PR just closes #76 .

Issue #43, opened 2019-10-15
extend modelling of vorsorgeaufwendungen


Eric-Sommer:
### Current and desired situation

By means of *Vorsorgeaufwendungen*, taxpayers can deduct a fair share of their social security contributions. This is a major tax deduction and needs to be properly modelled. In 2004 and 2010, legislation was changed. So far, we only have implemented the ruling since 2010. For earlier years, we don't deduct them *yet*.

### Proposed implementation

Three functions are needed.
- ruling before 2004
- ruling between 2004 and 2010
- a function comparing those three against each other, choosing the most profitable one.

### Considered alternatives

- None really :)

Issue #79, opened 2019-11-11
Split up param.yaml by group and rename groups


hmgaudecker:
### Current and desired situation

param.yaml is a monster and unreadable. We have already agreed to split it up in smaller files. This makes it concrete and suggests renaming groups to something comprehensible by the non-initiated.

### Proposed implementation

Use something close to the current groups (https://github.com/iza-institute-of-labor-economics/gettsim/pull/56#issue-335872991) with more sensible and more readable names. This table is ordered by proposed name so that it is clear which ones I propose to merge:

| Current           | Proposal        |
| ----------------- | --------------- |
| abgeltung         | abg_st          |
| alg               | arb_los_geld    |
| alg2              | arb_los_geld_2  |
| alh               | arb_los_hilfe   |
| kinderfreibetrag  | ek_st_abzuege   |
| zve               | ek_st_abzuege   |
| tarif             | ek_st_tarif     |
| tax_sched         | ek_st_tarif     |
| eg                | erziehungsgeld  |
| kindergeld        | kindergeld      |
| kiz               | kinderzuschlag  |
| ust               | mw_st           |
| rente             | rente           |
| soc_ins_contrib   | soz_vers_beitr  |
| sgbxii            | sozialhilfe     |
| sh                | sozialhilfe     |
| uhv               | unterh_vorsch   |
| wohngeld          | wohngeld        |

Please check carefully before we go for this (@mjbloemer, @Eric-Sommer, @stichnoth, your opinions woud be especially great to have!) - I may have missed something or misunderstood what is actually meant by a parameter.

Does it make sense to include *abgeltungssteuer* as a parametr of *ek_st_tarif*? It seems weird to have a group with just one element.

mjbloemer:
I think coarse groups are good. A few quick comments:

The names are really a matter of taste, so in that fashion:
- I don't like the *_st* abbreviations *ek_st*, *mw_st* (I prefer *est* and *ust* or *vat*).
- *arb_los_geld* sounds strange.
- Probably *sozialhilfe* is not a good category for the all the *sgbxii* parameters.
- *_abzuege* is good and more precise than zve.
- I would drop the *_beitr* in *soz_vers_beitr*. It is just a group with everything SSC related.

> Does it make sense to include abgeltungssteuer as a parametr of ek_st_tarif? It seems weird to have a group with just one element.

[...]

PR #85, opened 2019-11-12
review tests

### LINKED ISSUE #80: Doublecheck tests
### Bug description

In an earlier PR I came across some strange values in the test data. Therefore I included TODOs in 

- test_zve.py, 
https://github.com/iza-institute-of-labor-economics/gettsim/blob/0323de5170658415d2dba08694d0645c716ba213/gettsim/tests/test_zve.py#L82

- test_arbeitslosengeld_ui.py
https://github.com/iza-institute-of-labor-economics/gettsim/blob/0323de5170658415d2dba08694d0645c716ba213/gettsim/tests/test_arbeitslosengeld_ui.py#L52

- test_tax_sched.py
https://github.com/iza-institute-of-labor-economics/gettsim/blob/0323de5170658415d2dba08694d0645c716ba213/gettsim/tests/test_tax_sched.py#L51

We need to doublecheck the test data if this is just due to rounding errors, due to false formulas calculating the data or errors in the functions.

Eric-Sommer:

closes #80.

**Three tests were reviewed:**
1. `test_ui`: There is one failure due to a deviation of < 0.01€. I think we can accept this.
2. `test_zve`: The reason for deviation is when and how vorsorgeaufwendungen are rounded. I'm reviewing these tests in #43 anyway, therefore I'd rather avoid the double effort.
3. `tax_sched` tests were indeed partly flawed and are now corrected. 
    a. 2009 tax schedule parameters were wrong in the test. Beware! Even though the law was changed in 03/2009, it became effective for the whole tax year. So `2009-01-01` as date is fine. 
    b. `tu` value were calculated wrongly for the first three rows in the test.
    c. There is infact a different `soli`calculation for income tax and capital income tax. Tests and code now account for this.
    d. I had to set `check_dtype=False`because depending on how you convert `ods`to`csv`, integers may become floats or you get an error because `int32 != int64`.


gettsim/taxes/calc_taxes.py
```diff
@@ -15,32 +15,40 @@ def tax_sched(tax_unit, tb):
     """
 
     adult_married = (~tax_unit["child"]) & (tax_unit["zveranl"])
-    # create ts dataframe and copy three important variables
+
     for inc in tb["zve_list"]:
-        tax_unit["tax_" + inc] = tb["tax_schedule"](tax_unit["zve_" + inc], tb)
+        # apply tax tariff, round to full Euro amounts
+        tax_unit["tax_" + inc] = tb["tax_schedule"](tax_unit["zve_" + inc], tb).astype(
```
MaxBlesch: Use f string instead!

gettsim/tests/test_arbeitslosengeld_ui.py
```diff
@@ -49,5 +49,5 @@ def test_ui(input_data, tax_policy_data, year):
         out_cols=[OUT_COL],
         func_kwargs={"tb": tb},
     )
-    # TODO: THis should be reviewed.
+
```
MaxBlesch: Just add a comment, that we verified the correctness and this is due to rounding in ods!

hmgaudecker:
Down the road: let us please not do any rounding, anywhere. I know we deviate slightly from the law there, but the cost for calculating numerical standard errors can be enormous.

Eric-Sommer:
> Down the road: let us please not do any rounding, anywhere. I know we deviate slightly from the law there, but the cost for calculating numerical standard errors can be enormous.

so what now? fix it for now and remove rounding all over the place later on?

MaxBlesch:
ODS is rounding. I think we can't prevent that! We could write the output data from the functions now in the csv, as we now that these values are correct!

gettsim/taxes/calc_taxes.py
```diff
@@ -70,7 +78,7 @@ def abgeltung(tax_unit, tb):
                 tax_unit["gross_e5_tu"] - 2 * (tb["spsparf"] + tb["spwerbz"]), 0
             )
         )
-    return tax_unit_abgelt["abgst"]
+    return tax_unit_abgelt["abgst"].round(2)
```
MaxBlesch: As HMG pointed out, we shouldn't round in our code. Sorry didn't saw that in the first round!

gettsim/taxes/calc_taxes.py
```diff
@@ -128,4 +135,4 @@ def soli_formula(solibasis, tb):
         np.maximum(0.2 * (solibasis - tb["solifreigrenze"]), 0),
     )
 
-    return soli
+    return soli.round(2)
```
MaxBlesch: Same. No rounding in the code

hmgaudecker:
I would be okay leaving it as is for now. But do not introduce new rounding.

Maybe somebody could open an issue that implements removal of all rounding.

> ODS is rounding. I think we can't prevent that

As per the discussion @mjbloemer, @Eric-Sommer and myself had over Zulip (starting about [here](https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/test.20data)), this is not a great way to come up with expected results in the first place.

Once #79 is merged and we implement the downstream consequences (not every function takes the whole tb monster but an almost-minimal subset, ...), this should become much easier to do.

MaxBlesch:
Okay. I'll open an issue to remove rounding. I agree that this is not the best way to calculate the test data and in some near future we can also get rid of the ods files!

Issue #73, opened 2019-11-07
Convert remaining Excel-files to YAML


hmgaudecker:
### Current and desired situation

We mostly removed the dependence on Excel, but `pensions.xlsx` is still around.

### Proposed implementation

Same as in #54 

- [ ] Convert `pensions.xlsx` to YAML format
- [ ] Adjust data loading whereever we call `pd.read_excel`
- [ ] Remove `xlrd` as a dependency in `environment.yml` and `meta.yaml`
- [ ] Remove lines 28 and 29 (starting with `# household` and  `# wg.to_excel`) in wohngeld.py

Eric-Sommer:
There is a bit of overlap between the two files, namely in the pension contribution rate. You can remove this from the pensions file while you're at it and instead refer to the `tb` parameter (you need to double that one though).

hmgaudecker:
Would it make sense to wait for #79? It just does not make sense to have pensions  separate... Would they not just become part of _ges_renten_vers_ (or whatever we end up calling it)?

Eric-Sommer:
One could integrate them into `param.yaml`. I wouldn't do it now because
- the pension parameters contain not only legal parameters, but empirical values like sum of contributions, number of contributors etc.
- the pension stuff is far less mature than the other parameters. I collected all the values for the *Rentenformel* without quite knowing how they will be dealt with in the end. If one does dynamic simulation, lots of these values will be endogenously set from the model. Until we have a definitive plan on how to simulate future pensions, I feel it's easier to keep them apart.

hmgaudecker:
> the pension parameters contain not only legal parameters, but empirical values like sum of contributions

This seems perfectly fine to have in params.yaml. It is a value that is relevant for calculations, which will not change anymore. I do not care much whether it is a law or from somewhere else.

> number of contributors etc.

This does not seem to be relevant, or is it for the Nachhaltigkeitsfaktor etc.?

> the pension stuff is far less mature than the other parameters. I collected all the values for the _Rentenformel_ without quite knowing how they will be dealt with in the end. If one does dynamic simulation, lots of these values will be endogenously set from the model. Until we have a definitive plan on how to simulate future pensions, I feel it's easier to keep them apart.

Well, they **are** separate under the new structure, right?

Let us keep them in and maybe add notes expressing your uncertainty about some parameters during review.

Eric-Sommer:
> > number of contributors etc.
> 
[...]

PR #86, opened 2019-11-12
Move to yaml for pension tax data.

MaxBlesch:

We want to remove all excel dependencies from the gettsim package. The last data which is stored in a excel file is the tax data on pensions. Therefore I will adress #73 here.

### Todo

- [x] Convert `pensions.xlsx` to YAML format
- [x] Adjust data loading whereever we call `pd.read_excel`
- [x] Remove `xlrd` as a dependency in `environment.yml` and `meta.yaml`
- [x] Remove lines 28 and 29 (starting with `# household` and  `# wg.to_excel`) in wohngeld.py


MaxBlesch:
I would keep this for now in a separate object. As soon as we have splittet up params.yaml,  I will come up with an overall structure for all parameters.

hmgaudecker:
> I would keep this for now in a separate object. As soon as we have splittet up params.yaml, I will come up with an overall structure for all parameters.

Just wait on #79 before merging it. No rush.

gettsim/pensions.py
```diff
@@ -23,7 +23,7 @@ def pensions(person, tb, tb_pens):
     # meanwages is only filled until 2016
     yr = min(tb["yr"], 2016)
 
-    person = update_earnings_points(person, tb, tb_pens[yr])
+    person = update_earnings_points(person, tb, tb_pens, yr)
```
tobiasraabe: Can we change `yr` to `year`?

gettsim/pensions.py
```diff
@@ -49,22 +49,24 @@ def update_earnings_points(person, tb, tb_pens):
...
     person["EP"] += out
     return person
 
 
-def _ep_for_earnings(person, tb, tb_pens):
+def _ep_for_earnings(person, tb, tb_pens, yr):
     """Return earning points for the wages earned in the last year."""
     westost = "o" if person["east"] else "w"
-    return np.minimum(person["m_wage"], tb["rvmaxek" + westost]) / tb_pens["meanwages"]
+    return (
```
tobiasraabe: Do you need `np.minimum`? Are the entries vectors and you need the minimum element-wise?

Do not calculate on return.

gettsim/pensions.py
```diff
@@ -49,22 +49,24 @@ def update_earnings_points(person, tb, tb_pens):
...
+        np.minimum(person["m_wage"], tb["rvmaxek" + westost]) / tb_pens["meanwages"][yr]
+    )
 
 
-def _ep_for_care_periods(df, tb, tb_pens):
+def _ep_for_care_periods(df, tb, tb_pens, yr):
     """Return earnings points for care periods."""
     return 0.0
```
tobiasraabe: 0 is fine.

gettsim/pensions.py
```diff
@@ -90,7 +92,7 @@ def _regelaltersgrenze(person):
 
 def _rentenwert_until_2017(tb_pens, yr):
```
tobiasraabe: Isn't a single function for rentenwert better than two functions to centralize the logic?

gettsim/pensions.py
```diff
@@ -121,14 +123,11 @@ def _lohnkomponente(tb_pens, yr):
     the previous years. For details see
     https://de.wikipedia.org/wiki/Rentenanpassungsformel
     """
-    return tb_pens.loc["meanwages", yr - 1] / (
-        tb_pens.loc["meanwages", yr - 2]
+    return tb_pens["meanwages"][yr - 1] / (
```
tobiasraabe: That is way to complicated for a return statement. Compute intermediate values with good variables names.

gettsim/pensions.py
```diff
@@ -138,8 +137,8 @@ def _riesterfactor(tb_pens, yr):
     and the contributions to the pension insurance. For details see
     https://de.wikipedia.org/wiki/Rentenanpassungsformel
     """
-    return (100 - tb_pens.loc["ava", yr - 1] - tb_pens.loc["rvbeitrag", yr - 1]) / (
-        100 - tb_pens.loc["ava", yr - 2] - tb_pens.loc["rvbeitrag", yr - 2]
+    return (100 - tb_pens["ava"][yr - 1] - tb_pens["rvbeitrag"][yr - 1]) / (
```
tobiasraabe: Too complicated. Break it up.

gettsim/policy_for_date.py
```diff
@@ -71,3 +71,31 @@ def get_policies_for_date(year, tax_data_raw=None, month=1, day=1):
...
+            (ROOT_DIR / "data" / "pension_data.yaml").read_text()
+        )
+    pension_data = {}
+
+    # meanwages is only filled until 2016. The same is done in the pension function.
+    year = min(raw_year, 2016)
+
+    for key in raw_pension_data:
+        pension_data[key] = {}
+        data_years = list(raw_pension_data[key]["values"].keys())
```
tobiasraabe: Drop `.keys()`

gettsim/policy_for_date.py
```diff
@@ -71,3 +71,31 @@ def get_policies_for_date(year, tax_data_raw=None, month=1, day=1):
...
+        # For calculating pensions we need demographic data up to three years in the
+        # past.
+        for yr in range(year - 3, year + 1):
```
tobiasraabe: year

gettsim/policy_for_date.py
```diff
@@ -71,3 +71,31 @@ def get_policies_for_date(year, tax_data_raw=None, month=1, day=1):
...
+
+def get_pension_data_for_year(raw_year, raw_pension_data=None):
+    if not raw_pension_data:
+        raw_pension_data = yaml.safe_load(
+            (ROOT_DIR / "data" / "pension_data.yaml").read_text()
+        )
+    pension_data = {}
+
+    # meanwages is only filled until 2016. The same is done in the pension function.
+    year = min(raw_year, 2016)
```
tobiasraabe: More explanatory name.

hmgaudecker:
> I am in favor of removing the calculations on the return statements and use intermediate values with explanatory variable names instead. It is the decision between understanding the formula or reading it like a sentence and the second is easier in the long run.

I am not sure we need a strict rule on this. If it becomes too lengthy, that is a good idea indeed. And an even better idea **might** be to try to break up the calculation in two parts. But for relatively short calculations, I do not see any difference in readability.

mjbloemer:
> 
> > Do we need a note key for the change of empirical values?
> 
> Agreed. We should have links to the sources for the empirical values, too. 
> 
> But let us please continue further discussion like this in #86.

_Originally posted by @hmgaudecker in https://github.com/iza-institute-of-labor-economics/gettsim/issues/73#issuecomment-554379988_

Hm... the source and the definition of empirical values do not change over time. If the definition changed it would be a different empirical parameter. Do we take all the values from one single source/time series and only update this every year? We might have multiple sources for one single parameter but this could still be mentioned in a single description key for one "parameter".

gettsim/data/pension_data.yaml
```diff
@@ -0,0 +1,537 @@
+alpha:
```
hmgaudecker: Filename should be `ges_renten_vers`, seemed uncontroversial in the [current proposal](https://github.com/iza-institute-of-labor-economics/gettsim/issues/79#issuecomment-554283492) in #79.

gettsim/data/pension_data.yaml
```diff
@@ -0,0 +1,537 @@
...
+    2015:
+      note:
+      value: 29.21
+    2016:
+      note:
+      value: 30.45
+    2017:
+      note:
+      value: 31.03
+rvbeitrag:
```
hmgaudecker: Should be deleted, part of `param.yaml` / `soz_vers_beitr.yaml`.

hmgaudecker:
> Hm... the source and the definition of empirical values do not change over time. If the definition changed it would be a different empirical parameter. Do we take all the values from one single source/time series and only update this every year? We might have multiple sources for one single parameter but this could still be mentioned in a single description key for one "parameter".

Ah, I should have looked at the code before. Forget about my previous answer, I thought the note key was missing... 

Yes, here it makes more sense to add the note to the parameter directly. Source will typically stay the same and if there are two or three in there, also fine.

MaxBlesch:
This will be included in #111 .

Issue #83, opened 2019-11-11
API for tax units and household types


Eric-Sommer:
### Current and desired situation

The input data contain a number of demographic information derived from individual information. Examples include:

1. The definition of tax_units and their size.
2. The number of children in a given age (see #30) and the definition of a child itself.
3. The household type (Single, Single Parent, Couple without kids, Couple with kids, [other]). This also affects dummy variables, such as `married` and `zveranl`.

As these are currently dealt with in the data preparation, they are a black box to GETTSIM users. Moreover, it leads to more required input variables and thus more burden on the user.

### Proposed implementation

Create functions for each of the items above which return the variables `hhsize`, `hhsize_tu`, `child_num`, `hhtype` etc. based on the following variables only:

| Variable       | Explanation      |
|-----------|-------------------|
| `pid`       | Personal ID       |
| `female`    |                   |
| `age`       |                   |
| `ineducation` |         |
| `mother_id` | `pid` of mother |
| `father_id`  |  `pid` to Father                  |
| `child_id`    | `pid` to own child                  |
| `partner_id` |  `pid` to partner                   |

### Considered alternatives

leave as is, which is not feasible in the long run.

hmgaudecker:
In the long run, handling the data structures sensibly and efficiently will be a much bigger question. So, IIUC, `mother_id` and the subsequent variables are not part of the data for the moment, right? I think they probably should be, but that would be part of the "right" overhaul.

For now, would some simple Pandas constructs not do?
```
hhsize = household['age'].count()
nadults = household.query('age > 27 | age >=18 & ineducation == False)
...
```
(not that we should hard-code these ages in the long run :-))

Eric-Sommer:
Adding these ids is part of this task.

hmgaudecker:
As I said, this will be rather the level of a GEP than a task and require very careful design.

hmgaudecker:
Let's leave this to GEP 2.

PR #93, opened 2019-11-16
Fill documentation table

### LINKED ISSUE #91: Fill up documentation table on user data columns

The tables for [required input columns](https://gettsim.readthedocs.io/en/stable/getting-started.html#required-columns-in-input-data) and for [output columns](https://gettsim.readthedocs.io/en/stable/getting-started.html#columns-returned-by-the-simulator) are hardly filled in the second column.

It would be useful for picking up work on GEP 1 again and even more for my soon-to-be-drafted GEP 2 on internal data representation to have an overview of the actual names. @MaxBlesch, would be great if you could do that soonish with @Eric-Sommer's input as required.

### Proposed implementation

* Either just fill up the table for now. 
* In the long run, we probably want to have a YAML file that may have additional information structure (e.g., grouping of variables, units of measurement, ...)

I have no strong opinion for the moment what is easier; if it is simple to create a table from the YAML in the docs (is it, @tobiasraabe ?), maybe do it right immediately. Also saves the formatting stuff...

Eric-Sommer:

Filled some entries. I realized that 
1. there are lots of variables which should be calculated by us and not provided by the user: 
    1. `tu_id`  (#83) . 
    2. `hhsize`, `hhsize_tu`,`*_num`, `*_num_tu`,`hh_korr`  see #30.
    3. `alg_soep` doesn't make any sense. I once compared our UI simulation to observed variables.
    4. `byear`
    5. `child, `alleinerz`, `pensioner`
    6. `divdy` is redundant to `m_kapinc`
    7. `cnstyr` can be left blank, but if given, than the actual year.

2. There are also many variables which should not be in the default output, such as
    1. all the intermediate taxable income columns `zve_*
    2. `tax_*`
    3. allowances. `sonder`, `hhfreib`, `kifreib` etc.
    4. Same holds for `alg2_ek` stuff and `_basis` items. These refer to benefit amounts **before** being       checked against other, superior claims. They are useful for debugging though.


hmgaudecker:
Looks great, thanks!

@MaxBlesch: Could you please: 
* Go through the names and check for typos / capitalisation etc.
* Check whether we really need any of those left blank in the input
* Restrict the output to those that make sense 

That should go into 0.2.1; I think it is fine to deviate from the "no API changes" here. 

Once done, please open a new issue / PR so that Eric's  general points above do not get lost; the YAML converson in #91 should also find a home there.

MaxBlesch:
I cleaned out the table and just left variables which make sense. Also I removed some inputs which can be easily calculated in our code.

gettsim/benefits/kiz.py
```diff
@@ -96,6 +99,18 @@ def kiz(household, tb):
     return household
 
 
+def calc_min_income_kiz(household, tb):
+    # Are there kids in the household
+    if len(household[household["child"]]) > 0:
```
hmgaudecker: Let us please not rely on this actually being a Boolean value. Or have a check right in the begininng of the code.

docs/getting-started.rst
```diff
@@ -174,117 +158,71 @@ Note that if one of these columns exists, it will be overwritten.
...
+|gross_e5_tu        |Taxable Inc. from Capital, couple sum                | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|kindergeld_hh      |                                                     | Float      |
+|gross_e6_tu        |Taxable Inc. from Rents, couple sum                  | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|kindergeld_tu      |                                                     | Float      |
+|gross_e7_tu        |Taxable Inc. from Pensions, couple sum               | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|uhv                |                                                     | Float      |
+|abgst_tu           |Monthly Capital Income Tax Due, couple sum           | Float      |
```
hmgaudecker: Please be consistent in capitalisation.

I would actually suggest making everything lowercase, except for the very first letter.

docs/getting-started.rst
```diff
@@ -174,117 +158,71 @@ Note that if one of these columns exists, it will be overwritten.
...
+|wohngeld           |Monthly housing benefit, household sum               | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|m_alg2             |                                                     | Float      |
+|m_alg2             |Monthly social assistance, household sum             | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|dpi_ind            |                                                     | Float      |
+|dpi_ind            |Monthly disposable income, individual                | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|dpi                |                                                     | Float      |
+|dpi                |Monthly disposable income, household                 | Float      |
```
hmgaudecker: household sum ?

docs/getting-started.rst
```diff
@@ -174,117 +158,71 @@ Note that if one of these columns exists, it will be overwritten.
...
+|uhv                |Alimony Advance payment, individual to the child     | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|uhv_hh             |                                                     | Float      |
+|regelbedarf        |Household socio-economic *need*, incl. housing cost  | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|ekanrefrei         |                                                     | Float      |
+|regelsatz          |Household socio-economic *need*, lump-sum            | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|kiz_temp           |                                                     | Float      |
+|alg2_kdu           |Household Appropriate Housing Cost                   | Float      |
```
hmgaudecker: ??? Please look for a better translation, this sounds strange.

docs/getting-started.rst
```diff
@@ -174,117 +158,71 @@ Note that if one of these columns exists, it will be overwritten.
...
+|kindergeld_tu      |Monthly Child Benefit, household sum                 | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|regelbedarf        |                                                     | Float      |
+|incometax          |Monthly Income Tax Due, individual                   | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|regelsatz          |                                                     | Float      |
+|incometax_tu       |Monthly Income Tax Due, couple sum                   | Float      |
 +-------------------+-----------------------------------------------------+------------+
-|alg2_kdu           |                                                     | Float      |
+|uhv                |Alimony Advance payment, individual to the child     | Float      |
```
hmgaudecker: What is 'individual to the child' supposed to mean? This is the government paying because somebody else does not, right?

docs/getting-started.rst
```diff
@@ -98,69 +92,59 @@ Required columns in input data
...
+|alleinerz     |Dummy: Single Parent                                     | Bool        |
 +--------------+---------------------------------------------------------+-------------+
-|cnstyr        |                                                         | Int         |
+|eigentum      |Dummy: Owner-occupied housing                            | Bool        |
 +--------------+---------------------------------------------------------+-------------+
-|m_transfers   |                                                         | Int         |
+|cnstyr        |Constr. Yr of Dwelling (1: <1965,2:1966-2000,3:>2000)    | Int         |
 +--------------+---------------------------------------------------------+-------------+
-|hh_korr       |                                                         | Int         |
+|m_transfers   |Sum of monthly public/pricate transfers not simulated    | Int         |
```
hmgaudecker: private

What is it supposed to mean? We can have more than one row, can't we?

docs/getting-started.rst
```diff
@@ -98,69 +92,59 @@ Required columns in input data
...
+|ineducation   |Dummy: employment status "in education"                  | Bool        |
 +--------------+---------------------------------------------------------+-------------+
-|child11_num_tu|                                                         | Int         |
+|alleinerz     |Dummy: Single Parent                                     | Bool        |
 +--------------+---------------------------------------------------------+-------------+
-|cnstyr        |                                                         | Int         |
+|eigentum      |Dummy: Owner-occupied housing                            | Bool        |
 +--------------+---------------------------------------------------------+-------------+
-|m_transfers   |                                                         | Int         |
+|cnstyr        |Constr. Yr of Dwelling (1: <1965,2:1966-2000,3:>2000)    | Int         |
```
hmgaudecker: ??? Please more verbosity.

hmgaudecker: Ahh, construction ...

gettsim/benefits/kiz.py
MaxBlesch: I saw somewhere else that we want to have some kind of identifier saying: this individual is a child of the other. So we will change all these functions anyway. Therefore I think we should stick with a test for now!

docs/getting-started.rst
MaxBlesch: I guess this variable comes from SOEP where people are probably asked: Do you have other transfers?

docs/getting-started.rst
MaxBlesch: Changed to "assigned to the child"

docs/getting-started.rst
MaxBlesch: Changed to "Appropriate housing costs for a family in *need*"

docs/getting-started.rst
MaxBlesch: Fixed

docs/getting-started.rst
Eric-Sommer: the "individual" refers to the recipient and to the unit it should be assigned in the data set (as opposed to a household sum).

gettsim/benefits/kiz.py
Eric-Sommer: agreed to leave it for now (this holds almost all of the changes in this PR).
But `household['child'].sum()` would work for both types and is shorter.

gettsim/benefits/wohngeld.py
```diff
@@ -195,20 +193,29 @@ def _calc_wg_income_deductions_until_2015(household, tb):
     """ calculate special deductions for handicapped, single parents
     and children who are working
     """
+    household["child_below_11_tu"] = household.groupby("tu_id")["age"].apply(
+        count_child_below_11_tu
+    )
```
tobiasraabe: Remove ``count_child_below_11_tu``. Even if we are not concerned about performance for now, ``apply`` is essentially a for-loop and the function is not very clear. What about ``household.groupby("tu_id")["age"].transform(lambda x: x.le(11).sum())``. And should it be children instead of child?

hmgaudecker: Yeah, I guess the variable should be called "n_children_below_11_tu"

MaxBlesch: That looks a lot better. Adapted it!

gettsim/benefits/benefit_checks.py
```diff
@@ -130,7 +131,7 @@ def wealth_test(household, tb):
     # Wealth test for Wohngeld
     # 60.000 € pro Haushalt + 30.000 € für jedes Mitglied (Verwaltungsvorschrift)
     household.loc[
-        (household["hh_wealth"] > (60000 + (30000 * (household["hhsize"] - 1)))),
+        (household["hh_wealth"] > (60000 + (30000 * (household_size - 1)))),
```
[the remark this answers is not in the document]
MaxBlesch: Didn't see that before. But this is indeed more readable!

docs/getting-started.rst
MaxBlesch: Does "assigned to the child" work for everybody?

hmgaudecker: Just leave as 'individual' for now.

Btw, please add a note to the docs (this PR is as good as any) before this table on the data structure. I.e., they need to be in "fully" long format: Each household member gets his/her own row, regardless of age.

MaxBlesch: Inserted the note and changed the description to individual

PR #98, opened 2019-11-20
Update maintenance manual.

tobiasraabe:

Two adjustments to the maintenance manual

- Delete the note on labels for backport PRs
- longer names for backport branches

### DIFF
--- a/docs/how-to-maintain.rst
+++ b/docs/how-to-maintain.rst
@@ -98,15 +98,14 @@ critical bug was found, which should be fixed in both ``0.3.0`` and in ``0.2.1``
 
 1. Create a PR containing the bug fix which targets the master branch.
 2. Add a note to the release notes for version 0.2.1.
-3. Add a label ``backport-to-0.2.1`` to the PR.
-4. Squash merge the PR into master and note down the commit sha.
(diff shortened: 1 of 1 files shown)


docs/how-to-maintain.rst
```diff
@@ -98,15 +98,14 @@ critical bug was found, which should be fixed in both ``0.3.0`` and in ``0.2.1``
 
 1. Create a PR containing the bug fix which targets the master branch.
 2. Add a note to the release notes for version 0.2.1.
-3. Add a label ``backport-to-0.2.1`` to the PR.
-4. Squash merge the PR into master and note down the commit sha.
-5. Create a new PR against branch ``0.2``. Call the branch for the PR
-   ``backport-<#pr>`` where #pr is the PR number.
-6. Use ``git cherrypick -x <commit-sha>`` with the aforementioned commit sha to apply
+3. Squash merge the PR into master and note down the commit sha.
+4. Create a new PR against branch ``0.2``. Call the branch for the PR
```
hmgaudecker: Please use `[No.]` instead of `#`, else I would be afraid that cursory readers are tempted to actually use `#No` as in the branch name.

Issue #75, opened 2019-11-08
Add the wohnbedarf array as parameters


mjbloemer:
> While we are cleaning params here, do we want to add the `wohnbedarf` array into `param.yaml` here? Has also to do with alg2.

_Originally posted by @Eric-Sommer in https://github.com/iza-institute-of-labor-economics/gettsim/pull/69#issuecomment-551848848_

> [This one?](https://github.com/iza-institute-of-labor-economics/gettsim/blob/master/gettsim/benefits/kiz.py#L140)
> 
> 
> 
> Yes. But in a separate PR. 
> 
> I think these are the parents share for the housing cost (only used for KiZ, right?). 
> 
> There are probably multiple approaches to add these as parameters:
> 
> At ifo we do not use these numbers from the KiZ Merkblätter. They are functions from other parameters defined by a law and published in the Existenzminimumbericht.
> 
> Or is there a specific source for this array?

_Originally posted by @mjbloemer in https://github.com/iza-institute-of-labor-economics/gettsim/pull/69#issuecomment-551876768_

PR #111, opened 2019-11-23
Split up params.yaml

### LINKED ISSUE #79: Split up param.yaml by group and rename groups

MaxBlesch:

The data of all policies is stored in params.yaml. We want to split params.yaml according to the group for each variable. We want to handle every file in a own dictionary and therefore call every function only with the needed data! This PR will address the second part of https://github.com/iza-institute-of-labor-economics/gettsim/issues/79#issuecomment-555230502 and therefore closes #79 .
Also I will include

### Todo

- [x] Split up params.yaml
- [x] Adjust loading function.
- [x] Rewrite tax transfer functions and tests to handle new design. 

I also include #86 here and therefore the following will be done:

- [x] Convert `pensions.xlsx` to YAML format
- [x] Adjust data loading whereever we call `pd.read_excel`
- [x] Remove `xlrd` as a dependency in `environment.yml` and `meta.yaml`
- [x] Remove lines 28 and 29 (starting with `# household` and  `# wg.to_excel`) in wohngeld.py


MaxBlesch:
I agree, that `_params` sounds better than `_data`. Can you clarify, what you mean with just start with an underscore? So instead of `kindergeld_data`, just `_kindergeld`? I wouldn't like that, as this is not very self-explanatory to me!

MaxBlesch:
After this PR I would finish #86 and then rename all functions according to our new naming convention in https://github.com/iza-institute-of-labor-economics/gettsim/issues/79#issuecomment-554283492 . Or should do that all in this PR?

hmgaudecker:
Yes, that is what I mean. Given that we will use that structure everywhere, it will probably be obvious.

Particularly if you call by value and call the argument that is now `tb` `params`.

E.g., 

```python
def kindergeld(household, params):
  ...
  return out
```

and call it with 

```python
kindergeld(household=whatever, params=_kindergeld)
```

Not perfectly sure of this, though, and a `_params` suffix would be easier to rename.

hmgaudecker:
> After this PR I would finish #86 and then rename all functions according to our new naming convention in https://github.com/iza-institute-of-labor-economics/gettsim/issues/79#issuecomment-554283492 . Or should do that all in this PR?

I would suggest including #86 in this PR. The rest can wait for a little bit.

MaxBlesch:
I think we can move to your solution once we have the interface, that we have a dataset(household, tax unit, person) and paramters. As of now some functions, 
e.g. `ui` https://github.com/iza-institute-of-labor-economics/gettsim/blob/c64cc0dd4fcccdfe9c531cc76a30c35643a70aa9/gettsim/benefits/arbeitslosengeld.py#L6
need more than one object.

hmgaudecker:
We will never be able to fully avoid that we need multiple groups of parameters. In this case, the interface would just stay in place.

(this said, `ui` should really be called with a parameter `alg_entgelt`, separate concerns!)

MaxBlesch:
I changed now to `_params` or should I whenever there is just one params input, change to the general params formultion. I think, as it is now, it is more self-explanatory!

hmgaudecker:
We need to distinguish between cases here.

- Where we load the stuff and call the functions, we should use the "distinguishing" notation with the group names, whether we append `_params` or prepend `_` or do something else remains to be seen; fine to do the first for now.
- Within functions, `tb` gets replaced by plain `params` for the group the function belongs to according to #79. When we need stuff from other groups, disambiguate. This is far too long variable names now with little benefit, I can see somebody jumping up and down in Munich already :-)

Sorry, I thought that we were only talking about the first item above.

MaxBlesch:
I hope this solution is what you thoght! @hmgaudecker

gettsim/pensions.py
```diff
@@ -21,23 +18,23 @@ def pensions(person, tb, tb_pens):
 
     """
     # meanwages is only filled until 2016
-    yr = min(tb["yr"], 2016)
+    year = min(params["year"], 2016)
```
hmgaudecker: Raise an ImplementationError if year > 2016 rather than silently doing something wrong.

gettsim/pensions.py
```diff
@@ -104,63 +106,63 @@ def _rentenwert_from_2018(tb_pens, yr):
     # Hence, some calculations have been made
     # in the data preparation.
     # First the Lohnkomponente which depands on the wage development of last years.
```
hmgaudecker: Delete these comments (do not state the obvious)

gettsim/policy_for_date.py
```diff
@@ -19,55 +19,101 @@
...
+    for key in raw_pension_data:
+        data_years = list(raw_pension_data[key]["values"])
+        # For calculating pensions we need demographic data up to three years in the
+        # past.
+        for year in range(min_year - 3, min_year + 1):
+            past_data = [x for x in data_years if x.year <= year]
+            if not past_data:
                 # TODO: Should there be missing values or should the key not exist?
-                tax_data[key] = np.nan
+                pension_data[f"{key}_{year}"] = np.nan
```
hmgaudecker: Why store this stuff with such weird keys and not in DataFrames?

gettsim/social_insurance.py
```diff
@@ -26,185 +24,193 @@ def soc_ins_contrib(person, tb):
...
 
     # As there is only one household, we selcet west_ost dependent paramter in the
-    # beginning and place them in a seperate dictionary tb_ost.
+    # beginning and place them in a seperate dictionary params_ost.
     westost = "o" if person["east"] else "w"
-    tb_ost = {}
+    params_ost = {}
     for val in ["bezgr_", "mini_grenze", "kvmaxek", "rvmaxek"]:
-        tb_ost[val] = tb[val + westost]
+        params_ost[val] = params[val + westost]
```
hmgaudecker: See above. `f"{val}{westost}"`

gettsim/social_insurance.py
```diff
@@ -26,185 +24,193 @@ def soc_ins_contrib(person, tb):
...
     # Calculate accordingly the ssc
     if belowmini:
         person[OUT_COLS] = 0.0
     elif in_gleitzone:
-        # TODO: Before and in 2003 tb["midi_grenze"] is 0 and therefore we won't reach
-        #  this.
-        person = tb["calc_midi_contrib"](person, tb)
+        # TODO: Before and in 2003 params["midi_grenze"] is 0 and
+        #  therefore we won't reach this.
+        person = params["calc_midi_contrib"](person, params)
```
hmgaudecker: Do we store a function in `params` or how does this construct work?

gettsim/social_insurance.py
```diff
@@ -26,185 +24,193 @@ def soc_ins_contrib(person, tb):
     # a couple of definitions
 
     # As there is only one household, we selcet west_ost dependent paramter in the
-    # beginning and place them in a seperate dictionary tb_ost.
+    # beginning and place them in a seperate dictionary params_ost.
     westost = "o" if person["east"] else "w"
-    tb_ost = {}
+    params_ost = {}
```
hmgaudecker: This construct is hard to understand. Why call that `params_ost` if it is west or east???

gettsim/social_insurance.py
```diff
@@ -26,185 +24,193 @@ def soc_ins_contrib(person, tb):
...
+        )
     else:
         return grbetr_pv - ag_pvbeit
 
 
-def no_midi(person, tb):
+def no_midi(person, params):
     """Dummy function returning nothing
     """
     person[OUT_COLS] = 0.0
```
hmgaudecker: Do not rely on globals. Why not use `pass` ?

gettsim/tax_transfer.py
```diff
@@ -510,9 +542,96 @@ def _apply_squeeze_function(group, tax_func, level, func_args, func_kwargs):
...
-def calculate_tax_and_transfers(dataset, policy_year):
-    tax_data = get_policies_for_date(year=policy_year)
-    tax_data_pensions = pd.read_excel(ROOT_DIR / "data" / "pensions.xlsx").set_index(
-        "var"
+def calculate_tax_and_transfers(
+    dataset,
+    year,
+    day=1,
+    month=1,
+    arbeitsl_geld_2_params=None,
```
hmgaudecker: I would like to avoid these defaults at this point. This is our user-facing function for the moment. Let us not mess with it and leave the interface in place unless there are reasons I missed.

gettsim/taxes/calc_taxes.py
```diff
@@ -2,7 +2,7 @@
 import pandas as pd
 
 
-def tax_sched(tax_unit, tb):
+def tax_sched(tax_unit, params, e_st_abzuege_params, soli_st_params, abgelt_st_params):
```
hmgaudecker: Here it probably should be `e_st_params`. It is only separate from `e_st_abzuege_params` because otherwise it would be far too big.

gettsim/taxes/zve.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 
 
-def zve(tax_unit, tb):
+def zve(tax_unit, params, soz_vers_beitr_params, kindergeld_params):
```
hmgaudecker: Again, maybe use `e_st_abzuege_params` because the groups are not so obvious here.

gettsim/taxes/zve.py
```diff
@@ -126,11 +130,13 @@ def kinderfreibetrag(tax_unit, tb):
...
     else:
-        tax_unit.loc[~tax_unit["child"], "kifreib"] = 0.5 * tb["kifreib"] * child_num_kg
+        tax_unit.loc[~tax_unit["child"], "kifreib"] = (
+            0.5 * params["kifreib"] * child_num_kg
+        )
         return tax_unit
 
 
-def zve_nokfb(tax_unit, tb):
+def zve_nokfb(tax_unit, params):
```
hmgaudecker: Do we use `params`? If not, delete unless there are hardcoded numbers in there that we will remove soon (I only see the diff without the body).

gettsim/benefits/alg2.py
```diff
@@ -50,29 +50,37 @@ def regelsatz_alg2(household, tb):
     children_age_info["child_num"] = household["child"].sum()
     children_age_info["adult_num"] = len(household) - children_age_info["child_num"]
 
-    household["mehrbed"] = mehrbedarf_alg2(household, children_age_info, tb)
+    household["mehrbed"] = mehrbedarf_alg2(household, children_age_info, params)
     # 'Regular Need'
     # Different amounts by number of adults and age of kids
-    # tb['rs_hhvor'] is the basic 'Hartz IV Satz' for a single person
+    # arbeitsl_geld_2_params['rs_hhvor'] is the basic 'Hartz IV Satz' for a single person
```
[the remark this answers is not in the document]
MaxBlesch: Deleted!

gettsim/benefits/arbeitslosengeld.py
```diff
@@ -1,45 +1,66 @@
...
+    e_st_abzuege_params,
+    e_st_params,
+    soli_st_params,
+):
     """ Calculating the claim for the Arbeitslosengeld, depending on the current
     wage."""
     westost = "o" if person["east"] else "w"
     # Relevant wage is capped at the contribution thresholds
-    alg_wage = np.minimum(tb["rvmaxek" + westost], person["m_wage_l1"])
+    alg_wage = min(soz_vers_beitr_params["rvmaxek" + westost], person["m_wage_l1"])
```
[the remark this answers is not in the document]
MaxBlesch: Changed!

gettsim/benefits/arbeitslosengeld.py
```diff
@@ -1,45 +1,66 @@
...
 
     # We need to deduct lump-sum amounts for contributions, taxes and soli
-    alg_ssc = tb["alg1_abz"] * alg_wage
+    alg_ssc = params["alg1_abz"] * alg_wage
     # assume west germany for this particular calculation
```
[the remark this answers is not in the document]
MaxBlesch: Deleted!

gettsim/benefits/benefit_checks.py
```diff
@@ -77,7 +77,7 @@ def benefit_priority(household, tb):
     return household
 
 
-def wealth_test(household, tb):
+def wealth_test(household, arbeitsl_geld_2_params):
```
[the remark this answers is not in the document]
MaxBlesch: Changed to params!

gettsim/social_insurance.py
MaxBlesch: The global is defined in the beginning of the module and is just the five social security contributions.

hmgaudecker: Then at least fix the docstring.

gettsim/tax_transfer.py
MaxBlesch: I included day and month, because the code works for every date in the year, so I thought why should we just allow for the year and not allow for every date. The other defaults are if anybody wants to define own policy data. I can delete if you like....

gettsim/taxes/calc_taxes.py
MaxBlesch: Changed!

gettsim/taxes/zve.py
MaxBlesch: It is in there because we assign the function in the dictionary and therefor I think every possible function needs the same interface. `zve_abg_nokfb`uses params.

gettsim/taxes/zve.py
MaxBlesch: Changed

gettsim/tax_transfer.py
hmgaudecker: Yes, please do. We do not have a way for selecting things other than Jan 1 for the moment. So we should not give the user the impression that such a thing might work.

Same for the rest. Setting up an interface for reforms will be complex, see for example here for OpenFisca's approach: https://openfisca.org/doc/openfisca-python-api/reforms.html. We should not just add something because it is easy. 

For now, our scope is to provide an accurate depiction of the current (and past) Taxes and Transfers system. Next, an accurate fast depiction. Then allow users to modify.

gettsim/taxes/zve.py
hmgaudecker: I do not understand this comment.

hmgaudecker: Which dictionary, where?

gettsim/social_insurance.py
MaxBlesch: Yes. In Juli when we sat together with Eric, you said that we should get this year checks out of the function and assign externally functions to the dictionary depending on the year!

gettsim/benefits/kiz.py
```diff
@@ -99,24 +101,26 @@ def kiz(household, tb):
...
+            return params["a2kiz_minek_cou"]
     else:
         return 0
 
 
-def calc_kiz_ek(household, tb):
-    if tb["yr"] <= 2010:
+def calc_kiz_ek(household, params, arbeitsl_geld_2_params):
+    if params["year"] <= 2010:
         # not yet implemented
```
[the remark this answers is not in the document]
MaxBlesch: Deleted comment! It is implemented!

gettsim/social_insurance.py
hmgaudecker: Haha, that's why we need those GEPs...

gettsim/taxes/zve.py
hmgaudecker: Got it.

gettsim/tax_transfer.py
MaxBlesch: I deleted it. But @mjbloemer wanted to update the dates in the yaml files and policy for dates would actually select on the date. But you are right: We would pretend to be daily accurate but actually we are not!

gettsim/social_insurance.py
MaxBlesch: Reformulated docstring!

gettsim/social_insurance.py
MaxBlesch: So true!!!!

gettsim/benefits/unterhaltsvorschuss.py
```diff
@@ -1,17 +1,17 @@
-def uhv(tax_unit, tb):
+def uhv(tax_unit, params):
     """
     Since 2017, the receipt of this
```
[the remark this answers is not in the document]
MaxBlesch: I'll open an issue, that we should rename the docstring. Also I think we should define some standard rules for docstrings. The docsrtings right now are all there from the beginning. As we don't have this rules I actually didn't change any so far. But I edited at least the description of the output for now!

gettsim/data/abgelt_st.yaml
```diff
@@ -0,0 +1,14 @@
...
+    en:
+  description:
+    de:
+    en:
+  values:
+    1984-01-01:
+      value: 0
+      note:
+    2009-01-01:
+      value: .25
```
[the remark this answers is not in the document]
MaxBlesch: I'll add an issue tomorrow. Maybe we could collect there some general things we want to have in the yaml files. We do also have not that many descriptions yet. To be user friendly we need to work quite a bit on that. Also self-explanatory variables...

gettsim/benefits/wohngeld.py
```diff
@@ -181,15 +179,15 @@ def calc_wg_gross_income(household):
     return out
 
 
-def calc_wg_income_deductions(household, tb):
-    if tb["yr"] < 2016:
-        wg_incdeduct = _calc_wg_income_deductions_until_2015(household, tb)
+def calc_wg_income_deductions(household, params):
+    if params["year"] < 2016:
```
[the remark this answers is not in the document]
MaxBlesch: Changed!

gettsim/data/e_st.yaml
```diff
@@ -0,0 +1,1013 @@
+G:
```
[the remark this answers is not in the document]
MaxBlesch: Don't understand this. What is repeated?

gettsim/pensions.py
MaxBlesch: Deleted!

gettsim/policy_for_date.py
MaxBlesch: I was trying to avoid DataFrames because then I would need to use pandas?! I am really not sure about this and had my doubts as well. I thought this would be an elegant way...

gettsim/tax_transfer.py
hmgaudecker: Yes, of course we will eventually do this. But we do not at present.

gettsim/data/abgelt_st.yaml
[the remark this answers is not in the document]
hmgaudecker: Yes, though the latter will be for after GEP01. Will update that in a week or so.

gettsim/pensions.py
MaxBlesch: I can do that, but this will be a longer issue. I will need to adapt a lot of tests and etc.

gettsim/data/e_st.yaml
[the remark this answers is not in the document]
hmgaudecker: Forget about it. At a glance I thought `M` / `t_m` pairs etc. were the same. Will become clearer once we move to an array structure for the parameters as discussed at some point.

gettsim/pensions.py
hmgaudecker: As I said, then just open an issue for it. This is orthogonal to the PR anyhow.

gettsim/policy_for_date.py
hmgaudecker: Sort-of third Normal Form for data management: "No structure in variable names" :-)

gettsim/data/arbeitsl_hilfe.yaml
```diff
@@ -0,0 +1,145 @@
+ahsatz0:
+  name:
+    de: Höhe der Arbeitslosenhilfe im Verhältnis zum letzten Nettoentgelt bei Personen ohne Kinder im Sinne des EStG
+    en:
+  description:
+    de: § 195 SGB III bis 31.12.2004
+    en:
+  values:
+    1984-01-01:
+      value: .5600000000000001
```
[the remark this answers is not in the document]
MaxBlesch: Deleted this!

gettsim/data/e_st.yaml
```diff
@@ -0,0 +1,1013 @@
...
+t_b1:
+  name:
+    de:
+    en:
+  description:
+    de:
+    en:
+  values:
+    1984-01-01:
+      value: -73.76000000000001
```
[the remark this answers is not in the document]
MaxBlesch: Seems to be an export error! Deleted it!

gettsim/social_insurance.py
MaxBlesch: Changed!

gettsim/social_insurance.py
MaxBlesch: It saves east and west specific variables. I could either store a new value in the dictionary, change the name of the dictionary, just take westost with me or do the check every time!

hmgaudecker: Just leave for now and make an issue out of it that we should find a general treatment for east/west differences. Probably store these parameters always in arrays. Also see my comment on #75 here: https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/Plan.20until.20Christmas/near/181733445

gettsim/benefits/unterhaltsvorschuss.py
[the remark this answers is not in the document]
MaxBlesch: Opened #113

gettsim/data/abgelt_st.yaml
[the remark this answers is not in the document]
MaxBlesch: I added the leading zeros, so no need to open an issue!

gettsim/pensions.py
MaxBlesch: Opened #114

gettsim/social_insurance.py
MaxBlesch: Opened #115

gettsim/policy_for_date.py
MaxBlesch: Opened #116

PR #117, opened 2019-11-24
Correct kiz income deduction

### LINKED ISSUE #112: Correct Kiz income deduction
### Bug description

As of now, kids' income is summed up and subtracted from kiz claim.

### Correction

First, deduct children income (and alimony) individually for each child. From the rest, subtract parent income.

Eric-Sommer:

closes #112.

### Todo

- [x] correct income deduction. First the kids income, then parent income
- [x] introduce income share as parameter


gettsim/benefits/kiz.py
```diff
@@ -89,15 +96,11 @@ def kiz(household, params, arbeitsl_geld_2_params, kindergeld_params):
...
-        params["a2kiz"] * household["child_num_kg"]
-        - household["kiz_ek_anr"]
-        - household["uhv_tu"],
+        0, household["kiz_childinc_deducted"].sum() - household["kiz_ek_anr"]
     )
     household["kiz_temp"] = household["kiz"].max()
-    # Transfer some variables for eligibility check
-    # kiz["ar_base_alg2_ek"] = household["ar_base_alg2_ek"]
-    # kiz["n_pens"] = household["pensioner"].sum()
+    print(household[["kiz", "kiz_childinc_deducted", "kiz_ek_anr", "uhv", "child_kg"]])
```
mjbloemer: this is probably some leftover

Eric-Sommer:
> The changes for the correction of child income deduction look good to me. I did not test it.

I add your example with two kids earning 180 and 120€, respectively, to the tests.

gettsim/benefits/kiz.py
```diff
@@ -68,36 +67,37 @@ def kiz(household, params, arbeitsl_geld_2_params, kindergeld_params):
...
+    # (§6a (6) S. 3 BKGG)
+    household["kiz_ek_anr"] = np.maximum(
+        0,
+        params["a2kiz_withdrawal_rate"]
+        * (household["ar_alg2_ek_hh"] - household["kiz_ek_relev"]),
+    )
     # Dummy variable whether household is in the relevant income range.
     household["kiz_incrange"] = (
         household["kiz_ek_gross"] >= household["kiz_ek_min"]
     ) & (household["kiz_ek_net"] <= household["kiz_ek_max"])
```
mjbloemer: Is there a reason not to use `alg2_grossek_hh` and `ar_alg2_ek_hh` directly here? I see no other occurrences of `kiz_ek_gross` and `kiz_ek_net` anywhere.

In that case you can drop:
```python
    household["kiz_ek_gross"] = household["alg2_grossek_hh"]
    household["kiz_ek_net"] = household["ar_alg2_ek_hh"]
```

Eric-Sommer: This is how it was done before. I find it way more intuitive this way.

PR #118, opened 2019-11-24
Fix Windows on Azure.

tobiasraabe:

The tests on Windows did not run and reported success. See git push --set-upstream origin fix-windows-on-azure for an example.

The error was introduced in #100 when I changed from `pip install -e` to `conda develop .`. Apparently, the latter causes trouble, but I do not know why.

Azure is nice because it provides more resources than other providers, but the lack of an official support for conda is just annoying. In the long-run we should keep an eye on Github Actions which also support 10 simultaenous builds and also caches stages. This means, for example, that as long the environment is not changed, it is not recreated. This would speed up the test suite tremendously. Conda support is also premature at the moment (`https://github.com/actions/starter-workflows/issues/44`), but maybe this will change in future.

### DIFF
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -63,7 +63,7 @@ jobs:
 
   - script: |
       call activate gettsim
-      conda develop .
+      pip install -e .
       pytest
 
 
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -30,7 +30,6 @@ jobs:
 
   - bash: |
       source activate gettsim
-      conda develop .
       pytest --cov-report xml --cov=./
 
   - bash: |
@@ -63,7 +62,6 @@ jobs:
 
   - script: |
       call activate gettsim
-      pip install -e .
       pytest
 
 
@@ -99,5 +97,4 @@ jobs:
 
   - bash: |
       source activate gettsim
-      conda develop .
       pytest


hmgaudecker:
I see. Can you try the setup from over at [econ-project-templates](https://github.com/hmgaudecker/econ-project-templates)? The setting should not be all that different and tests do seem to run: https://dev.azure.com/hmgaudecker/econ-project-templates/_build/results?buildId=126

Problem with pip is that it works for now, but as soon as we introduce Numba as a dependency (let alone CuDF) it will fail.

tobiasraabe:
@hmgaudecker works just fine without the editable install.

hmgaudecker:
I cannot quite follow (what is an editable install?).

tobiasraabe:
A normal installation creates an immutable copy of the package somewhere in the environment site-packages folder. An editable installation creates a link to the source code and every change to the source code is available after re-importing the package. The commands are ``pip install -e .`` or ``conda develop .``.

hmgaudecker:
Okay, so would we need the editable version for anything in the tests?

tobiasraabe:
I do not think so. I think at some point the editable installation solved a problem for me where the tests could not import modules from the package because it was not registered as one (but for another project). And since then, I went with this approach.

hmgaudecker:
Okay, so can we just switch to non-editable conda?

tobiasraabe:
Actually, I removed the installation step for ``gettsim`` completely. ``pytest`` still finds the package and all modules. So, I think we should just go with it.

tobiasraabe:
I was puzzled as well. I had the change on my disk and thought I pushed it :).

PR #119, opened 2019-11-25
Remove child specific ALG II withdrawal.

### LINKED ISSUE #110: Child income treatment at ALG II
The current modelling of `einkommensanrechnungsfrei(household, tb)` replaces the `ekanrefrei` for children as follows:

````python
    # Children income is fully deducted, except for the first 100 €.
    household.loc[(household["child"]), "ekanrefrei"] = np.maximum(
        0, household["m_wage"] - 100
    )
````

Two issues here:

1. Say the Bedarfsgemeinschaft has a child with `m_wage=80` (`m_wage=0` for other household members), then the `ekanrefrei` would be 0. If there was a 100 Euro allowance and nothing more I think it should be 80. With `m_wage=120` for the child it should be 100. I think in the test data there is a household with a child `m_wage=200` which leads to the correct 100 euro anrechnungsfrei by accident so pytest passes.
2. What is the exact source for this rule? It should be added as a comment to the code and maybe the number 100 should be added as an additional parameter. Or is it just the "Grundfreibetrag" of 100 Euro, currently `a2grf` defined in § 11b (2) 1? If it is the identical parameter, why do the second and third intervals not apply for children too?

### PR DESCRIPTION
**Pull request by mjbloemer**

As discussed in #110, the current modelling of the child income specific ALG II withdrawal rule is wrong. This PR completely removes this rule. Closes #110.

### CODE REVIEW

**Comment by hmgaudecker** [comment]
Just linking relevant discussion: https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/alg2/near/181822183

**Comment by MaxBlesch** [comment]
I am already on it. Just will not do it today. But as this is such a minor change it isn't really a rush?!

**Comment by mjbloemer** [comment]
> I am already on it. Just will not do it today. But as this is such a minor change it isn't really a rush?!

no rush :)

**Comment by MaxBlesch** [comment]
I changed the calculation of `e_anr_frei` to an individual level. This seems now a lot more readable. Once this is merged we can tackle #87

File: gettsim/benefits/alg2.py [hunk 1]
```diff
@@ -221,58 +221,47 @@ def grossinc_alg2(household):
...
+    household.loc[:, cols] = household.groupby("pid")[cols].apply(
+        e_anr_frei_person, params, top_limit_2nd_interval
     )
 
     return household
+
+
+def e_anr_frei_person(person, params, top_limit_2nd_interval):
+    """Calculates the amount of income that is not deducted for each person."""
+    m_wage = person["m_wage"].iloc[0]
```
tobiasraabe: Maybe we can avoid this chained indexing. What is ``m_wage``? Why is it save to take just the first entry for each person?

And even though this function is straightforward and beautifully structured, ``apply`` is notoriously slow. Is there an easy way to write this without the groupby and apply? I see no real aggregation.

Eric-Sommer: I also don't see the benefit here. The calculation is on the individual level anyway and makes distinctions based on individual and household characteristics. Why an additional loop?

gettsim/benefits/alg2.py
```diff
@@ -221,58 +221,51 @@ def grossinc_alg2(household):
...
-        0, household["m_wage"] - 100
+def e_anr_frei(household, params):
+    """Calculate income not subject to transfer withdrawal for the household.
+
+    Determine the gross income that is not deducted. Withdrawal rates depend
+    on monthly earnings and on the number of children in the household. § 30 SGB
+    II. Since 01.04.2011 § 11b."""
+
+    # Calculate the number of children below the age of 18.
+    num_childs_0_18 = (household["child"] & household["age"].between(0, 18)).sum()
```
mjbloemer: does `.between(0, 18)` include 18? The law is
>  die entweder mit mindestens einem minderjährigen Kind in Bedarfsgemeinschaft leben oder die mindestens ein minderjähriges Kind haben

cf https://www.buzer.de/gesetz/2602/a170599.htm

gettsim/benefits/alg2.py
MaxBlesch: I did write an extra function on the individual level, because it allows to distinguish very clear the intervals of deduction. This is the only benefit. Otherwise we would slice the DataFrame everytime and then assign. Would you know a way, which combines readability and the fastness of slicing?

MaxBlesch: person is always a one row Dataframe or a series. Therefore `person["m_wage"]`only contains one element.

gettsim/benefits/alg2.py
MaxBlesch: It is `<=`. We need minderjährig? Then I change to `.between(0, 17)`

gettsim/benefits/alg2.py
tobiasraabe: Some thoughts:

1. You cannot get away with the indentation on the right-hand-side of the equation, so we focus on the left-hand-side.
2. I would write each condition on a separate line like ``mask = ....`` and then ``df.loc[mask]``
3. ``child0_18_num > 0`` and ``child0_18_num == 0`` become ``has_minor/underage_children`` and ``~has_minor_children`` which is easier to read.
4. I am a little bit torn regarding this: ``household["m_wage"].between(params["a2eg1"], params["a2eg3"])``. On one hand it uses the designated pandas syntax, but on the other hand ``params["a2eg1"] <= household["m_wage"] <= params["a2eg3"]`` seems more readable to me in this instance.

gettsim/benefits/alg2.py
```diff
@@ -221,58 +221,47 @@ def grossinc_alg2(household):
...
-        0, household["m_wage"] - 100
+    num_childs_0_18 = (household["child"] & household["age"].between(0, 18)).sum()
+
+    if num_childs_0_18 == 0:
+        top_limit_2nd_interval = params["a2eg2"]
+    else:
+        top_limit_2nd_interval = params["a2eg3"]
+
+    cols = ["m_wage", "ekanrefrei"]
+    household.loc[:, cols] = household.groupby("pid")[cols].apply(
```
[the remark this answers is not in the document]
hmgaudecker: Does that construct work when `household` is a slice of a bigger df? I.e., does it fall back on in-place assignment instead of creating or replacing a column?

gettsim/benefits/alg2.py
hmgaudecker: Best `< 18`. Who knows if we ever work at a monthly level and end up with fractions.

gettsim/benefits/alg2.py
hmgaudecker: Don't worry much about speed here. [Premature optimization...](https://stackify.com/premature-optimization-evil/) :-). Will get there soon.

Eric-Sommer: Speed is one issue. But we have individual assignment within the household all the time. See calculation of taxable income, or wohngeld income test. Do we want to loop over `pid` there as well? And remove them when *actually* optimize?

hmgaudecker: And yes, if we are already debating what "between" actually means, it is a useful thing to discard. In general, please avoid the less obvious Pandas constructs; remember we plan to also use CuDF and that does not implement the full set of Pandas methods.

MaxBlesch: Sorry, but whats the conclusion now? Leave it as implemented or change to Tobias sugestion. Could you do a final word @hmgaudecker ?

gettsim/benefits/alg2.py
MaxBlesch: Done

gettsim/benefits/alg2.py
```diff
@@ -221,58 +221,47 @@ def grossinc_alg2(household):
     )
 
 
-def einkommensanrechnungsfrei(household, params):
-    """Determine the amount of income that is not deducted. Varies withdrawal rates
-    depending on monthly earnings."""
+def e_anr_frei(household, params):
+    """Determine the amount of income that is not deducted. Withdrawal rates
```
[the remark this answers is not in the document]
MaxBlesch: Done by @mjbloemer

gettsim/benefits/alg2.py
```diff
@@ -221,58 +221,47 @@ def grossinc_alg2(household):
...
-    )
-    # Children income is fully deducted, except for the first 100 €.
-    household.loc[(household["child"]), "ekanrefrei"] = np.maximum(
-        0, household["m_wage"] - 100
+    num_childs_0_18 = (household["child"] & household["age"].between(0, 18)).sum()
+
+    if num_childs_0_18 == 0:
+        top_limit_2nd_interval = params["a2eg2"]
+    else:
+        top_limit_2nd_interval = params["a2eg3"]
```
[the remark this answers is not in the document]
MaxBlesch: Done

gettsim/benefits/alg2.py
hmgaudecker: Leave it as is. We will have a more generic way for these linear spline-type functions eventually (they just appear everywhere), I just would not worry at this point.

gettsim/benefits/alg2.py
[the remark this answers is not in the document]
tobiasraabe: Then, household is probably a view and it works but raises the infamous ``SettingWithCopyWarning``. The warning is independent of whether you use ``.loc[:`` or not

gettsim/tests/test_alg2.py
```diff
@@ -57,16 +57,13 @@ def input_data():
...
     df = year_data[INPUT_COLS].copy()
-    arbeitsl_geld_2_params = get_policies_for_date(year=year, group="arbeitsl_geld_2")
-    # if year <= 2010:
-    #     tb["calc_regelsatz"] = regelberechnung_until_2010
-    # else:
-    #     tb["calc_regelsatz"] = regelberechnung_2011_and_beyond
+    arbeitsl_geld_2_params = get_policies_for_date(
+        year=year, group="arbeitsl_geld_2", raw_group_data=arbeitsl_geld_2_raw_data
+    )
     for col in OUT_COLS:
```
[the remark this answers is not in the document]
MaxBlesch: Used this now, also in tax_transfer!

gettsim/benefits/alg2.py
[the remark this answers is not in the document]
MaxBlesch: Both raises at least in the test no such warning.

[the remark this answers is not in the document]
tobiasraabe: Yes, it is not clear whether ``groupby`` groups are copies or views but https://stackoverflow.com/questions/52711823 indicates they are most likely copies. Thus, no warning here.

My point was more general if you do not use groupby.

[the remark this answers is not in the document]
hmgaudecker: Are you sure, @tobiasraabe? My own model of how things work is different, but I am happy to adjust that! 

@MaxBlesch: Is this being called from a test that uses groupby apply on a bigger DataFrame? And did we not have code somewhere shutting warnings off?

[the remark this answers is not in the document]
MaxBlesch: It is called from `tax_transfer` test and there we have a larger dataset. This warning is not turned off. See https://github.com/iza-institute-of-labor-economics/gettsim/blob/issue110/tox.ini for the warnings turned off.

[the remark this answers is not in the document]
tobiasraabe:
```
In [1]: import pandas as pd

In [2]: import numpy as np

In [3]: df = pd.DataFrame(np.arange(25).reshape(5, 5))

In [4]: grouped = df.groupby(0)

In [5]: grouped.get_group(0).iloc[0, 0]
Out[5]: 0

In [6]: hex(id(df.iloc[0, 0]))
Out[6]: '0x15d491866a8'

In [7]: hex(id(grouped.get_group(0).iloc[0, 0]))
Out[7]: '0x15d491866c0'

In [9]: grouped.get_group(5).iloc[0, 0]
Out[9]: 5

In [10]: hex(id(df.iloc[1, 0]))
Out[10]: '0x15d49186750'

In [11]: hex(id(grouped.get_group(5).iloc[0, 0]))
Out[11]: '0x15d49186768'
```
The ``hex`` should be the memory address and it is different for the same number within the dataframe and the groupby. But, I am a little bit concerned whether this check is valid because the hex seem very similar.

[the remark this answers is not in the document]
hmgaudecker: This code will change; the tests show the current version works as it should. No need to spend a whole lot of effort on it now.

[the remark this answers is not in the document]
MaxBlesch: Okay! So I'll merge!

PR #131, opened 2019-12-14
proper modelling of Kinderfreibetrag

Eric-Sommer:
closes #104 

- [x] separate parameters for *sächliches Existenzminimum* and *BEA*.
- [x] numbers as in the law, not multiplied by two. `zve()` adjusted accordingly.

### DIFF
--- a/gettsim/data/e_st_abzuege.yaml
+++ b/gettsim/data/e_st_abzuege.yaml
@@ -20,7 +20,7 @@ kifreib_s_exm:
       note: Wert in Euro. Wert laut Gesetz 2102 DM.
     1996-01-01:
       value: 1601
-      note: Art. 1. G. v. 20.10.1995 BGBl. I S. 1250. Wert in Euro. Laut Gesetz 261 DM pro Monat.
+      note: Art. 1. G. v. 11.10.1995 BGBl. I S. 1250. Wert in Euro. Laut Gesetz 261 DM pro Monat.
     1997-01-01:
       value: 1767
       note: Wert in Euro. Laut Gesetz 288 DM pro Monat (ab 2000 3456 DM pro Jahr)
--- a/gettsim/data/e_st_abzuege.yaml
+++ b/gettsim/data/e_st_abzuege.yaml
@@ -8,25 +8,22 @@ kifreib_s_exm:
   values:
     1984-01-01:
       value: 110
-      note:
+      note: Wert in Euro. Der Betrag laut Gesetz ist 216 DM.
     1986-01-01:
       value: 635
-      note:
+      note: Art. 1 G. v. 26.06.1985 BGBl. I S. 1153. Wert in Euro. Der Betrag im Gesetz lautet 1242 DM.
     1990-01-01:
       value: 773
-      note: .
-    1994-01-01:
+      note: Wert in Euro. Wert laut Gesetz 1512 DM.
+    1992-01-01:
       value: 1049
-      note:
+      note: Wert in Euro. Wert laut Gesetz 2102 DM.
     1996-01-01:
       value: 1601
-      note: Art. 1. G. v. 20.10.1995 BGBl. I S. 1250
+      note: Art. 1. G. v. 20.10.1995 BGBl. I S. 1250. Wert in Euro. Laut Gesetz 261 DM pro Monat.
     1997-01-01:
       value: 1767
-      note: .
-    2000-01-01:
-      value: 1767
-      note: Art. 1 G. v. 28.12.1999 BGBl. I S. 2552
+      note: Wert in Euro. Laut Gesetz 288 DM pro Monat (ab 2000 3456 DM pro Jahr)
     2002-01-01:
       value: 1824
       note: Art. 1 G. v. 19.12.2000 BGBl. I S. 1790
@@ -61,7 +58,7 @@ kifreib_bea:
   values:
     2000-01-01:
       value: 774
-      note: Art. 1 G. v. 28.12.1999 BGBl. I. S. 2552
+      note: Art. 1 G. v. 28.12.1999 BGBl. I. S. 2552. Wert in Euro. Wert laut Gesetz 1512 DM.
     2002-01-01:
       value: 1080
       note:
--- a/gettsim/data/e_st_abzuege.yaml
+++ b/gettsim/data/e_st_abzuege.yaml
@@ -59,9 +59,6 @@ kifreib_bea:
     de: §32 (6) EStG, seit 2000.
     en:
   values:
-    1970-01-01:
-      value: 0
-      note: implizit
     2000-01-01:
       value: 774
       note: Art. 1 G. v. 28.12.1999 BGBl. I. S. 2552
--- a/gettsim/data/e_st_abzuege.yaml
+++ b/gettsim/data/e_st_abzuege.yaml
@@ -1,56 +1,76 @@
-kifreib:
+kifreib_s_exm:
   name:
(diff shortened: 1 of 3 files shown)


gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,76 @@
...
+  name:
+    de: Kinderfreibetrag. Betreuungs-, Erziehungs-, Ausbildungsfreibetrag.
+    en:
+  description:
+    de: §32 (6) EStG, seit 2000.
+    en:
+  values:
+    1970-01-01:
+      value: 0
+      note: implizit
```
mjbloemer: is that really necessary? I have a hunch that it would rather improve things for the reader to make this clear in the source code (e.g. add a year switch). Would the source code look too complicated if we did not (mis)use this parameter?

hmgaudecker: Agree. We need to have a standard way for dealing with parameters that are not around for certain years.

So rather have a check in the code for whether the parameter is there than to add zero.

gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,76 @@
...
     en:
   values:
     1984-01-01:
-      value: 221
+      value: 110
       note:
     1986-01-01:
-      value: 1270
+      value: 635
       note:
```
mjbloemer:
```suggestion
       note: Artikel 1 G. v. 26.06.1985 BGBl. I S. 1153. Wert in Euro. Der Betrag im Gesetz lautet 1242 DM.
```
see https://offenegesetze.de/veroeffentlichung/bgbl1/1985/34#page=1

add source (law) and DM to EUR calculation in the notes.

gettsim/taxes/zve.py
```diff
@@ -111,7 +111,11 @@ def kinderfreibetrag(tax_unit, params, kindergeld_params):
     nokfb_lower = tax_unit["zve_nokfb"].min()
 
     # Add both components for ease of notation.
-    kifreib_total = params["kifreib_s_exm"] + params["kifreib_bea"]
+    try:
+        kifreib_total = params["kifreib_s_exm"] + params["kifreib_bea"]
+        # 'kifreib_bea' does not exist before 2000.
+    except NameError:
+        kifreib_total = params["kifreib_s_exm"]
```
mjbloemer: I don't think this is good way to do these things:
* If you add a clear year condition in the source code it would be enough to look at the source code (no need to investigate the param files further).
* This type of check does not allow to model cases where a law becomes obsolete.
* This type of check does not allow to model cases where a law does not exist for a few years and has been introduced later (e.g. Soli only in 1991, 1992 and then again 1995).

Eric-Sommer: I think it's a way to deal with this particular issue. The alternative would be to introduce separate functions in `policy_for _date`, which again is stupid because there is no structural change in the law whatsoever. For larger changes like soli, one has to go the long way.

mjbloemer: This looks more like a quick hack to me. I would say it is a structural change.

What is wrong with adding it to `policy_for_date`? Is the problem that the `policy_for_date` way does not scale? (we will have a lot of this type of policy changes). What is wrong with a year switch? That would be easy to read and robust to the two other problems I mentioned.

Eric-Sommer: > What is wrong with adding it to `policy_for_date`? Is the problem that the `policy_for_date` way does not scale? (we will have a lot of this type of policy changes). 

Precisely. I can do a year switch although we wanted to avoid them.

hmgaudecker: In the long run, yes. For now, they are fine.

gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,70 @@
...
     en:
   description:
     de: § 32(6) EStG
     en:
   values:
     1984-01-01:
-      value: 221
-      note:
+      value: 110
+      note: Wert in Euro. Der Betrag laut Gesetz ist 216 DM.
```
mjbloemer: I find this in value of 216 DM in [B. v. 24.01.1984 BGBl. I S. 113](https://offenegesetze.de/veroeffentlichung/bgbl1/1984/5#page=1). That is just Bekanntmachung of EStG 1983. Not sure if this is for 1983 or 1984 or if this value is completely new.

gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,70 @@
...
+      value: 773
+      note: Wert in Euro. Wert laut Gesetz 1512 DM.
+    1992-01-01:
+      value: 1049
+      note: Wert in Euro. Wert laut Gesetz 2102 DM.
     1996-01-01:
-      value: 3203
-      note: .
+      value: 1601
+      note: Art. 1. G. v. 20.10.1995 BGBl. I S. 1250. Wert in Euro. Laut Gesetz 261 DM pro Monat.
```
mjbloemer:
```suggestion
      note: Art. 1. G. v. 11.10.1995 BGBl. I S. 1250. Wert in Euro. Laut Gesetz 261 DM pro Monat.
```

gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,70 @@
...
-      value: 1546
-      note: .
-    1994-01-01:
-      value: 2098
-      note:
+      value: 773
+      note: Wert in Euro. Wert laut Gesetz 1512 DM.
+    1992-01-01:
+      value: 1049
+      note: Wert in Euro. Wert laut Gesetz 2102 DM.
```
mjbloemer: I do not find the 2102 DM value.

https://offenegesetze.de/veroeffentlichung/bgbl1/1992/9#page=8 shows a few changes:
* replace "1512" by "2052",
* replace "3024" by "4104".

Should the DM value be 2052?

If so, please also add the BGBl. page to the note key.

gettsim/data/e_st_abzuege.yaml
```diff
@@ -1,56 +1,70 @@
...
+  description:
+    de: §32 (6) EStG, seit 2000.
+    en:
+  values:
+    2000-01-01:
+      value: 774
+      note: Art. 1 G. v. 28.12.1999 BGBl. I. S. 2552. Wert in Euro. Wert laut Gesetz 1512 DM.
+    2002-01-01:
+      value: 1080
       note:
```
mjbloemer: please add the note key

gettsim/taxes/zve.py
```diff
@@ -110,17 +110,24 @@ def kinderfreibetrag(tax_unit, params, kindergeld_params):
     # Find out who has the lower zve among partners
     nokfb_lower = tax_unit["zve_nokfb"].min()
 
-    diff_kifreib = nokfb_lower - (0.5 * params["kifreib"] * child_num_kg)
+    # Add both components for ease of notation.
+    if params["year"] >= 2000:
+        # 'kifreib_bea' does not exist before 2000.
```
mjbloemer: could you move this comment line before the if or before the else?

gettsim/data/e_st_abzuege.yaml
Eric-Sommer: No the DM values are correct. They refer to the couple though.

gettsim/data/e_st_abzuege.yaml
Eric-Sommer: I've looked for it for quite some time. No idea what changed it. It was first changed to 774€, but in the Neufassung from 2002 it's changed to 1080, meaning that it applies to the whole fiscal year.

PR #140, opened 2020-02-11
Generic functions

ppoepp:
- we **fixed errors** in YAML and generic_function
- we created a separate file with a **function to generate all three arrays**
- created a first version of a very simple **unit test**


gettsim/generic_functions.py
```diff
@@ -26,9 +26,10 @@ def fill_intercepts_at_lower_thresholds(
 
     intercepts_at_lower_thresholds = np.zeros(upper_thresholds.shape) * np.nan
     intercepts_at_lower_thresholds[0] = intercept_at_lowest_threshold
-    for i, up_thr in enumerate(upper_thresholds[1:]):
+    i = 1
```
hmgaudecker: Not required.

gettsim/generic_functions.py
```diff
@@ -26,9 +26,10 @@ def fill_intercepts_at_lower_thresholds(
...
+    for i, up_thr in enumerate(np.append(0, upper_thresholds[0:-1])):
```
hmgaudecker: This is a bit ugly. What is it supposed to do? 

Things like this should not be hard-coded. If we make too few assumptions somewhere else, we might need to add an explicit value for the domain of the function (>=0, ...  ?)

gettsim/generic_functions.py
```diff
@@ -57,6 +58,6 @@ def piecewise_linear(
 
     idx = np.searchsorted(upper_thresholds, value, side=side)
     intcpt = intercepts_at_lower_thresholds[idx]
-    out = intcpt + (value - intcpt) * rates[idx]
+    out = intcpt + (value - np.append(0, upper_thresholds)[idx]) * rates[idx]
```
hmgaudecker: Please avoid any use of np.append. It is slow and incomprehensible.

But this looks like we may want to introduce an explict 0 indeed if the domain is >= 0; and -inf if it is the real line. Then it is guaranteed that upper_thresholds[idx -1] is defined. We might have to special-case 0 in the former case, though.

gettsim/calc_e_anr_frei.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+import datetime
+
+import numpy as np
+import yaml
+from generic_functions import fill_intercepts_at_lower_thresholds
+from generic_functions import piecewise_linear
+
+
+def generate_arrays(dict_list):
```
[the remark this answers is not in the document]
MaxBlesch: Yes. We have right now the `policy_for_date` modules, for transferring the information in the yaml file to a policy given the desired date. Moving forward I suggest we use `generic_functions.py` as the corresponding module. Or setup an extra module. `calc_e_einr_frei.py` is definitely to specific!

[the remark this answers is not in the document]
hmgaudecker: Let us set up a dedicated module then.

ppoepp:
> Please add some tests with invalid input (some dict in the list missing a field, a string that is not convertible to a float, etc.) and provide some graceful error messages.


Thanks!  We will add the invalid test input next and try to find an alternative for np.append in the piecewise_linear function.

johannesgoldbeck:
> Please add some tests with invalid input (some dict in the list missing a field, a string that is not convertible to a float, etc.) and provide some graceful error messages.

I don't really understand, how we can achieve this with tests. Should we specify the error messages by writing exceptions into the function? Then we could test the outcome of the function with some flawed input for the correct Error message.

hmgaudecker:
The conversion function should raise a sensible error upon invalid input. Maybe you can draft a list of those first before writing code? Then we can discuss it first.

Then there should be one test for each way to trigger an error.

johannesgoldbeck:
How about checking for negative values and rates that are >1, besides missing values and unconvertables?

hmgaudecker:
Okay, let's see. Looking at it afresh, I think that the interface of `piecewise_linear` is too terse. We should make it clearer by adding `lower_thresholds`. Then `rates[i]` would correspond to the interval between `lower_thresholds[i]` and `upper_thresholds[i]`. That will make things much clearer; also when it comes to the range of the function. E.g., values < 0 might or might not be defined, we assume it implicitly so far. Implementation should be easier, too.

For the conversion function, we should thus require

1. The lowest `lower_threshold`.
2. The highest `upper_threshold`.
3. At least one of lower / upper threshold for everything in between (no mixing, always supply either upper or lower or both, but do not switch back and forth)
4. Either the lowest `intercept_at_lower_threshold` or all of them (but nothing in between).
5. All rates.

In the conversion, we should raise an error if

- Any of the above conditions is not supplied.
- Both `upper_threshold[i]` and `lower_threshold[i+1]` are supplied and they are not equal.
- The thresholds are not ordered (this is forcing discipline in the YAML file more than anything else).

Not sure about the restrictions on the rates. It is conceiveable that < 0 is how things are defined. > 1 probably does not make much sense. Anyhow, why don't you add them and we can always move them to specific conversions in case we decide at some point that they are too restrictive.

hmgaudecker:
@ppoepp & @johannesgoldbeck: Can you please incorporate the latest changes from this repo's `generic_functions` branch? 127143a842, feel free to disregard any of the changes starting in line 44.

johannesgoldbeck:
> 
> 
> Okay, let's see. Looking at it afresh, I think that the interface of `piecewise_linear` is too terse. We should make it clearer by adding `lower_thresholds`. Then `rates[i]` would correspond to the interval between `lower_thresholds[i]` and `upper_thresholds[i]`. That will make things much clearer; also when it comes to the range of the function. E.g., values < 0 might or might not be defined, we assume it implicitly so far. Implementation should be easier, too.
> 
> For the conversion function, we should thus require
> 
>     1. The lowest `lower_threshold`.
> 
>     2. The highest `upper_threshold`.
> 
>     3. At least one of lower / upper threshold for everything in between (no mixing, always supply either upper or lower or both, but do not switch back and forth)
> 
>     4. Either the lowest `intercept_at_lower_threshold` or all of them (but nothing in between).
> 
>     5. All rates.
> 
> 
> In the conversion, we should raise an error if
> 
>     * Any of the above conditions is not supplied.
> 
>     * Both `upper_threshold[i]` and `lower_threshold[i+1]` are supplied and they are not equal.
> 
>     * The thresholds are not ordered (this is forcing discipline in the YAML file more than anything else).
> 
> 
> Not sure about the restrictions on the rates. It is conceiveable that < 0 is how things are defined. > 1 probably does not make much sense. Anyhow, why don't you add them and we can always move them to specific conversions in case we decide at some point that they are too restrictive.

Ok, we are working on changing get_dict_of_arrays_piecewise_linear s.t. it is compatible with the new input structure.

But what exactly is the issue with piecewise_linear, in what sense does it need change?

hmgaudecker:
Add `lower_thresholds` to the interface. Then you can do something like:

```python
if (value < lower_thresholds[0]) or (value > upper_thresholds[-1]):
    out = np.nan
else:
    idx = np.searchsorted( ....)
    ... 
    out = ...
return out

You can get rid of that `np.append` - thing then.

**Comment by ppoepp** [comment]
[054785b](https://github.com/iza-institute-of-labor-economics/gettsim/pull/140/commits/054785b667609a4f9f686a4d24ea788940a49615)

Started to update `get_dict_of_arrays_piecewise_linear`. In the new YAML-structure, `references` is one of the keys that we need to get rid off first, before we can start to iterate. I am not really happy the way it's implemented at the moment, but accessing `rates` and `upper_thresholds` from the updated YAML-file works again. I think we can rewrite this more efficiently with fewer lines of code.

File: gettsim/generic_functions.py [hunk 5]
```diff
@@ -38,31 +38,41 @@ def fill_intercepts_at_lower_thresholds(
...
+        Args:
+            list_of_dicts: list_of_dicts specifying upper_thresholds and
+                  respective rates for a given legislature time.
+
+        Returns:
+            dict of 3 arrays for calc_e_anr_frei: upper_thresholds, rates and
+            intercepts_at_lower_thresholds.
+        """
+    keys_without_reference = list(list_of_dicts.keys())
+    keys_without_reference.remove("reference")
```
hmgaudecker: Sorry, my bad. Should have been a bit more verbose here. 

In the syntax of the YAML-file, I tried to be as close to a list as possible while maintaining the possibility to reference an element (see the examples below your main usecase; for kids or for the 2011 change). 

Let us please take the opposite route (picking the keys we want rather than discarding the irrelevant ones). So something like the following should do:

 ```python
keys = sorted([key for key in list_of_dicts.keys() if type(key) == int])
```

johannesgoldbeck:
I dont see how we can easily get rid of the np. append here.
I guess an array with all lower thresholds is basically: np.append(0, upper_thresholds[0:-1])
And if we have that, yes we can get rid of the append. But if the yaml file does not specify all lower thresholds, do we need to search for another way of creating this array with all lower thresholds?

hmgaudecker:
Again too terse, I see a pattern here :-) Feel free to drop by at some point to discuss in person, might be easier.

Anyhow, the way I thought about it is that the conversion function does everything that is required in order to make it simple on `piecewise_linear`.

In particular, we should generally assume that the function is defined everywhere between `lower_thresholds[0]` and `upper_thresholds[-1]`, so we can fill up `lower_thresholds[1:]` in the conversion function. Requiring the user to do that in the YAML file is too error-prone and tedious.

To be perfectly clear, the types of the input arguments to `piecewise_linear` should be as follows:

- `value` scalar float
- `lower_thresholds` 1-d float, size k
- `upper_thresholds` 1-d float, size k
- `rates` 1-d float, size k
- `intercepts_at_lower_thresholds` 1-d float, size k
- `side` ∈ {`left`, `right`}

johannesgoldbeck:
Done that. Is this a possible way to implement the lower_threshold array?
And what are important further steps besides writing new tests, that we already spoke of? implementing the piecewise_linear into alg2.py?

johannesgoldbeck:
So we don't really understand how to implement piecewise_linear into alg2.py to calculate ekanrfrei =(
How does the data specify the time, the wage was earned in?
And when does the Nettoquote need to be applied?

hmgaudecker:
(This is just from out of the top of my head, maybe @maxblesch can correct me if necessary)

- the year is part of the params dictionary
- I think you just need to pass Nettoquote * income into piecewise linear before 2015-10

johannesgoldbeck:
can we also access the dictionary defining the intervals through params["e_anr_frei"][values][data]?
I dont rly know how the objects "hosehold", "person" and "params" look like. Makes it hard to test the code as well.

hmgaudecker:
How about using the existing test suite and replacing the current functions with your implementation? Then set a breakpoint and inspect the objects.

(Sorry I cannot be of more help for the moment, only have my phone for checking)

PR #141, opened 2020-02-16
Move to GitHub Actions.

tobiasraabe:

There are several issues with Azure and Github Actions has some interesting features and develops faster.


hmgaudecker:
Looks good. A couple of things:

- I think we should run the pre-commits everywhere. As much as it is a test of the code, it is a test of whether setting up pre-commit works everywhere (and, in case of failure, fails everywhere). That is what brough Github Actions up in the first place thanks to @mjbloemer,.
- I set the secret, but CodeCov does not seem to upload anything yet? I did see a http 400 error but cannot find it anymore...
- Related: We should add to the docs on how to manually trigger the CI workflow and remove the corresponding instructions for Azure (while I am at it: I believe I saw the same docs section in estimagic, too).

tobiasraabe:
Try a commit with a failed build. The option is not available for successful workflows.

If it is a classic case of DRY, shouldn't we remove one of them? Maybe under "how to maintain" because the more narrow scope is "how to contribute"?

hmgaudecker:
1. Okay, please add that. "Note the option is only available for failed builds."
2. No, from a user perspective it is useful to have it there twice. In code, we would write a function. Here, a comment in both places noting the respective other one should do the job.

PR #146, opened 2020-02-24
Nicer formatting of header for GEPs.

hmgaudecker:

The header is ugly in the RTD Sphinx theme. This converts it into a table. See GEPs 1, 2, 3 for applications.


hmgaudecker:
Good catch! Could you please just go for it and approve once more?

Issue #89, opened 2019-11-13
Correct treatment of 'non-standard' households


Eric-Sommer:
### Bug description

As soon as a household does not fit into the four classical categories (Single, Single Parent, Couple, Couple with kids), things get complicated when it comes to the benefit system in particular. We haven't properly sorted this issue out yet. Examples include:

- [ ] how to treat the additional need for single parents in case there are more adults? In other words, how should a single parent be defined?
- [ ] Wohngeld and ALG2 apply different household concepts. You may have the core family eligible for ALG2, while there are other persons under the same roof eligible for Wohngeld (see [discussion](https://www.bbsr.bund.de/BBSR/DE/Veroeffentlichungen/Sonderveroeffentlichungen/2015/DL_Wohngeld.pdf?__blob=publicationFile&v=3), in particular section 5.4). This affects both the calculation of claims and a proper evaluation of different claims within in a household.
- [ ] The treatment of pensioners' eligibility for benefits is related to that. They are eligible to wohngeld, but not to ALG2, although "Grundsicherung im Alter" basically pays the same amount. Pensioners are also hardly covered in the tests as of now.

This issue can be solved by looking in particular at exemplary cases from the literature and applying them as test cases.

hmgaudecker:
Thanks. 
1. will be an interesting case to keep in mind when designing our data structures. I think that should be doable with the right structure on linkages between persons.
2. is tough. Is one a subset of the other at least? Else I do not see how we could ever get away without defining two different household ids. Not sure whether that is actually worth the pain. 
3. Same as 2. -- can you please edit towards "eligibility **for xxx**? 

Would be great if you could link some of the test cases you think are important.

hmgaudecker:
(I do not consider this a bug given that the state-of-the-art simulators do not seem to handle this correctly, either)

Eric-Sommer:
Yes, they are nested but apparently the other way round than I thought. In a nutshell, for ALG2, a household is the core family: mom, dad and dependent kids. The adult, working child under the same roof however counts to the same wohngeld household. As my link suggests, this is a growing phenomenon.

hmgaudecker:
(cold sweat disappearing... non-nestedness would have been **really** painful)

MImmesberger:
Partially closed by #694. All things that are not yet done are covered by more specific issues (thinking of #270 and #710).

PR #147, opened 2020-02-24
GEP 2 — Data structures

hmgaudecker:

Pin down the internal data structures.

So far, this is a very rough outline that is hopefully enough to serve as a basis for discussions.


docs/geps/gep-02.rst
```diff
@@ -0,0 +1,158 @@
...
+    We might get by without steuersubjekt (tax unit) by just tracing out relations
+    between people and eligibility; I leave it in for now.
+
+
+A: Strict adherence to 2nd normal form
+---------------------------------------
+
+* 3+ different tables (bedarfsgemeinschaft, steuersubjekt, person, ?).
+* Indexes would be ``jahr`` :math:`\times` {``bedarfsgemeinschaft``, ``steuersubjekt``,
+  ``person``}
```
Eric-Sommer: As it turned out, bedarfsgemeinschaft and steuersubjekt are essentially the same. So `haushalt`, `bedarfsgemeinschat`, `person` might be more appropriate

docs/geps/gep-02.rst
```diff
@@ -0,0 +1,158 @@
...
+* Very obvious what is the correct level meant
+* Never a need to append anything like ``_hh`` to a variable
+
+Pro B:
+
+* Most of the time, calculations at ``higher-up`` levels require information from the
+  person level (as opposed to aggregates only). E.g., for a lot of Arbeitslosengeld II -
+  stuff we have to check person-by person. One might as well just work with that
+  directly and avoid lots of merges.
+* More natural for people not used to the SQL-like way of doing things.
```
Eric-Sommer: Contra B: MultiIndices are not that straightforward to work with and complicate the syntax.

docs/geps/gep-02.rst
```diff
@@ -0,0 +1,158 @@
...
+* Index would be ``jahr`` :math:`\times` ``bedarfsgemeinschaft`` :math:`\times`
+  ``steuersubjekt`` :math:`\times` ``person``
+* Any variable relevant at a level higher than ``person`` will be filled in the same
+  way for all persons belonging to that unit.
+
+
+Pro A:
+
+* Very obvious what is the correct level meant
+* Never a need to append anything like ``_hh`` to a variable
```
Eric-Sommer: Why not? You'd still might calculate everything on person level first and the sum up.

docs/geps/gep-02.rst
hmgaudecker: Thanks for the reminder, should have looked up [that discussion](https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/Update.20Data.20Structures/near/180917151). `person` seems clear enough. Beyond that, my understanding so far: 

- `haushalt`
   - relevant for: `wohngeld`, ...  ? 
- `steuersubjekt` / `bedarfsgemeinschaft`
    - relevant for: `e_st`, `zu_verst_e`, `kindergeld`, `arb_los_geld_2`, `kinderzuschlag`, ...? 
    - 1 or 2 persons 
    - whether kids are relevant for `zu_verst_e` / `kindergeld` or not depends on age and income, but does not depend on being part of the `haushalt`
    - whether kids are relevant for transfers like  `arb_los_geld_2` and `kinderzuschlag` depends on age, income, and being part of the `haushalt`.

docs/geps/gep-02.rst
hmgaudecker: Fair enough, although I they would not be completely gone in A, either

docs/geps/gep-02.rst
hmgaudecker: ... and by summing up you got rid of the `person` dimension and just use a table that does not include it :-)

docs/geps/gep-02.rst
Eric-Sommer: Almost 100% agree, I edited your post.

hmgaudecker: No, I meant `haushalt`, if that is the only thing you changed. The thing is you have to live under the same roof, right?

Where this will lead to eventually, I think, is that users only specify the core adults for `steuersubjekt` / `bedarfsgemeinschaft` and we will figure out which kids to include in these things based on parent / child relationships, age, income, residence.

The problem is that these are endogenous to choices of many interesting models (labour supply of kids comes to mind), so it is essentially impossible for the user to specify them ex ante. How would you do that in a multinomial logit routine where different hours choices move a kid in or out being relevant for her parents' income tax statement?

Eric-Sommer: We'll have to check again in detail what the legal criteria are here. In IZAMOD, we don't care about kid's hours of work on the decision of whether or not they are in the `bedarfsgemeinschaft`. A kid is assinged to its parents' `bedarfsgemeinschaft` if it's below 18 or below 25 and in education. Hours of work /income matter for Kindergeld and Kinderfreibetrag, but you may have a child in the `bedarfsgemeinschaft` for which no kindergeld is paid. I don't see a major problem in allowing the `age <= 25` threshold to be flexible.

hmgaudecker: I do not quite understand what you are trying to get at, but maybe I was not clear, either. 

I think the basic problem is that this whole stuff is endogenous and we should be correct about it rather than relying on the user to make sure that a kid who earns too much next to being a student gets to be her own `steuersubjekt`. 

And yes, we should make things like that age threshold endogenous in the sense of moving it to the params files. These have changed in the past here and there, after all.

hmgaudecker:
Just started to update this and stopped right away. For the moment, it is clear that we use a dict of Series, each with an appropriate index. Whether we stick to that long-term seems too uncertain right now (e.g. #83 on relations between units, #89 on non-standard households, [related discussion Zulip](https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/Update.20Data.20Structures/near/257663837), speed, ...). 

So putting this to rest for the moment.

Issue #52, opened 2019-10-30
Monetary unit of parameters and data set


mjbloemer:
> > > I think every monetary value in param.xls is currently in Euro; DM values are converted to Euro (which makes it just a little bit harder when comparing to the written law).
> > 
> > 
> > Perfect. I have no strong opinion on whether we want to convert those values now, but eventually DM would seem to make more sense to me.
> 
> There is one exception, namely the tax tariff parameters before 2002. The others are in Euro; initially because SOEP delivers all incomes in Euro as well. I'd prefer to sticking to Euro all the way. If you do a long-term analysis, you'd report in Euro anyway. It's easier to convert incomes rather than parameters, as there are lots of parameters (shares) which must not be converted.

I think we are on the same page in the sense that internally, we will do everything in Euros. 

I disagree on parameters being more difficult to convert. That may be the case in Stata if you have to store all these animals in macros, but here? 

```python
def convert_dm_to_euro(dm):
    return dm / 1.95583

if unit == "Euro":
    val = raw_val
elif unit == "DM":
    val = convert_dm_to_euro(dm)
else:
   raise ValueError(f"Monetary Unit unknown: {raw_val}")
```

So this is only about how to store the input parameters and I think the idea of sticking close to the law is good there. As I wrote previously, whether we want to do that right away or do that at some later point (please open an issue in that case once we close this one, @mjbloemer lest we forget) I do not care much.

What we want to support in terms of input data is a wholly different question.

_Originally posted by @hmgaudecker in https://github.com/iza-institute-of-labor-economics/gettsim/issues/31#issuecomment-547781434_

Eric-Sommer:
Related to this: It might be worth to indicate whether a parameter is monetary or a (percentage) share or something else (e.g. an a age threshold)

hmgaudecker:
Also see the discussion here: https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/param/near/180365079

Eric-Sommer:
do we want to add this dimension in #148?

hmgaudecker:
Yes, I would say so. Can you add a note or start drafting a section there, enumerating the possible options?

hmgaudecker:
Might be a good thing for an RA not great in Python. @MaxBlesch, pls. keep in mind for 23/7

hmgaudecker:
[...]

PR #148, opened 2020-02-24
GEP 3 — Parameters of the Taxes and Transfers system.

hmgaudecker:

Pin down the handling of parameters specified by the law. This has been split off from GEP 1 (@ChristianZimpelmann was right, of course, that this should get its own GEP...).

So far, this is mostly about the YAML files. There is several things we have discussed over the last few weeks which I tried to synthesize here.

Relative to the discussion with @mjbloemer last Wednesday, I realised the `values` key next to `name` and `description` creates an unnecessary level of nesting. We can directly go into the dates. there.


docs/geps/gep-03.rst
```diff
@@ -0,0 +1,296 @@
...
+                upper_threshold: inf
+                rate: 0
+                note: Making this explicit.
+
+* Formulas: We want to allow for something like this if the law is specified that way
+  (e.g. the Steuertarif, IIRC):
+
+  .. code-block:: yaml
+
+      value: (2000 + 0.5 * x) * x
```
Eric-Sommer: not sure if this is such a good idea because it makes it more opaque where to look for a specific formula. Kind of an in-between case between parameter and policy structure. I see a use case though for soc sec contributions, where we use half the value in the law.

hmgaudecker: Sorry, this was a bit too terse. and in the wrong order. See the annotation below: 

    Will require that a function is specified in the ``type`` keyword so that we can get standard
    representations via SymPy or the like.

So this is just an alternative way to specify parameters for a formula that exists among our generic functions. E.g., we would not need to expand the example formula to get the coefficients of `a + b * x + c * x **2`. We would not use the formula directly, so the code stays the same.

docs/geps/gep-03.rst
```diff
@@ -93,7 +93,36 @@ entry is added.
                 Shares of income which do not lead to tapering of Arbeitslosengeld II
                 benefits.
 
-3. The YYYY-MM-DD key(s)
+3. The ``unit`` key informs on the unit of the values (Euro or DM if monetary, or some share).
```
MaxBlesch: I think Eric is right and we should add this information. I would use a different key though. I suggest we keep all that under the type key, which is, by the way, missing here so far. I can add it in the coming days. Possible values for type would then be: Euro, DM, share, percentage, piecewise...

hmgaudecker: I am afraid I do not follow your conversation. I do think we should add units, but I do not understand what the point of contention is / what the options are. Sorry...

MaxBlesch: Sorry, this is related to a private discussion on zulip. Thought this would wrap it up. So we have the `type` key. So far we only used it for piecewise functions. I suggest that we just keep this key and add there our information on units. But actually not a strong opinion. If you think we should have type, unit and time, I will adjust!

hmgaudecker: Thanks! 

I do have a strong opinion here: No key overloading, please! 

I.e., there is a lot of value in knowing exactly what is in a key without recurring to external information. If we were using the type key, we would need to know what kind of parameter we are talking about, which defies most of its purpose. 

Actually, we should make the unit an ENUM-style key (or categorical, in Pandas-speak), i.e., it should only be able to take on a number of specified values. E.g., DM, EUR, share, percent, ..., you name it.

hmgaudecker:
I just updated this. Looks pretty uncontroversial to me.

docs/geps/gep-03.rst
```diff
@@ -0,0 +1,319 @@
...
+        2:
+          upper_threshold: 1000
+
+
+The values of [YYYY-MM-DD]
+++++++++++++++++++++++++++
+
+The general idea is to make the replication of the laws very obvious. If the law
+includes a table, we will have a dictionary with keys 0, 1, 2, .... If the law includes
+a formula, the formula should be included and its parameters reference. Etc..
```
ChristianZimpelmann, line 216: referenceD?

ChristianZimpelmann (commented):
Looks great to me!

I made one comment and we also need to add the possibility to use ` access_different_date: vorjahr`. I will add a short paragraph about it.

ChristianZimpelmann:
> we also need to add the possibility to use ` access_different_date: vorjahr`. I will add a short paragraph about it.

I just pushed. As this possibility is only implemented in #275, make sure that this PR is merged before #148 or merge this part to main before (probably a bit complicated)

hmgaudecker:
The idea of these GEP's is to lay out general rules, less about a snapshot of current States. So do not worry about the order in which things will be merged.

This said, what do you think of the interface I propose here?

ChristianZimpelmann:
> This said, what do you think of the interface I propose here?

The parameter name would then be `xxx_vor_1y`? I am not sure if we will ever need something else than the previous year, but I like the new interface in general.

ChristianZimpelmann:
> xxx_t_min_1_y

Then I would go for `xxx_t_minus_1_y` as otherwise it is very hard to understand.

Let's go with it. Will update `grundrente` accordingly.

docs/geps/gep-03.rst
```diff
@@ -0,0 +1,342 @@
...
+        [...]
+      reference_period: week
+
+6. The (optional) ``access_prior_parameters`` can be used to make the parameter of a
+   previous point in time (relative to the date specified available in
+   ``set_up_policy_environment``) available within GETTSIM functions. It requires the
+   ``reference_period`` (one of ``y``, ``m``, ``w``, ``d``) and the ``number_of_lags``.
+
+   The lagged parameters will be available as
+   ``[param]_t_min_[number_of_lags]_[reference_period]`` next to ``[param]``.
```
ChristianZimpelmann, line 140:
```suggestion
   ``[param]_t_minus_[number_of_lags]_[reference_period]`` next to ``[param]``.
```

ChristianZimpelmann (approved):
just the suggestion mentioned above. Otherwise, it can be merged from my perspective.

docs/geps/gep-03.rst
```diff
@@ -0,0 +1,342 @@
...
+    kindergeld_stundengrenze:
+      name:
+        de: Wochenstundengrenze für Kindergeldanspruch
+        [...]
+      reference_period: week
+
+6. The (optional) ``access_prior_parameters`` can be used to make the parameter of a
+   previous point in time (relative to the date specified available in
+   ``set_up_policy_environment``) available within GETTSIM functions. It requires the
+   ``reference_period`` (one of ``y``, ``m``, ``w``, ``d``) and the ``number_of_lags``.
```
ChristianZimpelmann, line 137: And another thought: Should those be written out: ``year``, ``month``, etc.

hmgaudecker: I'd vote no because we would have to start thinking about singular / plural then...

hmgaudecker:
Cross-linking #52

docs/geps/gep-03.rst
```diff
@@ -0,0 +1,449 @@
...
+- In the YAML files, parameters may be specified as deviations from other values,
+  :ref:`see above <gep-3-deviation_from>`. All these are converted so that the relevant
+  values are part of the dictionary.
+- Similarly, values from other points in time (via ``access_prior_parameters``,
+  :ref:`see above <gep-3-access_prior_parameters>`) of ``[param]`` will be available as:
+  ``[param]_t_minus_[number_of_lags]_[reference_period[0].lower()]``.
+- Parameters for piecewise polynomials are parsed.
+- Parameters that are derived from other parameters are calculated (examples include
+  ``kinderzuschlag_max`` starting in 2021 or calculating the phasing in of
+  ``vorsorgeaufw_alter`` over the 2005-2025 period).
```
ChristianZimpelmann, line 432: Also mention rounding parameters again?

docs/geps/gep-03.rst
```diff
@@ -0,0 +1,449 @@
...
+   Example:
+
+   .. code-block:: yaml
+
+    rentenwert:
+      name:
+        de: Rentenwerte alte und neue Bundesländer.
+          [...]
+      access_prior_parameters:
+        - reference_period: Y
```
ChristianZimpelmann, line 160:
```suggestion
        - reference_period: Year
```

Based on description above, should be

ChristianZimpelmann (approved):
Just two comments.

Looks nice, thanks!

Issue #18, opened 2019-07-22
Calculate Lohnsteuer


MaxBlesch:
We want to implement the option, that taxes are simulated as Lohnsteuer and not as Einkommenssteuer, as it is of now. The following discussion leads the foundation:

Eric Sommer: 

As of now, our income tax calculation mimics the situation *after* the household filed the tax declaration and received the rebate. This is however not the same what the household pays when receiving wages. This is a fundamental choice one has to make.

1. Simulate Income Tax
     - income from all sources, compare tax due to child benefit payments
2. Simulate withholding tax (Lohnsteuer) on earnings
     - taxable earnings are wages minus
       - Werbungskostenpauschale (2019: 1000€)
       - Sonderausgabenpauschale (2019: 36€)
       - Vorsorgepauschale (similar, but not identical to "Vorsorgeaufwendungen")
     - child tax credit is irrelevant for Lohnsteuer, but not for Soli!
     - everybody receives child benefit.
     - for the remaining income sources, calculate income taxes as before.
     - There is no income splitting for married couples. Instead, the basic allowances might be distributed among couples ("Steuerklassen")

Currently, (1) is implemented, but there might be good reasons to opt for (2) as it more closely resembles the actual disposable income people are facing. It is questionable whether everybody anticipates the future tax rebate. One problem is that "Steuerklassen", which have a major impact on disposable income, are usually not available and would need to be assumed. A second problem is that withholding taxes are attributed to the individual. In order for them to have an impact on labor supply, one would need to drop the unitary household model and think about how common income is distributed among couples. 

On a sidenote, the withdrawal tax is the revelant item when calculating benefit amounts for ALG2 and Housing Benefit.

HMG:

Thanks for the heads-up! Tricky thing indeed, but I do not think that the decision should be that fundamental. We should be able to handle both cases in the code eventually, picking whichever one is relevant for a project. 

We might even be mixing both -- think of female labour supply with tax classes 3 (husband) / 5 (wife) -- she will look at a huge marginal tax rate when looking at her wage bill, but things might look much nicer when filing the tax return. 

As for data on Steuerklassen: We might not get the joint distribution of them with all relevant variables, but we can always get marginals (and some joints) from other data sets. That income tax database for sure, but maybe even Microcensus and EVS have something?

Eric Sommer:

> she will look at a huge marginal tax rate when looking at her wage bill, but things might look much nicer when filing the tax return.

[...]

PR #150, opened 2020-02-24
Lohnsteuer

### LINKED ISSUE #143: Steuerklassen

Steuerklassen are an important part of the tax code for married couples. GETTSIM implicitly assumes Steuerklasse IV everywhere at present.

While Steuerklassen have transient effects until the tax declaration, they can have important impacts on household bargaining (https://dx.doi.org/10.1016/j.jpubeco.2019.104048) and they have a first-order impact on unemployment benefit payments (anything else?).  

### Proposed implementation

* Require columns `e_st_klasse` and `e_st_klasse_faktor`
* Provide a function that generates them with defaults of 1 / missing (singles) and 4 / 0.5 (couples) to keep current behaviour
* Make it a mandatory parameter for functions calculating income taxes at the individual level.

Eric-Sommer:

closes #143 and #18. 

This PR models the German [Lohnsteuer](https://de.wikipedia.org/wiki/Lohnsteuer_(Deutschland)), i.e. taxes withheld from monthly earnings. This is as opposed to Einkommensteuer, for which income needs to be filed once per year.

Novel modelling elements include:
- Steuerklassen. Can be given by the user, but can also be internally assessed. Steuerklasse 6 is out of scope at least for now, because we don't distinguish between several jobs. Regarding the choice 5/3 versus 4/4 within couple, it is assumed that once one both partners earns income above the income tax threshold, they choose 4/4. This is, admittedly, a strong assumption. But in this realm, all assumptions are quite strong and assuming something is IMO better than making it a mandatory input. The new function is tested as well. 'Faktorverfahren' is also not modelled yet and left for future model development :smile: 
- Vorsorgepauschale: Similar to Vorsorgeaufwendungen, but not identical.
- Tests are created via the [API of the German Ministry of Finance](https://www.bmf-steuerrechner.de/). The API can be called when setting `reload_test_data=True` in `test_lohnsteuer`. Funny enough, there are problems:
  - In 2020 and 2021, the allowance for single parents was raised to 4008€. Apparently, this is increase was not considered in the API, as it was intended to be temporary. Since it became permanent in 2022, it is included. Hence, single parents are tested for 2022.
- Need for modelling health contribution rates in a different manner. This PR should wait until #323 has sorted this out.
- The solidarity surcharge on top of lohnsteuer is modelled as well. This required some re-arrangements in `soli_st.py`.
- I feel that using `st_tarif` via `import` is not the 'gettsim' way to go design-wise. I'm happy to be educated on this one.

Let me stress (again) what a great tool the DAG is! Once you get it, it makes modelling so much easier.

### Todo
- [x] calculate Lohnsteuer
  - [x] model Vorsorgepauschale. 
    - [ ] pre 2010
  - [x] collect 'ermäßigter Beitragssatz' §243 SGB V. This is the necessary one.
  - [x] model deductions based on steuerklassen
- [x] add tests for Lohnsteuer
- [x] add to `CHANGES.rst`.


hmgaudecker:
Just came across a useful resource

https://www.bundesfinanzministerium.de/Web/DE/Themen/Steuern/Steuerarten/Lohnsteuer/Programmablaufplan/programmablaufplan.html

Eric-Sommer:
I am kind of hitting a wall here. I calculate lohnsteuer via the income tax tariff:
https://github.com/iza-institute-of-labor-economics/gettsim/blob/ff6b60ebbaf69b8220405cb6dabafbb936cf6db5/gettsim/taxes/lohn_st.py#L74

The tricky thing: The basic allowance, i..e. the x-value of the first step in the piecewise linear function, [depends](https://de.wikipedia.org/wiki/Lohnsteuerklasse#Freibetr%C3%A4ge) on `steuerklasse`. Can this logic be implemented in the existing framework, @ChristianZimpelmann?

ChristianZimpelmann:
>

ChristianZimpelmann:
> I am kind of hitting a wall here. I calculate lohnsteuer via the income tax tariff:
> 
> https://github.com/iza-institute-of-labor-economics/gettsim/blob/ff6b60ebbaf69b8220405cb6dabafbb936cf6db5/gettsim/taxes/lohn_st.py#L74
> 
> The tricky thing: The basic allowance, i..e. the x-value of the first step in the piecewise linear function, [depends](https://de.wikipedia.org/wiki/Lohnsteuerklasse#Freibetr%C3%A4ge) on `steuerklasse`. Can this logic be implemented in the existing framework, @ChristianZimpelmann?

Can you use 

```
piecewise_polynomial(
        x=x,
        thresholds=params["eink_st_tarif"]["thresholds"],
        rates=params["eink_st_tarif"]["rates"],
        intercepts_at_lower_thresholds=params["eink_st_tarif"][
            "intercepts_at_lower_thresholds"
        ],
    )
    ```
    
directly? Then you should be able to adjust `thresholds` accordingly.

File: gettsim/demographic_vars.py [hunk 1]
```diff
@@ -327,3 +327,81 @@ def rentner_in_hh(hh_id: IntSeries, rentner: BoolSeries) -> BoolSeries:
...
+    alleinerziehend: BoolSeries,
+    bruttolohn_m: FloatSeries,
+    kind: BoolSeries,
+    eink_st_params: dict,
+    eink_st_abzuege_params: dict,
+) -> IntSeries:
+    """ Determine Lohnsteuerklassen (also called 'tax brackets')
+    They determine the basic allowance for the withdrawal tax
+
+    1: Single
```
ChristianZimpelmann, line 344: Great documentation

gettsim/demographic_vars.py
```diff
@@ -327,3 +327,81 @@ def rentner_in_hh(hh_id: IntSeries, rentner: BoolSeries) -> BoolSeries:
...
+    3: One spouse in married couple who receives allowance of both partners.
+       Only in case of Single-Earner Households
+    4: Both spouses receive their individual allowance
+    5: Other spouse in couple who receives no allowance
+    6: Additional Job...not modelled yet, as we do not
+    distinguish between different jobs
+
+    Parameters
+    ----------
+    tu_id: Tax Unit Identifier
```
ChristianZimpelmann, line 355: List of parameters should be in the same style as in other docstrings (e.g. just refer to the respective function or input variable)

gettsim/demographic_vars.py
```diff
@@ -327,3 +327,81 @@ def rentner_in_hh(hh_id: IntSeries, rentner: BoolSeries) -> BoolSeries:
...
+    Returns
+    ----------
+    steuerklasse: IntSeries
+        The steuerklasse for each person in the tax unit
+    """
+
+    bruttolohn_max = bruttolohn_m.groupby(tu_id).max()
+    bruttolohn_min = bruttolohn_m.groupby(tu_id).min()
+    erwachsener = ~kind
+    anz_erwachsene_tu = erwachsener.groupby(tu_id).transform("sum")
```
ChristianZimpelmann, line 377: the function `anz_erwachsene_tu` exists already and should be used as argument of this function.

ChristianZimpelmann: Or even better: just use `gemeinsam_veranlagt` which is just defined as `tu_id.replace(anz_erwachsene_tu) == 2`

gettsim/demographic_vars.py
```diff
@@ -327,3 +327,81 @@ def rentner_in_hh(hh_id: IntSeries, rentner: BoolSeries) -> BoolSeries:
...
+        & (anz_erwachsene_tu == 2)
+        & (gemeinsam_veranlagt)
+    )
+    cond_steuerklasse1 = (anz_erwachsene_tu == 1) & ~alleinerziehend
+    cond_steuerklasse2 = alleinerziehend
+    cond_steuerklasse3 = alleinverdiener_paar & (
+        bruttolohn_m > einkommensgrenze_zweitverdiener / 12
+    )
+    cond_steuerklasse4 = (anz_erwachsene_tu == 2) & (~alleinverdiener_paar)
+    cond_steuerklasse5 = alleinverdiener_paar & (
```
ChristianZimpelmann, line 396: Do these condition include tax units in which both partners earn below einkommensgrenze_zweitverdiener? Or do those couples just don't get a Lohnsteuerklasse?

gettsim/parameters/eink_st_abzuege.yaml
```diff
@@ -469,6 +469,55 @@ vorsorge_sonstige_aufw_max:
...
+    de: Mindestanteil an Beiträgen zu GKV und GPV, der abgesetzt wird
+  description: 39b (2) Nr.3 EStG
+  unit: share
+  2009-07-23:
+    scalar: 0.12
+    reference: Art. 1 G. v. 16.07.2009 BGBl. I S. 1959
+
+vorsorgepauschale_kv_max:
+  name:
+    de: Höchsbetrag für Beiträge zu GKV und PKV, der abgesetzt werden kann. Hängt ab von der Steuerklasse
```
ChristianZimpelmann, line 483: This sounds rather like a description. Please keep the name attribute short and move explanations to description.

See also #148

gettsim/parameters/eink_st.yaml
```diff
@@ -228,12 +228,12 @@ ertragsanteil:
     1:
       lower_threshold: 2005
       upper_threshold: 2021
-      rate_linear: 0.02
+      rate_linear: -0.02
```
ChristianZimpelmann, line 231: Could you explain shortly the rational behind the minus?

gettsim/social_insurance/krankenv.py
```diff
@@ -81,15 +81,39 @@ def krankenv_beitr_regulär_beschäftigt(
...
+    ----------
+    lohn_krankenv_regulär_beschäftigt
+        See :func:`lohn_krankenv_regulär_beschäftigt`.
+    soz_vers_beitr_params
+        See params documentation :ref:`soz_vers_beitr_params <soz_vers_beitr_params>`.
+    Returns
+    -------
+    Pandas Series containing monthly health insurance contributions on earnings
+    """
+    # This is essentially 'erm' plus half the top-up contribution.
```
ChristianZimpelmann, line 106: I would make this more verbose (e.g. what is erm?)

gettsim/social_insurance/krankenv.py
```diff
@@ -81,15 +81,39 @@ def krankenv_beitr_regulär_beschäftigt(
...
+    return (
+        soz_vers_beitr_params["soz_vers_beitr"]["ges_krankenv"]["erm"]
+        + (
+            soz_vers_beitr_params["soz_vers_beitr"]["ges_krankenv"]["an"]
+            - soz_vers_beitr_params["soz_vers_beitr"]["ges_krankenv"]["erm"]
```
ChristianZimpelmann, line 111: why is parameter `"erm"` loaded twice and first added and then substracted?

gettsim/taxes/eink_st.py
```diff
@@ -8,16 +8,14 @@ def st_kein_kind_freib_tu(
     anz_erwachsene_tu: IntSeries,
     eink_st_params: dict,
 ) -> FloatSeries:
-    """Taxes without child allowance.
+    """ Taxes without child allowance.
```
ChristianZimpelmann, line 11: I would remove all three changes in this docstring

gettsim/taxes/eink_st.py
```diff
@@ -49,27 +47,22 @@ def st_kind_freib_tu(
...
 
 
 def st_tarif(x: FloatSeries, params: dict) -> FloatSeries:
-    """The German Income Tax Tariff.
-
-    Modelled only after 2002 so far.
-    It's not calculated as in the tax code, but rather a gemoetric decomposition of the
-    area beneath the marginal tax rate function.
-    This facilitates the implementation of alternative tax schedules
+    """ The German Income Tax Tariff.
```
ChristianZimpelmann, line 55: no whitespace after `"""` by convention

gettsim/taxes/lohn_st.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def lohn_st_zve(
+    bruttolohn_m: FloatSeries,
+    steuerklasse: IntSeries,
+    eink_st_abzuege_params: dict,
+    vorsorgepauschale: FloatSeries,
+) -> FloatSeries:
+    """
+    Calculates taxable income (zve: zu versteuerndes Einkommen) for lohnsteuer
```
ChristianZimpelmann, line 17: Start description in same line as `"""` (also adjust for other new functions)

gettsim/taxes/lohn_st.py
```diff
@@ -0,0 +1,204 @@
+import numpy as np
+import pandas as pd
+
+from gettsim.piecewise_functions import piecewise_polynomial
+from gettsim.taxes.eink_st import st_tarif
+from gettsim.typing import FloatSeries
+from gettsim.typing import IntSeries
+
+
+def lohn_st_zve(
```
ChristianZimpelmann, line 10: How about `lohn_st_eink`?

gettsim/taxes/soli_st.py
```diff
@@ -39,16 +42,113 @@ def soli_st_tu(
...
-            intercepts_at_lower_thresholds=soli_st_params["soli_st"][
-                "intercepts_at_lower_thresholds"
-            ],
-        )
+        anz_erwachsene_tu * soli_tarif(st_per_individual, soli_st_params)
         + soli_st_params["soli_st"]["rates"][0, -1] * abgelt_st_tu
     )
 
     return out
+
```
ChristianZimpelmann, line 50: I think it might make more sense to move those functions to `lohn_st.py`.

gettsim/demographic_vars.py
```diff
@@ -327,3 +327,81 @@ def rentner_in_hh(hh_id: IntSeries, rentner: BoolSeries) -> BoolSeries:
     BoolSeries indicating pensioner in household.
     """
     return rentner.groupby(hh_id).any()
+
+
+def steuerklasse(
```
ChristianZimpelmann, line 332: I think it might make sense to move this function to `lohn_st.py`.

ChristianZimpelmann (changes requested):
I just had a look. Thank you for all this work!

- I made quite a lot of comments/suggestions. Let me know if you disagree somewhere or would like to discuss.
- What is this renaming of `dashboard/app` about?
- Re importing `st_tarif`: This is how functions that are called from within other functions are handled in other places of GETTSIM, as well. I agree that it doesn't fit perfectly to the DAG idea, but we don't need to fix this in this PR.
- Since the calculated Lohnsteuer doesn't interact with other taxes/transfers, I would suggest to keep the Lohnsteuer functions in a separate file instead of including them in several other files.

gettsim/demographic_vars.py
Eric-Sommer: In reality, those couples wouldn't pay lohnsteuer at all, so they wouldn't need one. If I'm not mistaken, they would be classified here as 4/4

gettsim/social_insurance/krankenv.py
Eric-Sommer: 'Ermäßigter Beitragssatz'. Properly explaining this should be done in #323 IMO

gettsim/social_insurance/krankenv.py
Eric-Sommer: again, this is a dirty version of what should be done properly in #323

Eric-Sommer:
Thanks for your very helpful comments, @ChristianZimpelmann 
> 
>     * What is this renaming of `dashboard/app` about?

I thought `bokeh serve` does not work with `app`. But that was nonsense, the change will be reverted...

> 
>     * Re importing `st_tarif`: This is how functions that are called from within other functions are handled in other places of GETTSIM, as well. I agree that it doesn't fit perfectly to the DAG idea, but we don't need to fix this in this PR.

OK for me (obviously)

>     * Since the calculated Lohnsteuer doesn't interact with other taxes/transfers, I would suggest to keep the Lohnsteuer functions in a separate file instead of including them in several other files.

Well it does not interact yet. It needs to be integrated, e.g. with Elterngeld and Arbeitslosengeld (I and II). I added the Soli part of Lohnsteuer into `soli_st.py` because it (now) shares a function with the 'standard' soli.

gettsim/demographic_vars.py
Eric-Sommer: OK

gettsim/demographic_vars.py
Eric-Sommer: done

gettsim/parameters/eink_st.yaml
Eric-Sommer: because it yields the correct rates :man_shrugging: 

```python
from gettsim.policy_environment import set_up_policy_environment
from gettsim.taxes.lohn_st import vorsorg_rv_anteil
policy_params, policy_functions = set_up_policy_environment("2021")
print(vorsorg_rv_anteil(policy_params['eink_st_abzuege']))
policy_params, policy_functions = set_up_policy_environment("2010")
print(vorsorg_rv_anteil(policy_params['eink_st_abzuege']))
```
yields `0.84` and `0.40`, respectively, which is correct

gettsim/parameters/eink_st_abzuege.yaml
Eric-Sommer: OK

gettsim/taxes/soli_st.py
```diff
@@ -39,16 +42,113 @@ def soli_st_tu(
...
+    out = (
+        (lohnsteuer_splittingtarif * (steuerklasse == 3))
+        + (lohnsteuer_basistarif * (steuerklasse.isin([1, 2, 4])))
+        + (lohnsteuer_klasse5_6 * (steuerklasse.isin([5, 6])))
+    )
+    return out
+
+
+def soli_tarif(st_per_individual: FloatSeries, soli_st_params: dict) -> FloatSeries:
```
ChristianZimpelmann, line 131: The function should start with an underscore as it is not part of the DAG, but only called within other functions.

hmgaudecker: See #333 -- not sure we'll even want it in the end.

gettsim/taxes/lohn_st.py
Eric-Sommer: done

JakobWegmann:
If I didn´t miss anything, parts of §39 Estg are missing for Steuerklasse V/VI (`lohnsteuer_klasse5_6`) in `lohn_st`:
"die Jahreslohnsteuer beträgt jedoch mindestens 14 Prozent des zu versteuernden Jahresbetrags, für den 11 480 Euro übersteigenden Teil des zu versteuernden Jahresbetrags höchstens 42 Prozent, für den 29 298 Euro übersteigenden Teil des zu versteuernden Jahresbetrags 42 Prozent und für den 222 260 Euro übersteigenden Teil des zu versteuernden Jahresbetrags 45 Prozent."
Perhaps, the 14% are already implemented. But I don´t understand for `lohnsteuer_klasse5_6` why the maximum and not the minimum is used

Eric-Sommer:
> If I didn´t miss anything, parts of §39 Estg are missing for Steuerklasse V/VI (`lohnsteuer_klasse5_6`) in `lohn_st`: "die Jahreslohnsteuer beträgt jedoch mindestens 14 Prozent des zu versteuernden Jahresbetrags, für den 11 480 Euro übersteigenden Teil des zu versteuernden Jahresbetrags höchstens 42 Prozent, für den 29 298 Euro übersteigenden Teil des zu versteuernden Jahresbetrags 42 Prozent und für den 222 260 Euro übersteigenden Teil des zu versteuernden Jahresbetrags 45 Prozent." Perhaps, the 14% are already implemented. But I don´t understand for `lohnsteuer_klasse5_6` why the maximum and not the minimum is used

This is a consequence of this weird rule:
https://github.com/iza-institute-of-labor-economics/gettsim/blob/4d844299072f3ae83cff158c02a6a4206c24af12/gettsim/taxes/lohn_st.py#L86-L93

It might happen that you pay less than 14% that way. the `max` function is needed because that way, you either pay 14% (in case 1.25/0.75 rule leads to a tax rate below 14%) or the correct amount higher than 14%. 

But I agree, the same would need to be done for the higher cut-offs and properly tested. The parameters are all included already.

JakobWegmann:
Makes sense! In any case, the more interesting part is the implementation of the nested 3  if-conditions 😄

hmgaudecker:
@JakobWegmann @Eric-Sommer -- I just merged main (a beast mostly due to #450) into this other beast, cannot really check in detail any more now whether I introduced a few additional issues.

But I'll be happy to work with you guys on ironing them out -- will be great to have lohnsteuer in GETTSIM!

JakobWegmann:
Thank you for the merge and the corrections.

Finally, this week all tests of the Lohnsteuer passed. However, the Soli is still missing. 

The calculation of daily/weekly/monthly/yearly Lohnsteuer works. We implemented it, as otherwise, the number of test cases would be quite small.

hmgaudecker:
> The calculation of daily/weekly/monthly/yearly Lohnsteuer works. We implemented it, as otherwise, the number of test cases would be quite small.

Great! 

> Finally, this week all tests of the Lohnsteuer passed. 

They don't here and it seemed to me like it was due to monthly variables being requested and yearly variables being around or so, but even better! 

> However, the Soli is still missing.

Cool. Should hopefully be a matter of applying its usual formula to the annual `lohnst` and then converting back to shorter time periods?

Btw, in terms of naming conventions, the names should become `lohnst` and `lohnst_soli` -- I'll be happy to do the search and replace once tests pass, though.

Thanks!

JakobWegmann:
> They don't here and it seemed to me like it was due to monthly variables being requested and yearly variables being around or so, but even better!

The tests we were working on are in `test_lohnsteuer2.py`. `test_lohnsteuer.py` are the tests Eric wrote that work with the BMF API.

> Cool. Should hopefully be a matter of applying its usual formula to the annual lohnst and then converting back to shorter time periods?

Unfortunately, it's more complicated. The Kinderfreibeträge are taken into account for the Soli, but not for the Lohnsteuer. So it's necessary to first calculate the Lohnsteuer without Kinderfreibeträge and then to recalculate the Lohnsteuer with Kinderfreibeträge to apply the Soli.

hmgaudecker:
Fine, thanks for the explanation!

I would suggest merging a simple version then,

- either ignoring Kinderfreibeträge in soli/Lohnsteuer
- or ignoring soli/Lohnsteuer altogether

We should strive to keep PRs smaller.


JakobWegmann ***@***.***> schrieb am Fr., 13. Jan. 2023,
12:06:

docs/gettsim_objects/variables_out.md
```diff
@@ -28,6 +28,10 @@ You can find their individual calculation in the documentation of all {ref}`func
   - Monthly child benefit
 * - {func}`eink_st_tu <_gettsim.functions.eink_st_tu>`
   - Income Tax on tax unit level
+* - {func}`lohnst_m` <_gettsim.functions.lohnst_m>`
+  - Withholding tax
+* - {func}`soli_st_lohnst <_gettsim.functions.soli_st_lohnst>`
```
hmgaudecker, line 33: Don't we want this function on a monthly basis? In that case, call it `soli_st_lohnst_m` and divide by 12 in the function body.

src/_gettsim/policy_environment.py
```diff
@@ -282,6 +284,40 @@ def _parse_einführungsfaktor_vorsorgeaufw_alter_ab_2005(date, params):
     return params
 
 
+def _parse_vorsorg_rv_anteil(date, params):
```
hmgaudecker: Please open an issue that this function will need adjustment for people who are receiving pension benefits while working (at least I'd think that'll be the case).

src/_gettsim/parameters/eink_st_abzuege.yaml
```diff
@@ -656,6 +656,75 @@ kinderbetreuungskosten_abz_maximum:
...
+    en: >-
+      Function for the share of pension insurance contributions deducted in the payroll
+      tax calculation
+  unit: share
+  reference: § 39 (4) EStG
+  note:
+    de: >-
+      Funktion erst ab 2005 vorhanden. Steigt zwischen 2005 und 2025 jährlich um 4
+      Prozentpunkte.
+    en: rises between 2005 and 2025 by 4 percentage points annually.
```
hmgaudecker, line 676: The jump in 2023 that we included in #476 for `einführungsfaktor` does not matter here? If it does, just copy the changes from there with appropriate adjustment.

src/_gettsim/parameters/lohnst.yaml
```diff
@@ -0,0 +1,34 @@
...
+    de: >-
+      Einkommensgrenzen für Steuerklasse 5 und 6, innerhalb derer
+      Mindest- und Maximalsteuern festgelegt sind.
+    en: >-
+      Thresholds for Steuerklasse 5 and 6, where
+      minimum and maximum tax rates apply.
+  reference: § 39b Absatz 2 Satz 7 EStG
+  note:
+    de: >-
+      Nur für 2022 bis jetzt implementiert
```
hmgaudecker, line 16: Wouldn't this render GETTSIM essentially unusable for years prior to 2022? I think we should expand this.

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+    Returns
+    -------
+    Base for Lohnsteuer for Steuerklasse 5 and 6
+
+    """
+
+    out = max(
+        2
+        * (
+            trunc(_eink_st_tarif(taxable_inc * 1.25, eink_st_params))
```
hmgaudecker, line 86: Please work with `@add_rounding_spec` instead of truncating here, if rounding is turned off, we want to avoid discontinuities everywhere.

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+    # add both RV and KV deductions. For KV, take the larger amount.
+    out = vorsorg_rv + max(vorsorg_kv_option_a, vorsorg_kv_option_b)
+    return out
+
+
+@add_rounding_spec(params_key="lohnst")
+def vorsorgepauschale_ab_2005_bis_2009() -> float:
+    """vorsorg_rv and vorsorg_kv_option_a are identical to after 2010."""
+
+    out = 0
```
hmgaudecker, line 284:
```suggestion
    out = 0.0
```

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+
+    # add both RV and KV deductions. For KV, take the larger amount.
+    out = vorsorg_rv + max(vorsorg_kv_option_a, vorsorg_kv_option_b)
+    return out
+
+
+@add_rounding_spec(params_key="lohnst")
+def vorsorgepauschale_ab_2005_bis_2009() -> float:
+    """vorsorg_rv and vorsorg_kv_option_a are identical to after 2010."""
```
hmgaudecker, line 282: This docstring refers to parameters not being used here and to a different time period than suggested by the function. Please fix; maybe get rid of this altogether? What is the point of returning 0, sounds like another interface should see adjustment ?

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+def lohnst_mit_kinderfreib(
+    lohnst_eink: float,
+    kinderfreib_für_soli_st_lohnst: float,
+    steuerklasse: int,
+    eink_st_params: dict,
+) -> float:
+    """Calculate Lohnsteuer with the tax base taking child allowance into account,
+    otherwise identical to :func:`lohnst_m`.
```
hmgaudecker, line 322: Can't we re-factor:
- `lohnst_m`
- `lohnst_mit_kinderfreib`

such that they use the same main body of the function? That is, define a helper function like `_eink_st_tarif` and call it internally?

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+        out = lohnsteuer_basistarif
+    elif steuerklasse == 3:
+        out = lohnsteuer_splittingtarif
+    else:
+        out = lohnsteuer_klasse5_6
+
+    return out
+
+
+# Possible ToDo: Determine Steuerklasse endogenously.
```
hmgaudecker, line 348: I am not sure we'll ever want this, seems to depend a lot on assumptions.

In any case, please get rid of the code snippet here and in case you think this should be discussed more, open an issue for it. You can paste the code there then.

src/_gettsim_tests/test_lohnst.py
```diff
@@ -0,0 +1,420 @@
...
+        zusatzbeitrag=soz_vers_params["beitr_satz"]["ges_krankenv"][
+            "mean_zusatzbeitrag"
+        ],
+    )
+
+    return hh
+
+
+@pytest.mark.skip()
+@pytest.mark.parametrize("year, column", itertools.product([2022], OUT_COLS))
```
hmgaudecker, line 389: I think it is a great thing to be able to download stuff directly, but the tests should always be based on a local file.

Please pull out all code calling the API to a module in `src/_gettsim_tests/test_data/original_testfaelle`. When ran, the module should save a file we can use on disk. 

Do I gather correctly that this is where `lohnsteuer_bmf_2022.xlsx` is coming from already?

hmgaudecker (changes requested):
Truly excellent work, many thanks!!! 

I made a lot of semantic changes directly, but tried not to touch anything that would change the logic of the code. Quite a few comments left, but most should be very easy to address.

hmgaudecker:
Forgot to mention a couple of things:
- I did not check in detail, but test coverage is fairly low. Any chance this could be increased with a reasonable amount of work?
- Should we have a consistency check on the input data regarding `steuerklassen`? E.g., no 1 if `gemeinsam_veranlagt` etc. -- at a minimum, we should open an issue for it, though we'll probably want to wait with the implementation until we made progress on #509 
- I think the docstring on `steuerklassen` mentions 1-5 as possible values, but the code is valid for 6 as well. We should mention somewhere that this is already thinking about a future date where GETTSIM might be able to distinguish between a primary and a secondary job.

src/_gettsim/parameters/lohnst.yaml
Eric-Sommer: parameters before 2022 are added

src/_gettsim/parameters/eink_st_abzuege.yaml
```diff
@@ -656,6 +656,75 @@ kinderbetreuungskosten_abz_maximum:
...
+      Höchsbetrag für Mindestvorsorgepauschale der Beiträge zu GKV und PKV, die von der
+      Lohnsteuer abgesetzt werden können.
+    en: >-
+      Maximum amount considered for Mindestvorsorgepauschale of contributions to health
+      care and care insurance that can be deducted from withholding tax
+  reference: 39b (2) Nr.3 EStG. Depends on Steuerklasse.
+  unit: Euro
+  time: year
+  2009-07-23:
+    stkl3: 3000
```
ChristianZimpelmann, line 725: I would prefer a more verbose label: `steuerklasse_3`

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+        - entlastung_freibetrag_alleinerz
+        - vorsorgepauschale,
+        0.0,
+    )
+
+    return out
+
+
+def _lohnsteuer_klasse5_6_basis(taxable_inc: float, eink_st_params: dict) -> float:
+    """Calculate base for Lst for Steuerklasse 5 and 6.
```
ChristianZimpelmann, line 59: Lst = Lohnsteuer?

Would be helpful to describe a bit more what you mean by "base".

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+    eink_st_params: dict,
+    lohnst_params: dict,
+    steuerklasse: int,
+) -> float:
+    """
+    Calculates Lohnsteuer (withholding tax on earnings), paid monthly by the employer on
+    behalf of the employee. Apply the income tax tariff, but individually and with
+    different exemptions, determined by the 'Steuerklasse'. Source: §39b EStG
+
+    Calculation is differentiated by steuerklasse
```
ChristianZimpelmann, line 106:
```suggestion
    Calculation is differentiated by steuerklasse:
```

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+def vorsorgepauschale_ab_2010(  # noqa: PLR0913
+    bruttolohn_m: float,
+    steuerklasse: int,
+    wohnort_ost: bool,
+    ges_krankenv_zusatzbeitrag: float,
+    ges_pflegev_zusatz_kinderlos: bool,
+    eink_st_abzuege_params: dict,
+    soz_vers_beitr_params: dict,
+) -> float:
+    """Calculate Vorsorgepauschale for Lohnsteuer valid since 2010 Those are deducted
```
ChristianZimpelmann, line 186:
```suggestion
    """Calculate Vorsorgepauschale for Lohnsteuer valid since 2010. Those are deducted
```

src/_gettsim/taxes/lohnst.py
```diff
@@ -0,0 +1,401 @@
...
+
+def kinderfreib_für_soli_st_lohnst(
+    steuerklasse: int,
+    anz_kinder_mit_kindergeld_tu: float,
+    eink_st_abzuege_params: dict,
+) -> float:
+    """Calculate Child Allowance for Lohnsteuer-Soli.
+
+    For the purpose of Soli on Lohnsteuer, the child allowance not only depends on the
+    number of children, but also on the steuerklasse
```
ChristianZimpelmann, line 296:
```suggestion
    number of children, but also on the steuerklasse.
```

src/_gettsim/taxes/lohnst.py
ChristianZimpelmann: Please post the code in that issue. This will be very helpful for the data preparation.

ChristianZimpelmann (commented):
A couple of (mostly minor) comments. Thank you very much 👍 Looks great!

src/_gettsim/taxes/lohnst.py
Eric-Sommer: Thanks. I guess this dosctring belongs to the (non-existent) function for up to 2004. We currently have `vorsorgepauschale` properly since 2010 only. @JakobWegmann: do we want to model the 2005 rule in this PR as well?

src/_gettsim_tests/test_lohnst.py
Eric-Sommer: > Do I gather correctly that this is where `lohnsteuer_bmf_2022.xlsx` is coming from already?

No, that file is directly provided by the BMF. And it does not always correspond to the API :sob: Overall, we regard the API as an additional source and hence skip it by default. It looks very attractive at first glance, but it's quite opaque what's happening there and when certain parameters are being updated. It could be two identical API call create different results when being called 6 months apart. Adding a result file to the repository would make things overly complicated in my view.

hmgaudecker: But then we should not keep this code as part of GETTSIM. This always adds a maintenance burden, too. How about a notebook somewhere else, we could link from the docs?

src/_gettsim/taxes/lohnst.py
hmgaudecker: Probably not, in that case, however, just open an issue for it and delete the code. Misleading docstrings are worse than no docs. :smile:

Eric-Sommer:
>     * I did not check in detail, but test coverage is fairly low. Any chance this could be increased with a reasonable amount of work?
> 
>     * Should we have a consistency check on the input data regarding `steuerklassen`? E.g., no 1 if `gemeinsam_veranlagt` etc. -- at a minimum, we should open an issue for it, though we'll probably want to wait with the implementation until we made progress on [Discussion of groupings of / relationships among individuals #509](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509)

I like a lot! Where would I add this?

hmgaudecker:
> > Should we have a consistency check on the input data regarding `steuerklassen`? E.g., no 1 if `gemeinsam_veranlagt` etc. -- at a minimum, we should open an issue for it, though we'll probably want to wait with the implementation until we made progress on [Discussion of groupings of / relationships among individuals #509](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509)
> 
> I like a lot! Where would I add this?

@ChristianZimpelmann, could you point Eric to the consistency checks on the data that you implemented at some point (IIRC). I think we should push along with this PR however, best open a new issue for that.

src/_gettsim_tests/test_lohnst.py
Eric-Sommer: Like a notebook in `src/_gettsim_tests`?

hmgaudecker: Honestly I do not see much of a value in keeping it as part of GETTSIM, worrying mostly about a maintenance burden. Dead code is always a drag, no matter what you do (take the current dashboard as a case in point, not sure how much time it cost me over the past three months to properly ignore it everywhere). 

IIUC, what we want to do with it is to call it once a year to check whether the outputs of GETTSIM make sense. We do not need anything internal for that, just the GETTSIM API (sic!). 

Happy to be convinced otherwise, though!

Eric-Sommer: I see your point. While these tools (BMF API, Param Dashboard) might be useful in some contexts, the maintenance burden speak for removing them from the 'core' gettsim. What do you think about outsourcing them to additional repositories in the @iza-institute-of-labor-economics organization?

hmgaudecker: Sounds great, will do (though dashboard will probably start over based on plotly/dash instead of bokeh.

src/_gettsim/taxes/lohnst.py
JakobWegmann: Something for the future :)

src/_gettsim_tests/test_lohnst.py
hmgaudecker: Here you go: https://github.com/iza-institute-of-labor-economics/gettsim-code-for-picking

Thanks!

ChristianZimpelmann:
> @ChristianZimpelmann, could you point Eric to the consistency checks on the data that you implemented at some point (IIRC). I think we should push along with this PR however, best open a new issue for that.

I haven't implemented anything in this direction, yet. Although I think it would be very helpful in the future.

#402 is the relevant issue, probably best to just add it there.

src/_gettsim/taxes/lohnst.py
Eric-Sommer: done :+1:

docs/gettsim_objects/variables_out.md
Eric-Sommer: is now on monthly basis

src/_gettsim/taxes/lohnst.py
Eric-Sommer: @JakobWegmann : I removed the `trunc` statements in the steuerklasse 5/6 calculation. Tests still work.

src/_gettsim/parameters/eink_st_abzuege.yaml
Eric-Sommer: they do matter in a similar vein. corrected and tested now

Eric-Sommer:
> 



> 



>     * I did not check in detail, but test coverage is fairly low. Any chance this could be increased with a reasonable amount of work?

This is only because the solidarity surcharge is untested. @JakobWegmann, does BMF also provide tests including soli?

JakobWegmann:
> > ```
> > * I did not check in detail, but test coverage is fairly low. Any chance this could be increased with a reasonable amount of work?
> > ```
> 
> 
> This is only because the solidarity surcharge is untested. @JakobWegmann, does BMF also provide tests including soli?

The Soli is also an outcome column in the excel table. I can‘t open the file right now but I think it is the last column. The second to last column is the „Bemessungsgrundlage“.

hmgaudecker (approved):
I think this is good to go now! Thanks a lot! 

@Eric-Sommer, could you please double-check the changes I just made when merging `main`?

PR #122, opened 2019-12-03
WIP: Visualize Params

Eric-Sommer:
### What problem do you want to solve?

closes #44. Visualizes the time-evolution of each and every parameter. There is a separate HTML file for each `yaml` file. Exact values, including legislation changes, can be made visible by hovering over the graph:

Screenshot from 2019-12-14 20-17-32

This PR could also serve as a cleanup/extension of the params, e.g. by linking to offenegesetze.de

### How to run
run `paramviz.py` in the `data` folder. You also need to create the folder `graphs` where results are stored.

### Todo
- [ ] Think about how to integrate into the documentation (http://docs.bokeh.org/en/1.3.2/docs/reference/sphinxext.html)
- [ ] Possible Extension, if feasible with sphinx: Instead of 17 separate files, create one file with 17 tabs.
- [x] Highlighting needs to be customized to some degree. No commas are shown now, which is bad for shares.

PR #152, opened 2020-03-01
Regroup and update wohngeld parameters

### LINKED ISSUE #144: Use groups of parameters in `wohngeld.yaml`
Having discussed this with @mjbloemer in the afternoon, the following might be a good choice:

```yaml
faktor:
    name:
        de: Faktor am Anfang der Wohngeldformel
        en:
    description:
        de: Anlage 2 WoGG zu §19 Abs. 2 WoGG
        en:
    1984-01-01:
        value: 1
        reference:
    2009-01-01:
        value: 1.08
        reference:
    2016-01-01:
        value: 1.15
        reference: §19 (1) aa) zuletzt geändert durch BGBl. Nr.38 vom 08.10.2015, S.1613

koeffizienten:
    name: Parameter a, b, c der Wohngeldformel
    1984-01-01:
        note: Parameter aus Regressionsanalyse der Wohngeldtabellen
        1:
            a: 0.0532
            b: 0.00135
            c: 0.0000632
        2:
            a:
            b:
            c:
        ...
        12:
            a:
            b:
            c:
```

~No big rush with implementing this, let us first make sure that the piecewise linear grouping works.~

MaxBlesch:
We want to regroup the parameters in wohngeld.yaml according to GEP-3 #148. Our starting point is the draft by HMG in #144, which will be closed with this PR. 

This PR also includes the 2020 updates by [Wohngeldstärkungsgesetz](https://www.buzer.de/s1.htm?a=&g=WoGSt%C3%A4rkG) made in #134, which is closed and will not be finished. Therefore this PR also addresses #124. 

Closes #144


MaxBlesch:
I started regrouping the values and came across something odd. In 1992 almost every parameter was updated except b for 1, 3, 5. Is this on purpose @Eric-Sommer ?

Eric-Sommer:
> I started regrouping the values and came across something odd. In 1992 almost every parameter was updated except b for household size 1, 3 and 5. Is this on purpose @Eric-Sommer ?

OK, one note of caution here. Before 2001, the Wohngeld formula was not expressed explicitly.  Instead, the WoGG provided [endless tables](https://offenegesetze.de/veroeffentlichung/bgbl1/1992/15#page=1). At one point, we OCR'd them and estimated the formula based on these values. This is where the parameters before 2001 come from. In the original param file, I also do not observe changes for the three parameters you mention.

> 
> Also, 2015 was only c changed for a household with 7 people. And only by 0.05e-6. Is this right?

This is most probably some typo or rounding error. In 2015, nothing changed.

gettsim/data/wohngeld_neu.yaml
```diff
@@ -177,6 +177,7 @@ koeffizienten:
           b: 0.000377
           c: 0.0000173
     2001-01-01:
+      reference: Art. 5 G. v. 22.12.1999 BGBl. I S. 2671
       value:
```
hmgaudecker: I think we should drop the `value` level here. Really only makes sense for scalars (maybe we should even call it `scalar`?), else it adds an unnecessary level of nesting.

hmgaudecker: But we should add a note that the keys refer to household sizes. Trivial, but useful for casual observers.

MaxBlesch: I am not sure. When we read the yaml file it is a lot easier to select `value`. Or is there a easy way, which I don't see?

MaxBlesch: I wouldn't add this below the date and instead a note at the very beginning of koeffizenten.

hmgaudecker: > I wouldn't add this below the date and instead a note at the very beginning of koeffizenten.

Fair enough.

hmgaudecker: > I am not sure. When we read the yaml file it is a lot easier to select `value`. Or is there a easy way, which I don't see?

Why would it? Instead of `[date]["value"][x]`, you just do `[date][x]` ? 

It probably makes sense to discard any non-integers in the transformation, though. As in #140:

```python
keys = sorted(key for key in list_of_dicts.keys() if type(key) == int)
```

MaxBlesch: Yes. So its tradeoff between nesting and discarding keys. I just found nesting more readable in the yaml file and also, as mentioned before, easier to process. But I do not have strong preferences on this, thus I will just change!

gettsim/policy_for_date.py
```diff
@@ -161,3 +161,32 @@ def load_ges_renten_vers_params(raw_pension_data, actual_date):
...
+            tax_data[param] = np.nan
+        else:
+            policy_in_place = tax_data_raw[param][np.max(past_policies)]
+            if "scalar" in policy_in_place.keys():
+                tax_data[param] = policy_in_place["scalar"]
+            else:
+                value_keys = sorted(
+                    key for key in policy_in_place.keys() if type(key) == int
+                )
+                tax_data[param] = {}
```
MaxBlesch: So thats the reason why I used value. I don't know a better way. Also googled quite a bit. Maybe you could point me in a better direction. @hmgaudecker

hmgaudecker: Not sure i understand where the problem is... We do need some specialised code here, hence the `type` attribute (holding `piecewise_linear` and friends). I have not looked at the Wohngeld formula, not sure how to best call that or whether it makes any sense to give it a generic name.

gettsim/benefits/wohngeld.py
```diff
@@ -264,3 +264,33 @@ def apply_wg_formula(household, params, household_size):
         )
 
     return wg_amount
+
+
+def regrouped_wohngeld_formel(household, params, household_size):
+    # The formula is only valid for up to 12 household members
+    household_size_max = min(household_size, 12)
+    koeffizenten = params["koeffizienten"][household_size_max]
```
hmgaudecker: Just use `min(household_size, 12) ` directly, the max-version is never used again and makes it harder to read IMHO.

gettsim/benefits/wohngeld.py
```diff
@@ -264,3 +264,33 @@ def apply_wg_formula(household, params, household_size):
...
+                    + (koeffizenten["b"] * household["M"])
+                    + (koeffizenten["c"] * household["Y"])
+                )
+                * household["Y"]
+            )
+        ),
+    )
+    # If more than 12 persons, there is a lump-sum on top.
+    # You may however not get more than the corrected rent "M".
+    if household_size > 12:
```
hmgaudecker: Is this correct? Even if `wg_amount` is 0, this is being added. Do we have a check for Wohngeld eligibility somewhere else then? In that case, do we even need to check whether `wg_amount` is negative?

Eric-Sommer: There are no hard critieria on wohngeld eligibility. This function computes the possible claim, which is checked later on against other benefits. a `max(0, wg_amount)` does not hurt here, though.

MaxBlesch:
So I started regrouping `wgmax...` and I came across some things I do not get. I hope @Eric-Sommer you can help me! My interpretation of the calculation of `max_rent` is the following:

Before 2009 construction of the building matters. These are three categories. We have 7 mietsstufen and the values differ for a household with up to 5 persons and above some payment for each additional person. This would make 6 * 7 * 3 = 126 different variables for me?! 

Since 2009 the year of building is assumed to be medium (so "m"). This means 42 of these 126 variables should be continued. This sometimes happens. But for example, the value for a 2 person household with mietstufe 7 and medium construction year starts in 2020?!

So I guess the variables are just not complete yet? Is there a good place to look them up? For regrouping it would really help to have a compete variable set!

Eric-Sommer:
We don't really assume medium, it's just that the law does not differentiate anymore between construction years. And mietstufe 7 was only introduced with the most recent Wohngeldreform in 2020.

hmgaudecker:
Cool. I think we should adjust the structure of parameters / the function in 2009 just to make that clear. Rather a few extra lines of code than a hack there.

And adding level 7 will be much more natural (indeed, trivial?) now!

MaxBlesch:
> We don't really assume medium, it's just that the law does not differentiate anymore between construction years. And mietstufe 7 was only introduced with the most recent Wohngeldreform in 2020.

I am aware of this assumption. But in general, there are a lot of combinations from anz_pers- mietstufe - baujahr not present? Is there a resource I can add those?

I will adjust the structure of the function accordingly!

gettsim/benefits/wohngeld.py
```diff
@@ -221,6 +238,7 @@ def _calc_wg_income_deductions_since_2016(household, params):
     and children who are working
     """
     workingchild = household["child"] & (household["m_wage"] > 0)
+
```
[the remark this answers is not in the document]
Eric-Sommer: Good catch! The ruling was changed in 2016: https://www.buzer.de/gesetz/8380/al52302-0.htm
. You get it if the child is in need of care at home. Before 2016, you needed 80% handicap degree. We basically assume that every child with handicap degree > 0 is in need of care at home, which is probably not true, but I don't really know. Anyway, the parameter deserves at least renaming.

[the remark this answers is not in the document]
MaxBlesch: I introduced a scalar parameter. So it is different, but still grouped under the same variable.

MaxBlesch:
So this is a first draft on the regrouping of the wohngeld parameters. First of all. Thanks to @Eric-Sommer, who completed a lot of the missing historic data. Of course, we do not have any tests for the new params. But I think for testing params we will someday have #122.

Maybe @Eric-Sommer you can review how I changed the params interface in wohngeld.py. Maybe there are even better structures.

I also included a new loading function in `policy_for_date.py`. I don't know if we want one loading function for all yaml groups, but this is a first draft, which I can probably reuse in #157. Maybe @hmgaudecker has some thoughts on that!

All feedback very welcome and appreciated!

docs/getting-started.rst
```diff
@@ -128,6 +128,8 @@ Required columns in input data
 +---------------------+--------------------------------------------------------+-------+
 | heizkost_m          | Monthly heating expenses                               | float |
 +---------------------+--------------------------------------------------------+-------+
+| mietstufe           | Level of rent in community                             | int   |
```
Eric-Sommer: maybe add "3 refers to average rent"

docs/getting-started.rst
```diff
@@ -145,7 +147,7 @@ Required columns in input data
 | immobilie_baujahr   | Construction year of dwelling                          | int   |
```
Eric-Sommer: Construction year is used in levels below. So drop the categories.

docs/getting-started.rst
MaxBlesch: Well that depends highly on the number of possible mietstufen in that year?

docs/getting-started.rst
MaxBlesch: Deleted it!

docs/getting-started.rst
Eric-Sommer: Since it's now mandatory we have to give some guidelines, i.e. the possible range. and "if in doubt, choose 3". Before 1985, Mietstufe 3 is for cities > 500.000 inhabitants.

MaxBlesch: Added some guideline. But I think will need sooner or later a more detailed documentation anyway. Then we can include such guidelines. We probably will have more, as not all datasets contain all variables.

gettsim/data/wohngeld.yaml
```diff
@@ -1,2513 +1,1646 @@
...
-    de: Parameter c (Haushalt mit 1 Person)
-    en:
-  description:
-    de: WoGG - Anlage 1 (zu § 19 Abs. 1)
-    en:
-  values:
-    1984-01-01:
-      value: 0.0000632
-      note:
+faktor:
```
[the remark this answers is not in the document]
MaxBlesch: Well, this name was picked by HMG. I think I agree with you, but given it is in the `wohngeld.yaml` I think it is not necessary to change right now. There is also `koeffizenten`, which is quite general.

gettsim/data/wohngeld.yaml
```diff
@@ -1,2513 +1,1646 @@
...
+    7: 702
+    8: 787
+    9: 872
+    10: 957
+    11: 1248
+    12: 1443
+
+freib_kinder:
+  name:
+    de: Monatlicher Freibetrag pro Einkommen eines Kind, das Kindergeld bekommt im Haushalt
```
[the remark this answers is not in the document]
MaxBlesch: Corrected!

MaxBlesch:
Okay. I will drop the whole folder `data_not_implemented`. It does not matter, that this data is lost?!

gettsim/benefits/wohngeld.py
```diff
@@ -205,13 +198,17 @@ def _calc_wg_income_deductions_until_2015(household, params):
     )
     workingchild = household["kind"] & (household["bruttolohn_m"] > 0)
```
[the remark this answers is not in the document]
MaxBlesch: I don't really get what to check here?

gettsim/data/wohngeld.yaml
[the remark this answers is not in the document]
hmgaudecker: Feek free to change, it is not like I have any idea of that. Was just an example (I hope, not that I would remember exactly). Thanks!

gettsim/benefits/wohngeld.py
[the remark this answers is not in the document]
Eric-Sommer: The actual ruling (as also noted in the params) requires that the child must be eligible for child benefit (hours or income-wise). If it's not, then there's no allowance on the wohngeld-relevant income.

Eric-Sommer:
> Okay. I will drop the whole folder `data_not_implemented`. It does not matter, that this data is lost?!

Ah, these are parameters that are not used at all? Then don't delete, I though this was just some intermediate step.

MaxBlesch:
> > Okay. I will drop the whole folder `data_not_implemented`. It does not matter, that this data is lost?!
> 
> Ah, these are parameters that are not used at all? Then don't delete, I thought this was just some intermediate step.

I would probably collect these things in that folder and once we are done with all yaml we will delete. Then one can easily restore, all at once!

gettsim/benefits/wohngeld.py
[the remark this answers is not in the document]
MaxBlesch: Okay. But do you want to check this again? Because wohngeld is one of the last functions called from `tax_transfer.py`. So use it as a pre-defined variable or check again?

gettsim/data/wohngeld.yaml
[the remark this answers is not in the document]
MaxBlesch: I added some clarification!

gettsim/benefits/wohngeld.py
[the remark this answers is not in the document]
Eric-Sommer: I'd simply call `kindergeld_anspruch_regel()`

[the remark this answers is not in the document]
MaxBlesch: Moved all eligibility checks to `kindergeld`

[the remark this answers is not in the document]
MaxBlesch: I think that is the most efficient way!

PR #153, opened 2020-03-02
Added example from discussion on naming arbeitsl_geld.

amageh:
Adds example of naming ``arbeitsl_geld`` from the discussion in #79 to the documentation in GEP 1.


hmgaudecker:
Looks great, thanks! 

[Would have been perfectly fine to just push to the gep-01 branch here, too]

PR #154, opened 2020-03-04
Reorganize configuration and conda files.

tobiasraabe:

- All yaml files should end with `.yaml`.
- All conda files should live in `.conda`


.conda/meta.yaml
```diff
@@ -23,7 +23,9 @@ requirements:
 
 test:
   commands:
-    - pytest
+    # We cannot call pytest directly because the source files rest in
+    # ../work_moved_gettsim_* and not in the cwd.
```
hmgaudecker: Is `work_moved_gettsim_*` for real?

docs/geps/gep-00.rst
```diff
@@ -178,9 +178,9 @@ more feedback and looking for opportunities to compromise.
 
 If the final comment period passes without any substantive objections, then the GEP can
 officially be marked ``Accepted``. You should send a follow-up message notifying the
-community (celebratory emoji optional but encouraged 🎉✨), and then update the GEP by
-setting its ``:Status:`` to ``Accepted``, and its ``:Resolution:`` header to a link to
-your follow-up message.
+community (celebratory emoji optional but encouraged 🎉✨), and then update the GEP
+by setting its ``:Status:`` to ``Accepted``, and its ``:Resolution:`` header to a link
```
hmgaudecker: Can you revert these, please? Or did they have any meaning?

tobiasraabe: Yes, because the emoticons are one character for all editors but apparently more for doc8 which causes an max line length exceeded error.

.conda/meta.yaml
tobiasraabe: the directory is called `work_moved_gettsim_` and then some number which changes for every build. I will clarify it with `work_moved_gettsim_[0-9]+`.

docs/geps/gep-00.rst
hmgaudecker: Fine, but why does it care if the file does not change otherwise in this PR ?

tobiasraabe: Formerly, you had to use `max-line-length=89` with `doc8` because it was treated like 88 characters. Since PyCQA took over the project, they fixed the bug. Then, you created the gep. Now, I changed the setting to 88 characters. That is why the error occurs now.

tobiasraabe:
I tried to create one which is why I noticed that not all yaml files were included in the release. You have to set the automatic upload to false. `anaconda_upload: false` in your `.condarc`.

tobiasraabe:
I do not merge the PR now, because I experiment with conda noarch (https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#noarch-packages) specification

docs/geps/gep-00.rst
hmgaudecker: Thanks for the explanation!

.conda/meta.yaml
tobiasraabe: Does not apply anymore.

tobiasraabe:
Should be finished now :)

Issue #151, opened 2020-02-29
Add Elterngeld features


MaxBlesch:
So far Elterngeld Plus and Partnerschaftsbonus are not implemented yet. This only make sense once we have Lohnsteuer and thus a better proxy for net income in the last year.

hmgaudecker:
This is also a really tough one when it comes to take-up. So far, we basically assume that everyone takes up all benefits they are eligible for, which is a poor assumption in some cases ([estimates for Kinderzuschlag are 30%](http://dip21.bundestag.de/dip21/btd/19/074/1907403.pdf)).

Even harder to think of here how households solve the labour supply - childcare (- leisure) - tradeoff here. Probably only relevant in a dedicated model; my hunch is that the standard Elterngeld claim still is the norm. So it is good to have this issue for remembering that there is a gap, but I would only tackle it if somebody actually needs it.

@Eric-Sommer started a related discussion on Zulip, [cross-linking that](https://gettsim.zulipchat.com/#narrow/stream/215726-Take-Up-and.20Discretion/topic/Which.20benefit.20to.20claim.3F/near/182159872).

MImmesberger:
Cross-linking this issue for relevant income for Elterngeld calculation: #692

PR #159, opened 2020-03-29
Cleanup

Eric-Sommer:

A couple of recent PRs (#151, #152) were not integrated perfectly, requiring some cleanup

1. Elterngeld now requires additional columns. These are added to the doc.
2. The crosswalk still falsely contains `immobilien_baujahr` in classes.
3. `wohngeld_neu.yaml` was only temporary and is now redundant.
4. There is an ambiguous use of `note` and `reference` in `wohngeld.yaml`.
5. (somewhat unrelated): Two age thresholds (kindergeld and arbeitslosengeld) are still hardcoded


gettsim/data/arbeitsl_geld.yaml
```diff
@@ -126,3 +126,26 @@ agsatz1:
...
       value: 0.67
       note: Prüfen!
+ag_stundengrenze:
+  name:
+    de: Stundengrenze für Bezug von Arbeitslosengeld
+    en: Working hours limit to receive unemployment benefit
+  description:
+    de: §138 (3) SGB III, früher auch §102 (2) AFG. Grenze, ab der nicht mehr Arbeitslosigkeit unterstellt wird.
+    en:
+  values:
```
MaxBlesch: We will have to reorganize this values structure as pointed out in GEP-3. I will change as soon as the new loading function is there!

PR #160, opened 2020-03-29
include wohnbedarf in params

### LINKED ISSUE #75: Add the wohnbedarf array as parameters

Eric-Sommer:

closes #75, moving a set of parameters from code to parameters.

Also introduces a dummy function for Kinderzuschlag before 2005.


gettsim/benefits/kinderzuschlag.py
```diff
@@ -192,79 +193,8 @@ def _calc_kiz_regel_since_2011(household, params):
...
-            [62.73, 71.93],
-            [52.88, 62.96],
-            [45.70, 56.04],
-            [40.24, 50.49],
-        ],
-        "2020": [[77, 84], [63, 72], [53, 63], [46, 56], [40, 50]],
-    }
-
-    return wohnbedarf[str(yr)]
+def kiz_dummy(household, params, arbeitsl_geld_2_params):
```
MaxBlesch: Maybe we could raise a simple warning saying that there was no kinderzuschlag and we didn't calculate anything instead!

PR #161, opened 2020-04-13
regroup kindergeld

Eric-Sommer:

Regroup Kindergeld params according to #148 and adjusted code accordingly.

### DIFF
--- a/gettsim/benefits/unterhalt.py
+++ b/gettsim/benefits/unterhalt.py
@@ -32,13 +32,13 @@ def uhv_since_07_2017(tax_unit, params, kindergeld_params):
     # Amounts depend on age
     tax_unit.loc[
         (tax_unit["alter"] < 6) & tax_unit["alleinerziehend"], "unterhaltsvors_m"
-    ] = (params["uhv6_amount"] - kindergeld_params["kgeld1"])
+    ] = (params["uhv6_amount"] - kindergeld_params["kindergeld"][1])
     tax_unit.loc[
         (tax_unit["alter"] >= 6)
         & (tax_unit["alter"] < 12)
         & tax_unit["alleinerziehend"],
         "unterhaltsvors_m",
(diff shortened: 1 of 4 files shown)


gettsim/taxes/kindergeld.py
```diff
@@ -13,21 +13,16 @@ def kindergeld(tax_unit, params):
...
     ).cumsum()
 
-    kg_amounts = {
-        1: params["kgeld1"],
-        2: params["kgeld2"],
-        3: params["kgeld3"],
-        4: params["kgeld4"],
-    }
-    tax_unit["kindergeld_m_basis"] = tax_unit["kindeld_anspruch"].replace(kg_amounts)
+    tax_unit["kindergeld_m_basis"] = tax_unit["kindeld_anspruch"].replace(
```
MaxBlesch, line 20: Maybe here a little comment on how this works w.r.t. the order of the children?

gettsim/benefits/unterhalt.py
```diff
@@ -32,13 +32,13 @@ def uhv_since_07_2017(tax_unit, params, kindergeld_params):
     # Amounts depend on age
     tax_unit.loc[
         (tax_unit["alter"] < 6) & tax_unit["alleinerziehend"], "unterhaltsvors_m"
-    ] = (params["uhv6_amount"] - kindergeld_params["kgeld1"])
+    ] = (params["uhv6_amount"] - kindergeld_params["kindergeld"][1])
```
MaxBlesch, line 35: We should write, why it is the amount for the first kid. I don't see it myself!

Eric-Sommer: that's the ruling...is already commented in line 29f.

MaxBlesch: Okay, sorry!

PR #163, opened 2020-04-15
Fix encoding issue with files containing umlauts.

tobiasraabe:

Some tests failed because they relied on params in yaml files which have not been read with the correct encoding. Use utf-8 for files containing umlauts.

FYI: @Eric-Sommer @hmgaudecker @MaxBlesch


hmgaudecker:
I thought things are automatically defaulting to utf-8 these days regardless of platform. Apparently that is not true, we should then make sure to do that explicitly **everywhere**.

MaxBlesch:
Thanks Tobias. I changed before to an with open setup and it didn't work. Thanks again. I will make sure it is right **everywhere**!!

PR #164, opened 2020-04-15
Regroup Kinderzuschlag Params

Eric-Sommer:
clean up kiz paramstill need to be done.


gettsim/benefits/kinderzuschlag.py
```diff
@@ -83,8 +83,8 @@ def kiz(household, params, arbeitsl_geld_2_params):
     household["kinderzuschlag_kindereink_abzug"] = household["kindergeld_anspruch"] * (
         np.maximum(
             0,
-            params["a2kiz"]
-            - params["a2kiz_withdrawal_rate_child"]
+            params["kinder_zuschlag"]
```
MaxBlesch, line 86: Lets try to write `kinderzuschlag` as identifier everywhere!

PR #165, opened 2020-04-16
Piecewise polynomial.

MaxBlesch:
We want to incorporate the more general idea of piecewise polynomial functions. This would allow reorganize the tax_tarif, which is piecewise quadratic.


gettsim/pre_processing/piecewise_functions.py
```diff
@@ -24,22 +28,48 @@ def piecewise_linear(
...
+    for pol in range(1, rates.shape[0] + 1):
+        out += rates[pol - 1, index_interval] * (increment_to_calc ** pol)
     return out
+
+
+def add_progressionsvorbehalt(param_dict, parameter):
+    """
+    If no quadratic rate is given, add rate according to the
+    progressionsvorbehalt formula.
+    """
```
Eric-Sommer, line 63: Please extend the comment on how you obtain the "quadratic rate". what is the quadratic rate after all, i.e. what is the general functional form? [Progressionsvorbehalt](https://de.wikipedia.org/wiki/Progressionsvorbehalt) means to apply the tax tariff to two different income components and apply the tax rate from the second calculation to the first income concept. I don't think this is what happens here so we should use a diffferent name.

gettsim/benefits/arbeitsl_geld.py
```diff
@@ -61,9 +62,7 @@ def proxy_net_wage_last_year(
     prox_ssc = soz_vers_pauschale * max_wage
 
     # Fictive taxes (Lohnsteuer) are approximated by applying the wage to the tax tariff
-    prox_tax = eink_st_params["st_tarif"](
-        12 * max_wage - werbungs_pausch, eink_st_params
-    )
+    prox_tax = st_tarif(12 * max_wage - werbungs_pausch, eink_st_params)
```
Eric-Sommer, line 65: I'd argue for keeping `eink_st_params["st_tarif"]`. If we implement older tariffs with different functional forms, we have to revert this step.

gettsim/taxes/eink_st.py
```diff
@@ -53,37 +56,8 @@ def st_tarif(x, params):
...
-                + ((params["t_s"] + params["t_m"]) / 2) * (params["S"] - params["M"])
-                + ((params["t_m"] + params["t_e"]) / 2) * (params["M"] - params["G"])
-            )
-        if x > params["R"]:
-            eink_steuer = eink_steuer + (params["t_r"] - params["t_s"]) * (
-                x - params["R"]
-            )
-        assert eink_steuer >= 0
-    return eink_steuer
+    eink_st = piecewise_polynominal(x, **params["eink_st_tarif"])
```
Eric-Sommer, line 62: What to the two stars indicate?

MaxBlesch: The two stars are for mapping keywords to values in the function. See https://www.geeksforgeeks.org/args-kwargs-python/ for details!

gettsim/benefits/arbeitsl_geld.py
MaxBlesch: Its only two places in the code. If we really implement it, we can still change. But I think with implementing the DAG soon, our function assignment will change in gereal!

gettsim/data/eink_st.yaml
```diff
@@ -1,1025 +1,172 @@
...
+      upper_threshold: 8820
+    1:
+      upper_threshold: 13769
+    2:
+      upper_threshold: 54057
+    3:
+      upper_threshold: 256303
+  2018-01-01:
+    deviation_from: previous
+    note: geändert zum 1.1.2018 durch Artikel 9 vom 23.12.2016, BGBL I, S. 3000
```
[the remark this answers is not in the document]
MaxBlesch: Yes it is. Will change!

gettsim/taxes/eink_st.py
hmgaudecker: Have not checked the example closely, but in general, try to be explicit and write them out. Makes it much easier to understand the code while reading, let alone debugging.

MaxBlesch: Replaced with explicit statement

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Sorry I misunderstood the concept. Will change!

MaxBlesch: Changed explanation and name to Progressionsfaktor.

gettsim/data/eink_st.yaml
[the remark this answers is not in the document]
MaxBlesch: Adapted to the template of reference.

gettsim/data/eink_st.yaml
```diff
@@ -1,1025 +1,172 @@
...
-    en:
-  description:
-    de:
-    en:
-  values:
-    1999-01-01:
-      value: 66312
-      note:
+  note:
+    de: Der quadratische Anstieg wird nach der Formel des Progressionsvorbehalt von gettsim berechnet.
```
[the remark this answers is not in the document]
MaxBlesch: Changed explanation.

gettsim/data/eink_st.yaml
```diff
@@ -1,1025 +1,172 @@
...
+      rate_linear: 0
+      intercept_at_lower_threshold: 0
+    1:
+      upper_threshold: 12739
+      rate_linear: 0.16
+    2:
+      upper_threshold: 52151
+      rate_linear: 0.2405
+    3:
+      upper_threshold: 61377
```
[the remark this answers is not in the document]
MaxBlesch: Sorry yes. I forgot to delete that. I set it to inf.

gettsim/data/eink_st.yaml
```diff
@@ -1,1025 +1,175 @@
...
+      intercept_at_lower_threshold: 0
+    1:
+      upper_threshold: 12739
+      rate_linear: 0.15
+    2:
+      upper_threshold: 52151
+      rate_linear: 0.2397
+    3:
+      upper_threshold: inf
+      rate_linear: 0.42
```
Eric-Sommer, line 61: Can't we just drop this threshold?

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -24,22 +28,48 @@ def piecewise_linear(
```
Eric-Sommer, line 28: The docstring mentions `linear function`, while the func itself is now named `piecewise polynominal`.

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -1,7 +1,11 @@
+import copy
+
 import numpy as np
 
+from gettsim.pre_processing.generic_functions import check_threholds
+
 
-def piecewise_linear(
+def piecewise_polynominal(
     value,
```
Eric-Sommer, line 9: polynomial

gettsim/pre_processing/piecewise_functions.py
Eric-Sommer: It's still not clear to me what this Progressionsfaktor actually is (I have a faint idea because I know the tax tariff, but this is impossible to understand without deeper knowledge.) Am I right that you infer the missing parameter by imposing equal tax dues from "below" and "above" each treshold?

gettsim/data/eink_st.yaml
MaxBlesch: Unfortnunatly not, as we need to specify that this rule is not capped! HMG said to specify this explicit and with implicit leaving it out.

hmgaudecker: Yeah, better explicit than implicit. I agree that for upper thresholds, I do not see much of a use case, but for lower thresholds leaving 0 or -inf implicit is often dangerous. Then rather require it everywhere.

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Changed docstring

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Corrected!

MaxBlesch:
I got the progressionsfaktor from Wikipedia: https://de.wikipedia.org/wiki/Einkommensteuer_(Deutschland) . Ist there an actual name for this? If not maybe a website to link in the docstring?

Eric-Sommer:
> I got the progressionsfaktor from Wikipedia: https://de.wikipedia.org/wiki/Einkommensteuer_(Deutschland) . Ist there an actual name for this? If not maybe a website to link in the docstring?

OK, then it's what I assumed. I don't think it's a legal name. Let's describe it as folllows:

"The marginal rates apply only at the thresholds and increase linearly in between them. For the explicit tax function, we need to add factors such that the tax liability is a smooth function (Progressionsfaktor). We infer them ourselves here even though they are mentioned in the law."

MaxBlesch:
But are they explicitly mentioned in the law or just their calculation? If they are explicitly mentioned I should write them explicitly as the quadratic rate!

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -1,45 +1,84 @@
...
-            calculates at the respective thresholds
+        x (float, >0): The value that the function is applied to.
+        lower_thresholds (1-d array): The lower thresholds of each interval.
+        upper_thresholds (1-d array): The upper thresholds each interval.
+        rates (n-d arry): The slope in each interval. Where n is the degree of the
+        polynomial function.
+        intercepts_at_lower_thresholds (1-d array): The intercepts at the lower
+        threshold of each interval.
         rates_modified: Boolean variable indicating, that intercepts can't be used
         anymore.
```
Eric-Sommer, line 28: used anymore => intercepts from the param file cannot be used (?)

MaxBlesch: This is for the case that we have an individual piecewise linear function. I implemented it for `eink_anr_frei` in 2005. There the rates are multiplied by the nettoquote and we can't use the intercepts.

PR #167, opened 2020-04-16
Regroup alg

Eric-Sommer:
clean up alg params

### DIFF
--- a/gettsim/benefits/arbeitsl_geld.py
+++ b/gettsim/benefits/arbeitsl_geld.py
@@ -22,16 +22,20 @@ def ui(
         soli_st_params,
         beit_bem_grenz=soz_vers_beitr_params[f"rvmaxek{westost}"],
         werbungs_pausch=eink_st_abzuege_params["werbungskostenpauschale"],
-        soz_vers_pausch=params["soz_vers_pausch"],
+        sozial_versich_pauschale_alg=params["sozial_versich_pauschale_alg"],
     )
 
     eligible = check_eligibility_alg(person, params)
 
     if eligible:
         if person["anz_kinder_tu"].sum() == 0:
-            person["arbeitsl_geld_m"] = alg_entgelt * params["agsatz0"]
+            person["arbeitsl_geld_m"] = (
+                alg_entgelt * params["arbeitsl_geld_satz_ohne_kinder"]
+            )
         else:
-            person["arbeitsl_geld_m"] = alg_entgelt * params["agsatz1"]
+            person["arbeitsl_geld_m"] = (
+                alg_entgelt * params["arbeitsl_geld_satz_mit_kindern"]
+            )
     else:
         person["arbeitsl_geld_m"] = 0.0
     return person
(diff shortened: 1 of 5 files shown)


gettsim/benefits/arbeitsl_geld.py
```diff
@@ -22,16 +22,20 @@ def ui(
         soli_st_params,
         beit_bem_grenz=soz_vers_beitr_params[f"rvmaxek{westost}"],
         werbungs_pausch=eink_st_abzuege_params["werbungskostenpauschale"],
-        soz_vers_pausch=params["soz_vers_pausch"],
+        sozial_versich_pauschale_alg=params["sozial_versich_pauschale_alg"],
```
MaxBlesch, line 25: We should use `arbeitsl_geld` as identifier.

gettsim/benefits/arbeitsl_geld.py
```diff
@@ -82,5 +86,5 @@ def check_eligibility_alg(person, params):
         (1 <= mts_ue <= 12)
         & (person["alter"] < 65)
         & (person["ges_rente_m"] == 0)
-        & (person["arbeitsstunden_w"] < params["ag_stundengrenze"])
+        & (person["arbeitsstunden_w"] < params["arbeitl_geld_stundengrenze"])
```
MaxBlesch, line 89: There is a s missing: `arbeitsl_geld`

gettsim/benefits/arbeitsl_geld.py
Eric-Sommer: corrected the argument's name to  `sozial_versich_pauschale`, because it's used also used for Elterngeld.

gettsim/benefits/arbeitsl_geld.py
```diff
@@ -22,16 +22,20 @@ def ui(
         soli_st_params,
         beit_bem_grenz=soz_vers_beitr_params[f"rvmaxek{westost}"],
         werbungs_pausch=eink_st_abzuege_params["werbungskostenpauschale"],
-        soz_vers_pausch=params["soz_vers_pausch"],
+        sozial_versich_pauschale=params["sozial_versich_pauschale_arbeitsl_geld"],
```
MaxBlesch, line 25: I just realized that we use `soz_vers` as identifier for social insurances.

gettsim/data/wohngeld.yaml
```diff
@@ -24,6 +24,7 @@ koeffizienten_berechnungsformel:
     note: The keys from 1 to 12 below refer to the household size.
     1984-01-01:
       note: Parameter aus Regressionsanalyse der Wohngeldtabellen
+      reference:
```
MaxBlesch, line 27: You don't have to delete, but we don't need to insert empty keys!

gettsim/benefits/arbeitsl_geld.py
MaxBlesch: In any case, we should stick to one abbreviation. I just saw the yaml file and orientated, therefore, my guess on the identifier.

gettsim/data/wohngeld.yaml
Eric-Sommer: I'm working on `paramviz` (#122) in parallel and I'd like to have references everywhere :)

hmgaudecker: Please don't. It is very easy to work around on the code side. That is where it should go. `dict.get("reference", None)` out of the top of my head.

Eric-Sommer: fair enough

PR #168, opened 2020-04-16
Regroup elterngeld

Eric-Sommer:

clean + regroup Elterngeld parameters

### Todo

- [x] Some parameter dates were wrong. As a result, one of the test fails. Needs to be corrected
- [x] What about `elg_geschw_bonus_max`? This parameter does not seem to exist


gettsim/benefits/arbeitsl_geld.py
```diff
@@ -22,16 +22,20 @@ def ui(
         soli_st_params,
         beit_bem_grenz=soz_vers_beitr_params[f"rvmaxek{westost}"],
         werbungs_pausch=eink_st_abzuege_params["werbungskostenpauschale"],
-        soz_vers_pausch=params["soz_vers_pausch"],
+        sozial_versich_pauschale_alg=params["sozial_versich_pauschale_alg"],
```
MaxBlesch, line 25: Identifier for social insurances is `soz_vers` and for `pauschale` there is `pausch` everywhere else. Also we use `arbeitsl_geld`.

gettsim/benefits/arbeitsl_geld.py
```diff
@@ -43,7 +47,7 @@ def proxy_net_wage_last_year(
     soli_st_params,
     beit_bem_grenz,
     werbungs_pausch,
-    soz_vers_pausch,
+    sozial_versich_pauschale_alg,
```
MaxBlesch, line 50: Same as before.

PR #170, opened 2020-05-01
Regrouping ges_renten_vers.yaml

MaxBlesch:
I just moved the variables which are not used to `data_not_implemented`. I think it makes not much sense to regroup this file in the way, I did it with the others. 

If I remember correctly @hmgaudecker wanted to discuss the general purpose of a simulated pension claim? or just keeping the update of the earning points?

If we keep the pension function in the current implementation, then I propose a different structure. The variables reflect exogenous data like meanwages, pension volumina etc. over the whole german population. This is not the same as the policy parameters we have in the other files. So lets move them to a different place, to have a clear separation. Also, we could have a script there, which transfers the current layout into a pickled DataFrame, which we then actually load and use as exogenous data. Maybe this dataframe would then just contain the relevant data, which is rentenwert and meanwages for updating the earning points.

@Eric-Sommer


hmgaudecker:
Yeah, the main problem is that stuff like the earnings giving you one Entgeltpunkt will historically come from the data, but they will be endogenous in lots of interesting models looking into the future.

So they should indeed be added from outside the model; we could have pre-processing step and/or require users to tuck it onto the data.

MaxBlesch:
I just realized some silent assignments we had so far. So if the year of the data was above 2017, we calculated the rentenwert manually, but with the parameters of 2016. I don't think, we should silently do that. I will add now a year check do the pensions function, s.t. that it returns nan if we are not in the year range, where we actually do have data! This range is 2005 til 2016.

PR #172, opened 2020-05-02
Regrouping Unterhalt

Eric-Sommer:

Regroup unterhalt parameters and outsource unused ones.

### DIFF
--- a/gettsim/data/unterhalt.yaml
+++ b/gettsim/data/unterhalt.yaml
@@ -1,420 +1,51 @@
-uhv6_share:
+mindestunterhalt:
   name:
-    de: Unterhaltsvorschuss für Kinder bis 6 Jahre. Anteil am Kinderfreibetrag.
-    en:
+    de: Mindestunterhalt für Kinder
+    en: Minimum Child Alimony
   description:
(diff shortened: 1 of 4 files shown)


gettsim/data/unterhalt.yaml
```diff
@@ -1,420 +1,51 @@
...
+  name:
+    de: Mindestunterhalt für Kinder
+    en: Minimum Child Alimony
+  description:
+    de: § 1612a BGB, § 1 Mindesunterhaltsverordnung. Mindestunterhalt für Kinder in Abhängigkeit des Alters.
+    en:
+  type: euro
+  time: month
+  2016-01-01:
+    6: 335
```
MaxBlesch, line 11: I would either add some more explanation: Saying kids under the age given or in between. Even better I think would be to give the keys not integer values and instead say: `unter_6`, `6_bis_12`etc.

Eric-Sommer: I went for the former (more concise keys)

MaxBlesch: Okay!

PR #176, opened 2020-05-13
Gep 04

MaxBlesch:

Pin down the layout and use of directed acyclic graphs in gettsim
.


hmgaudecker:
@MaxBlesch Can you please close and open anew, directly off master? The first bit is super confusing because it seems to be off the old GEP-01 branch?

PR #177, opened 2020-05-13
DAG rewrite: Social insurance

MaxBlesch:

We want to implement the dag for the social insurance contributions. Furthermore, we want to change the tests to the new interface and implement first wrapper for all other functions.

### Todo

- [x] Test for length of function names to fulfill gettsim's character limit on column names. Limit is 17 characters.
- [ ] Return dataframe
- [ ] Raise error as discussed below.


MaxBlesch:
So all tests for the tax and transfer functions call `compute_tax_and_transfer` and not the functions themself. We can now start to split all functions in their separate subfunctions to calculate each column. In the end we should reactivate the overall test to see if we have all columns, which are now defined as the desired outputs in `tax_and_transfer.py`. 

Furthermore I would rewrite `policy_for_date`, such that it returns one dictionary of several param groups and a function dictionary specifying year dependent policies, which we can then give to our dag function. 

Also we should decide if we want to offer a wrapper which just needs the dataframe and the policy year!

Maybe we also want to adapt  `compute_tax_and_transfer`, such that it takes a dataframe and yields also one. @tobiasraabe

hmgaudecker:
> It is not a particular test issue. There just will be different data structures. Every function name represents a data column. So if some user has the information on how much `arbeitslosengeld_m` an individual receives he does not want to calculate this variable. But it may be a dependency on some other variable, the user wants to calculate. Before this change, the DAG would have selected the function for the variable and not the given data.

And I believe this will be a common case. People do not RTFM and if they are overwhelmed, they more often than not do not bother toread warnings. 

> The rule implemented, just induces a priority of data over functions. We could send out a message to the user saying: "gettsim used `arbeitslosengeld_m` from the input data and didn't calculate it."

Let us please forget about messages for the reason cited above. 

I think we should require users to be explicit here. That is:

- We raise an error if a data column has the same name as a function in our taxes and transfers system.
- We specify that a column should override an existing function. This is exactly the same as for `user_functions`. Add another keyword `user_columns` ?

MaxBlesch:
I think that is a good way out! But then the error message should look something like:
"Your data provides "arbeitslosengeld_m" which is one of the calculated outputs. If you want your data to be used instead, please specify "arbeitslosengeld_m" in `user_columns`. "

Not sure about the exact phrasing, but the error message should definitely contain the `user_columns` alternative! If @tobiasraabe changes this layout, I will adapt the tests!

tobiasraabe:
Sounds good. I will add it to my todos.

hmgaudecker:
ooops, did not send that off earlier...

> But then the error message should look something like:

Oh yes, definitely!

Maybe, to be refined: 

```python
formatted_cols = ', '.join([c for c in overlapping_cols])
n_cols = len(overlapping_cols)

f"""Your data provides {formatted_cols}, which {'is' if  == 1 else 'are'} already present among
the functions of the taxes and transfers system. 

- If you want a data column to be used instead of calculating it within GETTSIM,
  please specify it among the *user_columns*.
- If you want a data column to be calculated internally by GETTSIM, do not pass it
  to {the_function_you_called_if_available} / GETTSIM.

"""
```

PR #178, opened 2020-05-14
GEP 4 - Directed Acyclic Graph

MaxBlesch:

Pin down the layout and use of directed acyclic graphs in gettsim.

### DIFF
--- a/docs/geps/gep-04.rst
+++ b/docs/geps/gep-04.rst
@@ -1,0 +1,272 @@
+=========================================
+GEP 4 — A DAG-based Computational Backend
+=========================================
+
++------------+---------------------------------------------------------------+
+| Author     | `Max Blesch <https://github.com/MaxBlesch>`_                  |
++            +---------------------------------------------------------------+
+|            | `Janos Gabler <https://github.com/janosg>`_                   |
++            +---------------------------------------------------------------+
+|            | `Hans-Martin von Gaudecker <https://github.com/hmgaudecker>`_ |
++            +---------------------------------------------------------------+
+|            | `Tobias Raabe <https://github.com/tobiasraabe>`_              |
++------------+---------------------------------------------------------------+
+| Status     | Draft                                                         |
++------------+---------------------------------------------------------------+
+| Type       | Standards Track                                               |
++------------+---------------------------------------------------------------+
+| Created    | 2020-05-05                                                    |
++------------+---------------------------------------------------------------+
+| Resolution | <url> (required for Accepted or Rejected or Withdrawn)        |
++------------+---------------------------------------------------------------+
+
+
+
+Abstract
+--------
+
+This GEP lays out the plan to build a DAG-based computational backend for gettsim. A
+directed acyclic graph (DAG) is common way to represent the relation between multiple
+tasks which depend on each other via inputs and targets. The same is true for gettsim
+where taxes and transfers depend on a multitude of observed variables in the data or on
+pre-computed values based on the data and parameters.
+
+
+Motivation and Scope
+--------------------
+
+The change is motivated by two primary reasons.
+
+1. gettsim should not only deliver a fixed state of a tax and transfer system, it should
+   also offer a comprehensive but simple API to alter the system.
+
+2. gettsim is slow because the current structure does not allow to model only a specific
+   part of the tax and transfer system while shutting all unrelated computation down.
+
+
+Setup of the DAG
+----------------
+
+Before we explain how the user interface changes, it is necessary to understand how the
+DAG is created and especially how the interdependency between variables is traced. We
+focus on the following graph as an example.
+
+.. image:: dag-example.png
+
+In this graph, nodes without any arrow pointing at them like ``bruttolohn_m`` and
+``wohnort_ost`` form the roots of the graph. Behind these nodes are variables which can
+be observed in a given data set. The data set inside the backend is not a
+``pandas.DataFrame``, but a dictionary of variable names and ``pandas.Series``'. At the
+start, the data dictionary look like this:
+
+.. code-block:: python
+
+    data = {"bruttolohn_m": pd.Series(...), "wohnort_ost": pd.Series(...)}
+
+All other nodes in the graph cannot be found in the data set. They are functions on
+input variables and parameters. The function are collected in a dictionary similar to
+the data dictionary where keys are the names of variables and values are functions.
+
+.. code-block:: python
+
+    functions = {
+        "regulär_beschäftigt": regulär_beschäftigt,
+        "krankv_beitr_bemess_grenze": krankv_beitr_bemess_grenze,
+    }
+
+
+For example, ``regulär_beschäftigt`` is defined like this:
+
+.. code-block:: python
+
+    def regulär_beschäftigt(bruttolohn_m, params):
+        return bruttolohn_m.ge(params["geringfügige_eink_grenzen"]["midi_job"])
+
+The function takes in ``bruttolohn_m`` which is a ``pandas.Series`` from the input data
+and ``params`` which is a dictionary of parameters. The result of the function is a
+``pandas.Series``.
+
+We can infer two things from this function.
+
+1. The name of the function is the name of its produced variable. The returned
+   ``pandas.Series`` will be associated with the name ``regulär_beschäftigt``.
+
+2. The names of the arguments identify which inputs are needed by this function. Before
+   the function is called, the backend looks at the names of its arguments, detects
+   whether all variables can be found in the data dictionary, and if so, passes the
+   arguments in the correct order to the function to receive the resulting variable
+   ``regulär_beschäftigt``.
+
+The same logic applies to ``krankv_beitr_bemess_grenze`` which looks like this:
+
+.. code-block:: python
+
+    def krankv_beitr_bemess_grenze(wohnort_ost, params):
+        pass
+
+Functions which not only depend on input data but also or solely on pre-computed inputs
+like ``lohn_krankv_regulär_beschäftigt`` are by no means different to the previous two
+functions.
+
+.. code-block:: python
+
+    def lohn_krankv_regulär_beschäftigt(
+        regulär_beschäftigt, krankv_beitr_bemess_grenze, params
+    ):
+        pass
+
+The difference is how the backend postpones the execution of this function until all
+input variables have been generated. Thus, ``lohn_krankv_regulär_beschäftigt`` would be
+executed in the third position and ``krankv_beitr_regulär_beschäftigt`` in the fourth.
+
+
+Usage and Impact
+----------------
+
+Having established how the graph is built and how the interdependency between functions
+is detected, this section lays out the changes to the user interface and other
+advantages of having a DAG-based backend.
+
+At first, a user takes the tax and transfer system as given which is defined by a
+collection of parameters and a dictionary of pre-defined functions like the ones in the
+previous section.
+
+If the user wants, for example, to alter definition of ``krankv_beitr_bemess_grenze``
+there exist multiple ways to achieve that.
+
+.. code-block:: python
+
+    def krankv_beitr_bemess_grenze(wohnort_ost, params):
+        out = np.where(
+            wohnort_ost,
+            params["beitr_bemess_grenze"]["ges_krankv"]["ost"],
+            params["beitr_bemess_grenze"]["ges_krankv"]["west"],
+        )
+
+        s = pd.Series(out, name="krankv_beitr_bemess_grenze", index=wohnort_ost.index)
+
+        return s
+
+1. The user can change the parameters responsible for the contribution assessment
+   ceiling.
+
+2. If the user wants to change the function completely to a version without any ceiling,
+   she writes a new function.
+
+.. code-block:: python
+
+    def keine_beitr_bemessungsgrenze(wohnort_ost):
+        return pd.Series(data=np.inf, index=wohnort_ost.index)
+
+Now, the new function must replace the old version. Thus, the interface function
+``tax_transfer`` receives a new argument called ``functions`` which accepts dictionaries
+(also paths to modules, etc.). If keys between the internal function dictionary and the
+the dictionary provided by the user overlap, the user's functions are preferred. If
+functions do not exist in the internal but in the user's dictionary, the user's
+functions extend the existing collection.
+
+Thus, the user would create the following dictionary
+
+.. code-block:: python
+
+    user_functions = {"krankv_beitr_bemess_grenze": keine_beitr_bemessungsgrenze}
+
+and pass it to
+
+.. code-block:: python
+
+    tax_transfer(..., functions=user_functions, ...)
+
+As the DAG-based backend offers many ways to alter the predefined tax and transfer
+system delivered with gettsim, a part of the documentation should be dedicated to this
+problem. We propose a collection of abstract patterns which can be followed and combined
+by users to achieve their goals.
+
+The second major change to the interface of ``tax_transfer`` is an argument called
+``targets`` which allows the user to specify the variables she is ultimately interested
+in. This allows to select a subset of the whole DAG which forms the tax and transfer
+system and executes only the computations defined in the subset.
+
+What are other benefits of using a DAG?
+
+- Visual debugging of the system
+- Pruning with ``targets``
+- Avoiding duplicate computations.
+
+
+Backward compatibility
+----------------------
+
+We do not expect any backward compatibility issues because, up to this moment, the tax
+and transfer system was not available for modification. The additional argument
+``functions`` and ``targets`` for ``tax_transfer()`` are optional and its defaults
+deliver the same results as before.
+
+
+Detailed description
+--------------------
+
+*Under development.*
+
+This section should provide a detailed description of the proposed change. It should
+include examples of how the new functionality would be used, intended use-cases and
+pseudo-code illustrating its use.
+
+
+Related Work
+------------
+
+- The `openfisca <https://github.com/openfisca/>`_ project uses an internal DAG as well.
+- Scheduling computations on data with task graphs is how `Dask
+  <https://docs.dask.org/>`_ splits and distributes computations.
+
+
+Implementation
+--------------
+
+*Under development.*
+
+This section lists the major steps required to implement the GEP.  Where possible, it
+should be noted where one step is dependent on another, and which steps may be
+optionally omitted.  Where it makes sense, each step should include a link to related
+pull requests as the implementation progresses.
+
+Any pull requests or development branches containing work on this GEP should be linked
+to from here.  (A GEP does not need to be implemented in a single pull request if it
+makes sense to implement it in discrete phases).
+
+
+Alternatives
+------------
+
+We have not found any alternatives which offer the same amount of flexibility and
+computational advantages.
+
+
+Discussion
+----------
+
+*Under development.*
+
+This section may just be a bullet list including links to any discussions regarding the
+GEP:
+
+- Links to relevant GitHub issues, pull requests.
+- Discussion on XXX
+
+
+References and Footnotes
+------------------------
+
+.. [1] Each GEP must either be explicitly labeled as placed in the public domain (see
+       this GEP as an example) or licensed under the `Open Publication License`_.
+
+.. _Open Publication License: https://www.opencontent.org/openpub/
+
+.. _#general/geps: https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/GEPs
+
+
+Copyright
+---------
+
+This document has been placed in the public domain. [1]_
--- a/docs/geps.rst
+++ b/docs/geps.rst
@@ -15,4 +15,5 @@ These GEPs are currently in place:
 
    geps/gep-00.rst
    geps/gep-01.rst
+   geps/gep-04.rst
    geps/gep-template


hmgaudecker:
@tobiasraabe, would be great if you could spend max half a day on this before you move on to greener pastures.

The focus should be on the new design, less on comparisons to the past. The latter is relevant only to a very small set of people, if at all. The new layout will be here to stay, I am sure!

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,185 @@
...
++------------+---------------------------------------------------------------+
+| Resolution | <url> (required for Accepted or Rejected or Withdrawn)        |
++------------+---------------------------------------------------------------+
+
+
+Abstract
+--------
+
+This GEP explains the DAG-based computational backend for gettsim which does not only
+increase performance, but, and more importantly, offers a way to make small changes to
```
janosg, line 28: "small changes" sounds like you cannot use gettsim to model radically different policies. You can make small or big changes to the tax system but the required change in code is almost as small as it can be.

janosg (approved):
Except for the small comment I think the GEP is very understandable and well written.

docs/geps/gep-04.rst
hmgaudecker: Fixed.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,177 @@
...
+
+The second problem calls for a real classic of computer science. We can visualize the
+relationship between functions and their input variables in a graph where nodes are
+variables in the data or computed by functions. Edges are pointing from from input
+variables to other variables which require them to be computed. See this `tutorial
+<../visualize_the_system.ipynb>`_ for some graphics.
+
+The resulting structure is a special kind of graph which is called a directed-acyclic
+graph (DAG). Directed and acyclic means that there exist no path of vertices which
+starts at some node :math:`\vega` and ends at the same node. Equivalently, a DAG has a
```
hmgaudecker, line 112: @tobiasraabe 

```
:math:`\vega`
```

does not render, do you remember what you had in mind here?

hmgaudecker: @tobiasraabe, any memories on that symbol?

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,177 @@
...
+The result of this function is again a :class:`pandas.Series` which has the name
+``kindergeld``, the same name as the function. Another function, say
+
+.. code-block:: python
+
+    def benefits(kindergeld, arbeitsl_geld_2, params):
+        pass
+
+would need to have ``kindergeld`` as a name for an input argument to request this
+:class:`pandas.Series`.
```
ChristianZimpelmann, line 100: Make sure somewhere here, that valid inputs are: parameter values, input variables, other functions, summations of other functions (with the new change proposed in #281)

ChristianZimpelmann (commented):
Just got to do the review.

It is well-written and I like the structure.

A short section about the automatic summation of values would fit nicely.

hmgaudecker:
@ChristianZimpelmann: Can you please re-write this so that it does not reference the situation of GETTSIM 0.2 or whatever the pre-DAG version was? Might be of historical interest, but it would be clearer if it just described the current implementation without reference to something only ancient people like myself or Eric might remember.

ChristianZimpelmann:
In commit `4b4bd385f9f617185cf725aff4de288bc51a8797` I started rewriting the GEP. However, I need some more input before finishing. In particular, I had the following two problems:

- The section names given in the [default Template](https://gettsim.readthedocs.io/en/stable/geps/gep-template.html#gep-template) seem to be chosen from the perspective of a new requested feature (Backward compatibility, Impact, Alternatives)
, but this GEP should be much more about the current state. How strongly should we deviate from the default structure?
- I was not sure how much detail we would like to give about the actual implementation. E.g., should we describe the interface of `compute_taxes_and_transfers` or do we describe the backend more abstract?

After some feedback, I can go over the GEP another time.

hmgaudecker:
> The section names given in the [default Template](https://gettsim.readthedocs.io/en/stable/geps/gep-template.html#gep-template) seem to be chosen from the perspective of a new requested feature (Backward compatibility, Impact, Alternatives), but this GEP should be much more about the current state. How strongly should we deviate from the default structure?

Yeah, right now we are using these guys more to structure our discussions / document the state rather than altering existing state. Just adapt the structure as you see fit, feel free to completely drop irrelevant sections.

> I was not sure how much detail we would like to give about the actual implementation. E.g., should we describe the interface of `compute_taxes_and_transfers` or do we describe the backend more abstract?

I think this is more about the backend, the interface of `compute_taxes_and_transfers` does not matter.

ChristianZimpelmann:
- I finished now the first draft of the GEP. Please let me know what you think.
- I also merged `gep-03` into this branch. Unfortunately, I forgot to squash merge.. sorry for the spam of commits!

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,210 @@
...
+
+.. code-block:: python
+
+    if year < 2009:
+        functions["sum_brutto_eink"] = sum_brutto_eink_mit_kapital
+    else:
+        functions["sum_brutto_eink"] = sum_brutto_eink_ohne_kapital
+
+
+However, all functions will be present in all years in principle.
```
ChristianZimpelmann, line 178: This is not true. In the DAG, only the function `sum_brutto_eink` would be present in all years and only this argument can be used in other functions. We should probably describe this feature in this GEP in more detail.

The problem case is rather when we don't want to subsume the two concepts in a common name (as it is probably the case for `arbeitsl_hile`) or if a function does no longer exist (see `kinderzuschlag`).

hmgaudecker: Sorry, my language was not precise enough.

You are right when it comes to the graph seen by the user. She would only see `sum_brutto_eink` indeed because this explicitly done in `set_up_policy_environment` / `load_reforms_for_date`. However, during the construction of the DAG as part of `compute_taxes_and_transfers`, the two underlying functions are present and one will just not be used. 

So my overall point is that it is pretty arbitrary right now what constitutes a policy that changes over time and thus ends up in `load_reforms_for_date` / `policy_functions` and what does not. I think we can and should be smarter than what we currently have.

docs/geps/gep-04.rst
```diff
@@ -186,7 +186,35 @@ This is not satisfactory; ideally, functions would only be present in years wher
...
 
+We implemented a small set of additional features that simplify the specification of
+certain types of functions of the tax and transfer system.
+
+Aggregation of variables
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Arguments of functions can also be aggregations of input variables or outputs of other
+functions on the tax unit or household level. In the example above, one argument of the
+function ``soli_st_tu`` is ``abgelt_st_tu``. This variable would be typically specified
```
hmgaudecker, line 201: I think a real-life example would be better here. I do not expect people to read through this and keep in mind the precise example from a few paragraphs ago, just in order to re-use it with a hypothetical case? (`abgelt_st_tu` is not fully obvious because of the exemption amount, which could be on individual or joint accounts, I guess).

docs/geps/gep-04.rst
```diff
@@ -186,7 +186,35 @@ This is not satisfactory; ideally, functions would only be present in years wher
     - Users putting together their policy environment based on functions from different years
     - ...
 
+Additional functionalities
+--------------------------
 
+We implemented a small set of additional features that simplify the specification of
+certain types of functions of the tax and transfer system.
+
+Aggregation of variables
```
hmgaudecker, line 196: Why call it aggregation if it is always the sum? Is it?

docs/geps/gep-04.rst
```diff
@@ -186,7 +186,35 @@ This is not satisfactory; ideally, functions would only be present in years wher
...
+function ``soli_st_tu`` is ``abgelt_st_tu``. This variable would be typically specified
+as a function on the tax unit level (``abgelt_st_tu``) as it is done above. However, if
+a function with this name does not exist, but a function on the individual level with
+the same name excluding the ``_tu`` suffix (``abgelt_st``), the output of this
+individual level function is just summed up on the tax unit level before it enters the
+function ``soli_st_tu``.
+
+This feature avoids the need to define a bunch of similarly looking functions that only
+aggregate values on certain levels. Similarly, functions with the ``_hh`` can be
+aggregations on the household level.
```
hmgaudecker, line 210: Just refer to all units defined in GEP-01?

As a general note on this section, I think I would prefer the motivation in the last paragraph first and describe the concept in an abstract faction. Then an example.

docs/geps/gep-04.rst
Eric-Sommer: > So my overall point is that it is pretty arbitrary right now what constitutes a policy that changes over time and thus ends up in `load_reforms_for_date` / `policy_functions` and what does not. I think we can and should be smarter than what we currently have.

Thinking about it, I'd put it this way: 

1. As a principle, any parameter should be exactly specified as in the law. Do not use parameters derived from actual ones, as it is currently done for social insurance contributions or as it has been done before with Kinderzuschlag.
2. As a consequence, any formal representation of a policy will be close to the text of the law. There are limits to this as e.g. the income tax law is written in a horrible manner which you don't want to see formalized.

Now to the decision on whether we need a new function
3. If a parameter change can be implemented by using the current function, don't create a new function. 
4. If one or several **new parameters** are introduced and can be added to the existing framework by setting to some 'neutral value' (e.g. 0, 1, null) for years prior to its introduction, one should do that instead of creating a new function. An example for this is the introduction of the 'Zusatzfreibetrag' for health insurance contributions (#323). Applying this rule would call for single function for `alleinerziehend_freib_tu`, as it would be sufficient to set `eink_st_abzuege_params["alleinerziehenden_freibetrag_zusatz"]` to 0 until 2014 https://github.com/iza-institute-of-labor-economics/gettsim/blob/426dc9a6a755311876e5920f7b3289bc52fd0157/gettsim/taxes/zu_verst_eink/freibetraege.py#L40-L92 
5. If this is not possible, set up a new function.

ChristianZimpelmann: That's very much what I had in mind! Sounds good

hmgaudecker: Sounds great! A couple of reactions:

> 1. Do not use parameters derived from actual ones, as it is currently done for social insurance contributions

At which point? In the parameters files? Yes. In the policy_params dict? Probably no. Both of those would be in the scope of GEP 3. Anything I am missing?

> 2. As a consequence, any formal representation of a policy will be close to the text of the law. There are limits to this as e.g. the income tax law is written in a horrible manner which you don't want to see formalized.

As a general rule, I agree. But I would not want to really constrain our implementations here. See, e.g., the proposal to change the calculation of social insurance contributions #332 

> 4. If one or several new parameters are introduced and can be added to the existing framework by setting to some 'neutral value' (e.g. 0, 1, null) for years prior to its introduction, one should do that instead of creating a new function. 

This is exactly what I'd hope we will be able to (largely) avoid by some clever re-design of how we deal with functions that change over time. Maybe that won't be possible, but it should be worth some brainstorming sessions.

Eric-Sommer: > At which point? In the parameters files? Yes. In the policy_params dict? Probably no. Both of those would be in the scope of GEP 3. Anything I am missing?

I was primarily referring to the parameter files.

> As a general rule, I agree. But I would not want to really constrain our implementations here. See, e.g., the proposal to change the calculation of social insurance contributions #332

Agreed, this is all guidelines, not the law😀. I'd just like to stress that aligning python code to legal code where possible has value in itself.

> This is exactly what I'd hope we will be able to (largely) avoid by some clever re-design of how we deal with functions that change over time. Maybe that won't be possible, but it should be worth some brainstorming sessions.

I'm just afraid this clever re-design migh lead to functions that have little in common with what can be read in the law. The last thing real-life

hmgaudecker: > I'm just afraid this clever re-design might lead to functions that have little in common with what can be read in the law.

My hope would be just the opposite -- being able to write functions that look like the law at a particular point in time, which are not littered with parameters becoming effective only years later etc.. Let's see, you are definitely raising great points!

docs/geps/gep-04.rst
ChristianZimpelmann: I restructured the section as suggested.

docs/geps/gep-04.rst
ChristianZimpelmann: So what's the plan? Keep the PR (and this comment) open until the brainstorming session (maybe during the workshop in April)?

hmgaudecker: Yes. No need to put the current situation into writing, GEPs are exactly the process for channeling the brainstorming.

In the beginning, they just end up being larger and taking longer than what they will eventually be (GEP 5 would be a good example of the latter, though our process was not ideal).

hmgaudecker:
Thanks! Just going through it. I do not quite understand this sentence:

> Arguments of functions can also be summations of outputs of other functions on the household level.

ChristianZimpelmann:
> Arguments of functions can also be summations of outputs of other functions on the household level.

I changed it to "The household level sum of individual level functions can be used as targets or as arguments of other functions."

Is that clearer?

hmgaudecker:
Thanks!

I streamlined it a bit and added more on what exactly is happening in terms of the graph (a node is added with parents `[column]` and `hh_id`). 

I also added a similar section on converting between time units of reference periods so we can get rid of all those hard-coded `12 * ...` bits.

Please have a look! https://gettsim--178.org.readthedocs.build/en/178/geps/gep-04.html#group-summation-of-variables

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,303 @@
...
+
+
+.. _gep-4-time-unit-conversion:
+
+Conversion between reference periods
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Similarly to :ref:`gep-4-group-sums`, GETTSIM will automatically convert values
+referring to different reference periods defined in :ref:`gep-1` (years (default, no
+suffix), months ``_m``, and days ``_t``).
```
ChristianZimpelmann, line 244:
```suggestion
suffix), months ``_m``, weeks ``_w``, and days ``_t``).
```

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,303 @@
...
+Conversion goes both ways and uses the following formulas:
+
++-----------+--------+---------+
+| time unit | suffix | factor  |
++-----------+--------+---------+
+| year      |        | 1       |
++-----------+--------+---------+
+| month     | `_m`   | 12      |
++-----------+--------+---------+
+| week      | `_w`   | 52.1775 |
```
ChristianZimpelmann, line 265: I get 365.25 * 7 ~ 52.1785. Should be consistent between week and day

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,303 @@
...
+        ...
+
+a node ``kindergeld_m_hh`` containing the household-level sum of ``kindergeld_m`` will
+be automatically added to the graph. Its parents in the graph will be ``kindergeld_m``
+and ``hh_id``.
+
+
+.. _gep-4-time-unit-conversion:
+
+Conversion between reference periods
```
ChristianZimpelmann, line 239: Great addition!

docs/geps/gep-04.rst
hmgaudecker: Excellent, thanks for checking! I confess I just used google... Fixed in f97ce.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,316 @@
...
+
+In the yaml-files corresponding to a particular tax / transfer, functions not present in
+all years will need to be listed with along with the dates for when they are active. See
+:gep-3-keys-referring-to-functions: for the precise syntax. That mechanism should be
+used for:
+
+1. Functions that are newly introduced.
+2. Functions that cease to be relevant.
+3. Functions whose interface changes over time.
+4. Functions whose body changes so much that
```
ChristianZimpelmann, line 199: I don't understand the difference between point 3 and 4.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,316 @@
...
+
+.. _gep-4-group-sums:
+
+Group summation of variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Individual-level measures often need to be summed up to a group level. <GEP-2 describes
+`gep-2-aggregation-functions`> how reductions are handled in principle.
+
+Because summations are very common, GETTSIM's DAG backend automates this process for
```
ChristianZimpelmann, line 225: We need to describe somewhere how users can aggregate functions. This won't be straightforward once we move to scalar. But probably best to wait until we know more about the implementation.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,316 @@
...
+4. Functions whose body changes so much that
+
+   - it is useful to signal that things have changed and/or
+
+   - it would be awkward to program the different behaviors in one block with case
+     distinctions.
+
+Needless to say, the different reasons may appear at different points in time for the
+same function.
+
```
ChristianZimpelmann, line 208: Should we require to have a decorator for those functions that indicates that a function doesn't exist everytime and in which yaml-file the specification is found?

This would be similar to the specification of rounding parameters. One difference is that the information in the `yaml`-files will be used already during `setup_policy_environment` and won't become part of the policy_params dictionary.

hmgaudecker: Nice idea! As we expand our temporal scope, I'd think that the number of functions present in all years will shrink quite dramatically. So not sure what the default will be in an ideal case. 

In any case, it will be good to have a clear idea of where things can be found. Whether that's via a decorator or a clear(er) mapping between the parameters/...yaml and the location of the function, let's see.

docs/geps/gep-04.rst
hmgaudecker: Yes, in GEP-2. The only thing relevant for the DAG is the automatic conversion because it injects nodes into the DAG. What other nodes do precisely, does not matter from the DAG's perspective (it will continue to work on columns, same behaviour as we have currently).

docs/geps/gep-04.rst
ChristianZimpelmann: Ok, makes sense!

hmgaudecker (commented):
*ignore this*

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,355 @@
...
+(single-responsibility principle). Another advantage is that each function is a
+potential entry point for a researcher to change the taxes and transfers system if she
+is able to replace this function with her own version.
+
+See the following example for capital income taxes.
+
+.. code-block:: python
+
+    def abgelt_st_tu(
+        zu_verst_kapital_eink_tu: FloatSeries, abgelt_st_params: dict
```
janosg, line 97: - With the switch to arrays/scalars, the example should maybe not use `FloatSeries` as type hint anymore.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,355 @@
...
+    ) -> FloatSeries:
+        """Calculate abgeltungssteuer per tax unit.
+
+        Parameters
+        ----------
+        zu_verst_kapital_eink_tu
+            See :func:`zu_verst_kapital_eink_tu`.
+        abgelt_st_params
```
janosg, line 105: I would prefer not to repeat the function name in the params argument. I.e. just `params` instead of `repeated_function_name_params`. This could either be achieved with the partial approach or with a similar approach as in my `dags` PR from yesterday. Here is a [test case](https://github.com/OpenSourceEconomics/dags/blob/2b081b9c92217db1732e81e9099436e78b038fcf/tests/test_dag.py#L124) that illustrates what I mean.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,355 @@
...
+already computed before a function that requires them is called.
+
+In order to calculate a set of taxes and transfers, GETTSIM builds a DAG based on three
+inputs provided by the user:
+
+ - Input data.
+ - A set of functions representing the taxes and transfers system, which consist of the
+   ones pre-implemented in GETTSIM and potentially user-written additional functions.
+   Parameters of the taxes and transfers system (see :ref:`gep-3` will already be
+   partialled into these functions, so they can be ignored in the following). These
```
janosg, line 169: The `partial` part is an implementation detail that I would not mention here.

janosg (commented):
I made a few comments that I think would improve the GEP. All of them are optional and should not block approval of the GEP.

docs/geps/gep-04.rst
ChristianZimpelmann: It is actually not `repeated_function_name_params` but the first part is associated to a parameter group (e.g. `eink_st`, `wohngeld`, etc.). Just a coincidence that the parameter group is the same as the function here.

I think we need to and should keep the parameter group in the argument. (sometimes even parameters from two parameter groups are used)

docs/geps/gep-04.rst
ChristianZimpelmann: sounds reasonable

hmgaudecker: Sure. Have not touched the type hints. Do type hints do anything for Numpy / Jax, btw?

docs/geps/gep-04.rst
hmgaudecker: Agreed that it is an implementation detail. Seems to me that ignoring those arguments in the DAG is important for the text, though. New version in [576d03a](https://github.com/iza-institute-of-labor-economics/gettsim/pull/178/commits/576d03a80db263d18b25d3b4bd6272b4e7f0cf11) better?

ChristianZimpelmann:
Two questions came to mind with respect to the aggregation function:

### Cumsum

We currently make use of a `cumsum` aggregation (in `kindergeld.py`) which I would like to implement in `aggregation.py`. In contrast to the other modes of aggregation, the order of household members is relevant in this case. So it might be helpful to include a key `sort_cols` to sort the household members before `cumsum` is applied (see an example below). Note that this sorting isn't relevant for the current use case in the code base and I would most likely not implement it already in #356. Also note that I am not sure if we can support "cumsum" in the fast jax implementation

```python
    aggregation_kindergeld = {
        "anz_erwachsene_tu": {"source_col": "kindergeld_anspruch", "aggr": "cumsum", "sort_cols": ["age"]},
    } 
```

### Rounding

Currently, we have no possibility to add the `@add_rounding_spec(params_key="XXX")` decorator to an aggregated function. 
So I would propose a new (optional) key `"rounding_spec_params"` in the `aggregation_` dictionary.

hmgaudecker:
Thanks, great points! 

> ### Cumsum

I would like to postpone this for the moment. Once we do kindergeld properly (assigning it to the kid / the parent receiving it), this should not matter anymore. If it does, we can tackle it then.

I'd say to just do some workaround for the moment. 


> ### Rounding

Do you have an example where that matters? That is, why would we ever do the rounding in the aggregation step?

ChristianZimpelmann:
> > ### Cumsum
> 
> I would like to postpone this for the moment. Once we do kindergeld properly (assigning it to the kid / the parent receiving it), this should not matter anymore. If it does, we can tackle it then.

Fine with me. It isn't needed at the moment either.

> > ### Rounding
> 
> Do you have an example where that matters? That is, why would we ever do the rounding in the aggregation step?

I am not sure if `unterhaltsvors_m_hh` is rounded (or e.g. only payed out above some threshold as it is the case for Wohngeld). If yes, this would be an example as we only calculate `unterhaltsvors_m` in the code directly. I don't see another example in the code right now.

I don't think we urgently need this functionality, but it would be an easy addition to the GEP. It feels like something is missing if we allow for some functions to be rounded, but not for others.

docs/geps/gep-04.rst
```diff
@@ -0,0 +1,356 @@
...
+        "haushaltsgröße_hh": {"aggr": "count"},
+    }
+
+The group identifier (``tu_id``, ``hh_id``) will be automatically included as an
+argument; for ``count`` no other variable is necessary.
+
+The output type will be the same as the input type. Exceptions:
+
+- Input type ``bool`` and aggregation ``sum`` leads to output type ``integer``.
+- Input type ``integer`` and agggregation :math:`\in \{` ``any``, ``all`` :math:`\}`
```
ChristianZimpelmann, line 257: Maybe add here:

- Input type ``integer`` and aggregation ``mean`` leads to output type ``float``
- Aggregation ``count``  always leads to output type ``integer``

Also: I used ``int`` instead of ``integer`` for type annotations so far.

hmgaudecker: Will do. Do we have examples of taking means of integers?

ChristianZimpelmann: > Do we have examples of taking means of integers?

No, we don't use ``mean`` aggregation at all at the moment.

hmgaudecker: Cool. Then we should only allow it for floats in the first place. Will always be able to change that down the road should it come up.

PR #180, opened 2020-05-17
Add error messages for conflicts between functions and data and data and user_columns.

hmgaudecker:
Please check whether this makes sense, previous formatting would be a problem for many entries.

I cared more about the look of the error message to the user than flake8 being happy here; need to find a workaround still.

### DIFF
--- a/gettsim/dag.py
+++ b/gettsim/dag.py
@@ -182,18 +182,19 @@ def fail_if_functions_and_columns_overlap(func_dict, data):
 
                     {formatted}
 
-                {'This is' if n_cols == 1 else 'These are'} already present among the functions of the taxes and
-                transfers system.
+                {'This is' if n_cols == 1 else 'These are'} already present among the
+                functions of the taxes and transfers system.
 
-                If you want {'this' if n_cols == 1 else 'a'} data column to be used instead of calculating it within
-                GETTSIM, please specify it among the *user_columns*.
+                If you want {'this' if n_cols == 1 else 'a'} data column to be used
+                instead of calculating it within GETTSIM, please specify it among the
+                *user_columns*.
 
-                If you want {'this' if n_cols == 1 else 'a'} data column to be calculated internally by GETTSIM, remove it from the *data* you pass to GETTSIM.
+                If you want {'this' if n_cols == 1 else 'a'} data column to be
+                calculated internally by GETTSIM, remove it from the *data* you pass to
+                GETTSIM.
 
-                {'' if n_cols == 1 else '''
-                You need to pick one option for each column that appears in the list
-                above.'''
-                }
+                {'' if n_cols == 1 else ''' You need to pick one option for each column
+                that appears in the list above.'''}
                 """
             )
         )
--- a/gettsim/dag.py
+++ b/gettsim/dag.py
@@ -90,7 +90,10 @@ def compute_taxes_and_transfers(
     )
 
     fail_if_user_columns_are_not_in_data(data, user_columns)
-    fail_if_user_functions_and_user_columns_overlap(data, func_dict)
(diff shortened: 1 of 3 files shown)


hmgaudecker:
Btw, do we have an meaningful error message if a user specifies something in `user_columns` that is not part of the data that she passes?

tobiasraabe:
Done.

gettsim/dag.py
```diff
@@ -182,18 +182,19 @@ def fail_if_functions_and_columns_overlap(func_dict, data):
...
                     {formatted}
 
-                {'This is' if n_cols == 1 else 'These are'} already present among the functions of the taxes and
-                transfers system.
+                {'This is' if n_cols == 1 else 'These are'} already present among the
+                functions of the taxes and transfers system.
 
-                If you want {'this' if n_cols == 1 else 'a'} data column to be used instead of calculating it within
-                GETTSIM, please specify it among the *user_columns*.
+                If you want {'this' if n_cols == 1 else 'a'} data column to be used
```
hmgaudecker, line 236: This is precisely what I wanted to avoid, sorry. The message the user is getting will look ugly because this line and others are too short. 

Maybe define extra variables upfront?

gettsim/dag.py
```diff
@@ -154,7 +155,54 @@ def create_function_dict(user_functions, internal_functions, user_columns, param
...
+                pass to GETTSIM.
+
+                {'' if n_cols == 1 else ''' You need to pick one option for each column
+                that appears in the list above.'''}
+                """
+            )
+        )
+
+
+def fail_if_user_functions_and_user_columns_overlap(data, func_dict):
```
hmgaudecker, line 205: Why `user_functions_and...` ? Usually, the functions will come from within GETTSIM? 

Although I do think we want to have a separate check on this one... :laughing: 

So `user_functions_and...` with a different body and `internal_functions_and...` for this one?

tobiasraabe:
Hi @hmgaudecker,

I reworked the formatting. I think the solution captures both benefits:

1. Being able to write exception messages in triple quotes in the source code and keeping the maximum line length.
2. Properly wrapping paragraphs in the cmd output.

I changed the formatting of the columns to a python list because I saw no benefit of the formatted version. It is not more readable because we are used to read lists. At the same time, the user can now copy-paste the column names for quick adjustments.

hmgaudecker:
**We** are used to reading lists, but the typical GETTSIM user (should it go where we would like it to go) is not :grinning:. So I do get the point about copying and pasting, but in light of the fact that you will not be able to see much if there are more than 5 long names I would prefer a newline-separated approach.

Maybe a compromise would be to do (no tabs, please):

```python
formatted = '",\n        ".join(unused_user_columns)
f"""
    [
        "{formatted}"
    ]
"""
```

tobiasraabe:
See your point. Here is how it looks like now:

```
ValueError: Your data provides the columns:

[
    "a",
    "b",
    "c",
    "d",
]

These are already present among the user functions of the taxes and transfers
system.

If you want a data column to be used instead of calculating it within GETTSIM,
please specify it among the *user_columns* or remove the function from
*user_functions*.

If you want a data column to be calculated by user functions, remove it from
the *data* you pass to GETTSIM.

You need to pick one option for each column that appears in the list above.
```

hmgaudecker:
Great! 

Looking at it like this, maybe you can indent the two paragraphs:
```
If you want a data column to be used instead of calculating it within GETTSIM,
please specify it among the *user_columns* or remove the function from
*user_functions*.

If you want a data column to be calculated by user functions, remove it from
the *data* you pass to GETTSIM.
```

by four spaces. 

Then go ahead with your own review and merge :grin: Thanks!!!

PR #182, opened 2020-05-20
Vectorize piecewise_polynomial.

tobiasraabe:

The function `piecewise_polynomial` only supports scalars as inputs for `x`. This PR enables the function to accept arrays and scalars.

### DIFF
--- a/gettsim/pre_processing/generic_functions.py
+++ b/gettsim/pre_processing/generic_functions.py
@@ -184,9 +184,9 @@ def create_intercepts(
     for i, up_thr in enumerate(upper_thresholds[:-1]):
         intercepts_at_lower_thresholds[i + 1] = fun(
             up_thr,
-            lower_thresholds,
-            upper_thresholds,
-            rates,
-            intercepts_at_lower_thresholds,
+            lower_thresholds=lower_thresholds,
+            upper_thresholds=upper_thresholds,
+            rates=rates,
+            intercepts_at_lower_thresholds=intercepts_at_lower_thresholds,
         )
     return intercepts_at_lower_thresholds
--- a/gettsim/pre_processing/generic_functions.py
+++ b/gettsim/pre_processing/generic_functions.py
@@ -183,7 +183,7 @@ def create_intercepts(
     intercepts_at_lower_thresholds[0] = intercept_at_lowest_threshold
     for i, up_thr in enumerate(upper_thresholds[:-1]):
         intercepts_at_lower_thresholds[i + 1] = fun(
-            up_thr,
+            x=up_thr,
             lower_thresholds=lower_thresholds,
             upper_thresholds=upper_thresholds,
             rates=rates,
--- a/gettsim/pre_processing/piecewise_functions.py
+++ b/gettsim/pre_processing/piecewise_functions.py
@@ -22,10 +22,6 @@ def piecewise_polynomial(
 ):
     """Calculate value of the piecewise function at `x`.
 
-    This function works only for floats. To make it accept arrays for `x`, the function
-    is vectorized with :func:`np.vectorize` below while keeping all other arguments
-    except `x` as they are. This only works if the other arguments are passed as kwargs.
-
     Parameters
     ----------
     x : float
--- a/gettsim/pre_processing/piecewise_functions.py
+++ b/gettsim/pre_processing/piecewise_functions.py
@@ -1,7 +1,18 @@
 import numpy as np
 
+from gettsim._numpy import numpy_vectorize
 
-def _piecewise_polynomial(
+
+@numpy_vectorize(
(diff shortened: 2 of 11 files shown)


gettsim/pre_processing/generic_functions.py
```diff
@@ -184,9 +184,9 @@ def create_intercepts(
     for i, up_thr in enumerate(upper_thresholds[:-1]):
         intercepts_at_lower_thresholds[i + 1] = fun(
             up_thr,
```
hmgaudecker, line 186: Why not use `x=up_thr` for consistency with the rest?

I do not really want to open a can of worms, but... The docstring of this function is incomprehensible and misleading (`piecewise_linear` does not seem to exist?). Holds for most docstrings in the module at a quick scan.

Also, all functions in here relate to some piecewise function. Why have two separate modules? This one seems superfluous.

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -1,31 +1,42 @@
...
-        x (float, >0): The value that the function is applied to.
-        lower_thresholds (1-d array): The lower thresholds of each interval.
-        upper_thresholds (1-d array): The upper thresholds each interval.
-        rates (n-d arry): The slope in each interval. Where n is the degree of the
-        polynomial function.
-        intercepts_at_lower_thresholds (1-d array): The intercepts at the lower
-        threshold of each interval.
-        rates_modified: Boolean variable indicating, that intercepts can't be used
-        anymore.
+    This function works only for floats. To make it accept arrays for `x`, the function
```
hmgaudecker, line 14: This function works only for scalar `x`. The public version `piecewise_polynomial` accepts 1-d arrays for `x`.

(rest is implementation detail and does not belong here but below)

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -50,4 +61,17 @@ def piecewise_polynomial(
     out = intercept_interval
     for pol in range(1, rates.shape[0] + 1):
         out += rates[pol - 1, index_interval] * (increment_to_calc ** pol)
+
     return out
+
+
+piecewise_polynomial = np.vectorize(
```
hmgaudecker, line 68: Would

```python
def piecewise_polynomial(...):
    return np.vectorize(...)
```
or something similar work? This way, we could have a proper docstring and would not run into problems with Sphinx.

gettsim/pre_processing/generic_functions.py
tobiasraabe: I will do `x=up_thr`.

I cannot answer the remaining questions.

gettsim/pre_processing/piecewise_functions.py
tobiasraabe: Sure, I can define it as a closure as you do. Or I can even write a full decorator which can receive excluded as an argument if we plan to have multiple similar functions like `piecewise_polynomial` in the future and decouple the vectorization from this function..

gettsim/pre_processing/piecewise_functions.py
tobiasraabe: ok.

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: Might be good to prepare for the latter usecase. 

I just took issue with the docstring. We should make sure that we only have one sensible one, which 
* stays clear of implementation details 
* and gets caught by Sphinx

gettsim/pre_processing/generic_functions.py
MaxBlesch: Yes, we can move that all to one module. Sorry about the docstrings! I will revisit them. Once we implemented all the changes for the dag and are sure about the final layout I will revisit! I'll create an issue!

hmgaudecker: https://github.com/iza-institute-of-labor-economics/gettsim/issues/183

hmgaudecker: I was too slow in typing, @MaxBlesch :grinning:

gettsim/pre_processing/piecewise_functions.py
tobiasraabe: Changed it to the decorator. What do you mean by "gets caught by Sphinx"? Should I include it in the docs? Currently, there is no API section.

hmgaudecker: Oh, we definitely want that eventually! So we should prepare for it, not necessary to implement it now.

hmgaudecker: (sorry I had missed this one)

PR #195, opened 2020-06-13
Reintroduce `renten_anspruch.py`

### LINKED ISSUE #185: Reintroduce `renten_anspruch.py`
For our first release with the directed acyclic graphs, we will not include the theoretical claim of pensions, as well as the update of earning points. I will introduce them back in, for some later update.

MaxBlesch:
In the first release of gettsim in the DAG layout, we excluded the claim of pensions and the update of earning points. We will introduce these back in and adapt the corresponding tests.

This closes #185


gettsim/data/exogene_renten_daten/lade_renten_daten.py
```diff
@@ -32,78 +29,73 @@ def lade_exogene_renten_daten():
...
-# def _rentnerquotienten(params, year):
-#     """The rentnerquotient is the relationship between pension insurance receivers and
-#     contributes. For details see
-#     https://de.wikipedia.org/wiki/Rentenanpassungsformel
-#     """
-#     return (params[f"rentenvol_{year}"] / params[f"eckrente_{year}"]) / (
-#         params[f"beitragsvol_{year}"]
-#         / (params[f"rvbeitrag_{year}"] / 100 * params[f"eckrente_{year}"])
-#     )
+def _rentenwert_from_2018(params, year):
```
MaxBlesch, line 32: @Eric-Sommer: The following code can be used to calculate the `rentenwert` since 2018. What is the difference before 2018. Do have less data here? Could we not just calculate it once and then save it for the following years?

Eric-Sommer: Please don't expect too much from the current implementation. Basically, the first function (until 2017) takes the observed real value. The other function calculates it for the given year based on the representative (Soep/eu-silc/evs) sample. In a dynamic setting, you would need to re-calculate this value each year.

MaxBlesch: So the values for `rentnerquotienten`, `nachhaltigkeitsfaktor`, `riesterfactor` and `lohnkomponente` are from the SOEP? The values for `rentenwert` before 2018 are official? Could we get official values for 2019 and 2020? In a second step we could provide functions to calculate these things!

Eric-Sommer: No they are all from official sources. For the future, you'd need to get aggregate pensions, average earnings etc. from your data. I updated the data to as recent as possible. Doesn't comply with our yaml style guide, though.

MaxBlesch:
Done with the pension claim. I left the functions for creating the `rentenwert` on simulated data in `gettsim/data/dynamic_pension_data/create_custom_pension_params.py`. These are not done and would depend on the exact setup of. I'll create an issue with low priority if you want @ChristianZimpelmann. I think this will become interesting once there is someone, who wants to do a pension study.

gettsim/renten_anspr.py
```diff
@@ -0,0 +1,139 @@
...
+    zugangsfaktor
+    entgeltpunkte_update
+    ges_renten_vers_params
+
+    Returns
+    -------
+
+    """
+
+    out = (entgeltpunkte_update * zugangsfaktor * rentenwert).clip(lower=0)
```
ChristianZimpelmann, line 29: The statement can be returned directly - no need to define out. (similar for the other functions here)

ChristianZimpelmann:
> Done with the pension claim. I left the functions for creating the `rentenwert` on simulated data in `gettsim/data/dynamic_pension_data/create_custom_pension_params.py`. These are not done and would depend on the exact setup of. I'll create an issue with low priority if you want @ChristianZimpelmann. I think this will become interesting once there is someone, who wants to do a pension study.

Sounds good! Would be great if you can do that.

ChristianZimpelmann:
Good work! 

Would be great if you can answer the first (maybe obvious) question. Also let me know if I should do the merge. In that case, I can also take care of the second comment (with the return statements)

ChristianZimpelmann:
Also one more question: Do we need to address the failure of  Codecov?

gettsim/data/exogene_renten_daten/lade_renten_daten.py
MaxBlesch: Thanks!

gettsim/renten_anspr.py
MaxBlesch: Well, that's actually a discussion, we had quite often and didn't come yet to a conclusion. We said before to avoid complicated calculations at the return statement for readability reasons. Here it is not complicated, but I find it more readable. Can change if you like!

gettsim/policy_environment.py
```diff
@@ -78,6 +79,9 @@ def set_up_policy_environment(date):
 
     functions = load_reforms_for_date(date)
 
+    # load exogenous pension data.
+    params["renten_daten"] = _load_parameter_group_from_yaml(date, "ges_renten_vers")
```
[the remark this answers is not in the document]
MaxBlesch: Was actually an artifact of the infrastructure I started with. Could catch. I removed it!

ChristianZimpelmann (approved):

PR #184, opened 2020-06-02
Visualize the DAG.

tobiasraabe:
### What problem do you want to solve?

The transition to the DAG-based backend helped to make the computations more modulare and also allowed users to inject code by user functions.

But, before users can inject a function, they need to know

- What does the DAG for the computation look like?
- Where do I want to inject code?
- What side-effects does it have?

These questions cannot be easily answered even with the source code at hand and, ideally, we do not want beginners to dig in the internals of gettsim.

Thus, we add a function to visualize the DAG and expose as much information as possible to the user.

- The graphic shows the DAG as it is used for the computations.
- If a node is a function, the tooltip exposes its source code.
- Allow to highlight certain nodes such that they can be found more easily.
- Allow many ways to select subsets of nodes
  - Define neighborhoods such that all nodes which are 1, 2, ... edges away from some given nodes are shown.
  - Allow to show only nodes before or after a certain node.
  - Investigate whether there is some API in networkx to do this.
  - Allow to simplify groups of variables to one dot.

*Current look*

https://gettsim--184.org.readthedocs.build/en/184/tutorials/visualizing_the_system.html

Closes #173.
Closes #175.

PR #198, opened 2020-06-16
Better loading of internal and user functions.

tobiasraabe:

- Better error messages when functions with the same name are loaded.
- Internal functions are loaded with `load_functions("gettsim.soz_vers.krankenv_pflegev")` so that functions have `func.__module__` set which facilitates linking to the docs.
- Tests for `function_loader.py`.

### Todo

- [x] Re-check with implementation in #184.


gettsim/config.py
```diff
@@ -7,10 +7,10 @@
 GEP_1_CHARACTER_LIMIT = 17
 
 INTERNAL_FUNCTION_FILES = [
```
hmgaudecker, line 9: `INTERNAL_FUNCTION_LOCATIONS` ? Most of these are directories, not files.

docs/tutorials/user_functions.ipynb
```diff
@@ -0,0 +1,444 @@
...
+    "user_functions"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Modules\n",
+    "\n",
+    "You can also pass a module as a user function. Assume you have import `abgelt_st.py`, you can pass all of its functions as user functions with"
```
hmgaudecker, line 323:
```suggestion
    "You can also pass a module name; all functions in this module will be loaded. Assume you have import `abgelt_st.py`, you can pass all of its functions as user functions with"
```

gettsim/config.py
tobiasraabe: Renaming to `PATHS_TO_INTERNAL_FUNCTIONS`?

hmgaudecker: Fine with me. Maybe a little note that this will take all modules in a directory or a single module?

hmgaudecker:
> If you do not think it is necessary to have it in 0.3.x, I would merge it with 0.4 and update the tutorial when another PR establishes the new functions names.

The only necessary thing are bugfixes in the taxes & transfers; this does seem orthogonal.

PR #201, opened 2020-06-17
Improve hh_freib and rename to alleinerziehend_freib

tobiasraabe:

There was an error in the calculation of `hh_freib` after 2014.

### Todo

- [x] Add a test case for single parents after 2014 in `test_zu_verst_eink.py`.
- [x] Rename `hh_freib` to https://de.wikipedia.org/wiki/Alleinerziehendenentlastungsbetrag


gettsim/pre_processing/policy_for_date.py
```diff
@@ -225,9 +225,11 @@ def load_reforms_for_date(policy_date):
         policy_func_dict["sum_brutto_eink"] = _sum_brutto_eink_ohne_kapital
 
     if year <= 2014:
-        policy_func_dict["hh_freib"] = _hh_freib_bis_2014
+        policy_func_dict["alleinerziehend_freib_tu"] = alleinerziehend_freib_tu_bis_2014
     else:
-        policy_func_dict["hh_freib"] = _hh_freib_seit_2015
+        policy_func_dict[
+            "alleinerziehend_freib_tu"
+        ] = alleinerziehend_freib_tu_seit_2015
```
tobiasraabe, line 232: Should we use `ab` instead of `seit`. The former is more common, right?

PR #202, opened 2020-06-17
Define piecewise functions on whole real line.

MaxBlesch:

Tight now, the piecewise functions are only defined on a certain part of the real line. In most cases this is just the positive part. We will require to define the function on the whole real line and adapt the tests and code,
### Todo

- [x] Extend definition in yaml files.
- [x] Adapt the creation of intercepts.
- [x] Create an error message if the function is in an area that is not defined.
- [ ] Create an issue for documenting the new design of piecewise, as well as defining how to make gettsim to raise an error if value is a non defined area.


gettsim/pre_processing/piecewise_functions.py
```diff
@@ -28,27 +28,42 @@ def piecewise_polynomial(
         The value of `x` under the piecewise function.
 
     """
+    # If no individual is transferred, we return an empty series
+    if x.empty:
```
hmgaudecker, line 32: Necessary?

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -28,27 +28,42 @@ def piecewise_polynomial(
...
+        return x
+    # # Check if missing data is given
```
hmgaudecker, line 34: Delete comments

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -28,27 +28,42 @@ def piecewise_polynomial(
...
+        include_lowest=True,
+        labels=range(num_intervals),
+    )
     # Create series with last threshold for each individual
     thresholds_individual = binned.replace(dict(enumerate(thresholds[:-1])))
     # Increment for each individual in the corresponding interval
     increment_to_calc = x - thresholds_individual
 
+    # Check if any value is in the lowest interval.
+    if 0 in binned.values:
```
hmgaudecker, line 55: Must not happen anymore as we check that we cover everything in (-inf, inf) while setting things up.

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -160,6 +177,10 @@ def check_threholds(parameter_dict, parameter, keys):
         )
     upper_thresholds[keys[-1]] = parameter_dict[keys[-1]]["upper_threshold"]
 
+    # Check if the function is defined on the complete real line
+    if (upper_thresholds[keys[-1]] != np.inf) | (lower_thresholds[0] != -np.inf):
+        raise ValueError(f"{parameter} needs to be defined on the compleate real line.")
```
hmgaudecker, line 182:
```suggestion
        raise ValueError(f"{parameter} needs to be defined on the entire real line.")
```

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -335,15 +356,10 @@ def create_intercepts(
...
-    lower_thresholds,
-    upper_thresholds,
-    rates,
-    intercepts_at_lower_thresholds,
-    rates_modified=False,
+def calculate_intercepts(
+    x, lower_thresholds, upper_thresholds, rates, intercepts_at_lower_thresholds,
 ):
-    """Calculate value of the piecewise function at `x`.
+    """This function is used to calculate the intercepts from the raw data.
```
hmgaudecker, line 362:
```suggestion
    """Calculate the intercepts from the raw data.
```

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -372,20 +386,14 @@ def piecewise_polynomial_float(
     if (x < lower_thresholds[0]) or (x > upper_thresholds[-1]) or np.isnan(x):
         return np.nan
     index_interval = np.searchsorted(upper_thresholds, x, side="left")
-    if rates_modified:
```
hmgaudecker, line 375: Can you explain why you moved it into the function that is called in every iteration? Sort of defies the purpose of the whole thing. Maybe fine for now given time pressure, but we should open an issue to fix it then.

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -372,20 +386,14 @@ def piecewise_polynomial_float(
...
-                intercept_interval += (rates[pol - 1, interval] ** pol) * (
-                    upper_thresholds[interval] - lower_thresholds[interval]
-                )
-
-    else:
-        intercept_interval = intercepts_at_lower_thresholds[index_interval]
+    intercept_interval = intercepts_at_lower_thresholds[index_interval]
 
     # Select threshold and calculate corresponding increment into interval
     lower_thresehold_interval = lower_thresholds[index_interval]
```
hmgaudecker, line 392:
```suggestion
    lower_threshold_interval = lower_thresholds[index_interval]
```

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -372,20 +386,14 @@ def piecewise_polynomial_float(
...
+
+    if lower_thresehold_interval == -np.inf:
```
hmgaudecker, line 394:
```suggestion
    if lower_threshold_interval == -np.inf:
```

gettsim/pre_processing/piecewise_functions.py
```diff
@@ -372,20 +386,14 @@ def piecewise_polynomial_float(
...
+        return intercept_interval
+
     increment_to_calc = x - lower_thresehold_interval
```
hmgaudecker, line 397:
```suggestion
    increment_to_calc = x - lower_threshold_interval
```

gettsim/pre_processing/policy_for_date.py
```diff
@@ -50,7 +50,7 @@
 from gettsim.taxes.zu_verst_eink.vorsorge import _vorsorge_bis_2004
 
 
-def get_policies_for_date(policy_date, policy_groups="all"):
+def get_policies_for_date(policy_date, policy_groups="all", groups=None):
```
hmgaudecker, line 53: Rest is unrelated to piecewise polynomial, right? Necessary in this PR?

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Yes. E.g. for `eink_anr_frei`, the function is called for individuals with kids and without. If we have in one of the groups no person, then it is an empty call and can not be processed otherwise!

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Still need to check if one value is in the interval starting at -inf. If we do not check the increment of the value over the lower threshold is value - (-inf) which yields nans!

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: All comments or you mean the extra #?

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: We need to tell the function if it can use the recalculated intercepts! If we have individual rates we can not use the intercepts anymore and need to calculate everything from the beginning!

gettsim/pre_processing/policy_for_date.py
MaxBlesch: We changed the interface before to `policy_groups`. I just introduced this to support the old interface. Given the time pressure, I thought, that this little change could be done quickly in here!

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: This belongs into the calling function then. Clearer and should be faster.

`df.loc[n_kinder > 0] =`

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: The commented-out code which I assume is a leftover from debugging

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: I guess I do not understand the concept of individual rates. Please leave as is for now and make an issue to revisit.

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Yes, we do restrict the call [here](https://github.com/iza-institute-of-labor-economics/gettsim/blob/defb9a869520c4373646cf27649bf7c2d23edd2a/gettsim/benefits/arbeitsl_geld_2/eink_anr_frei.py#L22). But maybe there are no individuals in one of the groups!

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: Please put into the issue to be created as per below. There should be a more elegant and clear way for this whole block of code, but this is not the right time to tackle it.

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: Seems weird. Why would it fail then? Again, add to that issue, please, there are more pressing things right now.

MaxBlesch: Okay. I put in the issue and delete it for now!

MaxBlesch:
I created #203, which captures the remaining issues!

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: Only delete if it is certain it won't fail.

gettsim/pre_processing/policy_for_date.py
hmgaudecker: Ahh, sorry. Just revert. I do not think anybody used that, we certainly did not in class.

I must confess I do not even know what it does exactly, although the new name allows me to guess :grinning:

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Yes I am certain!

gettsim/pre_processing/policy_for_date.py
hmgaudecker: Sorry for that misunderstanding somewhere on Zulip.

hmgaudecker: Entirely my fault for not digging closely enough into what you changed about the interface.

gettsim/pre_processing/piecewise_functions.py
MaxBlesch: Sorry. I deleted the one below! This is still there and has to be!

gettsim/pre_processing/policy_for_date.py
MaxBlesch: Reverted!

gettsim/pre_processing/policy_for_date.py
```diff
@@ -60,10 +60,11 @@ def get_policies_for_date(policy_date, policy_groups="all"):
     ----------
     policy_date : int, str, datetime.date
         The date for which the policy system is set up.
-    policy_groups : list, str
+    policy_groups, groups : list, str
         The group or a list of groups which parameters are loaded. If an invalid
         name is given, a list of all possible values is printed. Default is to load all
-        parameter groups.
+        parameter groups. The keyword group will be completly replaced in future
+        versions.
```
tobiasraabe, line 67: There exist a special rst role for deprecations. Scroll a little bit further down on this page (https://numpydoc.readthedocs.io/en/latest/format.html#sections). It is not necessary, but nice to try it out.

gettsim/pre_processing/piecewise_functions.py
hmgaudecker: Or `.array`, just used that one.

gettsim/pre_processing/policy_for_date.py
hmgaudecker: As I wrote on Zulip, nobody used this keyword for sure, no need to leave a trace of it.

PR #213, opened 2020-06-24
compute_taxes_and_transfers should always return a DataFrame.

### LINKED ISSUE #210: compute_taxes_and_transfers should always return a DataFrame
### Bug description

`compute_taxes_and_transfers` currently returns either a Series, a DataFrame, or a tuple with either one and the DAG. 

In order to minimise work on the calling code, it should always be a DataFrame.

Can be completed as part of #184, where we already had some of that discussion.

tobiasraabe:
Closes #210.


tobiasraabe:
Since this PR remove `return_dag` which is an backward-incompatible change, it could be released with 0.4. At the same, no one used the option so we might as well use 0.3.3.

hmgaudecker:
I think the whole block of PRs you are currently working on should go into 0.4 in the sense that none of these are bugfixes and taken together, they imply a large change.

PR #222, opened 2020-06-29
Fix coefficient from Wohngeld table, was missing a zero.

hmgaudecker:

Fix a number that was wrongly inputted.


hmgaudecker:
We need to improve our tests, e.g., add something like monotonicity (each additional hh member without income should weakly increase wohngeld receipt).

MaxBlesch:
Added test for policy reforms since 2009 and all mietsufen!

hmgaudecker:
> Added test for policy reforms since 2009 and all mietsufen!

Great, thanks! 

Just to be sure: Did you do the reverse test, i.e., does it catch the error if you undo b1e446e ?

MaxBlesch:
Yes it does!

gettsim/tests/test_wohngeld.py
```diff
@@ -78,15 +79,57 @@ def test_wg(input_data, year, column):
...
+            "brutto_eink_1": 0,
+            "brutto_eink_4": 0,
+            "brutto_eink_5": 0,
+            "brutto_eink_6": 0,
+            "eink_st_tu": 0,
+            "rentenv_beitr_m": 0,
+            "ges_krankenv_beitr_m": 0,
+            "behinderungsgrad": 0,
+        },
+        index=range(int((MAX_HH_SIZE * (MAX_HH_SIZE + 1)) / 2)),
```
tobiasraabe, line 119: Maybe obvious for a mathematician :)

```suggestion
        index=range(np.sum(range(MAX_HH_SIZE + 1))),
```

gettsim/tests/test_wohngeld.py
```diff
@@ -78,15 +79,57 @@ def test_wg(input_data, year, column):
...
+    df["p_id"] = df.index
 
+    return df
 
-@pytest.mark.parametrize("year, column", itertools.product([2013], TEST_COLUMN))
-def test_wg_no_mietstufe_in_input_data(input_data_2, year, column):
-    year_data = input_data_2[input_data_2["jahr"] == year]
-    df = year_data[INPUT_COLS].copy()
+
+POLICY_YEARS = [2009, 2016, 2020]
```
tobiasraabe, line 126: Module variables under imports.

gettsim/tests/test_wohngeld.py
tobiasraabe: Or is the index even necessary after you expanded the values in the dictionary?

gettsim/tests/test_wohngeld.py
MaxBlesch: Done!

gettsim/tests/test_wohngeld.py
MaxBlesch: Deleted explicit index!

PR #229, opened 2020-07-30
Backport 228 to 0.3.4 and release 0.3.4.

tobiasraabe:


tobiasraabe:
I am not an expert on backporting. We can discuss the whole workflow when it is not so urgent.

#228 was merged into master. To get the changes for 0.3.4, we need to backport to 0.3.

hmgaudecker:
Ah, okay, since there is no 0.4 released yet, I guess we could just have worked with 0.3 directly and then merged with master. But that is minor and I am not sure it would have been so much quicker.

PR #245, opened 2020-11-17
Improve Tutorials: Inconsistencies in Kinderbonus-Example

davpahl:

This pull request will be used to improve the parameters & policy_functions tutorials of the GETTSIM documentation. 

As discussed on Zulip, the wording of the parameters tutorial should be improved to avoid confusing the readers. Also, the parameters tutorial should link to the new example in the policy_functions tutorial.
In the policy_functions tutorial, the example on adding functions to GETTSIM needs to be replaced by using the `kinderbonus_m_hh` function.

Feedback  is very welcome!

(Please let me know if anything is not set up correctly, trying this for the first time)


davpahl:
Changes:

* policy_functions tutorial: add Kinderbonus example for "Adding new Function"
* parameters tutorial: link/refer to new functions example, wording
* remove `Untitled...` columns by adding `index_col=[0]` when importing data
* fix wrong link in basic usage tutorial

For some reason, the recent version of the tutorials throws an error when running the `compute_taxes_and_transfers` function, because `anz_minderj_hh` needs to be added to `columns_overriding_functions` or dropped from the data. Maybe something changed with the latest release of GETTSIM? (If I remember correctly, when we designed the tutorials in summer this did not happen)

* remove `anz_minderj_hh` from exemplary data to avoid using `columns_overriding_functions` before it is introduced in the tutorials (when needed, compute `anz_minderj_hh` via GETTSIM),

ChristianZimpelmann:
Thank you for your work! Looks great!

A few comments:

- Your changes are easier to review if you work on a new branch in gettsim instead of on your private branch.
- I think `p_id` would be the natural index for the data sets.
- Regarding the error: We indeed missed that. Can you please remove `anz_minderj_hh` from the data sets (This is already calculated internally) and make sure that then all tutorials ru througn? We should definitely do that every time before a release.
- I will ask in Zulip if the others have any comments.

davpahl:
Somehow I lost track of this for the last weeks, sorry for that! 
- I adjusted the datasets by defining ``p_id`` as the index of all the datasets  and dropped the column  ``anz_minderj_hh``.
- Also, I realised that we labeled the figures in the tutorials inconsistently, so I tried to adjust that. 
Is that what you had in mind @ChristianZimpelmann ?

ChristianZimpelmann (dismissed):

ChristianZimpelmann:
Looks good to me. Thank you very much, @davpahl!

ChristianZimpelmann:
I merged the master that included pull request #250 that fixed a problem with the CI.

Now the tests for 3.6 run through, but the others still throw an error in `test_soli_st.py`. That's strange as only files for the tutorials have been changed.

Does anybody have a quick idea? Otherwise I will have a look later this week.

hmgaudecker:
I have not checked closely, but it is likely that Pandas 1.2 breaks it locally on my machine (1.1.3-py39hd56b5f2_2 --> 1.2.0-py39hde0f152_1). 

Tests passed before upgrading all packages, they fail as in the CI now. Seems like Pandas is the only thing that is involved.

Aside: I think we can drop 3.6, but should support 3.9

ChristianZimpelmann (approved):

ChristianZimpelmann:
I fixed the error and will merge now

PR #252, opened 2020-12-18
Income Tax Reforms 2021/22

Eric-Sommer:
- reforms on tax tariff 
- raise in child tax deduction


gettsim/parameters/eink_st.yaml
```diff
@@ -178,6 +178,28 @@ eink_st_tarif:
       upper_threshold: 57051
     3:
       upper_threshold: 270501
+  2021-01-01:
+    deviation_from: previous
+    reference: Art. 1 G. v. 01.12.2020 BGBl. I S. 2616
```
mjbloemer, line 183: A nitpick on consistency: In #251 you added a note for the same law reference (2. FamEntLastG again here), and a full stop after the page number. I think mentioning "Zweites Familienentlastungsgesetz" is not necessary and rather a reference, not a note. What do we put in the note key? calculations?

mjbloemer (changes requested):

gettsim/parameters/eink_st.yaml
Eric-Sommer: I added a note because is has been done before like that: https://github.com/iza-institute-of-labor-economics/gettsim/blob/1f11c71fc0f63f6160fbae065bb7bd73c6a6aaa3/gettsim/parameters/eink_st.yaml#L114-L129

So `note` is the right key in my view and I could obviously add those here.

gettsim/parameters/eink_st_abzuege.yaml
```diff
@@ -406,6 +410,14 @@ alleinerziehenden_freibetrag:
   2015-01-01:
     scalar: 1908
     reference: Art. 1 G. v. 16.07.2015 BGBl. I S.1202
+  2020-01-01:
+    scalar: 4008
+    reference: Art. 1 Nr. 5 G. v. 29.06.2020 BGBl. I S.1512
+    note: Temporäre Corona-Hilfe für 2020 und 2021
+  2022-01-01:
```
mjbloemer, line 417: Parameter seems to be changed to be permanent 4008 https://www.bundestag.de/dokumente/textarchiv/2020/kw51-de-jahressteuergesetz-2020-812872, https://www.buzer.de/3_JStG_2020.htm.

We also don't have the additional 240 Euro:
> Für jedes weitere Kind im Sinne des Absatzes 1 erhöht sich der Betrag nach Satz 1 um 240 Euro je weiterem Kind.

Eric-Sommer: good catch. is changed. We have the €240 though:

https://github.com/iza-institute-of-labor-economics/gettsim/blob/079608c8ac3fd3bfe032f5d39ebb8bd8a3b3be55/gettsim/parameters/eink_st_abzuege.yaml#L410-L421

mjbloemer (approved):

Issue #88, opened 2019-11-12
Rounding is missing for several taxes / transfers


MaxBlesch:
### Bug description

Right now there are rounding functions in the code. We want to be as exact as possible and don't loose any information.

For example:
https://github.com/iza-institute-of-labor-economics/gettsim/blob/1b59c617b6b0c6f4acf845128988b60e207bc028/gettsim/taxes/calc_taxes.py#L138

### TODO:
- [ ] Remove all rounding functions.

hmgaudecker:
Well this is not "as exact as possible" in terms of GETTSIM's core mission of representing the German laws. They often state that things be rounded to the next Euro.

But the ensuing pain of getting correct numerical derivatives in choice models is more than worth deviating there IMHO.

Eric-Sommer:
The most flexible solution would be to insert a global Option whether to round or not. There is a use case for both and we do it in izamod as well.

Also note that `.as type(int)` is also a rounding operation

hmgaudecker:
> The most flexible solution would be to insert a global Option whether to round or not. There is a use case for both and we do it in izamod as well.

Yeah, eventually we may want to think about it again but I do not want to support the added complexity during the transition phase. @MaxBlesch, maybe you can add a standard comment in all places where you remove it, so it would be simpler to get back to the current state?

> Also note that `.astype(int)` is also a rounding operation

Yes. Ouch.

In general, I simply cannot imagine coming to substantively different conclusions economically depending on whether you round or not. Or does anyone have an example?

Eric-Sommer:
I don't have an economically meaningful example either. But I believe our model should also be able to deliver the correct Wohngeld payment for a given household. And in real life, wohngeld happens to be paid in steps of 5€. Let me just list the other real-life rounding incidents that come to my mind for later reference:

1. Wohngeld in steps of 5€.
2. Kinderzuschlag in steps of 5€.
3. Vorsorgeaufwendungen are rounded up.
4. Income tax is rounded down.

I agree with the transitional argument. We could also add a tag "rounding taken out".

Eric-Sommer:
I like the idea. One would need one additional general function for rounding down and rounding up to the next €.

Eric-Sommer:
Related: The output currently includes six digits. Which is fine for behavioral modelling, but we might need an option to round to cents.
![grafik](https://user-images.githubusercontent.com/44068467/86965236-811a5400-c167-11ea-83b1-001f2374df04.png)

[...]

PR #256, opened 2020-12-21
2021 wohngeld reform

Eric-Sommer:
Implements a wohngeld/housing benefit reform becoming effective 01/2021. Before, only rent exluding heating costs (Kaltmiete) were considered for the housing benefit. From 2021, heating costs are considered via a lump sum amount depending on household size.

The reform is tested and confirmed with [smart-rechner.de](https://www.smart-rechner.de/wohngeld/rechner.php). gettsim returns €283.60, which is the last step before rounding to full euros (see #88).
grafik


hmgaudecker:
Thanks -- so how much is missing here? Thinking about releasing the 2021 version.

Eric-Sommer:
For starters, the code does not run atm. I also discussed privately with @mjbloemer on what the reform actually does. We'd certainly need an additional test here. I'll have a look at it later tonight.

Eric-Sommer:
Should be fine now. Rent considered after is aligned with external sources, but there is a more general issue with relevant income, which will be addressed in a separate issue.

gettsim/transfers/wohngeld.py
```diff
@@ -489,7 +489,7 @@ def wohngeld_miete_ab_2009(
     return wg_miete
 
 
-def wohngeld_max_miete_ab_2021(
+def wohngeld_miete_ab_2021(
```
hmgaudecker, line 492: Docstring still says 2009. 

Why are there no tests for this set of functions? Seems reasonable easy that an error would slip in...

Given that the PR starts with that calculator, taking a couple of those examples including extreme ones / corner cases (hh_size in {5, 6, 13}) and making test cases out of them would seem like the natural place to start.

Eric-Sommer: Docstring is corrected.

There are two tests for 2021. The one I describe above and `test_increasing_hh_size()` is extended for 2021. To me, this confirms that the logic of the 2021 reform is captured correctly as it does differs only in the maximum value for the considered rent. Extreme cases (hh_size > 12) are an issue that would affect all wohngeld calculations and would deserve an own PR in my view. @mjbloemer has played around with wohngeld calculations and very large households.

hmgaudecker:
Okay, I was just surprised that the renaming of the function did not lead
to renaming anywhere in the tests.

In the long run, it would be good to also test functions like this
directly, but no prerequisite for merging this IMO.

Eric-Sommer <notifications@github.com> schrieb am So., 21. Feb. 2021, 10:51:

gettsim/transfers/wohngeld.py
hmgaudecker: Okay, I was just surprised that the renaming of the function did not lead
to renaming anywhere in the tests.

In the long run, it would be good to also test functions like this
directly, but no prerequisite for merging this IMO.

hmgaudecker: Testing each function directly will also make it easier to catch side-effects, see #272

hmgaudecker (approved):
Looks good to me from the code side, thanks! 

However, @mjbloemer, it would be great if you could have a quick look at the parameters.

mjbloemer:
> Looks good to me from the code side, thanks!
> 
> However, @mjbloemer, it would be great if you could have a quick look at the parameters.

parameters (heating cost) are fine. You could add the heating cost parameters for 2009 and 2010 which followed the same rule.

I have not looked at the calculations/structure.

hmgaudecker:
Great, feel free to merge and we will [tackle the tests](https://gettsim.zulipchat.com/#narrow/stream/212222-general/topic/Wohngeld/near/227247729) in a separate PR.

PR #258, opened 2020-12-21
2021 parents' share of living costs

Eric-Sommer:
This should be the last PR on 2021 parameter changes...

The recent Existenzminimumsbericht caused fictive living shares of parents to change slightly. Those are necessary to calculate Kinderzuschlag.


Eric-Sommer:
obsolete. This is implicitly done in #267

PR #259, opened 2020-12-25
Adding parameter visualization plot files for discussion

Si-Pf:
Draft plots of parameter visualizations for income tax rate, income tax deduction, kindergeld, social security contributions, alg1 and wohngeld. 

Add additional visualizations for discussion. Preview:

grafik
grafik
grafik
grafik
grafik

### Todo

- Some plots (i.e. wohngeld, income tax rate, ALG1) only run for a hardcoded year so far. I would like to add interactive sliders to let the user change the displayed year. With JavaScript interactions I am unable to get it running so far but I'm still looking into it. 

- Think about further parameter visualizations

### DIFF
--- a/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
+++ b/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
@@ -3,8 +3,8 @@ import pandas as pd
 from bokeh.layouts import column
 from bokeh.models import ColumnDataSource
 from bokeh.models import NumeralTickFormatter
+from bokeh.models import Panel
 from bokeh.models import Slider
-from bokeh.plotting import curdoc
 from bokeh.plotting import figure
 
 from gettsim import set_up_policy_environment
@@ -12,7 +12,9 @@ from gettsim.piecewise_functions import piecewise_polynomial
 from gettsim.taxes.eink_st import st_tarif
 
 
-def tax_rate_interact():
+def tax_rate_interact(plot_dict):
+    plot_dict = plot_dict["tax_rate"]
+
     def prepare_data(start, end):
         """
         For a given year span returns the policy parameters to plot income tax
@@ -56,13 +58,23 @@ def tax_rate_interact():
 
         return ColumnDataSource(dataset)
 
+    def plotstyle(p, plot_dict):
+        p.title.text = plot_dict["title"]
+        p.legend.location = plot_dict["legend_location"]
+        p.xaxis.axis_label = plot_dict["x_axis_label"]
+        p.yaxis.axis_label = plot_dict["y_axis_label"]
+        p.xaxis[0].formatter = NumeralTickFormatter(format=plot_dict["x_axis_format"])
+        p.yaxis[0].formatter = NumeralTickFormatter(format=plot_dict["y_axis_format"])
+        # p.xaxis.bounds = (lower_bound, upper_bound)
+
+        return p
+
     def setup_plot(src):
 
         p = figure(
             plot_width=800,
             plot_height=400,
             background_fill_color="white",
-            title="Tax rate per income",
             y_range=(-0.01, 0.5),
         )
         p.line(
@@ -89,13 +101,7 @@ def tax_rate_interact():
             legend_label="Marginal tax rate",
         )
 
-        p.xaxis[0].formatter = NumeralTickFormatter(format="0")
-        p.xaxis[0].axis_label = "Taxable income in €"
-
-        p.yaxis.axis_label = "Tax rate"
-        p.yaxis[0].formatter = NumeralTickFormatter(format="0%")
-
-        p.legend.location = "bottom_right"
+        p = plotstyle(p, plot_dict)
 
         return p
 
@@ -115,7 +121,7 @@ def tax_rate_interact():
     p = setup_plot(src)
 
     layout = column(year_selection, p)
-    curdoc().add_root(layout)
 
+    tab = Panel(child=layout, title="Tax rate per taxable income")
 
-tax_rate_interact()
+    return tab
--- a/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
+++ b/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
@@ -1,8 +1,9 @@
+from Scripts.plotstyle import plotstyle
+
 import numpy as np
 import pandas as pd
 from bokeh.layouts import column
 from bokeh.models import ColumnDataSource
-from bokeh.models import NumeralTickFormatter
 from bokeh.models import Panel
 from bokeh.models import Slider
 from bokeh.plotting import figure
@@ -58,25 +59,9 @@ def tax_rate_interact(plot_dict):
 
         return ColumnDataSource(dataset)
 
-    def plotstyle(p, plot_dict):
-        p.title.text = plot_dict["title"]
-        p.legend.location = plot_dict["legend_location"]
-        p.xaxis.axis_label = plot_dict["x_axis_label"]
-        p.yaxis.axis_label = plot_dict["y_axis_label"]
-        p.xaxis[0].formatter = NumeralTickFormatter(format=plot_dict["x_axis_format"])
-        p.yaxis[0].formatter = NumeralTickFormatter(format=plot_dict["y_axis_format"])
-        # p.xaxis.bounds = (lower_bound, upper_bound)
-
-        return p
-
     def setup_plot(src):
 
-        p = figure(
-            plot_width=800,
-            plot_height=400,
-            background_fill_color="white",
-            y_range=(-0.01, 0.5),
-        )
+        p = figure(plot_width=800, plot_height=400, y_range=(-0.01, 0.5),)
         p.line(
             x="income",
             y="tax_rate",
--- a/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/wohngeld_interact.py
+++ b/gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/wohngeld_interact.py
@@ -1,0 +1,169 @@
+# Import
+import numpy as np
+import pandas as pd
+from bokeh.layouts import column
+from bokeh.models import BasicTicker
+from bokeh.models import ColorBar
+from bokeh.models import ColumnDataSource
+from bokeh.models import LinearColorMapper
+from bokeh.models import NumeralTickFormatter
+from bokeh.models import Panel
+from bokeh.models import Slider
+from bokeh.palettes import Magma256
+from bokeh.plotting import figure
+from bokeh.transform import transform
+
+from gettsim import set_up_policy_environment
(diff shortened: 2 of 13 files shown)


Eric-Sommer:
Thanks a lot, this is a great start! I insert a few comments below

> 
> grafik

It is important to point out that the x axis in on *taxable* income. The marginal tax rate would also be nice to have. 

> grafik

When plotting raw parameters, one could include the reference as a tooltip:
https://github.com/iza-institute-of-labor-economics/gettsim/blob/da1ba614561284a2f46e081c4e21fba1f8e870a0/gettsim/parameters/paramviz.py#L117-L142

> grafik

This graph needs a continuous color scale to reflect smaller increments.

> grafik

Please review the labelling to distinguish between health and care insurance.

> grafik

>     * Some plots (i.e. wohngeld, income tax rate, ALG1) only run for a hardcoded year so far. I would like to add interactive sliders to let the user change the displayed year. With JavaScript interactions I am unable to get it running so far but I'm still looking into it.

I tried that at one point, too. I think it is easier to set up a bokeh server, though, with the downside of harder integration in github or jupyter. [This](https://towardsdatascience.com/data-visualization-with-bokeh-in-python-part-iii-a-complete-dashboard-dc6a86aa6e23) is an excellent tutorial on the matter.

>     * Think about further parameter visualizations

Some ideas:
  - average income tax rate for average full-time full-year employee (attention! average earnings != average taxable income) over time.
  - extend this to total payment burden (income tax, soli, social security)

Si-Pf:
Thanks for the feedback Eric. I think I addressed most of what you proposed. For the tooltips I still have some open questions, as I'm using the set_up_policy_environment I don't actually have data on the parameter reference. Also for the Kindegeld plot specifically, bokeh currently doesn't support hover tools for a step glyphs.  
Updated plots:
grafik
grafik
grafik

Plus the plots for which I forgot to add screenshots before:
grafik
For this one I tried to combine logical parameters into one plot but I'm not sure if there's a better way.. also what labelling convention makes sense?
grafik

For the interactive plots I added an interactive example of the tax rate plot, you can run it locally from the respective directory with bokeh serve --show tax_rate_interact.py :
grafik

hmgaudecker:
Thanks, Simon! Looks great! I will have specific comments eventually, but did not plan on working too much these days.

> I'm using the set_up_policy_environment I don't actually have data on the parameter reference

We should make a wrapper around it that includes the meta information. Not your job, @Si-Pf, but maybe you can include an empty tooltip already while you are at it?

The plan is to host the docs so that there will be a bokeh server available, IZA would be one option like this one here https://covid-19-impact-lab.iza.org/en/app

Eric-Sommer:
Thanks, @Si-Pf, this looks pretty neat already!

> For the tooltips I still have some open questions, as I'm using the set_up_policy_environment I don't actually have data on the parameter reference. Also for the Kindegeld plot specifically, bokeh currently doesn't support hover tools for a step glyphs.

I think that's why I used a circle graph on top, see the code bit in my previous comment. If I'm not mistaken, you have to use the `ColumnDataSource` type for the data.

> For this one I tried to combine logical parameters into one plot but I'm not sure if there's a better way.. also what labelling convention makes sense?

Why don't you use the `name` entries from the parameter files as graph labels?

On the tax deductions and allowances: In my view, it makes sense to focus on the most important ones, i.e. Grundfreibetrag (implicit from the tax schedule), Alleinerziehenden-Freibetrag, Beitrag für Erziehung und Ausbildung, Sächliches Existenzminimum des Kindes, Werbungskostenpauschale, Sparerpauschbetrag. Especially the Vorsorge stuff is hard to interpret on its own.
> grafik
> 
> For the interactive plots I added an interactive example of the tax rate plot, you can run it locally from the respective directory with bokeh serve --show tax_rate_interact.py :
> grafik

I'd suggest to increase the width of the graph substantially, as most of the action happened below 50k€ taxable income, which becomes hardly visible when plotting up to 300k€. The slider is potentially better placed on top of the graph.

Looking at the code, there is potential for refactoring, especially on the plotting side. As an example, the whole style settings could be done in one function:

```python

def plotstyle(p, legend_location, x_axis_label, y_axis_label):
    p.legend.location = legend_location
    p.legend.click_policy = "hide"
    p.xaxis.axis_label = x_axis_label
    p.yaxis.axis_label = y_axis_label

    return p


# call it 
p = figure(...)
p.step()
p = plotstyle(p, "top_right", "Year", "Benefit Amount in €")
```
The labels, axis positions, font sizes etc. could be defined in a separate dictionary.

You may also check out how to [combine](https://docs.bokeh.org/en/latest/docs/user_guide/layout.html) several plots easily.

Si-Pf:
Thanks Eric for the feedback! Especially the refactoring part, I'm sure there is a lot more potential to do things better, so I'm happy for all the best practice advice I can get. 

I continued with the interactive plots and combined them into a bokeh server application to get closer to the deliverable of an app. While doing that I refactored the plot styling for those plots. I guess the additional plots I will need to refactor later but I think it is most efficient to wait until we have an overall style we are happy with. 
This is what the app currently looks like: 
grafik
grafik
grafik

Unfortunately the loading time is quite bad because each data preparation function takes time and it adds up. I was thinking about excluding these functions into a pre-processing file and exporting the output to for example .csv but I'm not sure what the best approach might be. 

Look forward to your feedback.

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/wohngeld_interact.py
```diff
@@ -0,0 +1,169 @@
...
+
+from gettsim import set_up_policy_environment
+from gettsim.transfers.wohngeld import wohngeld_basis
+
+
+def wohngeld_tab(plot_dict):
+    plot_dict = plot_dict["wohngeld"]
+
+    # Prepare wohngeld data ALTERNATIVE
+    def prepare_data(sel_year, hh_size):
```
Eric-Sommer, line 24: could your performance problems be related to the fact that you define functions within functions? And by calling this function, you overwrite the definition?

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+        return tax_rate_dict_full
+
+    def make_dataset(sel_year, tax_rate_dict_full):
+        dataset = tax_rate_dict_full[sel_year]
+
+        return ColumnDataSource(dataset)
+
+    def plotstyle(p, plot_dict):
```
Eric-Sommer, line 61: Ideally, you would of course only need one function `plotstyle` which is used for every graph.

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/wohngeld_interact.py
Si-Pf: I don't think my functions overwrite their definitions. Should I generally avoid defining functions inside functions? 

The performance issues are not new, they were there even before moving the code snippets into functions.

Take the kindergeld data preparation which is rather simple, the loop over all 45 years takes quite long. It is like this with any multi year data preparation. This might be a general issue with the approach of running the set_up_policy_environment to generate the data instead of having the data ready upon running the plots. What do you think?
grafik
`def prepare_data(start, end):
    """
    Data preparation for kindergeld parameters. Returns a dataframe.

    Parameters:
    start (Int): Defines the start of the simulated period
    end (Int):  Defines the end of the simulated period
    """

    # Calculate simulation period
    years = range(start, end)

    # Data preparation for Kindergeld params
    kindergeld_df = pd.DataFrame()

    for i in years:
        policy_params, policy_functions = set_up_policy_environment(i)
        kindergeld_df[i] = policy_params["kindergeld"]["kindergeld"].values()

    kindergeld_df = kindergeld_df.transpose()

    kindergeld_labels = ["First child", "Second child", "Third child", "Fourth child"]
    kindergeld_df.columns = kindergeld_labels

    return kindergeld_df`

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
Si-Pf: Yes I agree. I will look into it and push it with the next update..

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/wohngeld_interact.py
Eric-Sommer: Without checking, I'd suspect that a nested `plotstyle()` is not accessible from another module. 

But if the slow performance is due to data processing, loading them once probably makes sense.

gettsim/parameter_vizualisation_simon/bokeh_app/Scripts/tax_rate_interact.py
Si-Pf: I moved the plot styling into a separate module, which can now be used for all plots. I'm still looking into the data preparation stuff but for now the start-up time of the server app is not ideal because it needs to recreate the data sources everytime when initiated.

hmgaudecker: Maybe I am misunderstanding something about the architecture. But I thought the server app would just run forever, so start-up time is a one-time cost? Or is a new instance created each time a user plays around with a plot?

Eric-Sommer: >  I'm still looking into the data preparation stuff but for now the start-up time of the server app is not ideal because it needs to recreate the data sources everytime when initiated.

But isn't it still cheaper to load all the data everytime from `.csv` rather than loop through all the `.yaml` files?

Si-Pf: > 
> 
> Maybe I am misunderstanding something about the architecture. But I thought the server app would just run forever, so start-up time is a one-time cost? Or is a new instance created each time a user plays around with a plot?

At least locally if I don't explicitly close the server it runs forever and I can just go back and play with the plot.. However if I understand the bokeh user guide correctly it might be different if the server is hosted:
> The application code is executed in the Bokeh server every time a new connection is made, to create the new Bokeh Document that will be synced to the browser. 
https://docs.bokeh.org/en/latest/docs/user_guide/server.html

hmgaudecker: Fine. Just don't worry about it. Remember [Knuth](http://wiki.c2.com/?PrematureOptimization).

That is, as long as we have a load_data() function upfront, we can always change its behaviour. E.g., create a version with all parameters and pickle it when firing up the server / updating GETTSIM. When a connection is made, just load the pickle. That sort of refactoring are near-trivial changes if the tasks are well separated (load all data first, then process it, then plot).

Si-Pf: Okay cool, we have a clear functional separation between data preparation, selecting from the prepared data, plotting and then updating the plot based on sliders.

hmgaudecker:
Superseded by #285.

PR #271, opened 2021-02-20
Paramaterize Wohngeld Wealth Check

### LINKED ISSUE #207: Hardcoded wealth thresholds

The values for the wealth check on housing benefits are hardcoded. We want to create a params entry!

Eric-Sommer:
closes #207 .


MaxBlesch (approved):
Thanks for cleaning this up and all the descriptional work! I added the option "inf" for scalars. Then we do not need to choose any large number.

PR #275, opened 2021-03-09
Implement Grundrente

davpahl:

Implement the Grundrentenzuschlag and the allowances of the Grundrente.

### Todo
- [x] Remove `ges_rente_m` as direct input variable. Introduce new input variable `entgeltpunkte`. Calculate `staatl_rente_m` (and `gesamte_rente_m`) based on `entgeltpunkte`.
- [x] Use `beitragspflichtiger_durchschnittslohn` instead of `durchschnittlohn` for calculation of entgeltpunkte.
- [x] For calculation of Entgeltpunkte, consider wage only up to Beitragsbemessungsgrenze
- [x] Implement that for the same wage, entgeltpunkte increase stronger in eastern states. 
- [x] Introduce grundsicherung im alter
- [x] Introduce `grundrentenzuschlag_m` based on new input variables `grundrentenzeiten`, `gr_bewertungszeiten`, and `entgeltpunkte_gr`
- [x] Update expected results for `test_renten_anspr` after various changes.
- [x] Introduce tests for Grundrentenzuschlag
- [x] Introduce tests for Grundsicherung
- [x] Introduce more generous means test for Grundsicherung when eligible for Grundrente
- [x] Adjust documentation, create_synthetic_data, and tutorials where necessary
- [x] extend/adjust ALG2 test for households with pensioners and non-pensioners at the same time.
- [x] add rounding where necessary
- [x] general allowance of 26 euro for income from wealth. 
- [x] check consistency with naming of ALG2 functions

The following "pension" features are still missing after this PR:
- Mütterrente
- Rente mit 63
- Wohngeld for retirees
- Separation in Entgeltpunkte and Entgeltpunkte Ost
- Grundsicherung bei Erwerbsminderung
- Correct handling of Gemischte Bedarfsgemeinschaften (one adult retired, one adult not retired yet). Current implementation: If all adults in household retired -> Grundsicherung im Alter; otherwise -> ALG 2. In reality, it seems to  be more complicated.
- Updates for grundrentenzeiten etc.


hmgaudecker:
Thanks! This will be super-important.

@Eric-Sommer @MaxBlesch @mjbloemer -- do you know how we have dealt with "Grundsicherung im Alter" so far? Are we just assuming it is the same as ALG II without the job search requirement?

Eric-Sommer:
In gettsim v.0.0.1, there used to be a check on pensioner status and the like. 
https://github.com/iza-institute-of-labor-economics/gettsim/blob/7e2822c5028081467113e68ad2ed8aa72ce4b373/src/analysis/tax_transfer_funcs/benefits.py#L756-L765

I don't see it in the current version, so I think ALG2 currently only checks potential claim against income and wealth. @davpahl : What happens if you feed a pensioner household into gettsim without any income? Does he/she receive ALG2?

davpahl:
No, in the master version ```rentner = True``` changes the value of ```arbeitsl_geld_2_m_hh``` from 783.5 to 0 keeping all other characteristics the same.

Eric-Sommer:
OK found it: 

https://github.com/iza-institute-of-labor-economics/gettsim/blob/1685e7b6acea5dc493a19069a727138a09dc8e36/gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2.py#L40

What might work is the following: create a new function `grundsicherung_alter_m_hh` which is identical to `arbeitsl_geld_2_m_hh` except for the fact that `rentner_in_hh` must be greater than 0. And please check how mixed household (pensioners and non-pensioners) are actually treated.

Eric-Sommer:
According to §7 SGB II, a person is entitled to ALG2 if he/she is employable *or* shares a household with an employable person. That means a 67year old pensioner and his 59year old wife would get ALG2 if their income is not sufficient. So rather to check for `rentner_in_hh > 0`, one should compare the number of adults with the sum of pensioners. If all are pensioners, they get their simulated ALG2 amount, but the transfer is called "Grundsicherung im Alter".

ChristianZimpelmann:
Grundsicherung im Alter is indeed similar to ALG-2, but there are important differences with respect to means tests for earnings and wealth. See fore example: https://www.sovd-sh.de/aktuelles/meldung/artikel/grundsicherung-und-hartz-4-was-sind-die-groessten-unterschiede

@davpahl, would you like to work on the Grundsicherung im Alter in this PR, as well? Then I would recommend copying `arbeitsl_geld_2_m_hh` as @Eric-Sommer suggests as a starting point and implementing the differences from there.

davpahl:
Thank you for your help! @ChristianZimpelmann yes, I can try that. 
Have you dealt with the tax-free part of pensions and "Versorgungsbezüge" in GETTSIM before? For implementing the Einkommensanrechnung of the Grundrente, I´m trying to add up the total income subject to taxation and the tax-free part of pensions / "Versorgungsbezüge".

ChristianZimpelmann:
That's great!

> Have you dealt with the tax-free part of pensions and "Versorgungsbezüge" in GETTSIM before?

Not as far as I know. But @MaxBlesch or @hmgaudecker should know better than me.

hmgaudecker:
Very unlikely. The ancestor of GETTSIM, IZAΨMOD, was heavily rigged towards static models of labour supply during prime age (correct me if I am wrong, @Eric-Sommer). So very little thought on pension-age specifics, so far!

Eric-Sommer:
That's right but the tax-free part of pensions (Ertragsanteil) is modelled. Versorgungsbezüge are not included

ChristianZimpelmann (commented):
Hi David,

I justed pushed a few minor changes and added some comments here. Some might be also helpful to further improve your programming style. Some more general comments:

- In longer functions (e.g. `einkommen_grundr` you could add 1-2 inline comments to make it easier to understand what the necessary steps are.
- Seven tests are still failing. I need to have closer look at that after my vacation.
- Altogether, very well done! 

Since there are a few changes to the input variables, I would merge the PR only after we made the relevant changes to the tutorials and possibly also make changes to how we specify input variables in our code. I can take care of those. But would be great if you can have another look at the comments and try to address them.

Thank you very much for your work, this is super helpful!!

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,399 @@
...
+    grundrentenzeiten
+        See basic input variable :ref:`grundrentenzeiten <grundrentenzeiten>`.
+    ges_renten_vers_params
+        See params documentation :ref:`ges_renten_vers_params <ges_renten_vers_params>`.
+
+    Returns
+    -------
+
+    """
+    # Categories for distinct cases
```
Eric-Sommer, line 212: can you spell out these four distinct cases?

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,399 @@
...
+    return out.round(4)
+
+
+def bonus_entgeltpunkte_grundr(
+    durchschnittl_entgeltpunkte_grundr: FloatSeries,
+    höchstwert_grundr_m: FloatSeries,
+    grundrentenzeiten: IntSeries,
+    ges_renten_vers_params: dict,
+) -> FloatSeries:
+    """Calculate additional Entgeltpunkte for pensioner.
```
Eric-Sommer, line 191: what is the legal source for this?

Eric-Sommer (dismissed):
great work, thanks! :muscle: :rocket: 

I did not check in detail the calculations for "Grundsicherung im Alter" and "Grundrente" in detail. But the tests are from a reliable source and are reproduced. Moreover, the logic of the functions transparent. Most minor comments address naming. Major points we need to address  in my view:
- create tests for Grundsicherung im Alter
- extend/adjust ALG2 test for households with pensioners and non-pensioners at the same time.

gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -307,161 +476,278 @@ rentenvolumen:
...
+    ost: 31.89
+  2020-07-01:
+    west: 34.19
+    ost: 33.23
+  2021-07-01:
+    deviation_from: previous
+    ost: 33.47
 
-rentenwert_ost:
+rentenwert_vorjahr:
```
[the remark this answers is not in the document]
hmgaudecker: Thanks both! Indeed sounds like we may want to add a processing step that loads the parameters from t-1. No reason that the policy function parameters as specified in the yaml-files have to enter the `policy_params` directly without changes (another example would be the income tax function).

gettsim/dag.py
```diff
@@ -268,19 +268,25 @@ def execute_dag(dag, data, targets, debug):
     visited_nodes = set(data)
     skipped_nodes = set()
 
+    # Check for cycles in DAG
+    cycles = list(nx.simple_cycles(dag))
+    if len(cycles) > 0:
+        raise ValueError(f"The DAG contains at least one cycle: {cycles}")
```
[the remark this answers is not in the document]
ChristianZimpelmann: No, but I will move this to a separate PR anyway

docs/gettsim_objects/input_variables.rst
```diff
@@ -46,11 +46,9 @@ household.
 +-------------------------+----------------------------------------------+-------------+
 | _`kapital_eink_m`       | Monthly capital income                       | FloatSeries |
 +-------------------------+----------------------------------------------+-------------+
-| _`ges_rente_m`          | Monthly pension payments of each individual  | FloatSeries |
```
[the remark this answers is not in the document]
ChristianZimpelmann: The public pension earnings and, therefore, the total pension earnings `gesamte_rente_m` are now an outcome variable.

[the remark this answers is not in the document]
hmgaudecker: Interesting case for the documentation. Most (static) applications would probably measure pension payments directly and thus not use the new machinery... Do we highlight somewhere that you could pass any node of the DAG, regardless of whether it is a root node or not?

gettsim/transfers/grundrente.py
ChristianZimpelmann: Very valuable comment.. I improved the readability of this function

gettsim/transfers/rente.py
```diff
@@ -0,0 +1,312 @@
...
+    Parameters
+    ----------
+    geburtsjahr
+        See basic input variable :ref:`geburtsjahr <geburtsjahr>`.
+
+    Returns
+    -------
+    """
+    # Create 65 as standard
+    out = geburtsjahr * 0 + 65
```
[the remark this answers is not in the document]
ChristianZimpelmann: This is necessary to get a full vector/pd.Series of 65's.

gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2.py
```diff
@@ -37,7 +38,7 @@ def arbeitsl_geld_2_m_hh(
         wohngeld_vorrang_hh
         | kinderzuschlag_vorrang_hh
         | wohngeld_kinderzuschlag_vorrang_hh
-        | rentner_in_hh
+        | (anz_erwachsene_hh == anz_rentner_hh)
```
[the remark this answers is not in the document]
ChristianZimpelmann: Currently, Grundsicherung im Alter is only payed if all adults in hh are retired and ALG2 otherwise.

See also the [discussion](https://gettsim.zulipchat.com/#narrow/stream/243798-arbeitsl_geld_2-etc.2E/topic/ALG.202.20if.20partner.20is.20retired/near/260008404) I opened on Zulip.

gettsim/transfers/rente.py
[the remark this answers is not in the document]
hmgaudecker: Something more explicit would be nicer indeed. `pd.Series(65, index=geburtsjahr.index, dtype=float)` or so?

Also, should the 65 not be a parameter?

docs/gettsim_objects/input_variables.rst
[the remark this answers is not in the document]
ChristianZimpelmann: Yes, that is also how we do it for some of the tests.

This feature is described in the advanced tutorial: https://gettsim.readthedocs.io/en/stable/tutorials/advanced_usage.html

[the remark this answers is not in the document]
ChristianZimpelmann: We might need to do this also at other places in the tutorials -- haven't adjusted those so far.

gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -307,161 +476,278 @@ rentenvolumen:
...
-    scalar: 25.31
+    west: 25.31
+    ost: 22.06
   2002-07-01:
-    scalar: 25.86
+    west: 25.86
+    ost: 22.70
   2003-07-01:
-    scalar: 26.13
-  2004-07-01:
```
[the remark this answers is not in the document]
ChristianZimpelmann: That is because the rentenwert did not change during these years. So no need to repeat them.

gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -307,161 +476,278 @@ rentenvolumen:
...
+    note: It didn't exist before 2021.
+  2021-01-01:
+    base: 0.0334
+    increment: 0.001389
+    reference: SGB VI § 76g Absatz 4 Satz 3, 4
+
+grundrentenzeiten:
+  description:
+    de: minimal benötigte Grundrentenzeiten um zur Grundrente berechtigt zu sein und Zahl an Grundrentenzeiten ab der zusätzliche Monate an Grundrentenzeiten nicht berücksichtigt werden
+    en: minimal number of Grundrentenzeiten required to be entitled to Grundrente and maximum number of months that are considered for Grundrentenzuschlag
```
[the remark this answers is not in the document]
ChristianZimpelmann: To my understanding the Grundrente consists of the Grundrentenzuschlag and a pension allowance for Grundsicherung im Alter (up to 223€).

gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -307,161 +476,278 @@ rentenvolumen:
...
+zugangsfaktor_veränderung_pro_jahr:
+  description:
+    de: Faktor um den sich der Zugangsfaktor erhöht/senkt für jedes Jahr, das länger/kürzer gearbeitet wird
+    en:
+  reference: §77 Abs. 2 Nr. 2 SGB VI
+  2001-01-01:
+    vorzeitiger_renteneintritt: 0.036
+    späterer_renteneintritt: 0.06
+
+staatl_rente_anr_frei:
```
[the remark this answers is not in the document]
ChristianZimpelmann: This refers to the usual public pension. These parameters were hard-coded before. I made the description a bit clearer.

gettsim/parameters/ges_renten_vers.yaml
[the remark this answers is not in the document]
ChristianZimpelmann: Sounds good! I opened an issue for this. Would be probably best to work on this issue before merging this PR.

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+        See :func:`rentenwert`.
+    tu_id
+        See basic input variable :ref:`tu_id <tu_id>`.
+    ges_renten_vers_params
+        See params documentation :ref:`ges_renten_vers_params <ges_renten_vers_params>`.
+    Returns
+    -------
+
+    """
+    alleinstehend_grundr = tu_id.replace(anz_erwachsene_tu < 2)
```
MaxBlesch, line 60: Shouldn't we have that as a separate function? Maybe we already check if tax units have a single or multiple persons?! If not it should be in demographic vars.

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+    )
+
+    # Calculate höchstwert
+    out = (
+        ges_renten_vers_params["höchstwert_grundrente"]["base"]
+        + ges_renten_vers_params["höchstwert_grundrente"]["increment"]
+        * months_above_thresh
+    )
+
+    # Round to 4 digits
```
MaxBlesch, line 184: You probably have an internal standard now how to implement rounding?We had an earlier discussion in #88 .

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+    -------
+
+    """
+    out = grundrentenzeiten.astype(float) * np.nan
+
+    # Case 1: Entgeltpunkte less than half of Höchstwert
+    below_half_höchstwert = durchschnittl_entgeltpunkte_grundr <= (
+        0.5 * höchstwert_grundr_m
+    )
+    out.loc[below_half_höchstwert] = durchschnittl_entgeltpunkte_grundr * (1 - 0.125)
```
MaxBlesch, line 223: Should we have that number hardcoded or in the parameter files?

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+
+    # Case 2: Entgeltpunkte more than half of Höchstwert, but below Höchstwert
+    cond = ~below_half_höchstwert & (
+        durchschnittl_entgeltpunkte_grundr < höchstwert_grundr_m
+    )
+    out.loc[cond] = (höchstwert_grundr_m - durchschnittl_entgeltpunkte_grundr) * (
+        1 - 0.125
```
MaxBlesch, line 230: Same as before

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+    brutto_eink_6: FloatSeries,
+    kapital_eink_minus_pauschbetr: FloatSeries,
+) -> FloatSeries:
+    """Income relevant for income crediting rule of Grundrentenzuschlag. The
+    Grundrentenzuschlag (in previous years) is not part of the relevant income and
+    does not lower the Grundrentenzuschlag.
+
+    The Deutsche Rentenversicherung uses the income of the year two to three years
+    ago to be able to use administrative data on this income for the calculation.
+
```
MaxBlesch, line 279: Whats the reason the following docstring is in german?

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+        See :func:`brutto_eink_6`.
+    kapital_eink_minus_pauschbetr
+        See :func:`kapital_eink_minus_pauschbetr`.
+    Returns
+    -------
+    """
+
+    # Earnings from self-employed work and rent income (assumption that they were
+    # the same last year)
+    earnings_self_empl_rent = brutto_eink_1 + brutto_eink_6
```
MaxBlesch, line 309: I think we should have a separate function for that.

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,391 @@
...
+        See :func:`zugangsfaktor`.
+    ges_renten_vers_params
+        See params documentation :ref:`ges_renten_vers_params <ges_renten_vers_params>`.
+
+    Returns
+    -------
+    """
+
+    # Calculate pension in the last year in case the subject was already retired
+    rentenwert_vorjahr = wohnort_ost.replace(
```
MaxBlesch, line 356: We have an extra function for rentenwert. For consistency then also this as an extra function?

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,323 @@
...
+    anz_erwachsene_hh
+        See :func:`anz_erwachsene_hh`.
+    anz_rentner_hh
+        See :func:`anz_rentner_hh`.
+    Returns
+    -------
+
+    """
+    out = grunds_ia_m_minus_eink_hh.clip(lower=0)
+    cond = anz_erwachsene_hh != anz_rentner_hh
```
MaxBlesch, line 30: This could be an extra function. Knowing that a household consists of pensioners is an interesting variable and may be used for other things as well.

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,323 @@
...
+    tu_id
+        See basic input variable :ref:`tu_id <tu_id>`.
+
+    Returns
+    -------
+
+    """
+
+    # Consider Elterngeld that is larger than 300
+    anrechenbares_elterngeld_m = (elterngeld_m - 300).clip(lower=0)
```
MaxBlesch, line 135: Hardcoded number!

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,323 @@
...
+
+    Returns
+    -------
+
+    """
+    earnings = bruttolohn_m + eink_selbst_m
+
+    # Can deduct 30% of earnings (but no more than 1/2 of regelbedarf)
+    earnings_after_max_deduction = earnings - arbeitsl_geld_2_params["regelsatz"][1] / 2
+    earnings = (0.7 * earnings).clip(lower=earnings_after_max_deduction)
```
MaxBlesch, line 185: Hardcoded 30%. Move to parameter file.

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,323 @@
...
+
+    """
+    # Can deduct 100 € of private pension completely
+    prv_rente_above_100 = (prv_rente_m - 100).clip(lower=0)
+
+    # Can deduct 30% of private pension above 100 (but no more than 1/2 of regelbedarf)
+    prv_rente_after_max_deduction = (
+        prv_rente_above_100 - arbeitsl_geld_2_params["regelsatz"][1] / 2
+    )
+    out = (0.7 * prv_rente_above_100).clip(lower=prv_rente_after_max_deduction)
```
MaxBlesch, line 214: hardcoded number.

MaxBlesch (dismissed):
Nice work. Just outlined some hardcoded parameters as well as operations which could be useful as an own function.

gettsim/transfers/grundrente.py
ChristianZimpelmann: As far as I understand, we are rounding whereever this is what is written in the law. This is the case here.

Thank you for the reminder to the issue -- I think it would be a good idea to make rounding optional at some point.

gettsim/transfers/grundrente.py
ChristianZimpelmann: I decided to just use the two variables directly. I don't see the need for aggregating them.

gettsim/transfers/grundrente.py
hmgaudecker: Just the other way round AFAIR. The law often requires rounding to the nearest Euro and we decided against it because it might lead to discontinuities in otherwise smooth functions. IZAΨMOD behaved differently and did the rounding.

See #88.

ChristianZimpelmann: Sorry, for missrepresenting the conclusion in #88 then! Let's continue the discussion over there.

gettsim/transfers/grundrente.py
MaxBlesch: Okay!

hmgaudecker:
Only now saw that -- I had to look into the code to understand `grunds_ia`. We should have enough room for `grunds_alter` or even `grunds_im_alter`, I'd think?

ChristianZimpelmann:
> Only now saw that -- I had to look into the code to understand `grunds_ia`. We should have enough room for `grunds_alter` or even `grunds_im_alter`, I'd think?

You mean the yaml-file `grunds_ia.yaml`. There, I totally agree. For function names, we have a few already to long ones, e.g. `anrechenbares_erwerbs_eink_grunds_ia_m`. So what is feasible depends a bit on how we decide in #314.

hmgaudecker:
I think for `anrechenbares_erwerbs_eink_grunds_ia_m`, the burden is elsewhere...

`anr_erw_eink_grunds_alter`

and probably the other way around (area of application first)

`grunds_alter_anr_erw_eink`

?

Please also watch out that parameters are named consistently with functions, `einkommensanr_grundrente` seems very different along several dimensions. I.e., `grundr_anr_eink` ? I know this is a different area, but generally we should use abbreviations everywhere for one thing or not at all. Helpful being able to search for `anr_eink` instead of `(einkommensanr|anr_eink|...)` and to keep N states low in one's head.

gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -240,6 +239,176 @@ beitragspflichtiger_durchschnittslohn:
     reference: V. v. 30.11.2020 BGBl. I S. 2612.
     note: Das vorläufige Durchschnittsentgelt.
 
+umrechnung_entgeltp_beitrittgebiet:
```
hmgaudecker, line 242:
```suggestion
umrechnung_entgeltp_beitrittsgebiet:
```

Please correct typo everywhere, only picked this one instance!

ChristianZimpelmann:
> Please also watch out that parameters are named consistently with functions, einkommensanr_grundrente seems very different along several dimensions. I.e., grundr_anr_eink ? I know this is a different area, but generally we should use abbreviations everywhere for one thing or not at all. Helpful being able to search for anr_eink instead of (einkommensanr|anr_eink|...) and to keep N states low in one's head.

Good catch! I will go over all function names another time once we decided on #314

johannesgeyer:
Great Job! I had a quick view on pieces related to 'Grundrente'. I did not find any error in Paramters; what I did not understand was if the upper threshold for allowances in 'Grundsicherung' is missing. It should be 50% of Regelsatz. This was the case here: `staatl_rente_anr_frei` and also here for income from work `erwerbs_eink_anr_frei` - but maybe that is coded somewhere else?

JakobWegmann:
I was also looking for that :) It´s directly implemented in the functions, e.g. in ` anrechenbare_staatl_rente_grunds_ia_m`:
` deducted_rent = deducted_rent.clip(upper=arbeitsl_geld_2_params["regelsatz"][1] / 2)`

johannesgeyer:
Minor point: Grundsicherung has a general allowance of 26 euro for income from wealth. Don't know if that changed over the years. Source is §43 SGB XII: "2) Zusätzlich zu den nach § 82 Absatz 2 vom Einkommen abzusetzenden Beträgen sind Einnahmen aus Kapitalvermögen abzusetzen, soweit sie einen Betrag von 26 Euro im Kalenderjahr nicht übersteigen." I did not see that in parameter files.

JakobWegmann:
Three very small thing:
For `prv_rente_anr_frei` the attribute name is wrong, as it applies to everyone and not only to individuals whose Grundrentenzeiten exceeds 33 years
Proposal for the docstring of `anrechenbare_staatl_rente_grunds_ia_m`: Calculate **public** pension earnings relevant for Grundsicherung im Alter.”
Proposal for the docstring of `anrechenbares_eink_grunds_ia_m_hh`: Aggregate income relevant for Grundsicherung **im Alter** on household level.

JakobWegmann:
Four small thing:
For `prv_rente_anr_frei` the attribute name is wrong, as it applies to everyone and not only to individuals whose Grundrentenzeiten exceeds 33 years. And it is about privat insuracne
Proposal for the docstring of `anrechenbare_staatl_rente_grunds_ia_m`: "Calculate **public** pension earnings relevant for Grundsicherung im Alter.”
Proposal for the docstring of  `anrechenbare_prv_rente_grunds_ia_m`: "Calculate **privat** pension earnings relevant for Grundsicherung im Alter"
Proposal for the docstring of `anrechenbares_eink_grunds_ia_m_hh`: "Aggregate income relevant for Grundsicherung **im Alter** on household level."

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,338 @@
...
+        lower=earnings_after_max_deduction
+    )
+
+    return earnings
+
+
+def anrechenbare_prv_rente_grunds_ia_m(
+    prv_rente_m: FloatSeries, arbeitsl_geld_2_params: dict, grunds_ia_params: dict,
+) -> FloatSeries:
+    """Calculate public pension earnings relevant for Grundsicherung of Grundsicherung
```
hmgaudecker, line 202:
```suggestion
    """Return private pension benefits which are considered in the calculation of Grundsicherung im Alter.
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,338 @@
...
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+    grunds_ia_params
+        See params documentation :ref:`grunds_ia_params <grunds_ia_params>`.
+
+    Returns
+    -------
+
+    """
+    deducted_rent = piecewise_polynomial(
```
hmgaudecker, line 220:
```suggestion
    prv_rente_m_amount_exempt = piecewise_polynomial(
```

rent ≠ pensions

And this way it really should be clear what is meant, even if there are a couple of extra parameters.

hmgaudecker:
Thanks for all your detailed comments!!!

hmgaudecker:
> Minor point: Grundsicherung im Alter has a general allowance of 26 euro for income from wealth. Don't know if that changed over the years. Source is §43 SGB XII: "2) Zusätzlich zu den nach § 82 Absatz 2 vom Einkommen abzusetzenden Beträgen sind Einnahmen aus Kapitalvermögen abzusetzen, soweit sie einen Betrag von 26 Euro im Kalenderjahr nicht übersteigen." I did not see that in parameter files.

Interesting -- how do these tax law people think about this differently from private pensions? Clearly a lack of economists around :smiling_face_with_tear:

johannesgeyer:
> Interesting -- how do these tax law people think about this differently from private pensions? Clearly a lack of economists around 🥲

Yes; it is difficult to understand the systematic idea...It follows a different "logic" than annuities from private pension insurance. 26 euro is 1% of the earlier wealth allowance of 2600 euro. The allowance is part of "Verwaltungsvereinfachung" ;) , More details (came into effect in 2016): https://dip.bundestag.de/vorgang/.../68551

ChristianZimpelmann:
Thank you very much for having a look at that and the great suggestions, @johannesgeyer and @JakobWegmann!

ChristianZimpelmann:
I just merged `main` to this PR. The new restriction on the length of function names was binding for some functions. While I could shorten some quite easily, the following are still too long:

```
E         anr_erwerbs_eink_grunds_im_alter_m       (34)
E         anr_kapital_eink_grunds_im_alter_m       (34)
E         regelbedarf_grunds_im_alter_vermögens_check_m_hh (48)
E         unter_vermögensfreib_grunds_im_alter_hh  (39)
E         vermögensfreib_grunds_im_alter_hh        (33)
```

We have the options to either reduce them or make them purely internal (by adding a leading underscore). Do you have ideas for the former?

gettsim/transfers/rente.py
```diff
@@ -0,0 +1,340 @@
...
+
+    Returns
+    -------
+
+    """
+    out = priv_rente_m + ges_rente_m
+    return out
+
+
+def ges_rente_m(
```
ChristianZimpelmann, line 25: Another open question is how to name public retirement earnings.

In line with the naming of other parts of the code, it would be `ges_rente_m`. The problem here is that there also exists the function `gesamte_rente_m` which it might be confused with.

ChristianZimpelmann: An alternative could be `staatl_rente_m` or `gesetzl_rente_m` (but both would be inconsistent to e.g. `ges_rentenv`, `ges_krankenv`

hmgaudecker:
Thanks!!!

> We have the options to either reduce them or make them purely internal (by adding a leading underscore). Do you have ideas for the former?

`regelbedarf_grunds_im_alter_vermögens_check_m_hh` and `unter_vermögensfreib_grunds_im_alter_hh` should become private or go altogether. Not sure what the benefit of the current implementation is over just writing

```python
    out.loc[vermögen_hh >= vermögensfreib_grunds_im_alter_hh] = 0
```
in line 404 of `grunds_im_alter.py`.  Can we think of some better name for `regelbedarf_grunds_im_alter_vermögens_check_m_hh` -- to my mind, the `regelbedarf` really is a parameter before applying any checks, thinking "subsistence minimum" here.

Also, is the implementation correct? That is, `mehrbedarf_behinderung_m` will be paid regardless of wealth, but reduced depending on income? 

`anr_erwerbs_eink_grunds_im_alter_m` and `anr_kapital_eink_grunds_im_alter_m` should drop the underscore after `erwerbs` and `kapital`, respectively (in case of `kapitaleink`, that will affect further modules. *Btw, are we talking about the same definition of capital income everywhere? E.g., income from renting out property would not be `kapitaleink_` for tax purposes, but maybe here?* Alternatively, one could do `erw_` and `kap_`, but we are using `erw_` as an abbreviation for "Erwachsene" a lot, so I'd sort of like to leave it as is.

Hence, my vote would be private variables in all cases. It's just a consequence of not having namespaces, what can you do? :shrug: 

Speaking of which: We also should check consistency of naming with `arbeitsl_geld_2`. E.g., `vermögensfreib_grunds_im_alter` vs. `vermögensfreibetrag` - there is a lot of overlap, of course, and it seems a bit like whoever came first wins in terms of not requiring a suffix... At least I suppose that there is a `mehrbedarf_behinderung_m_hh` also for `arbeitsl_geld_2`, which just does not happen to be implemented yet?

gettsim/transfers/rente.py
hmgaudecker: At a minimum, make `gesamte_rente_m` private, if someone really wants it, she can just replicate summing the private and public pensions.

This is opening a can of worms (and not for this PR), but we are dealing with this in an overly simplistic manner. E.g., we cannot simply merge all private pension payments (Betriebsrente, private Rentenversicherung pre-2004, Riester, Rürup, whatever) and apply health insurance contributions to it: https://www.test.de/Steuern-und-Sozialabgaben-Diese-Abgaben-zahlen-Sie-auf-Ihre-Rente-5542858-0/

hmgaudecker: > An alternative could be `staatl_rente_m` or `gesetzl_rente_m` (but both would be inconsistent to e.g. `ges_rentenv`, `ges_krankenv`)

Sorry, did not see this in time -- the problem is the `gesamte`, not this here. `staatl` would be wrong in a legal sense.

hmgaudecker: It is a convenient shorthand in the code but it is not central for users to have access to it.

ChristianZimpelmann:
Thank you for the detailed feedback!

> 
> `regelbedarf_grunds_im_alter_vermögens_check_m_hh` and `unter_vermögensfreib_grunds_im_alter_hh` should become private or go altogether. Not sure what the benefit of the current implementation is over just writing
> 
> ```python
>     out.loc[vermögen_hh >= vermögensfreib_grunds_im_alter_hh] = 0
> ```

Good point! I think we can then even get rid of `regelbedarf_grunds_im_alter_vermögens_check_m_hh`.

> Also, is the implementation correct? That is, `mehrbedarf_behinderung_m` will be paid regardless of wealth, but reduced depending on income?

Will check, sounds indeed strange.

> `anr_erwerbs_eink_grunds_im_alter_m` and `anr_kapital_eink_grunds_im_alter_m` should drop the underscore after `erwerbs` and `kapital`, respectively (in case of `kapitaleink`, that will affect further modules. _Btw, are we talking about the same definition of capital income everywhere? E.g., income from renting out property would not be `kapitaleink_` for tax purposes, but maybe here?_ Alternatively, one could do `erw_` and `kap_`, but we are using `erw_` as an abbreviation for "Erwachsene" a lot, so I'd sort of like to leave it as is.

Good catch, thanks!

> Speaking of which: We also should check consistency of naming with `arbeitsl_geld_2`. E.g., `vermögensfreib_grunds_im_alter` vs. `vermögensfreibetrag` - there is a lot of overlap, of course, and it seems a bit like whoever came first wins in terms of not requiring a suffix... 

That's right. I will check again.

> At least I suppose that there is a `mehrbedarf_behinderung_m_hh` also for `arbeitsl_geld_2`, which just does not happen to be implemented yet?

Yes, I will open an issue.

gettsim/transfers/rente.py
ChristianZimpelmann: I would say it is relevant for every user who is interested in any of these transfers and would like to find out how the inputs are calculated.

Would you propose to remove ``gesamte_rente_m`` and have ``priv_rente_m`` and ``ges_rente_m`` as separate arguments for the calculation of these functions instead? Otherwise, I don't think we should make the function private.

Maybe we can just rename it instead: `summe_ges_priv_rente_m`?

hmgaudecker: That is a very good suggestion!!! Explicit rather than implicit.

And yes, in the medium run I would think that we just use its parts — it will just be too simplistic to be useful in many places.

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,425 @@
...
+    return anr_eink_grunds_im_alter_m.groupby(hh_id).sum()
+
+
+def anr_eink_grunds_im_alter_m(
+    anr_erwer_eink_grunds_im_alter_m: FloatSeries,
+    anr_priv_rente_grunds_im_alter_m: FloatSeries,
+    anr_ges_rente_grunds_im_alter_m: FloatSeries,
+    sonstig_eink_m: FloatSeries,
+    vermiet_eink_m: FloatSeries,
+    anr_kap_eink_grunds_im_alter_m: FloatSeries,
```
ChristianZimpelmann, line 108: @hmgaudecker 
> Alternatively, one could do erw_ and kap_, but we are using erw_ as an abbreviation for "Erwachsene" a lot, so I'd sort of like to leave it as is.

I changed it to `kap_eink` and `erwer_eink` now.

I also checked that the allowance for capital income does not apply to rental income. So it is correct to keep it separate here.

hmgaudecker: Everywhere in the code? 

I really do not want to be searching for "erw_" in the codebase and decide on a case-by-case basis whether it is what I want to see or not...

hmgaudecker: Could we discuss changes like that a little longer? I am in favor of reverting.

ChristianZimpelmann: Yes, everywhere. 

I picked `erwer_eink ` to avoid confusion with `erw_`

> Could we discuss changes like that a little longer?

Ok, will do next time. We can easily revert it now. I, however, like the shorter version as it avoids two purely private functions.

hmgaudecker: I think the issue is with what we mean by "internal". In my view, it is definitely not "purely internal". Anybody (except for Stata users) could easily access them if they want. I agree that variable name length is not a great reason to exclude something from the list of directly documented functions. But the drawback are function names that are hard to parse. It would not occur to me to abbreviate `erwerbs` with `erwer`, other than this hard character limit. It will be used at many other points as well, is this local change really worth it? As for `kap_`, I would not want to rename it until we have answered:

*Btw, are we talking about the same definition of capital income everywhere? E.g., income from renting out property would not be kapitaleink_ for tax purposes, but maybe here?*

To summarize:

- There are clear trade-offs between the breadth of variables accessible to Stata users on the one hand and reasonable naming for all others and developers on the other hand. 
- How we solve them, we just have to do on a case-by-case basis
- In doing so, we should be sure to do global optimisation and not look locally at some portion of the code only -- hence the comment on different notions of capital income.

Thanks!!!

ChristianZimpelmann: > It will be used at many other points as well

For me this is also a reason to keep it short since it might lead to more private functions at other places.

But no need to discuss these variable names too much. Let's revert to `erwerbseink`!

> Btw, are we talking about the same definition of capital income everywhere? E.g., income from renting out property would not be kapitaleink_ for tax purposes, but maybe here?

As mentioned above (probably language was not precise enough), I checked that it does not cover rental income for Grundrente and Grundsicherung im Alter. I now also checked the other places where it is used. Everything seems to be consistent.

> I also checked that the allowance for capital income does not apply to rental income. So it is correct to keep it separate here.

Do you agree with ``kap_eink`` then?

ChristianZimpelmann: And thanks for the good summary! I am totally on the same page, but putting more weight on avoiding internal functions if possible (I think it is not only relevant for Stata users, but also a signal for Python users). I think having consistent names across modules is very important! But having abbreviations that one wouldn't come up usually is not a big problem as long as it works the other way around (when reading the abbreviation it is clear what is meant)

hmgaudecker: > > I also checked that the allowance for capital income does not apply to rental income. So it is correct to keep it separate here.

Thanks, I did miss it above indeed! My bad, apologies.

> Do you agree with `kap_eink` then?

Do you mind leaving it as is / changing to `kapitaleink` for the moment? I am happy to revisit at some other point, but I do like the rough symmetry in lengths of the different income types.

For this particular case, one might even make a case for keeping both private even if short enough -- might add room for confusion if one is there but not the other.

hmgaudecker: > I think it is not only relevant for Stata users, but also a signal for Python users

Fully agreed, of course. But if we make clear that the most important reason for making long variable names private is our love and care for Stata users, that signal might not be quite as strong as in other Python code...

> I think having consistent names across modules is very important! But having abbreviations that one wouldn't come up usually is not a big problem as long as it works the other way around (when reading the abbreviation it is clear what is meant)

For people who do not look at the codebase all the time, I'd argue that is different. This creates frictions when searching for things. 

But that is exactly why it needs good discussion.

ChristianZimpelmann: > Do you mind leaving it as is / changing to kapitaleink for the moment?

will do

We definitely need an agreement and better guidelines here. It is not feasible to discuss about each single column name in detail.

Btw. I also made the change `vermögensfreib` -> `vermög_freib` (globally). Would you like to revert this, too, for the moment?

hmgaudecker: > > Do you mind leaving it as is / changing to kapitaleink for the moment?
> 
> will do

Thanks!

> We definitely need an agreement and better guidelines here. It is not feasible to discuss about each single column name in detail.

Indeed. This said, changes that affect the entire code base and not just a single variable just need discussion and some form of consensus, no way around it. However, down the road it is pretty clear this a) won't happen that much in the first place and b) we can't change variable names everywhere because some concatenated variable name would end up being too long. They will only become more numerous.

> Btw. I also made the change `vermögensfreib` -> `vermög_freib` (globally). Would you like to revert this, too, for the moment?

No, perfectly fine. We agreed on that over on #314 and just did not do it there because we noticed the inconsistency too late. All good.

ChristianZimpelmann:
I improved the naming of a few functions to make them consistent with #312 and between ALG 2 and Grundsicherung im Alter. Some remarks:

- I didn't change the functions `freibetrag_vermögen_hh` as those are currently edited in #331 
- Potentially controversial name: `potenz_arbeitsl_geld_2_m_hh` for potential ALG 2 (before `wohngeld_vorrang` check). Alternative:  `arbeitsl_geld_2_vorläufig_m_hh`. But I think the word "potenzielles" is more precise. Happy to change it.

hmgaudecker:
>   I didn't change the functions `freibetrag_vermögen_hh` as those are currently edited in [Vermögensfreibetrag #331](https://github.com/iza-institute-of-labor-economics/gettsim/pull/331). WIll do later.

Great, that will definitely be merged first

>     Potentially controversial name: `potenz_arbeitsl_geld_2_m_hh` for potential ALG 2 (before `wohngeld_vorrang` check). Alternative:  `arbeitsl_geld_2_vorläufig_m_hh`. But I think the word "potenzielles" is more precise. Happy to change it.

To make it future-proof (#342), if anything it should be `arbeitsl_geld_2_potenz_m_hh`. `_vor_vorrang` would probably be too long?

ChristianZimpelmann:
> if anything it should be arbeitsl_geld_2_potenz_m_hh. ``_vor_vorrang`` would probably be too long?

I thought of "potenzielles Arbeitslosengeld 2" as one term. But doesn't matter because `arbeitsl_geld_2_vor_vorrang_m_hh` just about fits our threshold. Thanks for the idea!

ChristianZimpelmann:
I now ticked off all ToDos. 🎉 

@Eric-Sommer @hmgaudecker @MaxBlesch, would be great if you can have another look at it such that we can merge it next week :)

The PR really became way too large.. next time, we should split similar tasks in smaller chunks. It was also not the best decision to merge additional changes like #308 or #306 to this branch instead of to `main` first.

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,395 @@
...
+    does not lower the Grundrentenzuschlag.
+
+    The Deutsche Rentenversicherung uses the income of the year two to three years
+    ago to be able to use administrative data on this income for the calculation.
+
+    "It can be assumed that the tax office regularly has the data two years after
+    the end of the assessment period, which can be retrieved from the pension
+    insurance."
+
+    Warning: Currently, earnings of dependend work and pensions are based on the
```
Eric-Sommer, line 303:
```suggestion
    Warning: Currently, earnings of dependent work and pensions are based on the
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,444 @@
...
+    alle_erwachsene_sind_rentner_hh: BoolSeries,
+    vermögen_hh: FloatSeries,
+    grunds_im_alter_vermög_freib_hh: FloatSeries,
+) -> FloatSeries:
+    """Calculate Grundsicherung im Alter on household level.
+
+    # ToDo: There is no check for Wohngeld included as Wohngeld is
+    # ToDo: currently not implemented for retirees.
+
+    # ToDo: Grundsicherung im Alter is only payed if all adults in the household
```
Eric-Sommer, line 20:
```suggestion
    # ToDo: Grundsicherung im Alter is only paid if all adults in the household
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,444 @@
...
+        grunds_im_alter_erwerbseink_m
+        + grunds_im_alter_ges_rente_m
+        + grunds_im_alter_priv_rente_m
+        + sonstig_eink_m
+        + vermiet_eink_m
+        + grunds_im_alter_kapitaleink_m
+        + elterngeld_grunds_im_alter_m
+    )
+
+    # subtract taxes and social security contributions
```
Eric-Sommer, line 174:
```suggestion
    # subtract taxes and social security contributions
    # TODO: Change this to lohn_steuer
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,444 @@
...
+
+    return ges_rente_m - deducted_rent
+
+
+def grunds_im_alter_vermög_freib_hh(
+    anz_erwachsene_hh: IntSeries,
+    anz_kinder_hh: IntSeries,
+    grunds_im_alter_params: dict,
+) -> FloatSeries:
+    """Calculate maximum capital allowed for Grundsicherung im Alter.
```
Eric-Sommer, line 425:
```suggestion
    """Calculate maximum wealth not considered for Grundsicherung im Alter.
```

Eric-Sommer (commented):
thanks a lot! a couple of wording suggestions and a clarification question

docs/gettsim_objects/input_variables.rst
```diff
@@ -90,3 +88,13 @@ household.
 +-------------------------+----------------------------------------------+-------------+
 | _`entgeltpunkte`        | Earning points for pension claim             | FloatSeries |
 +-------------------------+----------------------------------------------+-------------+
+| _`g_r_bewertungsreiten` | Months with > 30% of average income          | IntSeries   |
```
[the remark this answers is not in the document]
ChristianZimpelmann: Grundrentenbewertungszeiten is just the term used in the law: https://www.sozialgesetzbuch-sgb.de/sgbvi/76g.html

gettsim/policy_environment.py
```diff
@@ -285,14 +309,24 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
         unnecessary keys.
 
     """
+
+    def subtract_years_from_date(dt, years):
```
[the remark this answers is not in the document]
ChristianZimpelmann: We don't have any type annotations in this module (and all the other files for the GETTSIM "infrastructure". So I would rather improve the docstring here.

docs/gettsim_objects/variables_out.rst
```diff
@@ -42,3 +42,5 @@ documentation of all :ref:`functions`.
 +-------------------------+-----------------------------------------------------------+
 | wohngeld_m_hh           | Monthly housing benefit per household                     |
 +-------------------------+-----------------------------------------------------------+
+| grunds_im_alter_m_hh    | Monthly subsistence payment for retirees per household    |
```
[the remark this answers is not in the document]
ChristianZimpelmann: As far as I understand, the Regelbedarf, as well as the considered income is always considered on the level of the household (Bedarfsgemeinschaft). Do you think it is different? And is it different to ALG 2 (because that one is similarly only calculated on the household level).

A problematic case is of course if one person in the Bedarfsgemeinschaft is retired and the other is not. In these cases GETTSIM currently makes the following simplification:

```
    # ToDo: Grundsicherung im Alter is only paid if all adults in the household
    # ToDo: are retired. Other households get ALG 2 (This is a simplification by
    # ToDo: GETTSIM)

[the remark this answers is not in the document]
Comment (Eric-Sommer) [reply docs/gettsim_objects/variables_out.rst:45]: you are right, sorry. I got confused between grundrente and grundsicherung im Alter here.

File: docs/gettsim_objects/input_variables.rst [hunk 34]
```diff
@@ -90,3 +88,13 @@ household.
...
 +-------------------------+----------------------------------------------+-------------+
+| _`g_r_bewertungszeiten` | Months with > 30% of average income          | IntSeries   |
++-------------------------+----------------------------------------------+-------------+
+| _`entgeltp_grundr`      | Entgeltpunkte in Grundrentenbewertungszeiten | FloatSeries |
++-------------------------+----------------------------------------------+-------------+
+| _`grundrentenzeiten`    | Number of months considered for Grundrente   | IntSeries   |
++-------------------------+----------------------------------------------+-------------+
+| _`priv_rente_m`         | Amount of monthly private pension            | FloatSeries |
++-------------------------+----------------------------------------------+-------------+
+| _`schwerbe_ausweis_g`   | Dummy: Schwerbehindertenausweis Merkzeichen G| BoolSeries  |
```
hmgaudecker, line 99: I find this abbreviation rather unnatural, in the middle of a syllable. How about `schwerbeh_ausweis_g` ? If too long, `schwerbeh_ausw_g` or `schwerb_ausweis_g`?

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+
+    # Wealth check
+    out.loc[vermögen_hh >= grunds_im_alter_vermög_freib_hh] = 0
+
+    # Only pay Grundsicherung im Alter if all adults are retired (see docstring)
+    out.loc[~alle_erwachsene_sind_rentner_hh] = 0
+    return out
+
+
+def grunds_im_alter_m_minus_eink_hh(
```
hmgaudecker, line 48: I find the name confusing, needed to check with `grunds_im_alter_m_hh` several times what is done here and there. I don't really think this requires a separate function; why not put income and wealth checks *(plus the "everybody retired" check, which is doomed anyhow)* into one function `grunds_im_alter_m_hh`  ?

If it stays, the name should be `grunds_im_alter_minus_eink_m_hh` for a start, but probably would see changes.

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+    grunds_im_alter_kapitaleink_m: FloatSeries,
+    elterngeld_m: FloatSeries,
+    eink_st_tu: FloatSeries,
+    soli_st_tu: FloatSeries,
+    anz_erwachsene_tu: IntSeries,
+    sozialv_beitr_m: FloatSeries,
+    tu_id: IntSeries,
+    grunds_im_alter_params: dict,
+) -> FloatSeries:
+    """Calculate income which is considered in the calculation of Grundsicherung im
```
hmgaudecker, line 120:
```suggestion
    """Calculate income considered in the calculation of Grundsicherung im
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+        grunds_im_alter_erwerbseink_m
+        + grunds_im_alter_ges_rente_m
+        + grunds_im_alter_priv_rente_m
+        + sonstig_eink_m
+        + vermiet_eink_m
+        + grunds_im_alter_kapitaleink_m
+        + elterngeld_grunds_im_alter_m
+    )
+
+    # subtract taxes and social security contributions
```
hmgaudecker, line 174:
```suggestion
    # subtract taxes and social insurance contributions
```
(search and replace for "social security" everywhere)

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+    return out
+
+
+def grunds_im_alter_erwerbseink_m(
+    bruttolohn_m: FloatSeries,
+    eink_selbst_m: FloatSeries,
+    arbeitsl_geld_2_params: dict,
+    grunds_im_alter_params: dict,
+) -> FloatSeries:
+    """Calculate earnings which are considered in the calculation of Grundsicherung im
```
hmgaudecker, line 192:
```suggestion
    """Calculate earnings considered in the calculation of Grundsicherung im
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+        lower=earnings_after_max_deduction
+    )
+
+    return earnings
+
+
+def grunds_im_alter_kapitaleink_m(
+    brutto_eink_5: FloatSeries, grunds_im_alter_params: dict,
+) -> FloatSeries:
+    """Calculate capital income which are considered in the calculation of Grundsicherung im
```
hmgaudecker, line 230:
```suggestion
    """Calculate capital income considered in the calculation of Grundsicherung im
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+    # Calculate and return monthly capital income (after deduction)
+    return capital_income_y / 12
+
+
+def grunds_im_alter_priv_rente_m(
+    priv_rente_m: FloatSeries,
+    arbeitsl_geld_2_params: dict,
+    grunds_im_alter_params: dict,
+) -> FloatSeries:
+    """Calculate private pension benefits which are considered in the calculation of
```
hmgaudecker, line 261:
```suggestion
    """Calculate private pension benefits considered in the calculation of
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+        upper=arbeitsl_geld_2_params["regelsatz"][1] / 2
+    )
+
+    return priv_rente_m - priv_rente_m_amount_exempt
+
+
+def _grunds_im_alter_mehrbedarf_behinderung_m_hh(
+    _grunds_im_alter_mehrbedarf_behinderung_m: FloatSeries, hh_id: IntSeries,
+) -> FloatSeries:
+    """Aggregate mehrbedarf for people with disabilities on household level.
```
hmgaudecker, line 299:
```suggestion
    """Aggregate additional allowance for people with disabilities on household level.
```

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+    )
+
+    priv_rente_m_amount_exempt = priv_rente_m_amount_exempt.clip(
+        upper=arbeitsl_geld_2_params["regelsatz"][1] / 2
+    )
+
+    return priv_rente_m - priv_rente_m_amount_exempt
+
+
+def _grunds_im_alter_mehrbedarf_behinderung_m_hh(
```
hmgaudecker, line 296:
```suggestion
def _grunds_im_alter_mehrbedarf_beh_m_hh(
```
(and similar below)

gettsim/transfers/grunds_im_alter.py
```diff
@@ -0,0 +1,445 @@
...
+    tu_id
+        See basic input variable :ref:`tu_id <tu_id>`.
+    Returns
+    -------
+
+    """
+    out = pd.Series(0, index=schwerbeh_ausweis_g.index, dtype=float)
+    hhsize_tu = tu_id.replace(hhsize_tu)
+
+    # mehrbedarf for disabilities = % of regelsatz of the person getting the mehrbedarf
```
hmgaudecker, line 342: So this is confusing. Are we talking only about people with `schwerbeh_ausweis_g` (impaired in movement, eysight, etc. AFAIU) or anyone with some registered disability? Adjust comment appropriately.

hmgaudecker (commented):
Whew! Great work!!!! :tada: :clap: Thank you!

Naturally, there are a few things but no game-changers, I think.

hmgaudecker:
> The PR really became way too large.. next time, we should split similar tasks in smaller chunks. It was also not the best decision to merge additional changes like #308 or #306 to this branch instead of to `main` first.

Yes. Although those are almost trivial and would not have happen had we thought about the deeper issue first: Implementing Grundrente required a better state of the pension system (was half-hearted at best) and Grundsicherung im Alter (was non-existent). Those should have been tackled first in separate PRs. 

That is, @davpahl did great work in his thesis, but my learning for the future is to limit the scope of theses much more.

gettsim/dag.py
```diff
@@ -55,6 +54,9 @@ def create_dag(
         dag, columns_overriding_functions
     )
 
+    # Check for cycles in dag
+    _fail_if_dag_contains_cycle(dag)
```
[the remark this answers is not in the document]
ChristianZimpelmann: I thought it doesn't hurt to check twice. But I agree that in the current workflow one check is sufficient.. will delete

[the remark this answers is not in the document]
hmgaudecker: Only hurts the person trying to make sense of the code and wondering whether there's a deeper reason! In general, better safe than sorry, but here I do vote for the other direction.

gettsim/taxes/zu_verst_eink/zu_verst_eink.py
```diff
@@ -17,8 +17,24 @@
...
+        See basic input variable :ref:`tu_id <tu_id>`.
+
+    Returns
+    -------
+
+    """
+    return sum_brutto_eink.groupby(tu_id).sum()
+
+
+def freibetraege(
```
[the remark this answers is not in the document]
ChristianZimpelmann: currently all files avoid non-English characters

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,397 @@
...
+
+    """
+    out = grundr_zuschlag_vor_eink_anr_m - grundr_zuschlag_eink_m
+    return out.clip(lower=0)
+
+
+@add_rounding_spec(params_key="ges_rentenv")
+def grundr_zuschlag_eink_m(
+    gemeinsam_veranlagt: BoolSeries,
+    eink_excl_grundr_zuschlag_m_tu: FloatSeries,
```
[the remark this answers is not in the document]
ChristianZimpelmann: Great idea!

Does `ges_rente_inkl_grundr_m` -> `ges_rente_nach_grundr_m` also make sense?

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,397 @@
...
+
+    """
+
+    # Calculate relevant income following the crediting rules using the values for
+    # singles and those for married subjects
+    # Note: Thresholds are defined relativ to rentenwert which is implemented by
+    # deviding the income by rentenwert and multiply rentenwert to the result.
+    # This will be easier once functions are defined on scalars.
+    einkommen_grundr = tu_id.replace(eink_excl_grundr_zuschlag_m_tu)
+    anr_eink_single = (
```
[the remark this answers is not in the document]
ChristianZimpelmann: See comment above: `# This will be easier once functions are defined on scalars.`

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,397 @@
...
+        See params documentation :ref:`ges_rentenv_params <ges_rentenv_params>`.
+
+    Returns
+    -------
+
+    """
+    out = pd.Series(0, index=grundrentenzeiten.index)
+
+    # Case 1: Entgeltpunkte less than half of Höchstwert
+    below_half_höchstwert = _durchschnittl_entgeltp_g_r_bewertungszeiten <= (
```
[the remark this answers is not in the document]
ChristianZimpelmann: `piecewise_polynomial` currently only allows for thesholds that are constant for all subjects. But I agree that this would be easier. Is it ok if I make a comment to revise when moving to scalars and to improve piecewise_polynomial in the process?

gettsim/transfers/grundrente.py
```diff
@@ -0,0 +1,397 @@
...
+        See basic input variable :ref:`tu_id <tu_id>`.
+
+    Returns
+    -------
+
+    """
+    return eink_excl_grundr_zuschlag_m.groupby(tu_id).sum()
+
+
+def eink_excl_grundr_zuschlag_m(
```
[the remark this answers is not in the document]
ChristianZimpelmann: No good test cases available as test cases usually just state "Einkommen X€". Also because we make a few assumptions here (because no information on last year income for most income sources).

Do you think it would be worth it to make up a (regression) test?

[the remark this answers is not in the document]
ChristianZimpelmann: Just realized that a few functions are affected by that. So I will create a few tests for it.

gettsim/transfers/kinderzuschl/kinderzuschl.py
```diff
@@ -136,10 +136,10 @@ def kinderzuschl_vorläufig_m_bis_06_2019(
 
     """
     out = pd.Series(0, index=kinderzuschl_eink_spanne.index)
-
-    out.loc[kinderzuschl_eink_spanne] = (
+    res = (
```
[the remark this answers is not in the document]
ChristianZimpelmann: No, should be same behavior.. I reverted it.

gettsim/transfers/rente.py
```diff
@@ -0,0 +1,301 @@
...
+    Returns
+    -------
+
+    """
+    out = priv_rente_m + ges_rente_m
+    return out
+
+
+@add_rounding_spec(params_key="ges_rentenv")
+def ges_rente_incl_grundrente_m(
```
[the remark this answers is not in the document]
ChristianZimpelmann: New docstring: 

```
"""Calculate total public pension including Grundrentenzuschlag. Is
    only active after 2021 when Grundrente is in place.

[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/rente.py:29]: New name `ges_rente_nach_grundr_m` (see question above)

File: gettsim/taxes/zu_verst_eink/zu_verst_eink.py (hunk 45 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/taxes/zu_verst_eink/zu_verst_eink.py:37]: No reason for that. Windows can handle that by now :smile:

File: gettsim/transfers/grundrente.py (hunk 47 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:71]: Cool. Make an explicit ToDo out of it? I missed it since it was part of such a large block. 

Line above should be "dividing".

File: gettsim/transfers/grundrente.py (hunk 48 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:238]: Sure.

File: gettsim/transfers/grundrente.py (hunk 46 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:35]: Oh, I only realise now what is going on here, I thought it would be actually something intermediate before `ges_rente_m` comes out. 

Please swap to "bis_2020" and "ab_2021" as in other cases where interfaces change over time. Will have a better mechanism for that soon!

File: gettsim/transfers/grunds_im_alter.py (hunk 43 above)
Comment (ChristianZimpelmann) [reply gettsim/transfers/grunds_im_alter.py:342]: Currently only Mehrbedarf for Ausweis G individuals is implemented.

@paulinaschroeder, do you know if there is any "Mehrbedarf" for individuals with disabilities, but no `schwerbeh_ausweis_g`?

File: gettsim/transfers/grundrente.py (hunk 49 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:287]: Cool, thanks! 

Only looked at the docstring in detail now. 

We will need to come up with something better for 

`grundr_zuschlag_eink_m`

and 

`eink_excl_grundr_zuschlag_m`

This is too confusing for my simple mind. Probably best to talk that over at some point!

File: gettsim/transfers/grundrente.py (hunk 46 above)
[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py:35]: It's a mixture of both: after 2021 you could still be interested in `ges_rente_vor_grundr_m`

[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py:35]: In that respect, I like the current names more. But would be also open to change it if you prefer.

File: gettsim/transfers/rente.py (hunk 51 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/rente.py:29]: See comment above -> `ges_rente_ab_2021_m`

File: gettsim/transfers/grundrente.py (hunk 46 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:35]: Fair enough. Shows we really need a new way to work with those changes over time...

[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:35]: Do as you see fit, also below.

File: gettsim/transfers/grundrente.py (hunk 49 above)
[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py:281]: Maybe just get rid of the second function?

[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:281]: Not sure, can we? I thought that is past income determining current eligibility and level. And the first function determines the tapering given concurrent income? So both seem necessary?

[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py:281]: `eink_vor_grundr_zuschlag_m` (former `eink_excl_grundr_zuschlag_m`) is total income relevant for Grundrentenzuschlag (before considerung allowances). `vor_grundr` just means that Grundrentenzuschlag is not part of this income, but all other sources.

`grundr_zuschlag_eink_m` is income which is deducted from Grundrentenzuschlag (after considering allowances).

The first one is just used in the second one, so we could put them together. I would, however, prefer if we find a good way to rename them. (Talking about it might be a good idea)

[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py:281]: Thanks! Just tried to understand this better, a couple of notes:
- We should strive to include references in docstrings of functions like `eink_excl_grundr_zuschlag_m`. That is, not fully obvious why these particular five types of income are considered and no help resorting to the references included with parameters. In the current state, I would have no idea where all that comes from.
- I do not understand this to-do, certainly not obvious where pension income enters `freibeträge`? 

    #ToDo: Freibeträge for income are currently not considered
    #ToDO: as freibeträge_tu depends on pension income

- I continue to be confused by the `[transfer]_eink` naming, i.e., without anything like `eink_anr`. Should reconsider in the big renaming thing.
- For now, I would actually suggest merging the two functions. We can reconsider when re-writing `grundr_zuschlag_eink_m` as part of the move to a scalar-based solution. Always easier to add variables ex post than removing some.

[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py:281]: > We should strive to include references in docstrings of functions like eink_excl_grundr_zuschlag_m. That is, not fully obvious why these particular five types of income are considered and no help resorting to the references included with parameters. In the current state, I would have no idea where all that comes from.

Agreed!

> I do not understand this to-do, certainly not obvious where pension income enters freibeträge?

through `ges_krankenv_beitr_m` -> `vorsorge` -> `freibeträge`. I added this information to the ToDo

> I continue to be confused by the [transfer]_eink naming, i.e., without anything like eink_anr. Should reconsider in the big renaming thing.

Alright!

> For now, I would actually suggest merging the two functions. We can reconsider when re-writing grundr_zuschlag_eink_m as part of the move to a scalar-based solution. Always easier to add variables ex post than removing some.

Will do.

File: gettsim/transfers/grunds_im_alter.py (hunk 43 above)
Comment (paulinaschroeder) [reply gettsim/transfers/grunds_im_alter.py:326]: No there isn’t. You have to have a Schwerbehindertenausweis mit Merkzeichen G to qualify as someone with a disability.
The only exception is if the disability requires a special diet (e.g. diabetes), then you might qualify for a different Mehrbedarf to compensate those special needs.

Comment (ChristianZimpelmann) [reply gettsim/transfers/grunds_im_alter.py:326]: Thanks! Then the adjusted docstring `Calculate mehrbedarf for individuals with disabled person's pass G.` is sufficient.

Comment (hmgaudecker) [reply gettsim/transfers/grunds_im_alter.py:326]: And the function name should become:

`_grunds_im_alter_mehrbedarf_schwerbeh_g_m` 

(inserting "schwer")

Thanks!

File: gettsim/transfers/grundrente.py (hunk 49 above)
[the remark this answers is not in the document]
Comment (ChristianZimpelmann) [reply gettsim/transfers/grundrente.py]: Done now.

Also added new tests as discussed: code coverage of the new code is 100% now.

**Review by hmgaudecker — COMMENTED** [review]
Excellent! Anything left from yesterday's to-do list? Some tests?

File: gettsim/transfers/grundrente.py (hunk 49 above)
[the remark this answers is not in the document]
Comment (hmgaudecker) [reply gettsim/transfers/grundrente.py]: Excellent! Then please check whether the documentation changes I suggested are correct!

**Comment by ChristianZimpelmann** [comment]
> Excellent! Anything left from yesterday's to-do list? Some tests?

No, all done now.

File: docs/gettsim_objects/input_variables.rst [hunk 52]
```diff
@@ -7,86 +7,97 @@ The table below gives an overview of all variables needed to run GETTSIM complet
...
++-------------------------+---------------------------------------------+--------------+
+| _`schwerbeh_g`          | Severerly handicapped, with flag "G"        | BoolSeries   |
++-------------------------+---------------------------------------------+--------------+
+| _`mietstufe`            | Level of rents in city (1: low, 3: average) | IntSeries    |
++-------------------------+---------------------------------------------+--------------+
+| _`immobilie_baujahr_hh` | Construction year of dwelling               | IntSeries    |
++-------------------------+---------------------------------------------+--------------+
+| _`vermögen_hh`          | Wealth of household                         | FloatSeries  |
++-------------------------+---------------------------------------------+--------------+
+| _`entgeltp`             | Earning points for pension claim            | FloatSeries  |
```
hmgaudecker, line 91:
```suggestion
| _`entgeltp`             | Earnings points for pension claim           | FloatSeries  |
```

hmgaudecker (approved):
Woohoo!

PR #285, opened 2021-04-11
Paramviz dashboard

Si-Pf:

Adding code for a bokeh server which visualizes certain parameters. Can be easily extended. 

grafik

### Todo

- [x] Add description texts
- [x] Fix labelling with tax deductions
- [x] Grundfreibetrag pre 2002 not in data
- [x] Accurate wohngeld calculations (maximum rent not accounted for atm)
- [ ] account for Mietstufe 
- [ ] *Optional*: Figure out hosting options


Eric-Sommer:
Awesome! Looks fantastic already!

A few comments:
- For me, the code runs only if `wohngeld_max_miete` is replace with `wohngeld_miete` (see #256).
- It's probably better to have a `step` plot also for the income tax deductions. But please use the option `mode="after"` throughout, as in #122. As of now, the line jumps before the dot. 
- don't hesitate to include the year 2021
- Could you add the exact date and reference of a parameter change into the hover box?
- The wohngeld heat map needs selectors for household size and mietstufe.
- What's the purpose of the "Regenerate Data" Button?
- Another group of parameters of interest are social security contributions for employees.

hmgaudecker:
Cool, thanks! Great stuff indeed! 

@Si-Pf, let us know whether and if so, how much, work you still want to put into this :smile:

Si-Pf:
@hmgaudecker I'd be willing to put in a couple more evenings to develop some of the improvements Eric suggested. I'd like to see this get implemented but I'll have to see how far I get in a reasonable timeframe. 
Especially getting the Wohngeld plot fixed up with two additional dynamic parameters will take some thought, the way it is currently set up we would need to create one additional dataframe (50x50 each) per year, hh_size and mietstufe (~1500) in the data preparation from which we can later select. This takes forever to run and will create quite a large file. There might be a better approach but I have no good idea yet.

Eric-Sommer:
I am happy to take over sooner or later, if you are giving up at some point :smile: . Wohngeld is certainly not crucial, but we could reduce complexity by e.g.
- sticking to household size 6 or lower
- producing the data only for a few years. Relevant changes occured in 1992, 2001, 2009, 2016, 2020, 2021

Si-Pf:
Hi Eric, I implemented a couple of your suggestions. The wohngeld_max_miete was fixed, the step functions updated, added the plot for social security contributions, the year 2021 is now included and the wohngeld heatmap now has selectors for the relevant years and hh_size - for Mietstufe I'm not sure how to include that, maybe you can take a look.  

The regenerate data button was an easy way to trigger the data preparation function but makes no sense in this context - if the data preparation changes one will need to run the function manually anyways. 

What is still missing is the reference in the tooltip, I don't have that data, maybe you can also take a look at that as well. 

I hope this helps to move this in the right direction, I'm happy to explain certain parts in more detail, but I will not be able to spend much more time coding myself.

Eric-Sommer:
Hi @Si-Pf, I worked on this a bit. It'd be nice if you could review this, hopefully you are not angry for what I've done to your code :smiley: 
The main additions, apart from adding descriptions and many cosmetic tweaks are:
- Wohngeld takes into account minimum income and maximum rent
- Labelling in deductions graph did not match the color of the lines
- Grundfreibetrag pre 2002 is not in parameter files, so I added it by hand.
- Mouse-over added to heatmap
- New Tab with ALG2 Regelsätze added

Si-Pf:
@Eric-Sommer looks awesome 👍 I especially like what you did with the heatmap. One thing I noticed was that you removed the muted color and muted alpha from the deductions plot - was that intentional? If so we should overrule the "mute" click policy we set for all plots with a legend. For the new ALG Regelsatz Plot I would move the legend to "bottom_right" in the plot_dict to minimize overlay. 
One other thing I noticed was that we now have an all_data.pickle in the dashboard directory and the App directory what is the idea behind that? 
Apart from these minor things I'd say this turned out really nice :)

Eric-Sommer:
Thanks @Si-Pf . I think I removed the "mute" stuff for testing without exactly knowing what I did. I'll bring it back. Two data files is obviously a bad idea. I'll reconcile that.

Si-Pf:
Yeah these mute settings allows you to mute or unmute specific lines by clicking on the legend which helps if you for example want to compare two specific lines - it should be the same color but a different alpha to be consistent with the other plots.

ChristianZimpelmann:
I really like it! Thank you for your work Eric and Simon! 

I moved the dashboard folder one level up since it is not really a core component of gettsim (similar to the documentation in `docs`). Also, I renamed a few directories and functions (e.g. made them lowercase).

A few additional comments:

- Should we include `Marginal tax rate incl Soli`?
- I would very much prefer if we added Grundfreibetrag pre 2002 to the parameter files instead of adding it by hand. It will be easiest to maintain if the visualisation represents one-to-one the GETTSIM parameters.
- Before we host it somewhere, we should add some more descriptions for each plot (no priority right now).
- Is there a way to automatically run `pre_processing_data` any time a parameter is changed? Otherwise, we should do this manually before any release.

Eric-Sommer:
> I moved the dashboard folder one level up since it is not really a core component of gettsim (similar to the documentation in `docs`). Also, I renamed a few directories and functions (e.g. made them lowercase).
:+1: 

> A few additional comments:
> 
>     * Should we include `Marginal tax rate incl Soli`?

Yes :smiley: . The fact that soli makes a difference for medium incomes in 2021 looks buggy to me by the way...

>     * I would very much prefer if we added Grundfreibetrag pre 2002 to the parameter files instead of adding it by hand. It will be easiest to maintain if the visualisation represents one-to-one the GETTSIM parameters.

In principle, I agree. The problem is that Grundfreibetrag is not a parameter on its own, but an element of the tax tariff function. I even tried to model the old tax tariffs, but it didn't work out. The alternative would be to create an additional parameter in `eink_st.yaml`, which is also not nice. As soon as the old tariffs are modelled, they should of course vanish from the bokeh code.

>     * Is there a way to automatically run `pre_processing_data` any time a parameter is changed? Otherwise, we should do this manually before any release.

I suggest to make re-generation of dashboard data compulsory if new parameters are added, e.g. in the yearly param update.

ChristianZimpelmann:
> In principle, I agree. The problem is that Grundfreibetrag is not a parameter on its own, but an element of the tax tariff function. I even tried to model the old tax tariffs, but it didn't work out. The alternative would be to create an additional parameter in `eink_st.yaml`, which is also not nice. As soon as the old tariffs are modelled, they should of course vanish from the bokeh code.

I see. Then I am ok with it as is.

ChristianZimpelmann (approved):
From my side, we are ready to merge. 

Feel free to add your contribution under `changes.rst`

PR #290, opened 2021-05-24
Align input variable list with tutorials

Eric-Sommer:

The tutorials, relying on the `synthetic_data` module, list a couple of variables that are not in the input list, leading to some confusion. This PR solves this.


ChristianZimpelmann (approved):
Looks great! Thank you!

PR #306, opened 2021-11-08
add the possibility to load the parameter value of a previous year.

### LINKED ISSUE #305: Make lagged policy parameters available

Some transfers are based on previous year's income. Therefore, sometimes access to previous year's policy parameters is necessary.

### Proposed implementation

- Add a processing step within `set_up_policy_environment ` that pre-processes the parameters in the .yaml files such that also policy parameters of the previous year become available.
- Make those easily accesible within GETTSIM functions (e.g.  `ges_renten_vers_params_last_year`)

Comment by @hmgaudecker: 

> another example [where pre-processing might be helpful] would be the income tax function

ChristianZimpelmann:
Closes #305 

Add the possibility to load the value of a parameter of the previous year when calling `set_up_policy_environment`


gettsim/policy_environment.py
```diff
@@ -362,6 +362,20 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
                     for key in value_keys:
                         tax_data[param][key] = policy_in_place[key]
 
+            # Also load earlier parameter values if this is specified in yaml
+            if "access_different_date" in raw_group_data[param]:
+                if raw_group_data[param]["access_different_date"] == "vorjahr":
+                    date_last_year = date - datetime.timedelta(days=365)
```
Eric-Sommer, line 368: this is not robust to leap years, isn't it? I'd suggest something like
```python
date_last_year = date(date.year - 1, date.month, date.day)
```

gettsim/tests/test_synthetic.py
```diff
@@ -1,50 +1,42 @@
...
-
-    # finally, run through gettsim
-    policy_params, policy_functions = set_up_policy_environment(2020)
-    # results = compute_taxes_and_transfers(df, policy_params, policy_functions,)
-    # assert len(results) == len(df)
+# import numpy as np
+# from gettsim import compute_taxes_and_transfers
+# from gettsim import set_up_policy_environment
+# from gettsim.synthetic_data.synthetic import create_synthetic_data
+# def test_synthetic():
```
Eric-Sommer, line 5: this test should be reactivated before merging

Eric-Sommer (commented):
Thanks @ChristianZimpelmann . I think your approach works for the grundrente case. I wonder however, whether we'd rather want to have a generic solution for every parameter. Right now, `access_different_date` needs to be specified for the parameter. Why not rather create a different set of parameters by invoking `set_up_policy_environment()` with a different year?

hmgaudecker:
> Thanks @ChristianZimpelmann . I think your approach works for the grundrente case. I wonder however, whether we'd rather want to have a generic solution for every parameter. Right now, `access_different_date` needs to be specified for the parameter. Why not rather create a different set of parameters by invoking `set_up_policy_environment()` with a different year?

This wholly depends on how often we are likely to need it. Invoking the whole beast is costly every time 'round.

Do (m)any other instances come to mind where we'd need the parameters of the previous year? In most cases, it seems to me like a question of individual data... (e.g., ALG I benefits and the like).

Thoughts, @Eric-Sommer @mjbloemer ?

gettsim/tests/test_synthetic.py
ChristianZimpelmann: This PR will be merged to the grundrente PR. I will definitelly reactivate it before merging to main branch.

ChristianZimpelmann:
Thank you for your comments, Eric!

> Why not rather create a different set of parameters by invoking `set_up_policy_environment()` with a different year?

I think we should hide this functionality from the user. Therefore, we shouldn't require them to call `set_up_policy_environment` twice or with two dates. 

Conversely, calling `set_up_policy_environment` for a given date should return a dict of all parameters that are necessary to run gettsim for this particular date (including parameters from the earlier year if necessary). 

The two alternatives I was thinking about were:

- adding parameters of last year if specified in .yaml (this is currently implemented)
- adding parameters of last year for all parameters by default

Since we will only need this for a small subset of parameters, I decided for the first solution.

gettsim/policy_environment.py
ChristianZimpelmann: good point, will adjust

ChristianZimpelmann: Done!

gettsim/tests/test_synthetic.py
ChristianZimpelmann: Test should run through now

Eric-Sommer (approved):
alright, you convinced me :smile:

ChristianZimpelmann:
Thank you very much, @Eric-Sommer 😄

PR #310, opened 2021-11-12
Mehrbedarf G

paulinaschroeder:

Add Mehrbedarf G - a Mehrbedarf for people with disabilities which are getting Grundsicherung im Alter und bei Erwerbsminderung - to the implementation of Grundsicherung im Alter

### Todo

- [x] add parameters for rate and threshold
- [x] add function for Mehrbedarf G
- [x] add function to aggregate on household level
- [x] implement in functions for grundsicherung


gettsim/parameters/ges_renten_vers.yaml
```diff
@@ -720,3 +720,19 @@ zugangsfaktor_veränderung_pro_jahr:
...
+  2006-12-07:
+    rate: 0.17
+
+mehrbedarf_g_grenze:
+  description:
+    de: Grenze des Behinderungsgrad ab der Menschen einen Schwerbehindertenausweis mit Merkzeichen G bekommen
+    en:
+  reference: § 3 Absatz 1 Nummer 7 SchwbAwV, § 229 Absatz 1 Satz 1 SGB IX, Anlage zu §2, Teil D, Nummer 1 VersMedV 
+  2008-12-15:
+    lower_threshold: 50
```
ChristianZimpelmann, line 738: Is there really a clear cutoff? In the laws I find different threshold for different types of handicaps.

ChristianZimpelmann (changes requested):
Thank you @paulinaschroeder, good job!

- I would take out the Einrichtungen. If we go this route, we needed to introduce a new input variable and include it in other files, as well. But can you please open an issue about this? You can also post your code there, such that we can reuse it once we address the issue.
- Please also add your contribution to the file `CHANGES.rst`
- We should also have a few tests for this Mehrbedarf. I would suggest duplicating the last line of `test_dfs_grunds_im_alter.csv` and adjusting those new lines such that they test for a couple of degrees of disability (we might have to add a new column `behinderungsgrad` to this test-data, too. Then make sure that tests are still running through.

gettsim/parameters/grunds_ia.yaml
```diff
@@ -121,3 +121,11 @@ prv_rente_anr_frei:
       2:
         upper_threshold: inf
         rate: 0.3
+
+mehrbedarf_g:
+  description:
```
ChristianZimpelmann, line 126: @LauraGergeleit, can you improve the documentation of this parameter in a similar fashion as you did for the other retirement parameters?

ChristianZimpelmann (changes requested):
Tests are running through now! 

Everything looks fine from my side with just one comment for @LauraGergeleit left.

Is there anything else from your perspective, @LauraGergeleit @paulinaschroeder?

paulinaschroeder:
> Is there anything else from your perspective, @LauraGergeleit @paulinaschroeder? Otherwise, we should be able to merge it soon.

I just fixed a calculation error in the test data. Otherwise everything looks good to me

ChristianZimpelmann (approved):

PR #312, opened 2021-11-29
GEP 01 updates.

### LINKED ISSUE #244: 17 character rule in GEP-1
We specified in [GEP-1](https://gettsim.readthedocs.io/en/stable/geps/gep-01.html) that there is a hard limit of 17 characters on all column names, which are in the user-facing API. Additionally, we have a test, which enforces this convention.

We will only enforce this limit on most import column names, but not everywhere.

- [ ]  Reformulate GEP-1.

- [ ] Rewrite test.

hmgaudecker:
Expand variable limit to 20 characters. Closes #244 .

Adjustments for time units and units of aggregation.

Remove instances of 'user-facing API.', pointless with DAG structure.

Adjust GEP 01 in some ways where reality came in the way of the original plan.


mjbloemer (approved):

docs/geps/gep-01.rst
```diff
@@ -101,8 +101,14 @@ pure ASCII option in German (this should always be fulfilled in the English vers
 there is demand, we could also support standardised variables like in `EUROMOD
 <https://www.euromod.ac.uk/>`_ or the `CNEF <https://cnef.ehe.osu.edu/data/>`_ standard.
 
-It might be useful to append an underscore plus one of {``j``, ``m``, ``w``, ``t``} to
-indicate the time unit.
+The default time unit is a year. If a column refers to a different time unit, an
+underscore plus one of {``m``, ``w``, ``t``} will indicate the time unite.
```
Eric-Sommer, line 105:
```suggestion
underscore plus one of {``m``, ``w``, ``t``} will indicate the time unit.
```

hmgaudecker: Thanks!

Eric-Sommer (approved):

ChristianZimpelmann (approved):
Looks great!

hmgaudecker:
Okay, time to come to a decision here. 

I made two changes. 

2bb498fa9e3ec76f1b489e35fd51c1ee6a7b6e6e:
- Using 32 characters for anything that users might be interested in, but typically will not be
- Adding internal columns that are purely helper variables (example would be `_zu_verst_eink_kein_kinderfreibetrag_tu`, which is only relevant internally for favorability check. Just to pre-empt questions: If someone wants this as a hack to get a policy environment without `kinderfreibetrag`, he should go elsewhere -- there is a correct way to achieve that and we do not want to mess with these hacks).

08075433cbea4880ab26fd473319c93b9f750a97:
- Remove a paragraph where we stated that we want to support English column names etc.. We may want to do that, but maybe rather as a separate project or some gists, not necessarily GETTSIM proper.
- Even if so, it is nothing fundamental enough to belong in a GEP, let alone the very first one :-)

docs/geps/gep-01.rst
```diff
@@ -33,11 +33,16 @@ a nutshell and without explanations, these conventions are:
...
   underscore, unless it is the last word.
-* Names should be long enough to be readable, but
+* Names should be long enough to be readable. However, we impose limits in order to make
+  GETTSIM usable in languages, which place limits on characters (Stata, in particular).
 
-  - for column names in the user-facing API, there is a hard limit of 17 characters;
-  - for other column names, there is a soft limit of 17 and a hard limit of 25
-    characters.
+  - Column names that are typically user-facing have a hard limit of 20 characters.
+    These columns are documented in `DEFAULT_TARGETS` in `gettsim/config.py`.
```
ChristianZimpelmann, line 40:
```suggestion
    These columns include input variables and default targets (documented in `STANDARD_DATA_TYPES` and `DEFAULT_TARGETS` in `gettsim/config.py`).
```

ChristianZimpelmann: We should probably rename `STANDARD_DATA_TYPES`

docs/geps/gep-01.rst
```diff
@@ -112,10 +124,11 @@ Parameters of the taxes and transfers system
...
-* There is a hierarchical structure to these parameters in that each of them is
+- There is a hierarchical structure to these parameters in that each of them is
   associated with a group (e.g., ``arbeitsl_geld``, ``kinderzuschlag``). These groups or
   abbreviations thereof do not re-appear in the name of the parameter.
-
+- Parameter names should be aligned with relevant column names. This makes searching for
+  identifiers easier and less error-prone.
 
 
 Python Identifiers (Functions, Variables)
```
ChristianZimpelmann, line 134: A bit contradictory that we mention functions here (which typically are also columns/potential targets and should be written in German.

Maybe mention helper functions instead?

ChristianZimpelmann (approved):
I agree with the changes. Just made a few minor comments/suggestions.

Having internal functions is not ideal, but should be a good solution if we follow this guideline: "Internal variables should be used sparingly."

docs/geps/gep-01.rst
```diff
@@ -86,24 +91,31 @@ in English. For column names, we always allow a pure ASCII option, see the next
...
+Across variations that include the same identifier, this identifier should not be
+changed, even if it leads to long variable names (e.g., ``kinderfreibetrag``,
+``_zu_verst_eink_kein_kinderfreibetrag_tu``). This makes searching for identifiers
+easier and less error-prone.
+
+The default time unit is a year. If a column refers to a different time unit, an
+underscore plus one of {``m``, ``w``, ``t``} will indicate the time unit.
+
+The default unit a column refers to is an individual. In case a household or tax unit is
+the relevant unit, an underscore plus one of {``hh``, ``tu``} will indicate the level of
```
mjbloemer, line 115: Should we rename `tu` to something in German? It is not exactly an institutional feature like the variables names but  time units (`t`) is also in German (or should we change that to `d`?).

Plus, should we state that time unit is appended always before `hh` or `tu` if both are used?

mjbloemer:
I only added one little comment on `_tu` and `_t`.

I have seen some problems with code highlighting and Umlauts (e.g. github https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/gettsim/transfers/benefit_checks/vermoegens_checks.py), but I don't think this is a problem of GETTSIM. We should ignore this.

So overall, I agree with the GEP Update.

mjbloemer (approved):

docs/geps/gep-01.rst
hmgaudecker: > Should we rename tu to something in German? It is not exactly an institutional feature like the variables names but time units (t) is also in German (or should we change that to d?).

The latter, which is also used in the parameters. Good catch! 

> Plus, should we state that time unit is appended always before hh or tu if both are used?

Yip. In fact, it is what the sentence below your quote states :smile:

docs/geps/gep-01.rst
hmgaudecker: Great. Using 'other functions' now.

docs/geps/gep-01.rst
hmgaudecker: Indeed. Suggestions?

docs/geps/gep-01.rst
```diff
@@ -86,24 +91,32 @@ in English. For column names, we always allow a pure ASCII option, see the next
 Column names (a.k.a. "variables" in Stata)
 ------------------------------------------
 
-We impose a hard limit of 17 characters for all column names that are part of the API,
-i.e., those that are an input to or an output of GETTSIM's main simulation functions.
+We impose a hard limit of 20 characters for all column names that typically user-facing.
```
Eric-Sommer, line 94:
```suggestion
We impose a hard limit of 20 characters for all column names that typically returned to users.
```

PR #316, opened 2021-12-13
implemented decorator for rounding.

### LINKED ISSUE #88: Rounding is missing for several taxes / transfers

ChristianZimpelmann:

Implement optional rounding.

Closes #88 

### Todo

- [x] Implement decorator to round functions
- [ ] Implement way to set globally whether outputs should be rounded or not (where applicable)
- [ ] Write tests for rounding decorator
- [ ] If it is a relevant contribution, put it in `CHANGES.rst`.

### DIFF
--- a/gettsim/functions_loader.py
+++ b/gettsim/functions_loader.py
@@ -112,7 +112,6 @@ def _convert_paths_and_strings_to_dicts_of_functions(sources, allow_imported_mem
                 spec.loader.exec_module(out)
             elif isinstance(source, str):
                 out = importlib.import_module(source)
-
             functions_defined_in_module = {
                 name: func
                 for name, func in inspect.getmembers(
@@ -131,14 +130,3 @@ def _convert_paths_and_strings_to_dicts_of_functions(sources, allow_imported_mem
 
 def _is_function_defined_in_module(func, module):
     return func.__module__ == module
-
-
-def _format_duplicated_functions(duplicated_functions, functions, source):
-    """Format an error message showing duplicated functions and their sources."""
-    lines = []
-    for name in duplicated_functions:
-        lines.append(f"'{name}' is defined in")
-        lines.append("    " + inspect.getfile(functions[name]))
-        lines.append("    " + inspect.getfile(source[name]))
-
-    return "\n".join(lines)
--- a/gettsim/shared.py
+++ b/gettsim/shared.py
@@ -1,5 +1,8 @@
+import functools
 import inspect
 import textwrap
+
+import numpy as np
 
 
 def format_list_linewise(list_):
@@ -57,3 +60,45 @@ def get_names_of_arguments_without_defaults(function):
             argument_names_without_defaults.append(parameter)
 
     return argument_names_without_defaults
+
+
+def round_out(base, round_d, direction):
+    """Decorator to round the output of a function.
+
+    Parameters
+    ----------
+    base : float
+        Precision of rounding (e.g. 0.1 to round to the first decimal place)
+    round_d : bool
+        Whether rounding should be applied
+    direction : str
+        Whether the series should be rounded up, down or to the nearest number
+
+    Returns
+    -------
+    results : pandas.Series
+        Series with (potentially) rounded numbers
+    """
+
+    def inner(func):
+        # Make sure that signature is preserved
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            out = func(*args, **kwargs)
+            if round_d:
+                if direction == "up":
+                    return base * np.ceil(out / base)
+                elif direction == "down":
+                    return base * np.floor(out / base)
+                elif direction == "nearest":
+                    return base * (out / base).round()
+                else:
+                    raise ValueError(
+                        "direction must be one of 'up', 'down', or 'nearest'"
+                    )
+            else:
+                return out
+
+        return wrapper
+
+    return inner
--- a/gettsim/transfers/grundrente.py
+++ b/gettsim/transfers/grundrente.py
@@ -1,10 +1,12 @@
 import numpy as np
 
+from gettsim.shared import round_out
 from gettsim.typing import BoolSeries
 from gettsim.typing import FloatSeries
 from gettsim.typing import IntSeries
 
 
+@round_out(base=0.01, round_d=True, direction="nearest")
 def grundr_zuschlag_m(
     grundr_zuschlag_vor_eink_anr_m: FloatSeries, anrechenbares_eink_gr_m: FloatSeries
 ) -> FloatSeries:
@@ -23,7 +25,7 @@ def grundr_zuschlag_m(
 
     """
     out = grundr_zuschlag_vor_eink_anr_m - anrechenbares_eink_gr_m
-    return out.clip(lower=0).round(2)
+    return out.clip(lower=0)
 
 
 def anrechenbares_eink_gr_m(


gettsim/interface.py
```diff
@@ -531,6 +538,78 @@ def _root_nodes(dag):
...
+        # Make sure that signature is preserved
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            out = func(*args, **kwargs)
+            if direction == "up":
+                return base * np.ceil(out / base)
+            elif direction == "down":
+                return base * np.floor(out / base)
+            elif direction == "nearest":
+                return base * (out / base).round()
```
janosg, line 569: I prefer if each function has just one return statement in the last row. I.e. just assign a variable inside the if conditions and return it once at the end.

gettsim/interface.py
```diff
@@ -94,6 +98,9 @@ def compute_taxes_and_transfers(
     }
     _fail_if_targets_not_in_functions(functions, targets)
 
+    if rounding:
```
janosg, line 101: We should add a comment that it is very important to handle the rounding before the partialling. Just to prevent bugs in future refactoring.

gettsim/interface.py
```diff
@@ -531,6 +538,78 @@ def _root_nodes(dag):
...
+    -------
+    functions_new : dict of callable
+        Dictionary mapping function names to callables for whose outputs appropriate
+        rounding is applied.
+
+    """
+    functions_new = {}
+    for f_name, func in functions.items():
+        if hasattr(func, "__roundingspec__"):
+            try:
```
janosg, line 597: Instead of a try-except block I would just check that all required keys are there and raise a ValueError if not. 

If you stick to the try-except, you should handle `SystemExit` and `KeyboardInterrupt` separately. Here is an [example](https://github.com/OpenSourceEconomics/estimagic/blob/1507ed18e68958cb3278ed97e224a6abd189b201/src/estimagic/optimization/internal_criterion_template.py#L158).

janosg (commented):
Very nice PR. Just a few minor stylistic things. 

Since the implementation deviates quite a bit from what has been discussed in #88 (but I think in a very good way!) you should probably explain your implementation (either in #88 or the PR description).

docs/how_to_guides/disable_rounding.ipynb
```diff
@@ -0,0 +1,665 @@
...
+     "execution_count": 9,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from gettsim.shared import add_rounding_spec\n",
+    "\n",
+    "@add_rounding_spec(base=0.01, direction=\"up\")\n",
+    "def your_own_function():\n",
```
hmgaudecker, line 598:
```suggestion
    "def flat_tax(income, tax_params):\n",
    "    return income * tax_params['rate']\n",
```

We should stay a little closer to real-world examples

docs/how_to_guides/disable_rounding.ipynb
```diff
@@ -0,0 +1,665 @@
...
+      ]
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plt.plot(\"entgeltpunkte\", \"grundr_zuschlag_m\", data=results.join(data).query(\"entgeltpunkte < 10.01\"))"
```
hmgaudecker, line 532:
```suggestion
    "plt.plot(\"entgeltpunkte\", \"grundr_zuschlag_m\", data=results.join(data).query(\"entgeltpunkte < 10.01\"));"
```
Nicer output

hmgaudecker:
Looks great, thanks! Just a couple of tiny notes and one big wish
- Could you rebase on the `main` branch instead of `grundrente` ?
- The how to only talks about transfers, should be "taxes and transfers". Pretty much all tax deductions are rounded up to the nearest Euro etc. -- maybe also use taxes as an example in the tutorial.
- It would be great if rounding could just be specified in the yaml files. That's where it belongs, really. This would require some mechanism for programmatically decorating functions (much like using `importlib` for importing modules) and I am not sure whether that is possible. @janosg, any ideas?

ChristianZimpelmann:
> Looks great, thanks! Just a couple of tiny notes and one big wish
> 
> * Could you rebase on the `main` branch instead of `grundrente` ?
> * The how to only talks about transfers, should be "taxes and transfers". Pretty much all tax deductions are rounded up to the nearest Euro etc. -- maybe also use taxes as an example in the tutorial.
> * It would be great if rounding could just be specified in the yaml files. That's where it belongs, really. This would require some mechanism for programmatically decorating functions (much like using `importlib` for importing modules) and I am not sure whether that is possible. @janosg, any ideas?

Thank you for the remarks and suggestions which I like. Just for the big wish:
Why do you think that it belongs to the parameter files? Currently, we never round parameters, but always targets/functions, which are at the moment not represented in the .yaml-files. Hence, I think from the user perspective it would make more sense to specify the rounding in the same file where the functions/targets are specified. 

From the implementation perspective, it would be fairly easy to have a separate parameter file which specifies all functions that should be rounded. Then we wouldn't need to add attributes to the function, but would just loop over those specified functions in `compute_taxes_and_transfers` instead of all annotated functions. But as lined out before, I think it would be less straight-forward for users. If they would like to check which functions are rounded or if they would like to supply their self-written function to `compute_taxes_and_transfers`, they need to have a look at both the function code and the respective .yaml file and/or `policy_params` dictionary.

hmgaudecker:
My thinking was that rounding is specified as a parameter in the law, just like any other number. 

So for the example, we might have something like this in `ges_renten_vers.yaml`

```yaml
rounding:
- grundr_zuschlag_m:
  - direction: up
  - base: 0.01
  - reference: Whatever law
```

All in one place. Definitely no separate file!

As a user, why would I ever care which functions are rounded? Either I want the law, or fairly smooth outcomes. Can't imagine a mix of the two being relevant and if so, they need to dig deep into stuff, anyhow.

janosg:
I agree with Christian that implementing the optional rounding with a `yaml` file would even be simpler than the current implementation because we would not need the decorator that adds the `__roundingspec__`. However, I am not sure if we should do it. 

One nice thing about the current implementation is, that it plays really well with user defined functions that also need optional rounding. If all rounding information needs to be stored in a yaml file, how would that work for user defined functions? Would they have to load the yaml file, add more entries to the dictionary and then pass it as a function argument?

If we need the yaml file for documentation, it could easily be generated by reading the information from the decorated functions and combining it into a dictionary. 

If there absolutely needs to be a yaml source file, but we want to maintain the same behavior for user functions we could allow both ways. i.e. check if `__roundingspec__` exists or there is an entry in the yaml file.

ChristianZimpelmann:
> If all rounding information needs to be stored in a yaml file, how would that work for user defined functions? Would they have to load the yaml file, add more entries to the dictionary and then pass it as a function argument?

The rounding parameters could be just part of the already existing `policy_params` dictionary which could be manipulated by the user.

But I still think that from the user perspective writing a function and setting the attribute (via a decorator) is more straight-forward than writing a function and manipulating the `policy_params` dictionary (something like `policy_params["ges_renten_vers"]["grundr_zuschlag_m"]["direction"] = "up"`). (and e.g. make sure that there is no typo in the function name)

But it also makes sense that it is kind of a parameter in the law.. I need to think another time what I think is more straight-forward.

hmgaudecker:
I was actually thinking of a combination. Leave the current implementation, but make it implicit.

Much like I'd usually write  verbatim

```python
from path.to.module import function
```

I can achieve the same thing programmatically using importlib. I was thinking about whether we the same is possible here: Read rounding from the yaml file, if relevant, add it, else don't.

I am not sure we need to think about users defining their function with optional rounding. Seems like an empty set to me. Either, they could pass in either one, depending on the use case. If it is actual laws to be implemented, we are talking more at the level of PRs in the first place.

ChristianZimpelmann:
I added a short and rough draft of a GEP about optional rounding: https://github.com/iza-institute-of-labor-economics/gettsim/blob/10341dfc8c496ed5b6d11a9cae93729f515dc76d/docs/geps/gep-05.rst

@hmgaudecker, @janosg, would be great if you can have a look and make changes/additions.

I am not sure whether the way I describe Option 3 is what you had in mind, @hmgaudecker? This seems to me the best way how to combine option 1 and 2.

ChristianZimpelmann:
- Functionality is already implemented in #343. 
- Rounding specs for Grundrente functions is trivial, can be done in #275 directly.
- Tutorial `docs/how_to_guides/disable_rounding.ipynb` can be adjusted to new functionality and added in a separate PR (I will keep track).

PR #317, opened 2021-12-13
Define age as a function.

### LINKED ISSUE #211: Age / time / cohort effects: GETTSIM version

We currently require:

- Date of birth
- Age
- Policy year

which can lead to discrepancies.

We should require only two and calculate the third.

### Proposed implementation

Drop requiring `alter` and always require a parameter `datum` to be passed to `compute_taxes_and_transfers` (= same for everyone). Then calculate `alter` based on the other two.

`datum` would typically be the same as the one supplied to `get_policy_for_date`, but it may differ for simulations etc.

### Considered alternatives

- could become part of `params`, too. Then we'd need to differentiate between "policy_date" and "evalutation_date" there.

LauraGergeleit:
Eliminate discrepancies which are subject to requiring date of birth, age and policy year. Only the date of birth and policy year should be required and the age should be calculated. Therefore, age now has been defined as a function of date of birth and an evaluating_date, which shall be passed to 'compute_taxes_and_transfers'.

Closes #211 

### Todo

- [ ] Define the input variable evaluating_date and pass it to 'compute_taxes_and_transfers'.


hmgaudecker:
Thanks! Cannot really check right now, but this seems like a good opportunity to think about precision of time variables. Entire years will not be enough for many variables. 

That is, how do we handle rounding? Many policies will be relevant at the level of the month (e.g., `kindergeld` will be paid for entire months only), sometimes only years will be relevant. 

Where will the rounding be done? E.g., record age in years + fractions and round in the functions? Or are we storing separate values for age (age_m, age)? ... 

We could of course use this as a prototype and then extend later, no strong opinion, but I thought I'd bring it up.

ChristianZimpelmann:
I have two more questions:

### How should the two different dates we are using be named:

1. the date for which the policy system is set up (parameters, functions, etc.): This is currently just called `date`. I would change it to `policy_date`.
2. the date for which the data set is evaluated. In #211, @hmgaudecker proposed `evaluation_date`, but I am not sure if the distinction is clear enough.

### What is/are the typical use case(s) where `policy_date` and `evaluation_date` differ?

What I can think of is that we only have data for 2018, but would like to run it in 2021. But don't we would like to keep age, age at retirement, etc. constant and vary `birth_year`? Then we should consider having `alter` as input and calculating `birth_year` via `evaluation_date. I am not sure here as from the perspective of data cleaning, having birth_year as constant and in more detail seems to make more sense.

What do you think? (also tagging @Eric-Sommer @mjbloemer

hmgaudecker:
Cool. 

Yes, I'll believe right away we only use full years right now. The question is: if we touch this anyhow, why don't we do it "right". 

So it is great to have this discussion.

Re policy date vs. evaluation date: This is actually a great example for why having only age would have been a good thing. Past tense, because I believe this is true for almost all things that have been part of GETTSIM so far (am I correct on this one, @mjbloemer @Eric-Sommer ?)

Enter Pensions. Here laws often take effect based on an age/cohort combination. The question is how we want to deal with it in the end and what we would like to allow the user to choose.

In the end, `year = cohort + age` and we probably want to use only two of these, calculating the third.

Seems like a good example to 
- check what others are doing, e.g. OpenFisca, EUROMOD, ...
- discuss this during the next developer call

mjbloemer:
> Re policy date vs. evaluation date: This is actually a great example for why having only age would have been a good thing. Past tense, because I believe this is true for almost all things that have been part of GETTSIM so far (am I correct on this one, @mjbloemer @Eric-Sommer ?)

Currently there are already elements where the birth year is important, e.g.: https://github.com/iza-institute-of-labor-economics/gettsim/blob/7e9a386b2320139e32023d42926aabdb0a223bcb/gettsim/parameters/arbeitsl_geld_2.yaml#L277

ChristianZimpelmann:
> Re policy date vs. evaluation date: This is actually a great example for why having only age would have been a good thing.

I agree that if we decide to have age as input and, hence, `geburtsjahr` etc. only as internal functions, `evaluation_date` seems a bit clearer.

I also think it will be beneficial to check out EUROMOD and OpenFisca.

hmgaudecker:
Thanks, everyone! We are really opening a can of worms here... We'll need to map out the use cases very clearly and design the interface accordingly, I think.

Eric-Sommer:
> ### What is/are the typical use case(s) where `policy_date` and `evaluation_date` differ?
> 
> What I can think of is that we only have data for 2018, but would like to run it in 2021. 

This is the typical use case. Others are ex-post decomposition studies, like [this one](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=1913617). The easiest thing to do is to use `alter` as is, i.e. apply the 2021 regime to the 2018 population. Otherwise, you'd have to worry about what to do with newborns, how to deal with entering retirement etc.. From that perspective, I have sympathy for keeping `alter` fixed (so I agree with @hmgaudecker). 

In general, we are talking about _what happens if evaluation date and policy date differ?_ In IZAMOD, we did not touch any demographics, but uprated monetary income components with the CPI. According to the gettsim philosophy, we would leave such data mingling to the user, so users should make sure their data do follow a sensible age structure before entering gettsim. And if users can mingle with age, they can equivalently do so with birth date.

hmgaudecker:
The more I think about it, the more complex it becomes...

If you do not need to worry about small kids or migration, Eric's example would actually mean that specifying birth date and evaluation year would be ideal. Just make everyone three years older and birth-date based rules like those linked by Maximilian above apply as they are.

In other cases, you may want to use 2018 population structure. Then you would want to shift birth dates. 

I agree most of this should be left to the user, but I think if in some cases we specify age-based rules and in other cases cohort-based rules, it won't be possible for her to lay out all cases. It just becomes complicated once you think of dynamic models...

hmgaudecker:
Probably will be useful to start over at some point, what do you think, @ChristianZimpelmann @LauraGergeleit ?

*(no rush in replying)*

ChristianZimpelmann:
> Probably will be useful to start over at some point, what do you think, @ChristianZimpelmann @LauraGergeleit ?

Agreed!

Issue #127, opened 2019-12-12
param style guidelines


mjbloemer:
Let us have a short guideline how to format the description and note key in the parameter yaml files.

Initial thoughts:

> I have not much to say about note style guidelines. This really is a detail. We can have a separate discussion on that, sure (about the comma, the full stop and things like "Artikel.."). [Here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/69) I basically just follow the principle to make it is as easy as possible to look up the parameter in the law over on offenegesetze.de and buzer.de plus just have a consistent citing style (it is the citing style  buzer.de uses btw. So it is even very easy to copy and paste the citation).

_Originally posted by @mjbloemer in https://github.com/iza-institute-of-labor-economics/gettsim/pull/74_

> Well, that sounds like a guideline, but it would need to be made consistent in all files. Actually, in order to make full-text search work, one would need to replace the month to words. ('September' instead of '.09.'). Using this bit: `2017 BGBl. I S. 3525` should however lead to the correct bit.

_Originally posted by @Eric-Sommer in https://github.com/iza-institute-of-labor-economics/gettsim/pull/74_

Eric-Sommer:
Add-on: The main purpose is to enable full-text search at [offenegesetze.de](offenegesetze.de). It seems to be enough to stick to the format `[YYYY] BGBL. I S. [SSSS]`. What's before or after is of less importance.

mjbloemer:
suggestion/opinion:

the `name` key
* should have a short (!) name without stating the realm (e.g. "ALG II" or "Kinderzuschlag") again
* is not a sentence
* should always have a `de` and `en` name

the `description` key
* should give a good and full explanation of the parameter
* show the § and Gesetzbuch/Paragraph (history) of that parameter ([Example](https://github.com/iza-institute-of-labor-economics/gettsim/blob/79ca7cedd79566c453f41d4bd6ddddfb81731219/gettsim/data/arbeitsl_geld_2.yaml#L18): this parameter was initially defined in § 30 SBG II, later in § 11b SGB II).
* It should mention bigger amendments/Neufassungen and be as helpful as possible to make sense of that parameter.
* if an english translation is added it should state the same information.

for the `note` key:
* style `Artikel [n] [G./V./B.] v. [DD.MM.YYYY] BGBl. I S. [SSSS]. [optional note].`
[...]

PR #323, opened 2022-01-08
Align GKV parameters with law

Eric-Sommer:
So far, we have defined contribution rates separately for employer and employee, doing our own calculations regarding the top-up ('Zusatfreibetrag'). This is now done explicitly.

- [x] Parameters are specified explicitly as in the law, i.e. general rate and additional rate.
- [x] Actual employee/employer split is done in functions
- [x] Introduce new parameter *ermäßigter Beitragssatz* as it is required for Lohnsteuer
- [x] adjust dashboard 

We can distinguish four phases:
1. until 06/2005, contribution rates differed between entities, but were equally shared between employeer and employee. In general, rates hardly differed among the main entities, but smaller/regional entities offered lower rates and sometimes went bancrupt as a consequence.
2. from 07/2005 to 2008, the contribution rates differed between entities, and employees had to pay a top-up rate (also differing by entity)
3. From 2009 to 2018, the general contribution rate was fixed and the top-up rate differed between entities.
4. Since 2019, employees and employers share the contributions, i.e. `(general rate + top-up rate)/2`. Top-up rate still varies between entities

Applying strict gettsim philosopy would mean that we do not impose any parameter if it's not statutory. Therefore, the parameter `allg` is set to `null`. Apart from these considerations, the remaining difference between regimes is whether contributions are split between employer and employee or not. This is reflected by the two functions calculating the actual contributions.

Screenshot from Dashboard:
grafik


hmgaudecker:
We might want to wait a little bit with this one. IIRC, there have been numerous changes over the years about which parameters are specified in the law etc. (at some point 50-50 determined within bands where we would have to pick some empirical value, then AG specified and AN sort-of-free, ... lots of different things)

Depending on what we do with the functions over the years (ref [GEP PR](https://gettsim--178.org.readthedocs.build/en/178/geps/gep-04.html#level-of-the-dag-and-limitations), [Zulip discussion](https://gettsim.zulipchat.com/#narrow/stream/309998-GEPs/topic/GEP.2004/near/267191436)), we may actually want to have different parameter names by policy in effect.

hmgaudecker:
Thanks!!! I think it would be fine to implement this now with the following changes, as discussed on Zoom.

- Have different functions by policy regime, using the machinery in gettsim/policy_environment.py even more than what you do already
- Use the exact parameters as specified in the law (as you did), using longer names for keys like "ag" -- just write out.
- During periods where amounts are set by sickness funds (nineties big time, but also the Zusatzbeitrag in recent years), require a user column. The content and name of that will be different over different regimes (e.g. `ges_krankenv_beitragssatz` in the nineties which includes employer / employee (parity, IIRC), `ges_krankenv_zusatzbeitrag` in recent years.
- Clearly document the shifts in the law, in params-yaml but also function docstrings.

mjbloemer:
There is a parameter 2005-7 to 2014 in § 249 SGB V (definitely not a user column). Maybe call it `ges_krankenv_nur_arbeitnehmer`. During that time the employer has to pay half of the `ges_krankenv_beitragssatz` (allgemeiner Beitragssatz) which has been reduced by that parameter (I think a typical or the only value was 0.009) and the employee pays the rest. I sketched it here: https://github.com/iza-institute-of-labor-economics/gettsim/issues/127#issuecomment-568175694.

hmgaudecker:
Yeah, I won't claim to remember the exact details of every policy regime there was. And there certainly is no need for a user column in every year; whether there is one and which one is required will depend on the policy regime.

At present, there should be a user column, see, e.g. https://www.gkv-spitzenverband.de/service/krankenkassenliste/krankenkassen.jsp

Although admittedly, the differences are not very large. Was a very different situation back in the day...

Eric-Sommer:
I added some clarifications in the yaml file and the initial comment of this PR. I did some research on the situation before 2008. There was considerable competition and smaller entities were able to attract lots of contributors via low rates. Sometimes this lead to bankruptcies of these entities. Among the larger players, contributions apparently did not differ that much.

hmgaudecker:
Thanks. I do not know where the bankruptcy stories come from or why they are relevant? Thing is, there were large price differences and people could switch freely. That is the nature of competition. We also don't assume anything on private health insurance contributions. 

We just should not be putting anything there but require a user column. We have a clear philosophy on this.

mjbloemer:
What exact variable would be the user column? The contributions in Euro (do we have that for private health insurance at the moment?) or the contribution rates? Something like the latter might be important for applications where GETTSIM is used for counterfactuals and for discrete choice models. Do the differences boil down to differences in rates or were there more complex GKV pricing schemes?

hmgaudecker:
> The contributions in Euro (do we have that for private health insurance at the moment?) or the contribution rates? 

The sweet thing about the DAG is -- it does not matter! In principle, we should have a function converting the rates, gross wage, and whether in marginal employment / midi job to monetary quantities. At present, we might be using the rates directly here and there, might require some trivial refactoring.

> Something like the latter might be important for applications where GETTSIM is used for counterfactuals and for discrete choice models.

Agreed. I can think of datasets / applications where either one would be better suited.

> Do the differences boil down to differences in rates or were there more complex GKV pricing schemes?

As far as I remember—have not taught my Health Econ course in a couple of years, so a bit rusty—only the rates differed. At some point I think sickness funds were experimenting with no-claims refunds, but that was stopped by law. I might be making this up, though.

Eric-Sommer:
> > Do the differences boil down to differences in rates or were there more complex GKV pricing schemes?
> 
> As far as I remember—have not taught my Health Econ course in a couple of years, so a bit rusty—only the rates differed.

That's my understanding as well.

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -45,31 +58,28 @@ soz_vers_beitr:
...
     ges_krankenv:
-      ag: 0.0644
-      an: 0.0644
+      allg: null
   1989-01-01:
     deviation_from: previous
     ges_krankenv:
-      ag: 0.0645
-      an: 0.0645
+      allg: 0.129
```
ChristianZimpelmann, line 65: did it really exist just for 1 year?

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -45,31 +58,28 @@ soz_vers_beitr:
...
     ges_krankenv:
-      ag: 0.0625
-      an: 0.0625
+      allg: null
   1991-01-01:
     deviation_from: previous
     ges_krankenv:
-      ag: 0.064
-      an: 0.064
+      allg: null
```
ChristianZimpelmann, line 73: date entries can be deleted if nothing was changed.

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -91,71 +100,65 @@ soz_vers_beitr:
...
     ges_rentenv: 0.0975
-  2006-01-01:
+  2005-07-01:
     deviation_from: previous
     ges_krankenv:
-      ag: 0.0666
-      an: 0.0756
+      allg: null
+      zus: 0.009
+    note: Allgemeiner Beitragssatz wird nicht mehr gleichmäßig aufgeteilt, Einführung eines Zusatzbeitrags
```
ChristianZimpelmann, line 149: Adjust note. If I understand correctly, this is also the first time that general contribution rates are defined (and part of the GETTSIM parameters)

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -67,11 +67,13 @@ def ges_krankenv_beitr_m_tu(
     return ges_krankenv_beitr_m.groupby(tu_id).sum()
 
 
-def _ges_krankenv_beitr_reg_beschäftigt(
+def _ges_krankenv_beitr_reg_beschäftigt_nicht_pari(
     bruttolohn_ges_krankenv_beitr_m: FloatSeries, soz_vers_beitr_params: dict
 ) -> FloatSeries:
     """Calculates health insurance contributions for regular jobs
 
+    Between 07/2005 and 12/2019,
```
ChristianZimpelmann, line 75: Isn't it 12/2018?

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -258,7 +316,8 @@ def ag_beitr_ges_krankenv_midi_job(
     bruttolohn_m_in_gleitzone = bruttolohn_m.loc[in_gleitzone]
     return (
         bruttolohn_m_in_gleitzone
-        * soz_vers_beitr_params["soz_vers_beitr"]["ges_krankenv"]["ag"]
+        * soz_vers_beitr_params["soz_vers_beitr"]["ges_krankenv"]["allg"]
```
ChristianZimpelmann, line 319: Is it correct that the "Zusatzbeitrag" is not considered here under all regimes?

ChristianZimpelmann (commented):
Thank you @Eric-Sommer for moving this forward! I added a couple of comments. The largest open points:
- It seems like the regime before 2005 is not fully implemented. If I understand correctly, we wanted to introduce an input variable for this time period and would, therefore, need functions that calculate e.g. `_ges_krankenv_beitr_reg_beschäftigt` based on an this input variable instead of based on the general parameters (which you have already deleted). Then we should include the input parameter in `input_variables.rst` (add information on in which time period this input column is needed) and `config.py`.
- Didn't we also plan to make the Zusatzbeitrag rate a user-provided input column as it differs over individuals? 

To be honest, for the moment, I would be also fine with keeping the "average rates" for Zusatzbeitrag and pre-2005 rates as parameters. A user with more information can always override the respective columns and it would save us quite some work refactoring (although this solution is not perfectly clean as it is not a proper "parameter"). But @hmgaudecker seems very convinced? Either way, we should be consistent.

hmgaudecker:
>     It seems like the regime before 2005 is not fully implemented. If I understand correctly, we wanted to introduce an input variable for this time period and would, therefore, need functions that calculate e.g. `_ges_krankenv_beitr_reg_beschäftigt` based on an this input variable instead of based on the general parameters (which you have already deleted). Then we should include the input parameter in `input_variables.rst` (add information on in which time period this input column is needed) and `config.py`.

There were multiple regime changes before 2005, too. Probably quite homogenous until early 90s, then plenty of reforms, including that period with lots of competition. By 2005, the red-green govt had already curbed competition quite a bit.

>     Didn't we also plan to make the Zusatzbeitrag rate a user-provided input column as it differs over individuals?

Yes, I think so. Although again, probably different regimes.

> To be honest, for the moment, I would be also fine with keeping the "average rates" for Zusatzbeitrag and pre-2005 rates as parameters. A user with more information can always override the respective columns. Although this solution is not perfectly clean as it is not a proper "parameter", it would save us quite some work refactoring and wouldn't make GETTSIM more effortfull to use as all these rates need to be added to the input data set. This would be easier at some later stage when we allow for default values of input variables. But @hmgaudecker seems very convinced? Either way, we should be consistent.

Fine. This definitely is an improvement over the status quo and we do not need to fix everything here. In the long run, I do feel strongly about doing it the "right" way :wink:

hmgaudecker:
Btw, we should aim to merge by Monday.

ChristianZimpelmann:
> I would be also fine with keeping the "average rates" for Zusatzbeitrag and pre-2005 rates as parameters.

@Eric-Sommer, would be great if you can implement this. Please make sure that the pre-2005 rates are put into a separate parameter and add to that parameter description that it is just the average rate and can in fact differ between individuals.

gettsim/parameters/soz_vers_beitr.yaml
Eric-Sommer: no, this should be removed as well...

gettsim/parameters/soz_vers_beitr.yaml
Eric-Sommer: No, the general ones were introduced in 2009.

gettsim/social_insurance_contributions/ges_krankenv.py
Eric-Sommer: correct

gettsim/social_insurance_contributions/ges_krankenv.py
Eric-Sommer: This is the employer part, why should it depend on "zusatzbeitrag"?

dashboard/app/main.py
```diff
@@ -95,9 +95,9 @@
         f"""This graph depicts contribution rates to the four main branches of
         social insurance. With the exception of health insurance from 2006 to
         2018, contributions are shared equally between employer and employee.
-        The graph shows only the employees share for those branches. The
-        additional health care contribution rate for employees used to
+        The additional health care contribution rate for employees used to
         vary across health insurance funds; we assume the national average.
+        Before 2008, there was no general health contribution rate.
```
hmgaudecker, line 100:
```suggestion
        In the period 1993-2007, competition between sickness funds meant there was not one contribution rate. GETTSIM provides an average.
```

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -4,39 +4,57 @@ soz_vers_beitr:
...
+                    durchschnitt - average contribution rate, until 2008
+                    allgemein - general statutory contribution rate, since 2009
+                    ermäßigt - reduced rate
+                    zusatz - average top-up contribution rate
   unit: share
   1984-01-01:
     ges_krankenv:
-      ag: 0.0572
-      an: 0.0572
+      durchschnitt: 0.1144
```
hmgaudecker, line 21: Move to `allgemein` until 1992. See, e.g., https://www.aok-bv.de/hintergrund/gesetze/index_15084.html

gettsim/policy_environment.py
```diff
@@ -11,6 +11,57 @@
 from gettsim.config import ROOT_DIR
 from gettsim.piecewise_functions import check_thresholds
 from gettsim.piecewise_functions import get_piecewise_parameters
+from gettsim.social_insurance_contributions.eink_grenzen import (
+    midi_job_bemessungsentgelt_since2009,
```
hmgaudecker, line 15:
```suggestion
    midi_job_bemessungsentgelt_ab_2009,
```

and similar for everything below.

gettsim/policy_environment.py
```diff
@@ -11,6 +11,57 @@
...
+)
+from gettsim.social_insurance_contributions.eink_grenzen import (
+    midi_job_bemessungsentgelt_until2008,
```
hmgaudecker, line 18:
```suggestion
    midi_job_bemessungsentgelt_ab_2005_bis_2008,
```

Note: 2005 is from the top of my head, might well be wrong.

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -190,31 +327,116 @@ def ges_krankenv_rente(
...
+    Parameters
+    ----------
+    ges_krankenv_rente
+        See :func:`ges_krankenv_rente`.
+    soz_vers_beitr_params
+        See params documentation :ref:`soz_vers_beitr_params <soz_vers_beitr_params>`.
     Returns
     -------
-    Pandas Series containing monthly health insurance contributions for pension income.
+    Pandas Series containing monthly health insurance contributions for self employed
```
hmgaudecker, line 373:
```suggestion
    Monthly health insurance contributions for pension income.
```

hmgaudecker (commented):
Looks great! Main thing is to distinguish 1984-1992 from 1993-2005, else wording and naming conventions (already thinking a bit about #355, but don't worry much about that yourself, externalities caused by me after all.

hmgaudecker:
@Eric-Sommer, if you find time today to tackle the more substantive things I noted, would be great, but don't worry! Else would be great if @ChristianZimpelmann could finish it first thing tomorrow morning. Thanks!

gettsim/policy_environment.py
Eric-Sommer: OK. Gleitzone has been introduced in 2003

Eric-Sommer:
@ChristianZimpelmann : I implement the suggestions by you and @hmgaudecker. Not all tests pass now though. I would need to have a closer look...

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -101,61 +119,59 @@ soz_vers_beitr:
...
-      an: 0.0785
+      durchschnitt: 0.139
     arbeitsl_v: 0.021
     ges_rentenv: 0.0995
   2008-01-01:
     ges_krankenv:
-      ag: 0.07
-      an: 0.079
+      allgemein: 0.14
+      zusatz: 0.009
```
ChristianZimpelmann, line 173: This Zusatz-parameter is a different parameter than zusatz below (which differs over individuals). The `0.009` rate was apparently called Sonderbeitrag. So I propose to rename it to `sonderbeitrag` and the one below to `mean_zusatzbeitrag`.)

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -164,19 +180,22 @@ soz_vers_beitr:
...
+      ermäßigt: 0.140
     arbeitsl_v: 0.014
-  2010-01-01:
+  2009-07-01:
     deviation_from: previous
     ges_krankenv:
-      ag: 0.07
-      an: 0.079
+      allgemein: 0.14
+      ermäßigt: 0.134
```
ChristianZimpelmann, line 191: Parameter ermäßigt is never used in the code.

Do I remember correctly that we need it for #150 ?

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -189,32 +208,34 @@ soz_vers_beitr:
...
       zusatz_kinderlos: 0.0025
     arbeitsl_v: 0.015
     ges_rentenv: 0.0935
   2016-01-01:
     deviation_from: previous
-    note: Bekanntmachung Bundesanzeiger 30.12.2015 und  http://www.bmg.bund.de/ministerium/meldungen/2015/gkv-zusatzbeitrag.html
+    note: Bekanntmachung Bundesanzeiger 30.12.2015
     ges_krankenv:
-      an: 0.084
+      zusatz: 0.011
```
ChristianZimpelmann, line 223: Is this the sum of Sonderbeitrag and `mean_zusatzbeitrag`? If yes, we should separate them as one is the same for everybody and the other one is not.

ChristianZimpelmann (changes requested):
Thank you @Eric-Sommer! There are a few open comments/suggestions. We decided to not merge it today yet.

If you have some time the next days, it would be great if you have another look at it. Otherwise, I will get back to it at the end of the week.

I pushed a small commit renaming the functions.

Btw., [this}(https://www.krankenkassenzentrale.de/wiki/paritaetische-finanzierung#:~:text=CSU%20und%20SPD-,2019,vollen%20Beitragskosten%20zur%20gesetzlichen%20Krankenversicherung.) seems a good overview over the different regimes (although no official source)

gettsim/parameters/soz_vers_beitr.yaml
Eric-Sommer: yes correct

gettsim/parameters/soz_vers_beitr.yaml
Eric-Sommer: [Wikipedia](https://de.wikipedia.org/wiki/Gesetzliche_Krankenversicherung#Beitragssatz_und_Zusatzbeitrag) calls it 'fester Zusatzbeitrag'. I see your point, but this would require even more functions, wouldn't it?

hmgaudecker: N functions does not matter :smile:

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -220,12 +536,48 @@ def _ges_krankenv_beitr_midi_job_m(
...
+    ag_beitr_ges_krankenv_midi_job
+        See :func:`ag_beitr_ges_krankenv_midi_job`.
+    Returns
+    -------
+
+    """
+    return ges_krankenv_beitr_midi_job_m - ag_beitr_ges_krankenv_midi_job
+
+
+def selbstständig_ges_krankenv(selbstständig: bool, in_priv_krankenv: bool) -> bool:
```
ChristianZimpelmann, line 571: Is this used anywhere?

ChristianZimpelmann:
Thank you, @Eric-Sommer! I restructured the code and fixed a few bugs. Most important changes:

- Only the functions `ges_krankenv_beitr_satz` and `_ges_krankenv_beitr_satz_arbeitg` differ over years. All other functions use those functions as inputs and can stay the same over time.
- I use the code snippet below to incorporate years with `sonderbeitrag` and those with `mean_zusatzbeitrag` in the same functions (they never exist both). Similar for `allgemein` and `mean_allgemein`. Alternative would be to have four different versions of `ges_krankenv_beitr_satz` instead of the current two.

```python
    # From Juli 2007 until 2014, Sonderbeitrag is in place
    # From 2015 on, a Zusatzbeitrag is in place which differs between
    # insurance entities
    if "sonderbeitrag" in beitr_satz_params:
        zusatzbeitrag = beitr_satz_params["sonderbeitrag"]
    elif "mean_zusatzbeitrag" in beitr_satz_params:
        zusatzbeitrag = beitr_satz_params["mean_zusatzbeitrag"]
    else:
        zusatzbeitrag = 0
```

Open points:
- The tests for 2022 are not running through, yet. Problem seems to be that they are based on a calculator by Barmer which has a different Zusatzbeitrag than `mean_zusatzbeitrag`. We need a different source for test cases.
- We need to check again the parameters and the docstrings of all functions
- Decide if my solution above is too "hacky".

@Eric-Sommer, would you like to have a look again?

gettsim/social_insurance_contributions/ges_krankenv.py
Eric-Sommer: Probably a reminiscent of what is `ges_krankenv_beitr_selbst_m()` by now.

Eric-Sommer:
I like your approach a lot, @ChristianZimpelmann! I looked through the docstrings. Beyond, I found an onlince calculator that lets you specify the 'Zusatzbeitrag'. Now the 2022 test works as well.

ChristianZimpelmann:
Thanks, @Eric-Sommer! Great job.

I forgot to mention above: I also changed the following variable name:

- ``_ges_krankenv_beitr_midi_job_m`` -> `_ges_krankenv_midi_job_arbeitn_m`  (in line with ``_ges_krankenv_midi_job_arbeitg_m`` and ``_ges_krankenv_midi_job_sum_arbeitn_arbeitg_m``)

@mjbloemer, would you also like to have a look?

@hmgaudecker, do you have a strong opinion against the code snippet above? I think it actually make the code clearer instead of having 4 functions as it is just one parameter replacing another (before this PR it was even called the same, but it is much better now that we call it differently).

Otherwise, the only open point is to tell codecov that the dashboard/ directory should be ignored. I will continue to work on this.

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -34,31 +31,127 @@ def ges_krankenv_beitr_m(
...
+    Parameters
+    ----------
+    soz_vers_beitr_params
+        See params documentation :ref:`soz_vers_beitr_params <soz_vers_beitr_params>`.
+
+    Returns
+    -------
+
+    """
+    beitr_satz_params = soz_vers_beitr_params["beitr_satz"]["ges_krankenv"]
```
hmgaudecker, line 63:
```suggestion
    params = soz_vers_beitr_params["beitr_satz"]["ges_krankenv"]
```
This is just a short function, so fine have a shorthand to remove context. 

The prior version is actually confusing without seeing this line because it is the same as the outer dict without the `ges_krankenv` specifier.

hmgaudecker:
> @hmgaudecker, do you have a strong opinion against the code snippet above? I think it actually makes the code clearer instead of having 4 functions as it is just one parameter replacing another (before this PR it was even called the same, but it is much better now that we call it differently).

You mean `ges_krankenv_beitr_satz_bis_2018` Fine with me with the trivial change I just suggested.


> Otherwise, the only open point is to tell codecov that the dashboard/ directory should be ignored. I will continue to work on this.

Feel free to ignore. Current dashboard is doomed, anyhow. Very likely plotly/dash, but with the recent announcement of pyscript (e.g. https://betterdatascience.com/pyscript-intro/) it seems all the more helpful not to make that a pressing concern. Let's see how things pan out over the coming months.

PR #324, opened 2022-01-08
GEP 5 - optional rounding

ChristianZimpelmann:
GEP for the implementation of optional rounding of taxes or transfers. 

This should help us to decide about the best implementation and describe the interface. Afterwards we will implement optional rounding in #316.

I moved the GEP from #316 to this new PR. See there, for some earlier discussions.


docs/geps/gep-05.rst
```diff
@@ -0,0 +1,124 @@
...
+  to be a floating point number.
+- The parameter ``direction`` has to be one of ``up``, ``down``, or ``nearest``.
+- The reference must contain the reference to the law, which specifies the rounding.
+
+The rounding parameters become part of the dictionary ``policy_params`` once
+``set_up_policy_environment`` is called in the same way as other policy parameters.
+
+.. code-block:: yaml
+
+       rounding:
```
hmgaudecker, line 58: Add which file this is and requirements on structure. At which level is rounding specified, for example.

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,124 @@
...
+         grundr_zuschlag_m:
+           2020-01-01:
+             base: 0.01
+             direction: nearest
+             reference: Whatever law
```
hmgaudecker, line 63: Please use actual example and add the law in the text above, too.

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,124 @@
...
+           grundr_zuschlag_vor_eink_anr_m: FloatSeries, anrechenbares_eink_gr_m: FloatSeries
+       ) -> FloatSeries:
+           ...
+           return out
+
+The decorator adds the attribute ``__rounding_parameter_file__`` to the function. When
+calling ``compute_taxes_and_transfers`` with ``rounding=True``, GETTSIM loops over all
+functions. If a function has a rounding attribute, the rounding is applied based on the
+specification in the argument ``policy_params``. In the example above, the rounding
+would be based on ``policy_params["ges_renten_vers"]["rounding"]["grundr_zuschlag_m"]``.
```
hmgaudecker, line 82: Describe what happens if we specify only one but not the other.

- Specify in yaml but forget the decorator
- Specify decorator but forget in yaml
- If one thing has no consequences in terms of errors (probably the case for the first), specify a strategy for writing tests so as to minimize the chances of wrong assumptions regarding the behavior.

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,124 @@
...
+happened to write a function that requires an intermediate variable to be rounded, split
+it up and create a new function returning exactly the potentially rounded value.
+
+
+User-written functions
+~~~~~~~~~~~~~~~~~~~~~~
+
+For self-written functions, the user needs to add the rounding parameters to
+``policy_params`` and decorate the respective functions with ``add_rounding_spec`` in
+the same way as demonstrated above.
```
hmgaudecker, line 94: Add an actual example here and walk through it, just like above.

hmgaudecker (commented):
Looks great! As always, a few requests for enhancements / more detail...

hmgaudecker:
Thanks for the update!  I made some changes, 0fd9c1a should be pretty uncontroversial. 

In e2fe5f5, I suggest some deeper renaming and argument change. That is, we do not really need (or want, I believe) the reference to the parameters **file** in the decorator or its argument. I changed that so it refers to the argument of the function. In the example case, one could even think about leaving it out. But in general, one might have several `xxx_params` this depends on, so I think we should just make it mandatory. Can you please check whether this makes sense @ChristianZimpelmann @janosg ?

Sorry for not suggesting this in the previous iteration, it occurred to me only now in all its clarity. 

Finally, I commented out the  "user-written rounding" section for the moment. In a way, it was too much tutorial-style, I think, and less about the actual implementation. I'll be happy to update that once the previous part is settled.

I really think these iterations show the value in the GEP process, by the way!

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,198 @@
...
+Note that GETTSIM only allows for optional rounding of functions' results. In
+case one is tempted to write a function requiring an intermediate variable to be
+rounded, the function should be split up so that another function returns the
+quantity to be rounded.
+
+Error handling
+~~~~~~~~~~~~~~
+
+In case a function has a ``__rounding_arg__``, but the respective
+parameters are missing in ``policy_params``, an error is raised.
```
ChristianZimpelmann, line 123: I just realized that this would also happen if rounding of a variable would change over time such that the variable isn't rounded at some point in time. For this time period, rounding parameters would be missing.

I would hence suggest that we also just test that all decorated functions have rounding parameters specified (for at least one point in time), but don't raise an error.

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,198 @@
...
+.. code-block:: python
+
+    @add_rounding_spec(func_arg="ges_rentenv_params")
+    def höchstwert_grundr_zuschlag_m(
+        grundrentenzeiten: IntSeries, ges_rentenv_params: dict
+    ) -> FloatSeries:
+        ...
+        return out
+
+The decorator adds the attribute ``__rounding_arg__`` to the function. When
```
ChristianZimpelmann, line 100: Great change to use the name of the parameter dict instead of the file!

But can you explain the rational behind the name func_arg? Why is it different than the attribute later?

docs/geps/gep-05.rst
hmgaudecker: This might be a case where I would it is better to require explicit setting of rounding to `None` or the like in the parameters file for periods where it does not exist. If I see the `rounding_spec` decorator, I expect it to take effect and do not want to hide required action at a distance.

That is, for the GETTSIM codebase, it is no big deal either way, I am thinking more about user-provided changes here where I would want to see errors.

docs/geps/gep-05.rst
hmgaudecker: My thinking was 
- `func_arg` as an argument to the decorator to make clear the name has to come from the function, `arg` is just too generic.
- for `__rounding_arg__`, I thought it was clearer without the `func`. I changed that from `__rounding_parameter_file__`, but it probably should have been `__rounding_spec__` in the first place? That is, it will contain base/direction, as in the (otherwise outdated) implementation in #316?

docs/geps/gep-05.rst
ChristianZimpelmann: We could allow for ``direction = None``

docs/geps/gep-05.rst
ChristianZimpelmann: > func_arg as an argument to the decorator to make clear the name has to come from the function

I don't understand. To my understanding, the name comes from the name of the parameter dictionary.

> That is, it will contain base/direction, as in the (otherwise outdated) implementation in implement decorator for rounding. #316?

No, I won't contain the parameters as those needed to be read (including the correct date) from the parameter file when the function is read.

docs/geps/gep-05.rst
hmgaudecker: direction / base or both should be `None`, I'd say.

docs/geps/gep-05.rst
ChristianZimpelmann: So I would prefer ``__rounding_parameter_file__`` for both the attribute name and the argument of the decorator.

hmgaudecker: > No, I won't contain the parameters as those needed to be read (including the correct date) from the parameter file when the function is read.

I do not understand this sentence.

hmgaudecker: > So I would prefer `__rounding_parameter_file__` for both the attribute name and the argument of the decorator.

My point is that with what I am suggesting, you only need to look at this code snippet to understand what is going on.

```python
    @add_rounding_spec(func_arg="ges_rentenv_params")
    def höchstwert_grundr_zuschlag_m(
        grundrentenzeiten: IntSeries, ges_rentenv_params: dict
    )
    pass
```

The rounding specification will be in the dictionary `ges_rentenv_params`, which is an argument to the function. The fact that it comes from some parameters file is irrelevant at this point and downward misleading for user-provided functions / ex post changes to the parameters dictionary.

docs/geps/gep-05.rst
ChristianZimpelmann: I added a short paragraph about that.

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,198 @@
+.. _gep-5:
+
+=============================================
+GEP 5 — Optional Rounding of Variables
+=============================================
+
++------------+------------------------------------------------------------------+
+| Author     | `Christian Zimpelmann <https://github.com/ChristianZimpelmann>`_ |
++            +------------------------------------------------------------------+
+|            | `Janos Gabler <https://github.com/janosg>`_                      |
```
ChristianZimpelmann, line 10: @hmgaudecker, would you also like to be added?

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,182 @@
...
+``höchstwert_grundr_zuschlag_m`` is rounded to the next-lowest fourth decimal
+point instead of to the nearest. In that case, the rounding parameters net to changed as follows
+
+.. code-block:: python
+
+       policy_params["ges_rentenv"]["rounding"]["höchstwert_grundr_zuschlag_m"][
+           "direction"
+       ] = "down"
+
+If a user would like to add user-written functions which should be rounded, she needed
```
ChristianZimpelmann, line 149: We could outline this also in more detail, but I don't think it is necessary.

docs/geps/gep-05.rst
hmgaudecker: No, I just added annoying comments, few constructive stuff!

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,182 @@
...
+missing in ``policy_params``, an error is raised.
+
+Note that if the results have to be rounded in some years, but not in others (e.g.
+after a policy reform) the rounding parameters (both ``"base"`` and ``"direction"``)
+can be set to ``None``. This allows that the rounding parameters are found and no error
+is raised, but still no rounding is applied.
+
+In case rounding parameters are specified and the function does not have
+``__rounding_params_key__`` attribute of the functions is missing does not
+lead to an error during execution. This will never happen in the GETTSIM
```
hmgaudecker, line 130:
```suggestion
lead to an error. This will never happen in the GETTSIM
```

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,182 @@
...
+codebase, however, due to a suitable test.
+
+User-specified rounding
+~~~~~~~~~~~~~~~~~~~~~~~
+
+If a user wants to change rounding of a specified function, she needs to adjust the
```
hmgaudecker, line 136:
```suggestion
If a user wants to change rounding of a specified function, she will need to adjust the
```

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,182 @@
...
+rounding parameters in ``policy_params``.
+
+Suppose one would like to specify a reform in which
+``höchstwert_grundr_zuschlag_m`` is rounded to the next-lowest fourth decimal
+point instead of to the nearest. In that case, the rounding parameters net to changed as follows
```
hmgaudecker, line 141:
```suggestion
point instead of to the nearest. In that case, the rounding parameters will need to be changed as follows
```

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,182 @@
...
+
+.. code-block:: python
+
+       policy_params["ges_rentenv"]["rounding"]["höchstwert_grundr_zuschlag_m"][
+           "direction"
+       ] = "down"
+
+If a user would like to add user-written functions which should be rounded, she needed
```
hmgaudecker, line 149:
```suggestion
If a user would like to add user-written functions which should be rounded, she will need
```

docs/geps/gep-05.rst
```diff
@@ -0,0 +1,187 @@
...
+  rounding.
+
+In the same way as other policy parameters, the rounding parameters become part
+of the dictionary ``policy_params``.
+
+A function to be rounded must be decorated with ``add_rounding_spec``. This decorator
+indicates that the output should potentially be rounded. ``add_rounding_spec`` takes
+one required argument: ``params_key`` points to the key of the policy parameters
+dictionary containing the rounding parameters relating to the function that is
+decorated. In the above example, the rounding specification for
```
Eric-Sommer, line 87: Just for clarification: Is the 'key of the policy parameters dictionary' identical to the name of the `yaml` file?

hmgaudecker:
Yes, that is how it works.

Eric-Sommer ***@***.***> schrieb am Mi., 26. Jan. 2022, 20:02:

PR #340, opened 2022-01-25
Update docs on release process

hmgaudecker:
Trivial updates and clarifications for parts where I stumbled when setting up 0.4.2

### DIFF
--- a/docs/gettsim_developer/how-to-maintain.rst
+++ b/docs/gettsim_developer/how-to-maintain.rst
@@ -28,7 +28,7 @@ The branching model for GETTSIM is very simple.
    deleted.
 
 
-.. _releases:
+.. _release_major_minor:
 
 How To Release Major or Minor Versions
 --------------------------------------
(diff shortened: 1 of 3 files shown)


docs/gettsim_developer/how-to-maintain.rst
```diff
@@ -103,21 +110,31 @@ Procedure
...
+3. Squash merge the PR into main and note down the commit sha.
 4. Create a new PR against branch ``0.2``. Call the branch for the PR
    ``backport-pr[No.]-to-0.2.1`` where ``[No.]`` is the PR number.
 5. Use ``git cherrypick -x <commit-sha>`` with the aforementioned commit sha to apply
    the fix to the branch. Solve any merge conflicts, etc..
 6. Add the PR to the milestone for version ``0.2.1`` so that all changes for a new
    release can be collected.
-7. The release process for a patch version works as above in :ref:`releases` to release
-   ``0.2.1``; just that it is released off the maintenance branch, not off master.
+7. The release process for a patch version works as above in :ref:`releases`, steps
```
ChristianZimpelmann, line 126: Is there a reason why this is under "How to Release Patched Versions". I would have expected it directly under "How To Release Major or Minor Versions"

ChristianZimpelmann (approved):
Just one comment. Otherwise, very clear!

docs/gettsim_developer/how-to-maintain.rst
hmgaudecker: What do you mean by "this"? Everything *does* appear above, too -- I just tried to make the process for patches more self-contained relative to what it was. Both of us ended up confused at some point during 0.4.1 / 0.4.2 :smile:

ChristianZimpelmann: I mean point 7. (The release process for a patch version)

Patch versions are usually just "normal" releases and not necessarily backports to previously released versions (which to my understanding is the main topic of this latter section).

hmgaudecker: > Patch versions are usually just "normal" releases and not necessarily backports to previously released versions (which to my understanding is the main topic of this latter section).

There is the misunderstanding, I think. We should (continue to) think about them as patches = backports, not as adding new features. Thinking a bit ahead here when we reach more of a stationary state, I agree that, e.g., 0.4.1 was different.

Anyhow, we might switch to something like the year-indication anyhow, so leaving it as is for now and no need for big discussions ATM.

PR #342, opened 2022-02-02
Clarifications to GEP-01

hmgaudecker:

A couple of issues came up while working on #314, #275. This adds corresponding clarifications to GEP-01. No substantial changes.


docs/geps/gep-01.rst
```diff
@@ -108,6 +120,10 @@ changed, even if it leads to long variable names (e.g., ``kinderfreib``,
 ``_zu_verst_eink_kein_kinderfreib_tu``). This makes searching for identifiers
 easier and less error-prone.
 
+If names need to be concatenated for making clear what a column name refers to (e.g.,
```
ChristianZimpelmann, line 123: Repetition of third bullet above.

docs/geps/gep-01.rst
```diff
@@ -128,8 +144,10 @@ general naming considerations here.
 - There is a hierarchical structure to these parameters in that each of them is
   associated with a group (e.g., ``arbeitsl_geld``, ``kinderzuschlag``). These groups or
   abbreviations thereof do not re-appear in the name of the parameter.
-- Parameter names should be aligned with relevant column names. This makes searching for
-  identifiers easier and less error-prone.
+- Parameter names should be generally be aligned with relevant column names. However,
+  since the group is not repeated for the parameter, it is often better not to
+  abbreviate them (e.g., ``arbeitsl_geld_2["vermögensfreibetrag"]`` for the parameter
+  and ````arbeitsl_geld_2_vermög_freib_hh`` for a column derived from it).
```
ChristianZimpelmann, line 150: Not sure about that. Having the same term to search for seems to be relevant.

ChristianZimpelmann (changes requested):
Very important additions!

docs/geps/gep-01.rst
hmgaudecker: Yes, on purpose. Some people might read closer up there, some down here. I did not see a way to shortening it in the tl;dr bit nor a point in expanding it down here.

docs/geps/gep-01.rst
hmgaudecker: Yeah, I thought I'd put the idea out there. The sentence *"No substantial changes."* was not good.

There are pros and cons either way. Readability of the yaml files is also important (will lose some of that as we restrict ourselves to 32 characters in many places :wink:) and it is not bad if there is some distinction between column names/parameters if they are not exactly the same. Example above as a case in point.

Let's postpone until the next (=hopefully last) big round of renaming. I won't suggest we merge this before that happens.

docs/geps/gep-01.rst
ChristianZimpelmann: alright

docs/geps/gep-01.rst
ChristianZimpelmann: Alright, let's postpone the discussion. Since each parameter should have both a `name` and `description` key, I would hope that readability would still be okay. But I need to check how much abbreviations there would be

hmgaudecker: Yes, name might be enough, in case we display that e.g. in dashboard. Should review how well we are doing on that (consistency etc.) when going through the other stuff.

hmgaudecker: Upon inspection, I think would leave it in. Check the [new example in 4983585](https://github.com/iza-institute-of-labor-economics/gettsim/pull/342/commits/498358554024c004caaeea7425da1504ea31f3d8), I would find it weird to write ``vermög_grundfreib`` instead of ``vermögensgrundfreibetrag`` just to make sure we use the same abbreviations. 

I'd also think that it is reasonably clear it is just a possibility and we may just as well use the abbreviations we use for the functions.

ChristianZimpelmann: Fine with me!

hmgaudecker:
Clarifications re concatenated column names, dealing with conflicting objectives, names for columns vs parameters.

PR #343, opened 2022-02-03
Implement optional rounding

ChristianZimpelmann:

Implement the possibility for optional rounding as specified in [GEP-05](https://gettsim--324.org.readthedocs.build/en/324/geps/gep-05.html). 

And implement optional rounding for one examplary function (`eink_st_tu`)

Addresses #88.

#316 was started earlier and will implement rounding for all Grundrente and Grundsicherung im Alter variables after this PR is merged.

### Todo

- [x] Implement decorators functions
- [x] Implement way to set globally whether outputs should be rounded or not (where applicable)
- [x] Add rounding parameter and decorator for `eink_st_tu`
- [x] Write tests for rounding decorator

### Open ToDos after PR

- Reintroduce rounding where applicable

### DIFF
--- a/gettsim/interface.py
+++ b/gettsim/interface.py
@@ -2,6 +2,7 @@ import functools
 import textwrap
 import warnings
 
+import numpy as np
 import pandas as pd
 
 from gettsim.config import DEFAULT_TARGETS
@@ -25,6 +26,7 @@ def compute_taxes_and_transfers(
     targets=None,
     columns_overriding_functions=None,
     check_minimal_specification="ignore",
+    rounding=True,
     debug=False,
 ):
     """Compute taxes and transfers.
@@ -51,6 +53,8 @@ def compute_taxes_and_transfers(
     check_minimal_specification : {"ignore", "warn", "raise"}, default "ignore"
         Indicator for whether checks which ensure the most minimal configuration should
         be silenced, emitted as warnings or errors.
+    rounding : bool, default True
+        Indicator for whether rounding should be applied as specified in the law.
     debug : bool
         The debug mode does the following:
 
@@ -93,6 +97,11 @@ def compute_taxes_and_transfers(
         k: v for k, v in functions.items() if k not in columns_overriding_functions
     }
     _fail_if_targets_not_in_functions(functions, targets)
+
+    # Note: It is crucial that rounding is applied before partialling the
+    # functions. Otherwise, the attributes __roundingspec__ are lost.
+    if rounding:
+        functions = _add_rounding_to_functions(functions)
 
     # Partial parameters to functions such that they disappear in the DAG.
     functions = _partial_parameters_to_functions(functions, params)
@@ -517,6 +526,84 @@ def _root_nodes(dag):
         has_no_parents = len(list(dag.predecessors(node))) == 0
         if has_no_parents:
             yield node
+
+
+def _add_rounding_for_one_function(base, direction):
+    """Decorator to round the output of a function.
+
+    Parameters
+    ----------
+    base : float
+        Precision of rounding (e.g. 0.1 to round to the first decimal place)
+    round_d : bool
+        Whether rounding should be applied
+    direction : str
+        Whether the series should be rounded up, down or to the nearest number
+
+    Returns
+    -------
+    results : pandas.Series
+        Series with (potentially) rounded numbers
+    """
+
+    def inner(func):
+        # Make sure that signature is preserved
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            out = func(*args, **kwargs)
+            if direction == "up":
+                rounded_out = base * np.ceil(out / base)
+            elif direction == "down":
+                rounded_out = base * np.floor(out / base)
+            elif direction == "nearest":
+                rounded_out = base * (out / base).round()
+            else:
+                raise ValueError("direction must be one of 'up', 'down', or 'nearest'")
+            return rounded_out
+
+        return wrapper
+
+    return inner
+
+
+def _add_rounding_to_functions(functions):
+    """Add appropriate rounding of outputs to functions
+
+    Parameters
+    ----------
+    functions : dict of callable
+        Dictionary of functions which are either internal or user provided functions.
+
+
+    Returns
+    -------
+    functions_new : dict of callable
+        Dictionary mapping function names to callables for whose outputs appropriate
+        rounding is applied.
+
+    """
+    functions_new = {}
+    for f_name, func in functions.items():
+        if hasattr(func, "__roundingspec__"):
+            rounding_spec = func.__roundingspec__
+
+            # Check if expected parameters are present in roundingspec
+            if not ("base" in rounding_spec and "direction" in rounding_spec):
+                raise ValueError(
+                    "If roundingspec is set to a function, both"
+                    " 'base' and 'direction' are expected"
+                )
+
+            # Add rounding
+            base = rounding_spec["base"]
+            direction = rounding_spec["direction"]
+            functions_new[f_name] = _add_rounding_for_one_function(base, direction)(
+                func
+            )
+
+        else:
+            functions_new[f_name] = func
+    return functions_new
 
 
 def _partial_parameters_to_functions(functions, params):
--- a/gettsim/interface.py
+++ b/gettsim/interface.py
@@ -213,12 +213,12 @@ def prepare_functions_and_set_up_dag(
     _fail_if_targets_not_in_functions(all_functions, targets)
 
     # Partial parameters to functions such that they disappear in the DAG.
-    partialled_functions = _partial_parameters_to_functions(all_functions, params)
+    partialed_functions = _partial_parameters_to_functions(all_functions, params)
 
     # Create DAG and perform checks which depend on data which is not part of the DAG
     # interface.
     dag = create_dag(
-        functions=partialled_functions,
+        functions=partialed_functions,
         targets=targets,
         columns_overriding_functions=columns_overriding_functions,
         check_minimal_specification=check_minimal_specification,
--- a/gettsim/interface.py
+++ b/gettsim/interface.py
@@ -1,3 +1,4 @@
+import copy
 import functools
 import textwrap
 import warnings
@@ -17,6 +18,13 @@ from gettsim.shared import format_list_linewise
 from gettsim.shared import get_names_of_arguments_without_defaults
 from gettsim.shared import parse_to_list_of_strings
 from gettsim.typing import check_if_series_has_internal_type
+
+
+class KeyErrorMessage(str):
+    """Subclass str to allow for line breaks in KeyError messages"""
+
+    def __repr__(self):
+        return str(self)
 
 
 def compute_taxes_and_transfers(
@@ -70,7 +78,7 @@ def compute_taxes_and_transfers(
 
     """
 
-    # Set defaults for some parameters
+    # Set defaults for some parameters.
     targets = DEFAULT_TARGETS if targets is None else targets
     targets = parse_to_list_of_strings(targets, "targets")
     columns_overriding_functions = parse_to_list_of_strings(
@@ -78,7 +86,7 @@ def compute_taxes_and_transfers(
     )
     params = {} if params is None else params
 
-    # Create the dag
+    # Create the dag and perform.
     dag = set_up_dag(
         data,
         params,
@@ -90,7 +98,7 @@ def compute_taxes_and_transfers(
     )
 
     # We delay the data preparation as long as possible such that other checks can fail
-    # before this
+    # before this.
     data = data.copy(deep=True)
     data = _process_data(data)
     data = _reduce_data(data)
@@ -118,7 +126,8 @@ def set_up_dag(
     check_minimal_specification,
     rounding,
 ):
-    """Load all potential functions of the DAG and create the DAG.
+    """Load all potential functions of the DAG, perform checks on data and functions, and
+    create the DAG.
 
     Parameters
     ----------
@@ -160,7 +169,7 @@ def set_up_dag(
     for funcs, name in zip([internal_functions, functions], ["internal", "user"]):
         _fail_if_functions_and_columns_overlap(columns, funcs, name)
 
-    # Create one dictionary of functions and perform check
+    # Create one dictionary of functions and perform check.
     functions = {**internal_functions, **functions}
     _fail_if_datatype_is_false(data, columns_overriding_functions, functions)
     _fail_if_columns_overriding_functions_are_not_in_functions(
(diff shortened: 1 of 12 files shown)


gettsim/tests/test_rounding.py
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    calc_result = compute_taxes_and_transfers(
+        data=data, functions=[test_func], params=rounding_specs, targets=["test_func"],
+    )
+    np.array_equal(calc_result["test_func"].values, np.array(exp_output))
+
+
+@pytest.mark.parametrize(
+    "base, direction, input_values, exp_output",
```
hmgaudecker, line 110:
```suggestion
    "base, direction, input_values_exp_output, _ignore",
```

Then re-use the parametrization from `test_rounding`.

gettsim/tests/test_rounding.py
```diff
@@ -0,0 +1,187 @@
...
+    [
+        (5, "up", [100.24, 100.78], [100.24, 100.78]),
+        (0.1, "down", [100.24, 100.78], [100.24, 100.78]),
+        (0.001, "nearest", [100.24, 100.78], [100.24, 100.78]),
+    ],
+)
+def test_no_rounding(base, direction, input_values, exp_output):
+    """Check if no rounding if disabled"""
```
hmgaudecker, line 118:
```suggestion

```
(obvious)

gettsim/tests/test_rounding.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Define function that should be rounded
+    @add_rounding_spec(params_key="params_key_test")
+    def test_func(income):
+        return income
+
+    data = pd.DataFrame([{"p_id": 1}, {"p_id": 2}])
+    data["income"] = input_values
```
hmgaudecker, line 126:
```suggestion
    data["income"] = input_values_exp_output
```

gettsim/tests/test_rounding.py
```diff
@@ -0,0 +1,187 @@
...
+    rounding_specs = {
+        "params_key_test": {
+            "rounding": {"test_func": {"base": base, "direction": direction}}
+        }
+    }
+
+    calc_result = compute_taxes_and_transfers(
+        data=data, functions=[test_func], params=rounding_specs, targets=["test_func"],
+    )
+    np.array_equal(calc_result["test_func"].values, np.array(exp_output))
```
hmgaudecker, line 136:
```suggestion
    np.array_equal(calc_result["test_func"].values, np.array(input_values_exp_output))
```

hmgaudecker (commented):
Looks great, thanks! Just a number of small things!

gettsim/tests/test_rounding.py
ChristianZimpelmann: What would be the function of `_ignore`? If I understand your comment correctly, I could re-use the parametrization without an additional keyword.

ChristianZimpelmann: Ah, got it now :)

gettsim/interface.py
```diff
@@ -545,11 +717,15 @@ def _partial_parameters_to_functions(functions, params):
...
         }
+        if partial_params:
+            partial_func = functools.partial(function, **partial_params)
 
-        partialed_functions[name] = (
-            functools.partial(function, **partial_params)
-            if partial_params
-            else function
-        )
+            # Make sure that rounding parameter attribute is kept
```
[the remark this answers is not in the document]
ChristianZimpelmann: Changed the comment to:

```
            # Make sure that rounding parameter attribute is transferred to partial function.
            # Otherwise, this information would get lost.
            
            ```
            
            Clearer?

File: gettsim/interface.py [hunk 6]
```diff
@@ -519,6 +586,111 @@ def _root_nodes(dag):
...
+                    raise KeyError(
+                        "Both 'base' and 'direction' are expected as rounding "
+                        "parameters in the parameter dictionary. At least one of them "
+                        f"is missing at ['{params_key}']['rounding']['{task}']."
+                    )
+
+                # Add rounding
+                base = rounding_spec["base"]
+                direction = rounding_spec["direction"]
+                dag.nodes[task]["function"] = _add_rounding_to_one_function(
```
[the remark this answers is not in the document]
ChristianZimpelmann: It fails within `_add_rounding_to_one_function`. I would prefer that as is it is easier to adjust when we add more potential values for `direction`.

ChristianZimpelmann:
Great comments! Thanks!

I addressed almost all of the comments, but left a few open to give more information or to disagree with the proposed changes (in two cases). Can change it if I couldn't convince you. :)

gettsim/interface.py
```diff
@@ -65,13 +69,87 @@ def compute_taxes_and_transfers(
...
+
+    if not debug:
+        results = results[targets]
+
+    results = _reorder_columns(results)
+
+    return results
+
+
+def set_up_dag(
```
hmgaudecker, line 120: Only realised this just now: It is pretty confusing to have one function `set_up_dag`, which calls `create_dag`. Can we do better? No time for brainstorming right now, sorry.

gettsim/interface.py
```diff
@@ -65,13 +69,87 @@ def compute_taxes_and_transfers(
...
+    data,
```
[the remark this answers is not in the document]
hmgaudecker: But almost trivially, no? Go through all nodes of the DAG, check if it is a function or a data column, do the check if the latter. 

Given that the function is new, I would go that far in case I am not missing anything.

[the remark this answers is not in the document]
ChristianZimpelmann: Usual data columns are trivial. The problem is data columns that ovewrite functions. For those, the expected data type, depends on the return type of the function, but the function is not part of the DAG and doesn't exist outside of the function as it is implemented right now.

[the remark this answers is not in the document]
hmgaudecker: Okay, just so that I understand: we would need to update the dict of expected data types defined for the input columns (can't look up right now) while looking at columns overriding functions?

What about passing (example, untested)

`{col: dtype for col, dtype in data.dtypes}`

for the moment?

I'd just like to signal that it has nothing to do with the data, really.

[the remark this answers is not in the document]
ChristianZimpelmann: I now restructured `compute_taxes_and_transfers` which now calls (among other) two functions:

- ``check_data_check_functions_and_merge_functions``
- ``prepare_functions_and_set_up_dag``

``prepare_functions_and_set_up_dag`` no longer requires ``data`` as argument. ``check_data_check_functions_and_merge_functions`` still does so (among others because `check_if_series_has_internal_type` requires the pd.series and not just the dtypes). 

But this structure should make clear that creating the DAG has nothing to do with data as was your concern if I understand correctly.

gettsim/interface.py
```diff
@@ -65,55 +77,182 @@ def compute_taxes_and_transfers(
...
+        columns_overriding_functions=columns_overriding_functions,
+        check_minimal_specification=check_minimal_specification,
+        rounding=rounding,
+    )
+
+    # Do some checks.
+    _fail_if_root_nodes_are_missing(dag, data)
+    _fail_if_more_than_necessary_data_is_passed(dag, data, check_minimal_specification)
+
+    # We delay the data preparation as long as possible such that other checks can fail
```
hmgaudecker, line 111: Comment does not seem quite up-to-date? We may want to do the `_process_data` step earlier with other checks? Looks a bit like it should be the first one, really.

E.g., `data_cols = list(data.columns)` in `check_data_check_functions_and_merge_functions` now implicitly assumes that the input data is a DataFrame.

gettsim/interface.py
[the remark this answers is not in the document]
hmgaudecker: Cool. I just removed creating variables just to pass them to the function, then it makes sense to my weird mind, too :wink:

gettsim/parameters/eink_st.yaml
```diff
@@ -239,3 +239,12 @@ ertragsanteil:
       upper_threshold: inf
       rate_linear: 0.0
       intercept_at_lower_threshold: 1
+
+rounding:
```
hmgaudecker, line 243: Oh, so you did add some rounding? Sorry I missed that in the previous round. Be precise then in changes.rst

gettsim/policy_environment.py
```diff
@@ -358,9 +361,52 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+
+    Returns:
+        dictionary: Rounding parameters for the specified date
+    """
+    out = {}
+    rounding_parameters = ["direction", "base"]
+
+    # Load values of all parameters at the specified date
+    for function_name in rounding_spec.keys():
+        dates = sorted(
```
hmgaudecker, line 391:
```suggestion
        policy_dates = sorted(
```

gettsim/policy_environment.py
```diff
@@ -358,9 +361,52 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+            key
+            for key in rounding_spec[function_name].keys()
+            if isinstance(key, datetime.date)
+        )
+        past_policies = [x for x in dates if x <= date]
```
hmgaudecker, line 396: I find the logic (too) hard to follow here. Why not get `policy_in_place` in one step? Also, the name `past_policies` suggests that none of them is relevant anymore, maybe that's what is confusing me.

gettsim/shared.py
```diff
@@ -57,3 +57,27 @@ def get_names_of_arguments_without_defaults(function):
             argument_names_without_defaults.append(parameter)
 
     return argument_names_without_defaults
+
+
+def add_rounding_spec(params_key):
+    """Decorator to add the parameter key where rounding specifications are
```
hmgaudecker, line 63:
```suggestion
    """Decorator adding the location of the rounding specification to a function.
```

gettsim/shared.py
```diff
@@ -57,3 +57,27 @@ def get_names_of_arguments_without_defaults(function):
...
+    found to a function.
```
hmgaudecker, line 64:
```suggestion
```

gettsim/tests/test_interface.py
```diff
@@ -155,3 +157,38 @@ def test_consecutive_internal_test_runs():
 
     with pytest.warns(UserWarning, match="Repeated execution of the test suite"):
         test("--collect-only")
+
+
+def test_partial_parameters_to_functions():
```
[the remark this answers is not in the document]
hmgaudecker: I guess you could just define it outside of the tests to avoid writing it twice? Of course you could not call it `test_func` then :smile:

gettsim/policy_environment.py
hmgaudecker, line 396:
```suggestion
        past_policies = [d for d in policy_dates if d <= date]
```

hmgaudecker (approved):
Left open a couple of small things, but basically done! Thanks!!!

gettsim/interface.py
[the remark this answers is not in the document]
hmgaudecker: Cool. Just a small follow-up re the expected data left now.

gettsim/tests/test_interface.py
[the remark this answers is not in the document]
ChristianZimpelmann: good point!

gettsim/policy_environment.py
ChristianZimpelmann: Good point about the naming! I improved naming of variables and Documentation.

> Why not get policy_in_place in one step?

To handle cases in which `policy_dates_before_date` is empty. Then taking the maximum leads to an error. (Could be also handled with try except, but I like the current version a bit more.

Alternative would be to fail already here if this happens, but I wouldn't since the user can still change `policy_params` themselves.

gettsim/policy_environment.py
```diff
@@ -358,9 +361,55 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+        # Find all specified policy dates before date.
+        policy_dates_before_date = sorted(
+            key
+            for key in rounding_spec[function_name].keys()
+            if isinstance(key, datetime.date) and key <= date
+        )
+
+        # If any rounding specs are defined for a date before the specified
+        # date, copy them to params dictionary.
+        # If not, no key for this function name is created (this will
```
hmgaudecker, line 401: I don't understand the second part ("If not, ...")

gettsim/policy_environment.py
```diff
@@ -358,9 +361,55 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+        for several functions and several dates
+
+    Returns:
+        dictionary: Rounding parameters for the specified date
+    """
+    out = {}
+    rounding_parameters = ["direction", "base"]
+
+    # Load values of all parameters at the specified date.
+    for function_name in rounding_spec.keys():
```
hmgaudecker, line 390: Iterate over `.items()` since we are accessing `rounding_spec[function_name]` a couple of times. 

Maybe:

```suggestion
    for function_name, this_rounding_spec in rounding_spec.items():
```
?

gettsim/policy_environment.py
```diff
@@ -358,9 +361,55 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+    Parameters
+    ----------
+    date : datetime.date
+        The date for which the policy system is set up.
+    rounding_spec : dictionary
+        The dictionary that contains rounding parameters
+        for several functions and several dates
+
+    Returns:
+        dictionary: Rounding parameters for the specified date
```
hmgaudecker, line 384:
```suggestion
        dictionary: 
          - Keys: Functions to be rounded.
          - Values: Rounding parameters for the specified date
```

gettsim/policy_environment.py
```diff
@@ -358,9 +361,55 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
 
+def _load_rounding_parameters(date, rounding_spec):
+    """Load rounding parameters for a specific date from a dictionary.
+
+    Parameters
+    ----------
+    date : datetime.date
+        The date for which the policy system is set up.
+    rounding_spec : dictionary
+        The dictionary that contains rounding parameters
```
hmgaudecker, line 380:
```suggestion
          - Keys: Functions to be rounded.
          - Values: Rounding parameters for all dates
```

gettsim/policy_environment.py
```diff
@@ -358,9 +361,55 @@ def _load_parameter_group_from_yaml(date, group, parameters=None):
...
+        for several functions and several dates
```
hmgaudecker, line 381:
```suggestion
```

hmgaudecker (approved):
Nice, just documentation left, essentially.

gettsim/policy_environment.py
ChristianZimpelmann:
```

        # If no appropriate rounding specs are found for the requested date, 
        # no key for this function name is created (this will raise an error 
        # later if the user doesn't adjust params df before calling 
        # `compute_taxes_and_transfers`).

Comment (hmgaudecker) [reply gettsim/policy_environment.py:401]:
```
        # If no appropriate rounding specs are found for the requested date, 
        # the function will not appear in the returned dictionary. 
        # Note this will raise an error later unless the user adds an appropriate
        # rounding specification to the parameters... 
```

*(not sure about the wording in the very last bit, but we should be precise there. I don't think there is a "params df", for example?)*

ChristianZimpelmann: Sounds good: `... to the parameters dictionary.`

PR #344, opened 2022-02-03
Correct Kinderfreibetrag and Kindergeld for Single Parents

### LINKED ISSUE #158: Correct child benefit for single parents

Currently, single parents receive the full child benefit (Kindergeld) but only half the child tax allowance (Kinderfreibetrag). This is wrong at least for those receiving alimony payments for their child(ren).

### Proposed implementation

Those single parents should instead receive only half the child benefit amount. If they receive child alimony *advance* (Unterhaltsvorschuss), however, they are paid the full amount.

**EDIT:** We probably have to fix two things:

- [ ] For the *Günstigerprüfung* (Kindergeld versus Kinderfreibetrag), only half the child benefit is considered, because the other half is (virtually) paid to the other parent.
- [ ] Alimony payments would increase if child allowance hits. We therefore might need to increase alimonies to avoid jumps in disposable income.

Eric-Sommer:

closes #158

### Todo

- [x] Single Parents receive only half Kinderfreibetrag
- [x] make sure they receive Kindergeld even after the Kinderfreibetrag threshold
- [x] treat Unterhaltsvorschuss appropriately
- [ ] extend relevant tests with necessary variables.
- [ ] create test cases for cases mentioned in #158 


hmgaudecker:
Superseded by #463 - @Eric-Sommer, please let me know if that was a misjudgement!

Issue #205, opened 2020-06-18
Take a fresh look at Vorsorgeaufwendungen


hmgaudecker:
### Current and desired situation

While working on #197 I noted a couple of things, mostly on the structure of this whole thing

- The differentiation between different years (-2004, 2005-09, 2010-19, 2020+) mostly (only?) concerns old-age provisions, which are calculated separately. 
- In particular, we are conflating policy changes (2005, 2010) with favourability checks. Not pretty right now, needs to be straightened.
- Ideally, we would have a function `vorsorge` that is public and hide most of the differences from the user. Certainly the four regimes seem a bit too much.
- Private health insurance contributions are not taken into account in `vorsorge_ab_2020`.

hmgaudecker:
Done

PR #349, opened 2022-02-25
Hardcoded numbers

LauraGergeleit:
What problem do you want to solve?

Replacing hard-coded numbers in the code by including parameters in the yaml-Files. Closes #338. Closes #345.

PR #351, opened 2022-02-25
Trivial changes to old age provisions

hmgaudecker:

Make working on a B.Sc. thesis related to #205 easier.

Two things:
- rename `altervorsorge_aufwend` -> `vorsorge_alter_aufwend` (note missing "s" in original name, I think the current version is most in line with #342, but we might decide for `altersvorsorge` eventually, can be done during large PR)
- Move to correct module (`vorsorge.py`)


ChristianZimpelmann (approved):
Looks great!

I was missing the change to `freibeträge.py`, but there it is.

Can be merged as soon as tests run through.

hmgaudecker:
These lines in vorsorge.vorsorge_alter_aufwend:

```py
    einführungsfaktor = 0.6 + 0.02 * (
        min(eink_st_abzuege_params["datum"].year, 2025) - 2005
    )
```

just caught my attention in conjunction with #349.

I think we should re-factor so that `eink_st_abzuege_params['einführungsfaktor_vorsorge_alter_aufwend']` is calculated outside the graph (same for everybody, should be part of params dictionary) in a similar way to `kinderzuschl_max` since 2021. 

@ChristianZimpelmann, could you give that a shot?

Should we also rename to `eink_st_abzüge` here or wait with that one?

ChristianZimpelmann:
I will have a look at it next week.. will be part of #349 then probably. 

> Should we also rename to eink_st_abzüge here or wait with that one?

Let's wait. I put it on the list.

PR #352, opened 2022-03-02
Add lower boundary to ges_rente_zugangsfaktor

paulinaschroeder:

Add lower boundary to ges_rente_zugangsfaktor so it won't be negative for pensioners without jahr_renteneintr.


hmgaudecker:
Thanks!

> pensioners without jahr_renteneintr

I am not sure we want to support any missing data? My understanding was that this would likely be slowing things down extremely. Is that correct, @janosg ?

ChristianZimpelmann:
The question if and potentially how we support missings in columns is definitely relevant (`jahr_renteneintr` just doesn't exist for subjects that aren't retired yet). In my opinion, however, the PR is not directly affected by this question as also other (erroneous) values in `jahr_renteneintr` could lead to negative values which we should take care of. 

Can we merge it @hmgaudecker?

hmgaudecker:
Cool, I think I was just confused by the initial description then!

hmgaudecker:
This said, it might be good to add a couple of test cases in case that does not take more than 15min. E.g., people who are not pensioners should be no problem as per

```py
    out.loc[~rentner] = 0
```

I.e., would be great if the test specifies by what "erroneous values" entail.

ChristianZimpelmann:
If I remember correctly, it happend when some subjects had `jahr_renteneintr == 0`?

> This said, it might be good to add a couple of test cases in case that does not take more than 15min.

What kind of test cases? Those including `erroneous values` like `jahr_renteneintr < geburtsjahr`? I don't think we should include tests like that. At some later point, we should instead implement checks on input variables to find out if they are reasonable.

ChristianZimpelmann (approved):

hmgaudecker:
Well, as always, tests would be there to define the behaviour we expect. If we want to support crazy values for `jahr_renteneintritt` (crazy in one direction), fine. The tests don't have to stay forever and I agree we should have checks eventually.

There would have been less misunderstandings here if a test case had been around and the initial description had been more precise (which is trivial if you are just describing a test).

PR #361, opened 2022-03-29
Set up development environment as part of installation.

hmgaudecker:

Directly run `pip install -e .` as part of `conda env create`

### Todo

- [x] Update docs.
- [x] Make sure this does not conflict with CI setup anywhere.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,8 +6,6 @@ dependencies:
   - pip
   - anaconda-client
   - bokeh
-  - conda-build
-  - conda-verify
   - jupyterlab
   - nbsphinx
   - networkx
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -28,5 +26,7 @@ dependencies:
   - sphinx-automodapi
   - sphinx-copybutton
   - sphinx_rtd_theme
+  - sphinx_rtd_theme
```
ChristianZimpelmann, line 29:
```suggestion
```

Or is there a reason that we have this twice?

ChristianZimpelmann (approved):
Thanks!

PR #364, opened 2022-03-30
Speed up tests

### LINKED ISSUE #363: Ignore dashboard directory in tests
### Bug description

When running pytest from the command line, it will import all Python scripts. Triggers setting up the dashboard data, which takes 80s on my fairly fast machine.

There should be a possibility to just ignore that in a quick 'n' dirty way -- will move to nice setup of https://github.com/pytask-dev/cookiecutter-pytask-project eventually.

hmgaudecker:
Closes #363 

pytest would import the dashboard code, which sets up the data in cwd. However, it is present in the dashboard directory already. Just doing a try/except to pick the right one.

### DIFF
--- a/dashboard/app/main.py
+++ b/dashboard/app/main.py
@@ -121,8 +121,11 @@ plot_dict = {
     p: {a: attribute_dict[p][counter] for counter, a in enumerate(plot_attributes)}
     for p in plot_list
 }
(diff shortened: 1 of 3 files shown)


ChristianZimpelmann (approved):
Great change!

PR #369, opened 2022-04-06
Some visual adjustments for paramter files

ChristianZimpelmann:

Some visual adjustments for paramter files


hmgaudecker:
Thanks! We should get some pre-commit hook to style the yaml files, except for the `rate` thing this should be unnecessary eventually.

ChristianZimpelmann:
> We should get some pre-commit hook to style the yaml files

Good point! I will keep it in mind.

I take your point as acceptance of this PR as changes are very minor (haven't made any further changes)

PR #374, opened 2022-04-08
Move calculation of `midi_job_faktor_f` to separate function.

ChristianZimpelmann:

Tidy up some parameter files. Move `midi_job_faktor_f` to separate function. The latter will be relevant when the proposed changes to Midijobs will be realized.


MaxBlesch (approved):
Very nice work!

PR #387, opened 2022-05-09
Lohnsteuer 2

JakobWegmann:

Today, @LauraGergeleit and I created the branch lohnsteuer_2 to adapt the Lohnsteuer calculation to the new naming, coding and commenting conventions on main. This is now partially in conflict with branch lohnsteuer where in parallel @Eric-Sommer also adapted to the new conventions.

### Todo

- [ ] Merging lohnsteuer_2 in lohnsteuer allows to compare the changes. Optimally, they would be identical
- [ ] Ensure, that no code exists as duplicate. I think, code duplicates can only pass tests in policy_environment.py and the parameter files

### DIFF
--- a/gettsim/taxes/lohn_st.py
+++ b/gettsim/taxes/lohn_st.py
@@ -27,17 +27,19 @@ def lohn_st_eink(
     """
     # WHY IS THIS 1908??
     entlastung_freibetrag_alleinerz = (steuerklasse == 2) * eink_st_abzuege_params[
-        "alleinerziehenden_freibetrag"
+        "alleinerz_freibetrag"
     ]
 
-    werbungskosten = [
-        eink_st_abzuege_params["werbungskostenpauschale"] if stkl != 6 else 0
-        for stkl in steuerklasse
-    ]
-    sonderausgaben = [
-        eink_st_abzuege_params["sonderausgabenpauschbetrag"] if stkl != 6 else 0
-        for stkl in steuerklasse
-    ]
+    if steuerklasse == 6:
+        werbungskosten = 0
+    else:
+        werbungskosten = eink_st_abzuege_params["werbungskostenpauschale"]
+
+    if steuerklasse == 6:
+        sonderausgaben = 0
+    else:
+        sonderausgaben = eink_st_abzuege_params["sonderausgabenpauschbetrag"]
+
     # zu versteuerndes Einkommen / tax base for Lohnsteuer
     out = max(
         12 * bruttolohn_m
@@ -90,11 +92,12 @@ def lohn_st(lohn_st_eink: float, eink_st_params: dict, steuerklasse: int) -> flo
         lohn_st_eink * eink_st_params["eink_st_tarif"]["rates"][0][1],
     )
 
-    out = (
-        (lohnsteuer_splittingtarif * (steuerklasse == 3))
-        + (lohnsteuer_basistarif * (steuerklasse.isin([1, 2, 4])))
-        + (lohnsteuer_klasse5_6 * (steuerklasse.isin([5, 6])))
-    )
+    if (steuerklasse == 1) | (steuerklasse == 2) | (steuerklasse == 4):
+        out = lohnsteuer_basistarif
+    elif steuerklasse == 3:
+        out = lohnsteuer_splittingtarif
+    else:
+        out = lohnsteuer_klasse5_6
 
     return out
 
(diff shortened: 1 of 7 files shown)


gettsim/tests/test_lohnsteuer2.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.mark.parametrize("year, column", itertools.product(YEARS, OUT_COLS))
+def test_lohnsteuer_2(input_data, year, column):
+    year_data = input_data[input_data["jahr"] == year]
+    df = year_data[INPUT_COLS].copy()
+    policy_params, policy_functions = set_up_policy_environment(date=year)
+
+    result = compute_taxes_and_transfers(
+        data=df,
+        params=policy_params,
```
Eric-Sommer, line 48: this needs an addtional option `columns_overriding_functions = ['steuerklasse']` because we provide it in the (test) data. Apart from that, could you integrate that test case to the existing `test_lohnsteuer`?

gettsim/tests/test_data/lohn_st_2.csv
```diff
@@ -0,0 +1,2 @@
+hh_id,tu_id,p_id,jahr,steuerklasse,anz_kindergeld_kinder_tu,bruttolohn_m,wohnort_ost,ges_krankenv_zusatz,alter,hat_kinder,lohn_st,lohn_st_soli
```
Eric-Sommer, line 1: You specify `ges_krankenv_zusatz` in the data. We don't have a solution for this at the moment. The parameter corresponds to the average additional contribution rate and this is what the test should assume without providing information on that in the data.

Eric-Sommer (commented):
Looks good to me, thanks @JakobWegmann and @LauraGergeleit, especially for getting rid of numpy statements. I refactored a tiny bit.

gettsim/taxes/lohn_st.py
```diff
@@ -177,7 +178,7 @@ def vorsorgepauschale_2005_2010() -> float:
     return out
 
 
-def vorsorg_rv_anteil(eink_st_abzuege_params: dict):
+def vorsorg_rv_anteil(eink_st_abzuege_params: dict, jahr: int):
```
Eric-Sommer, line 181: Why did you add `jahr` as argument here and where is it handed to the function?

gettsim/tests/test_lohnsteuer2.py
hmgaudecker: Just curious: how could `steuerklasse` possibly be a function?

gettsim/taxes/lohn_st.py
JakobWegmann: I think this is just a mistake of us. We thought that since `piecewise_polynomial` requires `jahr `, it should also be an argument of `vorsorg_rv_anteil`.

gettsim/tests/test_data/lohn_st_2.csv
JakobWegmann: Is it technically possible to merge lohnsteuer_2 into lohnsteuer while ignoring that `lohn_st_2.csv` and `test_lohnsteuer_2.py` do not pass any test? We created them two weeks ago as dummy files, and they are not even passing the tests on branch lohnsteuer_2 anymore. Otherwise they can just be deleted.

hmgaudecker: No worries about failing tests before merging into `main`!

gettsim/taxes/lohn_st.py
JakobWegmann: grafik

What is the right way to call the year as an input for the piecewise_polynomial? 

Before it was `x=pd.Series(eink_st_abzuege_params["datum"].year)`. To apply to the new conventions we replaced it by `x=jahr` which obviously is not working.

ChristianZimpelmann: The parameter `vorsorge_pauschale_rv_anteil` should be handled in `policy_environment.py` similar to e.g. `_parse_einführungsfaktor_vorsorgeaufw_alter_ab_2005`. That way, `eink_st_abzuege_params` would contain `vorsorge_pauschale_rv_anteil` for the correct year as an entry.

JakobWegmann: Should I also change the function name to `_parse_vorsorge_pauschale_rv_anteil`?

JakobWegmann: Could you please [check](https://github.com/iza-institute-of-labor-economics/gettsim/pull/387/commits/1f8e9c6178d1b711a76fa3df67111566f2271709) whether these are the changes you had in mind? I tried to stick to `_parse_einführungsfaktor_vorsorgeaufw_alter_ab_2005` as close as possible.

ChristianZimpelmann: The Parameters section in the docstring is not up-to-date. Otherwise, it looks great!

PR #388, opened 2022-05-10
Supported groupings

LauraGergeleit:
### What problem do you want to solve?

Renaming households (_hh) to bedarfsgemeinschaft (_bg). Defining supported groupings in config.py and referring to them in interface.py instead of using the hardcoded groups there.
Addresses issue #379.

PR #392, opened 2022-05-11
Fix rtol in tests and associated bugs

### LINKED ISSUE #217: Revisit test cases of vorsorgeaufwendungen and zu_verst_eink
### Bug description

In #216 we encountered some deviation of the test results in `test_vorsorge.py` and `test_zu_verst_eink.py`. We need to check if these are driven by a bug in the calculation of `vorsorge` or if there are multiple bugs.

### LINKED ISSUE #391: Some tests are not strict enough
This is related to #389. As of now, four of our tests using `pandas.testing.assert_series_equal` use extremely relaxed tolerances, in particular `rtol=1`. These tests are: `test_eink_st`, `test_elterngeld`, `test_vorsorgeaufw`, `test_zu_verst_eink`.

What do I mean by "extremely relaxed"? Regarding the tolerance parameters `rtol` and `atol`, the [pandas documentation](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_series_equal.html) for `assert_series_equal` refers to the documentation of `math.is_equal` from the standard library [here](https://docs.python.org/3/library/math.html#math.isclose). According to the latter documentation, two numbers `a` and `b` are considered equal if

```python
abs(a-b) <= max(rtol * max(abs(a), abs(b)), atol)
```

is `True`. This implies that the test passes for `rtol=1` and arbitrarily small settings of `atol` as long as `a` and `b` have the same sign (if `a==0` then the comparison is true for all `b` and vice versa). 

Interpreting the condition above as a function of `a`, `b` for fixed parameters `rtol` and `atol`, the plotted "truth function" (with `True`interpreted as 1 and `False` as 0) looks like this (`rtol=1`, `atol=1e-8`):

graph

This is not how we want this function to look. Instead, we want the "truth function" to have this (stylized) shape (with `rtol=1e-1, atol=1e-1`, to make it more obvious what the parameters do): A narrow "ridge of truth", which gets a bit wider for larger (in absolute terms) values of `a` and `b` 

graph2

With `rtol=1`, e.g., 1000 and 0 are considered to be equal, which explains why the tests passed in #389 even though the calculated "ALG I" was completely wrong.

Try this:

```Python
import pandas as pd
from pandas.testing import assert_series_equal

a = pd.Series([0, 1000])
b = pd.Series([1000, 0])
c = pd.Series([0, -1000])

assert_series_equal(a, b, rtol=1, atol=1e-8)
assert_series_equal(b, c, rtol=1, atol=1e-8)

d = pd.Series([1e-3, 1000.0])
e = pd.Series([-1e-3, 1000.0])

assert_series_equal(d, e, rtol=1, atol=1e-8)
```

The first two tests pass, the third fails. 

Most (all?) outcome variables are non-negative by construction. Thus, using `rtol=1` in `test_eink_st`, `test_elterngeld`, `test_vorsorgeaufw`, and `test_zu_verst_eink` implies that these tests _cannot_ fail, which is probably not what we want. 😉  

The hard part is to find a more appropriate setting for `rtol`, which is likely test-specific. I tried to replace `rtol=1` with `rtol=1e-3`, which leads to 19 failing tests. Even a quite relaxed `rtol=1e-1` leads to a couple of failing tests. Anyway, simply changing `rtol` to a "reasonable value" is only part of the solution. We will also have to check why these tests fail for more appropriate settings of `rtol`.

ChristianZimpelmann:

- Remove `rtol=1` in test cases as this is way to lenient.
- Fix some bugs in the code that haven't been detected so far.
- add `rtol=0` to all `assert_series_equal` calls in the tests to avoid that the default of `1e-5` is used.
- Closes #391 
- Closes #217

### Notes
- Fixed bug in `elterngeld.py` that was introduced during major restructuring.
- `test_eink_st` is still off in the decimal places. We should be able to hit this more precisely using the `rounding` functuality.

### Todo

- [x] `test_elterngeld`
- [x] `test_eink_st`
- [x] `test_vorsorgeaufw`
- [x] `test_zu_verst_eink`
- [x] add `rtol=0` to all `assert_series_equal` calls 


ChristianZimpelmann:
Quick update:

- three testcases in `vorsorgeaufw.csv` are off by ~5€ and we couldn't find out why. Problem is that the source of the test cases is behind a paywall. @LauraGergeleit now added a few new test cases which run through without a problem. I adjusted the expected results for the three cases in vorsorgeaufw.csv for now and included a note such that we don't forget. In the medium term we should replace these tests completely.
- in `zu_verst_eink.csv` even more test cases fail. Here, no source is mentioned at all (@Eric-Sommer, @MaxBlesch, do you have any idea?). As we had set `rtol=1` for quite some time, it is not easy to find out if the test cases make sense at all. I think we should replace them as soon as possible.

Eric-Sommer:
>     * in `zu_verst_eink.csv` even more test cases fail. Here, no source is mentioned at all (@Eric-Sommer, @MaxBlesch, do you have any idea?). As we had set `rtol=1` for quite some time, it is not easy to find out if the test cases make sense at all. I think we should replace them, too.

Setting up proper tests is certainly easier than trying to figure out what I had in mind three years ago :smile:

gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -258,15 +264,23 @@ def eink_st_sonderausgaben_ab_2012(
...
         eink_st_abzuege_params["kinderbetreuungskosten_abz_maximum"],
     )
+    sonderausgaben_betreuung = (
+        anz_kinder_tu
+        * abziehbare_betreuungskosten
+        * eink_st_abzuege_params["kinderbetreuungskosten_abz_anteil"]
+    ) / anz_erwachsene_tu
+
+    if gemeinsam_veranlagt_tu:
+        pauschale = eink_st_abzuege_params["sonderausgabenpauschbetrag"]["couple"]
```
ChristianZimpelmann, line 274: We should compare it only to ["sonderausgabenpauschbetrag"]["single"] as `sonderausgaben_betreuung` is also considered on the individual level. Or am I mistaken?

gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -258,15 +264,23 @@ def eink_st_sonderausgaben_ab_2012(
...
+    else:
+        pauschale = eink_st_abzuege_params["sonderausgabenpauschbetrag"]["single"]
 
     if kind:
         out = 0.0
+    elif sonderausgaben_betreuung > pauschale:
+        out = sonderausgaben_betreuung
```
ChristianZimpelmann, line 281: We should add a test case for which `sonderausgaben_betreuung > pauschale` equals to true. This seems to be never the case.

ChristianZimpelmann (commented):
Just two comments left.

Thanks for the great job!

gettsim/taxes/zu_verst_eink/freibetraege.py
hmgaudecker: I think it should be at the couple level, but that is mostly guessing from my own tax return.

To avoid the confusion I just had, it would be useful to define a variable

`sonderausgaben_gesamt = sonderausgaben_betreuung`

and do the comparisons below in terms of that one. Just to prepare for the future already.

ChristianZimpelmann: But `sonderausgaben` are currently defined on the individual level like all other freibeträge.

It might make sense to change it to the tax unit level, but I would not like to do it in this PR and rather postpone until we decided on #388.

ChristianZimpelmann: > To avoid the confusion I just had, it would be useful to define a variable...

I like this!

LauraGergeleit: > We should compare it only to ["sonderausgabenpauschbetrag"]["single"] as `sonderausgaben_betreuung` is also considered on the individual level. Or am I mistaken?

I think you are right. The `sonderausgaben_betreuung` was also scaled down to the individual level by dividing it by the number of 'anz_erwachsene_tu'. So I adjusted the comparison

> To avoid the confusion I just had, it would be useful to define a variable
> 
> `sonderausgaben_gesamt = sonderausgaben_betreuung`
> 
> and do the comparisons below in terms of that one. Just to prepare for the future already.

I changed the structure of the function a bit and added the suggested variable, so that we can add further items to the 'Sonderausgaben' in the future (e.g. donations, Kirchensteuer, etc.). That was the idea or did I misunderstand it?

gettsim/taxes/zu_verst_eink/freibetraege.py
LauraGergeleit: I added two test cases, which I had to construct myself since I, again, couldn't find an entire calculation example online. So I referred to one previous example from one of the online calculators. The second test case includes one child which is above the age threshold of 14.

gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -231,54 +233,71 @@ def eink_st_sonderausgaben_bis_2011(
...
     details here https://www.buzer.de/s1.htm?a=10&g=estg.
     Parameters
     ----------
     betreuungskost_m
         See basic input variable :ref:`betreuungskost_m <betreuungskost_m>`.
-    kind
-        See basic input variable :ref:`kind <kind>`.
-    anz_kinder_tu
-        See :func:`anz_kinder_tu`.
+    anz_kinder_bis_13_hh
```
ChristianZimpelmann, line 251: This should be based on the tax unit, i.e. `anz_kinder_bis_13_tu`

gettsim/taxes/zu_verst_eink/freibetraege.py
ChristianZimpelmann: Looks great!

gettsim/taxes/zu_verst_eink/freibetraege.py
ChristianZimpelmann: Looks good! Just one more comment

gettsim/taxes/zu_verst_eink/freibetraege.py
LauraGergeleit: You're right! I changed the aggregation in 'aggregation_demographic_vars'.

ChristianZimpelmann (commented):
Well done! Thank you very much @LauraGergeleit !

ChristianZimpelmann:
@hmgaudecker, would you like to have a final look? (I don't think it is necessary)

hmgaudecker:
Looks wonderful, thanks for the great work!

PR #399, opened 2022-06-21
Use dags package for implementation of DAG backend.

ChristianZimpelmann:

- The DAG backend is currently implemented in `dag.py`.
- The [dags](https://github.com/OpenSourceEconomics/dags) package provides most of the functionality and will be used to avoid repetition in code across different projects.

### Todo

- [x] Use dags for `compute_taxes_and_transfers`
- [x] Find out whether and how dags can be used for `plot_dag`
- [x] Decide whether dags should be used for aggregation functions, too
- [x] Put contribution in `CHANGES.rst`.


ChristianZimpelmann:
In GETTSIM, we currently proceed in 2 steps:
  1. Build the DAG (as networkx.DiGraph object)
  2. Execute the DAG.
- Dags, currently, does not return the full DAG, but a concatenated function where "The arguments of the combined function are all arguments of relevant functions that are not themselves function names, in alphabetical order."

I think it would be beneficial to keep the functionality to have the DAG as object within GETTSIM.

- Most important reason: allow for the visualisation of the DAG. 
- Also, some operations, e.g. rounding, are currently applied to the functions in the DAG after it is created, but before it is executed. But we probably can and should do these operations before building the DAG anyway.

I, hence, suggest that I adjust dags such that it allows to return the DAG. This would likely just split the [`concatenate_functions`](https://github.com/OpenSourceEconomics/dags/blob/96d9dce05a7cfbd11dae3ca520df01379312dda3/src/dags/dag.py#L13) function in two parts which then could be called separately.

@janosg @hmgaudecker, what do you think?

janosg:
- I would hope that gettsim developers contribute their visualization code to `dags`. This would then use the internal `dags` functions to build the dag object and plot it. Then plotting would not be a reason to have a dag object. 
- I would also not do anything outside the dag. `dags` provides a few decorators/higher order functions to rename arguments of functions. When using these it usually becomes very easy to do everything inside the dag. 

Having said that: I see no problem in adding a `return_type` argument to `concatenate_functions` or having a separate function to create a DAG object.

hmgaudecker:
I would prefer to refactor `concatenate_functions` so that the first part (up to the pruning of the dag, [currently line 54](https://github.com/OpenSourceEconomics/dags/blob/96d9dce05a7cfbd11dae3ca520df01379312dda3/src/dags/dag.py#L54)) is in a separate function `set_up_dag` or so. 

I'd prefer that to changing the return types. Calling the `set_up_dag` part twice will have close to zero overhead.

Once that is in place, it will be trivial to have the plotting code in either location, once we have something in GETTSIM we can decide whether it is general enough to port it to dags or not.

Finally, the above sounds like a might good idea to swap the order of applying decorators and creating the DAG right away, before doing anything else.

ChristianZimpelmann:
> I would prefer to refactor `concatenate_functions` so that the first part (up to the pruning of the dag, [currently line 54](https://github.com/OpenSourceEconomics/dags/blob/96d9dce05a7cfbd11dae3ca520df01379312dda3/src/dags/dag.py#L54)) is in a separate function `set_up_dag` or so.

That's what I have in mind. Will do! Sounds all very good!

However, I am not sure whether I understand correctly what you mean by decorators/higher order functions. The most important higher order changes we are currently doings are:
- create aggregation functions based on `aggregations_specs` (or automatically) -> needs to be done before DAG is build
- partial policy parameters to functions -> needs to be done before DAG is build
- apply rounding to some functions (as wrapper of the raw function) -> is currently done inside the DAG, but could be done before creating the DAG (which I think is cleaner).
- take care of `columns_overriding_functions` -> currently done while creating the DAG, but AFAICT could be done before

janosg:
> Also, some operations, e.g. rounding, are currently applied to the functions in the DAG after it is created, but before it is executed. But we probably can and should do these operations before building the DAG anyway.

What I meant is that I would not modify the dag after it is created. Instead I would always modify a built-in or user provided function dict and let the dag do the rest. E.g. if a function, say `kindergeld` needs rounding and the original function dict is `{"kindergeld": unrounded_implementation}`, I would rename the key "kindergeld" to "kindergeld_unrounded" internally, then add a function that takes `kindergeld_unrounded` as input and does the rounding under the key "kindergeld". 

This second step often requires renaming arguments of functions. This is where the decorators are helpful.

hmgaudecker:
Thanks!  2 questions

> * create aggregation functions based on aggregations_specs (or automatically) -> needs to be done before DAG is build

Based on `aggregation_specs` (known) seems clear. But the automatic thing I am not sure about -- basically that should happpen during creation of the DAG on demand? I.e., if `bruttolohn_m` exists and `bruttolohn_m_hh` is requested somewhere, create that function? Do you see a clever way on how to do that other than building a DAG first and then looking for such patterns, setting up a new dict with the required functions?

> *    take care of columns_overriding_functions -> currently done while creating the DAG, but AFAICT could be done before

Yes, I guess we could just remove functions that are part of the columns? Maybe easiest if `_harmonize_functions` was exposed?

janosg:
> basically that should happpen during creation of the DAG on demand? I.e., if bruttolohn_m exists and bruttolohn_m_hh is requested somewhere, create that function?

I would have simply created all those functions and let the dag figure out which of them are needed. Creating such functions is almost cost free compared to running calculating the taxes and transfers.

ChristianZimpelmann:
> What I meant is that I would not modify the dag after it is created.

Ok, that is in line with what I meant. 👍 

> Based on aggregation_specs (known) seems clear. But the automatic thing I am not sure about -- basically that should happpen during creation of the DAG on demand? I.e., if bruttolohn_m exists and bruttolohn_m_hh is requested somewhere, create that function? Do you see a clever way on how to do that other than building a DAG first and then looking for such patterns, setting up a new dict with the required **functions?**

We currently just look at the set of all functions and all arguments of those functions to determine the automatic aggregation specs. 

Janos' solution to create all potential aggregation function seems also fine, but might not be necessary.

> Yes, I guess we could just remove functions that are part of the columns?

Yes.

ChristianZimpelmann:
> Yes, but that is the core functionality of dags and we would just be repeating it.

I am not sure as the calculation of

ChristianZimpelmann:
> > We currently just look at the set of all functions and all arguments of those functions to determine the automatic aggregation specs.
> 
> Yes, but that is the core functionality of dags and we would just be repeating it.

To find out the set of all functions and all arguments, yes. But determining the automatic aggregation specs based on the naming convention seems very specific to GETTSIM. I would rather not add this functionality to dags unless another project would also use it.

hmgaudecker:
> To find out the set of all functions and all arguments, yes.

This is what I meant. How could we do the second step without repeating this first step? I'd think just creating all functions would require much less code on GETTSIM's side and certainly much less sophisticated code.

hmgaudecker:
First step: 

> We currently just look at the set of all functions and all arguments of those functions

Second step:

> determine the automatic aggregation specs.

> And are you proposing to create aggregation functions for all functions (for all supported groupings) independent of whether the aggregated function is requested somewhere and of whether it is defined somewhere explicitly?

Yes, anything that is not required will be weeded out by the DAG. According to @janosg, should be very quick!

environment.yml
```diff
@@ -28,7 +29,7 @@ dependencies:
   - sphinx-automodapi
   - sphinx-copybutton
   - pip:
-    - kaleido
-    - plotly
-    - bump2version
-    - -e .
+      - kaleido
```
hmgaudecker, line 32: We should be consistent with indentation of yaml-files everywhere. Is that what this change is about?

gettsim/interface.py
```diff
@@ -632,12 +696,23 @@ def _reorder_columns(results):
...
-    missing_nodes = []
-    for node in _root_nodes(dag):
-        if node not in data and "function" not in dag.nodes[node]:
-            missing_nodes.append(node)
+def _fail_if_root_nodes_are_missing(dag, data, functions):
+    root_nodes = set(_root_nodes(dag))
+
+    # Identify functions that are part of the DAG, but do not depend
+    # on any other function
+    funcs_based_on_params_only = [
```
hmgaudecker, line 704: Can you think of a simpler way to do this check, @janosg ?

hmgaudecker (commented):
Thanks, looks great! 

> I addressed all points we discussed with one exception (which I don't want to tackle in this PR): the creation of aggregation functions is unchanged and does not use the dags package.

Can you explain a bit more? The discussion above is a couple of months old and I fail to make directly sense of this sentence in relation to the above discussion.

Need to run now and will be on a different machine next time 'round, so sending of some very small comments.

environment.yml
ChristianZimpelmann: Good point to keep it consistent. I reverted the change and adjusted my code formatter.

ChristianZimpelmann:
Thanks!

> Can you explain a bit more? The discussion above is a couple of months old and I fail to make directly sense of this sentence in relation to the above discussion.
I see that it wasn't easy to understand :D This is about the automatic generation of aggregation functions. 

Currently, we check for all arguments which:
- end with `_hh` or `_tu`
- no function or data column with this name exists
- a function or data column with this name excluding the group-suffix exists
Then we build aggregation functions for exactly those functions.

If I understood correctly, you proposed to create aggregation functions for all functions and data column (for all groupings) and then let dags decide whether those aggregation function is needed.

I am not sure if this is really an improvement and either way would like to not include it in this PR.

gettsim/interface.py
janosg: I think the the arguments of the concatenated function would already exclude the functions that only depend on params. So you could skip the construction of `funcs_based_on_params_only`. If you still partial in params before you build the dag, this would be simpler. Otherwise you have to filter out the `params` related arguments similar to line 708. 

But I don't find the current solution bad either.

ChristianZimpelmann: Thanks!

We use `funcs_based_on_params_only` only within this function so the concatenated function does not play a role, The function checks whether all root nodes are in the input data. Since we partial in params before building the DAG, functions which only have params as argument are recognized as root nodes.

So, I don't see a better solution (but am of course open for suggestions)

hmgaudecker (commented):
Thanks! I find it extremely hard to follow `compute_taxes_and_transfers` by now (certainly not only caused by this PR!!!). 

I think it would be helpful to sit down next week with you guys @ChristianZimpelmann @janosg and see what can reasonably be done to simplify this (e.g., this comment "Note: Needs to be done after rounding such that dags recognizes partialled parameters." tells me that dags support for partialled parameters could potentially be improved?). 

*(One issue is that contents of variables are constantly changing, e.g., `data` or `all_functions`)*

gettsim/functions_loader.py
```diff
@@ -9,11 +9,11 @@
 def load_user_and_internal_functions(functions):
```
hmgaudecker, line 9:
```suggestion
def load_user_and_internal_functions(user_functions_raw):
```
Why drop the specific name?

gettsim/interface.py
```diff
@@ -105,73 +102,93 @@ def compute_taxes_and_transfers(
     params = {} if params is None else params
     aggregation_specs = {} if aggregation_specs is None else aggregation_specs
 
+    # Perform several checks on data.
+    data = _process_and_check_data(
+        data=data, columns_overriding_functions=columns_overriding_functions
+    )
+
     # Load functions.
```
hmgaudecker, line 110:
```suggestion
    # Load dictionaries with functions.
```

gettsim/interface.py
```diff
@@ -105,73 +102,93 @@ def compute_taxes_and_transfers(
     params = {} if params is None else params
     aggregation_specs = {} if aggregation_specs is None else aggregation_specs
 
+    # Perform several checks on data.
```
hmgaudecker, line 105:
```suggestion
```
Don't state the obvious.

gettsim/interface.py
```diff
@@ -105,73 +102,93 @@ def compute_taxes_and_transfers(
...
+    prepared_results = _prepare_results(results, data, debug)
 
-    return results
+    return prepared_results
 
 
-def check_data_check_functions_and_merge_functions(
-    user_functions,
-    internal_functions,
+def load_and_check_functions(
```
hmgaudecker, line 163: I think we want to move this to `functions_loader.py` and probably remove `load_user_and_internal_functions` entirely. Would help with balancing size (1300 / 200 lines) and the division seems rather arbitrary atm.

Unless I am missing something, could you implement that before I continue with the review? Done for the day, anyhow.

hmgaudecker (commented):
Thanks! See suggestion for moving a (set of) function(s) before I'll continue.

ChristianZimpelmann:
Thanks, good idea!

I moved the respective functions.

hmgaudecker:
In the process of changing the docstring, I also changed 

    load_reforms_for_date -> load_functions_for_date

in 14afe031f1. Hope that is uncontroversial?

hmgaudecker (commented):
Almost there, I think, excellent!!! 

Please double-check my commits, I think they all should be fairly uncontroversial, adding some clarity / consistency (I hope). 

I did not go through interface.py in detail yet. Two things, first minor, second important imho.

- Can we get by without that `KeyErrorMessage` class? I find it distracting and I'd hope that by re-using `format_errors_and_warnings`. Could you check, please?
- The module is very large and likely to grow. This might be a good point to think about the order of the functions (reproduced just below). I do not see any particular structure in that order right now, which makes the module difficult to navigate. Could you make a suggestion here? We can discuss and then implement later. Thanks!

Functions in `interface.py`:

- `compute_taxes_and_transfers`
- `set_up_dag`
- `_process_and_check_data`
- `_round_and_partial_parameters_to_functions`
- `_create_input_data`
- `_prepare_results`
- `_fail_if_columns_overriding_functions_are_not_in_dag`
- `_convert_data_to_correct_types`
- `_fail_if_group_variables_not_constant_within_groups`
- `_fail_if_columns_overriding_functions_are_not_in_data`
- `_reorder_columns`
- `_fail_if_root_nodes_are_missing`
- `_reduce_to_necessary_data`
- `_fail_if_pid_is_non_unique`
- `_fail_if_duplicates_in_columns`
- `_root_nodes`
- `_add_rounding_to_one_function`
- `_add_rounding_to_functions`

gettsim/tests/test_data/grundrente_proxy_rente.csv
```diff
@@ -6,4 +6,4 @@
 "5","5","5","2021","20","False","0","2014","70","0","1951","1","True","624.31","FALSE","480","6","36","0","0","0","0","0","0","0","0","0","0","240","made up, regression test"
 "6","6","6","2021","20","False","0","2020","70","0","1951","1","True","830.82","FALSE","480","6","36","0","0","0","0","0","0","0","0","0","0","240","made up, regression test"
 "7","7","7","2021","20","False","0","1996","90","0","1931","1","True","683.8","FALSE","480","6","36","0","0","0","0","0","0","0","0","0","0","240","made up, regression test"
-"8","8","8","2021","20","False","0","2033","50","0","1971","1","False","0.0","FALSE","300","6","63","0","0","0","0","0","0","0","0","0","0","120","made up, regression test"
+"8","8","8","2021","20","False","0","2033","50","0","2033","1","False","0.0","FALSE","300","6","63","0","0","0","0","0","0","0","0","0","0","120","made up, regression test"
```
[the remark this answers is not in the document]
ChristianZimpelmann: Seems to be a mistake.. I reverted it.

ChristianZimpelmann:
> - Can we get by without that `KeyErrorMessage` class? I find it distracting and I'd hope that by re-using `format_errors_and_warnings`. Could you check, please?

Apparently, `KeyError` is always wrapped by quotes and all line breaks, therefore, are ignored. See [this post](https://stackoverflow.com/questions/46892261/new-line-on-error-message-in-keyerror-python-3-3).

An alternative to using the `KeyErrorMessage` class would be to raise the error as follows:

```python
try:
    raise KeyError("This is a \n Line break")
except KeyError as e:
    print(e.args[0])
```

My prefered alternative would be to keep `KeyErrorMessage`, but moved it to the `shared` module.

ChristianZimpelmann:
Proposal for order of functions is below.

# public functions
- `compute_taxes_and_transfers`
- `set_up_dag`

# data checks and data processing before calculations

(`_process_and_check_data`, `_convert_data_to_correct_types`, and `_create_input_data` and all functions called therein)

- `_process_and_check_data`
- `_convert_data_to_correct_types`
- `_create_input_data`
- `_fail_if_duplicates_in_columns`
- `_fail_if_group_variables_not_constant_within_groups`
- `_fail_if_columns_overriding_functions_are_not_in_data`
- `_fail_if_pid_is_non_unique`
- `_root_nodes`
- `_fail_if_root_nodes_are_missing`
- `_reduce_to_necessary_data`

# function processing

(`_round_and_partial_parameters_to_functions` and and all functions called therein)
(also add `_fail_if_columns_overriding_functions_are_not_in_dag` at the end of this block which is called in `set_up_dag`)

- `_round_and_partial_parameters_to_functions`
- `_add_rounding_to_functions`
- `_add_rounding_to_one_function`
- `_fail_if_columns_overriding_functions_are_not_in_dag`

# data processing after calculations
- `_prepare_results`
- `_reorder_columns`

gettsim/shared.py
```diff
@@ -1,6 +1,38 @@
...
 import textwrap
 
+from gettsim.config import SUPPORTED_GROUPINGS
+
+
+class KeyErrorMessage(str):
+    """Subclass str to allow for line breaks in KeyError messages"""
+
+    def __repr__(self):
+        return str(self)
```
hmgaudecker, line 11: Codecov complains this is not being tested. Useful to add a case, in this PR or elsewhere?

gettsim/shared.py
```diff
@@ -59,24 +122,23 @@ def get_names_of_arguments_without_defaults(function):
...
 
-    def inner(func):
-        func.__rounding_params_key__ = params_key
-        return func
+    return out
 
-    return inner
+
+def rchop(s, suffix):
+    # ToDO: Replace by removesuffix when only python >= 3.9 is supported
```
hmgaudecker, line 139: I think we can safely assume this for the next release. Can you add an issue to remove support for 3.8 and add support for 3.11? Leave a note there to delete this function.

hmgaudecker (approved):
Great! Just a couple of little notes left. 

Small comment on the commit message in 52f97cb5: It should be "Merge branch 'main'." or so -- whether that required resolving merge conflicts is irrelevant. This way I needed to open it in a GUI to understand what was happening.

ChristianZimpelmann:
> Small comment on the commit message in https://github.com/iza-institute-of-labor-economics/gettsim/commit/52f97cb5053810e45abbaffea23ad52ca75c6663: It should be "Merge branch 'main'." or so -- whether that required resolving merge conflicts is irrelevant. This way I needed to open it in a GUI to understand what was happening.

Thanks! Will consider next time.

I added four new test cases and opened the issue. Will merge as soon as tests have run through.

PR #403, opened 2022-06-29
Visualization with plotly

effieHAN:
Replace Bokeh with plotly for visualization. 
The changes are made in the documentation file and visualization py file.


gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+
+    Parameters
+    ----------
+    functions : str, pathlib.Path, callable, module, imports statements, dict
+        Functions can be anything of the specified types and a list of the same objects.
+        If the object is a dictionary, the keys of the dictionary are used as a name
+        instead of the function name. For all other objects, the name is inferred from
+        the function name.
+    source_code str,
+        Experimental feature which makes the source code of the functions or tbe node
```
ChristianZimpelmann, line 357: List the potential values of the argument here. I would rather call it `hover_info`. Please also move the argument to the end of the argument list (not the second position)

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+    check_minimal_specification : {"ignore", "warn", "raise"}, default "ignore"
+        Indicator for whether checks which ensure the most minimal configuration should
+        be silenced, emitted as warnings or errors.
+    selectors : str or list of str or dict or list of dict or list of str and dict
+        Selectors allow to you to select and de-select nodes in the graph for
+        visualization. For the full list of options, see the tutorial about
+        `visualization <../docs/tutorials/visualize.ipynb>`_. By default, all nodes are
+        shown.
+    orientation :str,default "v"
+        decide whether the graph is horizontal or vertical
```
ChristianZimpelmann, line 375:
```suggestion
        Whether the graph is horizontal or vertical
```

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+    showlabels : bool, default True
+        Annotate the graph with labels.
```
ChristianZimpelmann, line 377:
```suggestion
        Whether the graph is annotated with labels.
```
(similar for showlegend)

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+        Indicator for whether checks which ensure the most minimal configuration should
+        be silenced, emitted as warnings or errors.
+    selectors : str or list of str or dict or list of dict or list of str and dict
+        Selectors allow to you to select and de-select nodes in the graph for
+        visualization. For the full list of options, see the tutorial about
+        `visualization <../docs/tutorials/visualize.ipynb>`_. By default, all nodes are
+        shown.
+    orientation :str,default "v"
+        decide whether the graph is horizontal or vertical
+    showlabels : bool, default True
```
ChristianZimpelmann, line 376: I would use `show_labels` and `show_legend`

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
     if 1 <= order:
         for successor in dag.successors(node):
             yield from _kth_order_successors(dag, successor, order=order - 1)
+
+
+def plot_dag(
```
ChristianZimpelmann, line 336: Move this function again to the top as it is the function most interesting for users looking at the code.

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+            marker={
+                "showscale": False,
+                "reversescale": True,
+                "color": "red",
+                "size": 15,
+            },
+        )
+    else:
+
+        raise ValueError("Please choose right argument specification")
```
ChristianZimpelmann, line 573: Be more specific about which argument is wrong, which argument value was given, and which argument values are allowed.

gettsim/visualization.py
ChristianZimpelmann: I just saw that you currently just implemented it as a boolean variable which doesn't fit the description here.

Maybe boolean makes sense if we show labels on hover-mode whenever `source_code = False`?

Or what was your idea?

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+                marker={
+                    "showscale": False,
+                    "reversescale": True,
+                    "color": "red",
+                    "size": 15,
+                },
+            )
+
+    elif source_code is False:
+        fig.add_scatter(
```
ChristianZimpelmann, line 556: Is it necessary and expected that `fig.add_scatter(` is always called three times? I am not sure if all `if statement` are indented as intended.

gettsim/visualization.py
```diff
@@ -571,3 +331,246 @@ def _kth_order_successors(dag, node, order):
...
+            showlegend=showlegend,
+            marker={
+                "showscale": False,
+                "reversescale": True,
+                "color": "red",
+                "size": 15,
+            },
+        )
+
+    elif (len(names) <= 10) or ((len(names) > 10) and (showlabels is True)):
```
ChristianZimpelmann, line 528: Mention this in the docstring of the function: "If 10 or less nodes are plotted, labels are always displayed."

ChristianZimpelmann (changes requested):
Hi,

Thank you very much for that! 

- I made a few comments and suggestions. 
- Please add plotly and all other necessary packages (ipywidgets, ..?) to the environment file `environment.yml`. This is the reason that the tests are failing.
- Please add your contribution to `changes.rst`
- I get the error below when running the HowTo and no plot is shown. Do you have any idea? (Might be also an error on my end .. I haven't used plotly so far)
- I haven't had a closer look at the HowTo for now. Let's just leave it as is for now until we agreed on the function itself.

```
C:\.conda\envs\gettsim\lib\site-packages\jupyter_client\session.py:718: UserWarning:

Message serialization failed with:
Out of range float values are not JSON compliant
Supporting this message is deprecated in jupyter-client 7, please make sure your message is JSON-compliant

**Comment by LauraGergeleit** [comment]
Hi ! :)

I just took a look at the HowTo Guide and when running it through I always got the error below.
I google the error I got this: "The Python "KeyError: 0" exception is caused when we try to access a 0 key in a a dictionary that doesn't contain the key. To solve the error, set the key in the dictionary before trying to access it or conditionally set it if it doesn't exist." (Maybe that helps) 

`---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-20-7b4cbae92021> in <module>
----> 1 plot_dag(
      2     functions=policy_functions,
      3     selectors=[{"node": "zu_verst_kapitaleink_tu", "type": "neighbors"}],
      4     orientation ='v',
      5 );

~\projects\gettsim\gettsim\visualization.py in plot_dag(functions, source_code, targets, columns_overriding_functions, check_minimal_specification, selectors, orientation, showlabels, showlegend)
    424     # Even if we do not show the source codes , we need to remove the functions.
    425     dag = _replace_functions_with_source_code(dag)
--> 426     layout_df = _create_pydot_layout(dag, orientation)
    427     # prepare for the nodes dataframe including their url
    428     names = layout_df.index

~\projects\gettsim\gettsim\visualization.py in _create_pydot_layout(dag, orientation)
    154     # Remap layout from integers to labels.
    155     integer_to_labels = dict(zip(dag_w_integer_nodes.nodes, dag.nodes))
--> 156     layout = {
    157         integer_to_labels[i]: np.array(integer_layout[str(i)])
    158         for i in integer_to_labels

~\projects\gettsim\gettsim\visualization.py in <dictcomp>(.0)
    155     integer_to_labels = dict(zip(dag_w_integer_nodes.nodes, dag.nodes))
    156     layout = {
--> 157         integer_to_labels[i]: np.array(integer_layout[str(i)])
    158         for i in integer_to_labels
    159     }

KeyError: '0'`

**Comment by sofyaakimova** [comment]
Hi we have made changes accordingly, except for the issue with if indentation ` and `fig.add_scater `. 

We consider it important to call` fig.add_scater`  three times as first it draws edges and second the nodes. With respect to nodes it depends on the `hover_info` argument. When `hover_info` is false we need just name of the node, so it refers to the list of the function names. However, when we additionally want to display docstrings of the functions, we have to match the names and the docstrings. It becomes more complex as names in this case are just displayed next to the nodes while hover info includes the docstring. These two cases requires 2 different `fig.add_sacter` functions. That’s how we realize different options. 
If anything is still not clear or if you have any suggestions, please do let us know

@LauraGergeleit we are not completely sure what is the reason for the problem you face, as for us code works on both of computers. We can just give some suggestion that may help. We found that, in previous versions of environment this part of code ```
 layout = {
    157         integer_to_labels[i]: np.array(integer_layout[str(i)])
```
works only if it is displayed as follows```
 layout = {
    157         integer_to_labels[i]: np.array(integer_layout(i))
``` 
Probably the error you receive is also connected to this issue. Could you please try the new version and see if that works now or not?

**Comment by hmgaudecker** [comment]
I changed it back to the original version without the string-conversion. @sofyaakimova @effieHAN, do you remember what made you change that? Runs without it in a freshly created environment -- if that is different for any one of you, we'll need to investigate further.

@LauraGergeleit, my hunch is that the changed code did not work for you because of a missing kernel restart. Could you test again, please?

**Comment by effieHAN** [comment]
> As far as we remembered, we changed the code to string-conversion because in the new environment we got a KeyError (see attached screenshot), and Christian suggested we convert it to string. For us, it would run if we use the original version.
error msg

**Comment by hmgaudecker** [comment]
Thanks! Will be hard to track down what exactly the difference relative to the current version is, let's continue watching it whether any error like that pops up again.

**Comment by sofyaakimova** [comment]
I just tried the latest version and on my computer it works now without string conversion. No idea what caused the problem before.

**Comment by LauraGergeleit** [comment]
> I changed it back to the original version without the string-conversion. @sofyaakimova @effieHAN, do you remember what made you change that? Runs without it in a freshly created environment -- if that is different for any one of you, we'll need to investigate further.
> 
> @LauraGergeleit, my hunch is that the changed code did not work for you because of a missing kernel restart. Could you test again, please?

I guess you were right, now it runs through!

File: gettsim/visualization.py [hunk 9]
```diff
@@ -95,25 +57,16 @@ def plot_dag(
...
-    node_kwargs : dict
-        Additional keyword arguments passed to :class:`bokeh.models.Circle`. For
-        example, change the color with ``{"fill_color": "orange"}``.
+    orientation :str,default "v"
+         Whether the graph is horizontal or vertical
+    show_labels : bool, default True
+        Whether the graph is annotated with labels.
+    show_legend: bool, default False
+        Whether the graph include legend.
+    hover_info: bool, default as false
```
ChristianZimpelmann, line 66: I would rather call it `hover_source_code`.

gettsim/visualization.py
```diff
@@ -95,25 +57,16 @@ def plot_dag(
...
-        Additional keyword arguments passed to :class:`bokeh.models.LabelSet`. For
-        example, change the fontsize with ``{"text_font_size": "12px"}``.
-    node_kwargs : dict
-        Additional keyword arguments passed to :class:`bokeh.models.Circle`. For
-        example, change the color with ``{"fill_color": "orange"}``.
+    orientation :str,default "v"
+         Whether the graph is horizontal or vertical
+    show_labels : bool, default True
+        Whether the graph is annotated with labels.
+    show_legend: bool, default False
```
ChristianZimpelmann, line 64: Setting this to True did not add a legend for me. Can you include an example in the Tutorial?

gettsim/visualization.py
```diff
@@ -95,25 +57,16 @@ def plot_dag(
...
-    label_kwargs : dict
-        Additional keyword arguments passed to :class:`bokeh.models.LabelSet`. For
-        example, change the fontsize with ``{"text_font_size": "12px"}``.
-    node_kwargs : dict
-        Additional keyword arguments passed to :class:`bokeh.models.Circle`. For
-        example, change the color with ``{"fill_color": "orange"}``.
+    orientation :str,default "v"
+         Whether the graph is horizontal or vertical
+    show_labels : bool, default True
+        Whether the graph is annotated with labels.
```
ChristianZimpelmann, line 63: Explain here that this is only relevant for less than 10 nodes and for more nodes it is never shown.

gettsim/visualization.py
```diff
@@ -156,76 +109,161 @@ def plot_dag(
...
-    graph_renderer.edge_renderer.visible = False
-    for (
-        _,
-        (start_node, end_node),
-    ) in graph_renderer.edge_renderer.data_source.to_df().iterrows():
-        (x_start, y_start), (x_end, y_end) = _compute_arrow_coordinates(
-            layout[start_node], layout[end_node]
+    # prepare for the edges dataframe
+    df = pd.DataFrame(list(dag.edges))
+    if len(df) == 0:
```
ChristianZimpelmann, line 137: Please add a test case for which this statement is True.

gettsim/visualization.py
```diff
@@ -156,76 +109,161 @@ def plot_dag(
...
+        x=edge_x,
+        y=edge_y,
+        line={"width": 0.5, "color": "blue"},
+        hoverinfo="none",
+        mode="lines",
+        showlegend=False,
+    )
+    # choose the different option for plotting
+
+    if (len(names) > 10) or (show_labels is False):
```
ChristianZimpelmann, line 200: Same here.. add a test case that hits this if statement.

gettsim/visualization.py
```diff
@@ -156,76 +109,161 @@ def plot_dag(
...
-    tools = [BoxZoomTool(), ResetTool()]
-    tools.append(TapTool(callback=OpenURL(url="@url")))
-    if tooltips:
-        tools.append(HoverTool(tooltips=TOOLTIPS))
-
-    plot.add_tools(*tools)
+    elif (len(names) <= 10) or ((len(names) > 10) and (show_labels is True)):
+        mode = "markers+text"
+
+    elif (len(names) > 10) and (show_labels is True):
```
ChristianZimpelmann, line 221: This elif statement will never be hit since in case `(len(names) > 10)` is True the first statement evaluates to True.

Please check these nested if statements again carefully.

gettsim/visualization.py
```diff
@@ -156,76 +109,161 @@ def plot_dag(
...
+        mode = "markers +text"
+
+    if hover_info is True:
```
ChristianZimpelmann, line 224: shorter: 
```suggestion
    if hover_info:
```

gettsim/visualization.py
```diff
@@ -156,76 +109,161 @@ def plot_dag(
...
-        source = ColumnDataSource(
-            pd.DataFrame(layout).T.rename(columns={0: "x", 1: "y"})
-        )
-        labels = LabelSet(
-            x="x",
-            y="y",
-            text="index",
-            source=source,
-            **{**LABEL_KWARGS_DEFAULT, **label_kwargs},
+    elif hover_info is False:
```
ChristianZimpelmann, line 245:
```suggestion
    elif not hover_info:
```

ChristianZimpelmann (changes requested):
Thank you for the work so far!

I made a couple of comments.

gettsim/visualization.py
sofyaakimova: Hi Christian, we made changes accordingly to all of your comments. As for the legend argument, we deleted it in this version because we think it does not add additional value. Since all the nodes are the same color and the node names are shown anyway, either directly or as hover info.

gettsim/visualization.py
```diff
@@ -202,25 +197,26 @@ def plot_dag(
...
-            marker={
-                "showscale": False,
-                "reversescale": True,
-                "color": "red",
-                "size": 15,
-            },
-        )
-
+        hover_info = "text"
     elif (len(names) <= 10) or ((len(names) > 10) and (show_labels is True)):
```
ChristianZimpelmann, line 201: wouldn't just `else` be enough here?

gettsim/visualization.py
```diff
@@ -156,76 +109,158 @@ def plot_dag(
...
+        hoverinfo="none",
+        mode="lines",
+        showlegend=False,
+    )
+    # choose the different option for plotting
+
+    if (len(names) > 10) or (show_labels is False):
+        mode = "markers"
+        hover_info = "text"
+    elif (len(names) <= 10) or ((len(names) > 10) and (show_labels is True)):
```
ChristianZimpelmann, line 201: Wouldn't `else` be enough here?

I like the simplification that within the if statements only the parameters are set!

gettsim/visualization.py
```diff
@@ -95,25 +56,17 @@ def plot_dag(
...
-        example, change the fontsize with ``{"text_font_size": "12px"}``.
-    node_kwargs : dict
-        Additional keyword arguments passed to :class:`bokeh.models.Circle`. For
-        example, change the color with ``{"fill_color": "orange"}``.
+    orientation :str,default "v"
+         Whether the graph is horizontal or vertical
+    show_labels : bool, default True
+        Whether the graph is annotated with labels.
+        By default when number of nodes is smaller than 10 all names are displayed
+        next to the node.  Otherwise, names are displayed next to the node only when
```
ChristianZimpelmann, line 64:
```suggestion
        next to the node. Otherwise, names are displayed next to the node only when
```

gettsim/visualization.py
```diff
@@ -156,76 +109,158 @@ def plot_dag(
...
+        x=edge_x,
+        y=edge_y,
+        line={"width": 0.5, "color": "blue"},
+        hoverinfo="none",
+        mode="lines",
+        showlegend=False,
+    )
+    # choose the different option for plotting
+
+    if (len(names) > 10) or (show_labels is False):
```
ChristianZimpelmann, line 198: Here, a small comment would be good.

ChristianZimpelmann (commented):
Good job, thanks! I pushed a commit with small changes to tutorial and the docstring and made very few new comments. I have one bigger open question

- in case I select `hover_source_code=True` what happens to the node name/label in case those are not shown for all nodes (`(len(names) > 10) or (show_labels is False)`). If I see correctly, we don't show the name of the node at all in this case? Can we improve on the current behavior? But it is not super important as it is only an experimental feature.

Otherwise, we will be able to merge very soon!

gettsim/visualization.py
```diff
@@ -400,19 +410,21 @@ def _create_pydot_layout(dag):
...
         layout[k] = (v - (max_ + min_) / 2) / ((max_ - min_) / 2).clip(1)
 
-    return layout
+    if orientation == "v":
 
+        layout_df = np.transpose(pd.DataFrame.from_dict(layout))
 
-def _compute_arrow_coordinates(start, end, scalar=0.05):
-    """Compute arrow coordinates.
+    elif orientation == "h":
```
ChristianZimpelmann, line 417: Can you also add a test for horizontan orientation and for `hover_source_code=True`?

LauraGergeleit:
Hi! 
I looked through the HowTo-Guide again, now that it runs through, and checked for any mistakes left. Here are some small things I came across: 

## Introduction:
- `This function**s** has two arguments and one of them passes parameters to the function.` (get rid of the s)
- `Each node is either a function**s** or a variable.` (get rid of the s)
- `By clicking on a node, you are redirected to the documentation of **gettsim** and if the variable is computed by a function, the function's documentation is displayed.` (write GETTSIM in capital letters)
- `Functions inside GETTSIM are a little bit special. Take for example abgelt_st_tu which is documented here. The signature of the function is` (this sentence is double and can be deleted at the end of the paragraph)
- `Here is the complete signature**.**` (my suggestion would be to use a colon instead)

## Selectors:
- `3. "select" specifies whether the nodes should be selected or de-selected. If you do not specify "select" it is assumed to be True.**Let us go through the keys of the dictionary one by one.**` (this part is double and can be deleted)

## De-selecting Nodes:
- `For a simple and silly example, we want to reproduce the graph with the single node for kapitaleink_brutto_tu **after**, but starting from the last plot which also showed zu_verst_kapitaleink_tu.` ('after' makes no sense here, I think it can just be deleted)
- ` The second key in the de-selects "zu_verst_kapitaleink_tu".` (something is missing in that sentence, I would suggest: _The second selector or dictionary in the list de-selects "zu_verst_kapitaleink_tu"._)

## Ancestors and Descendants:
- `To select "zu_verst_kapitaleink_tu" which is the calculated taxable capital income per tax unit and all its ancestors, do the following**.**` (my suggestion would be to use a colon instead)

## Orientation:
- `By default orientation of graph is vertical.` (This sentence seems a bit incomplete. I would suggest: _By default, the orientation of the graph is vertical._)

## Labels:
- `By default when number of nodes is at most 10 all names are displayed next to the node.` (I would suggest the following to make this sentence easier to read: _By default, when the number of nodes is at most 10, all names are displayed next to the respective node._)

## Hover info (source code):
- `It is also possible to address the source code of the function. By setting plot_dag(..., hover_source_code=True)` (I would combine these two sentences to one: _It is also possible to address the source code of the function when hovering over the node by setting plot_dag(..., hover_source_code=True)_)

I hope that helps. In case you have questions, left me know and please note that some of the points are just my suggestions.

ChristianZimpelmann:
I just had one more idea. Could we please change the possible values of the argument `show_labels` as follows:
- `True` -> show labels (independent of number of nodes)
- `False` -> don't show labels (independent of number of nodes)
- `None` (default) -> show labels if number of nodes <= 10

If you think that makes sense, please change it in code, docstring, and tutorial

sofyaakimova:
Hi Christian, we made the changes you and Laura made correspondingly. Could you please have a look?
There is also one thing that grabbed my attention. In the tutorial, we have: "By clicking on a node, you are redirected to the documentation of gettsim and if the variable is computed by a function, the function's documentation is displayed" in the introductory part. However, we haven't still implemented this feature, as the click function that allows us to redirect to external links does not work inside our function and we couldn't find the solution. 
So, should we for now delete this sentence from our tutorial?

ChristianZimpelmann:
Dear Sofya,

Thank you very much! Sorry for the late response. I was very busy this week, but will be faster from now on.

> So, should we for now delete this sentence from our tutorial?

That sounds great!

I have two requests:

- Can you resolve all comments above that you have addressed?
- Can you run precommit hooks over the code.. one tests seems to be failing because of a trailing whitespace. Let me know if you have questions about it.

Best, Christian

gettsim/visualization.py
```diff
@@ -194,17 +195,27 @@ def plot_dag(
...
-        # Otherwise, they are displayed next to the nodes.
+    # When show_lebels = None and number of nodes >10
+    # labels are shown when hovering over it.
+    # Same happens when show_labels is False,
+    # Otherwise, labels are displayed next to the nodes.
 
-        mode = "markers"
-        hover_info = "text"
-    else:
+    if show_labels is None:
```
ChristianZimpelmann, line 203: You can combine the conditions more efficiently:

if show_labels or (show_labels is None and len(names) <= 10):
    asdf
else:
    asdf

gettsim/visualization.py
```diff
@@ -58,11 +58,12 @@ def plot_dag(
...
-        Whether the graph is annotated with labels next to each node, in case the
-        number of nodes is at most 10. Otherwise, names are displayed next to the node
-        only when hovering over it. For more than 10 nodes, this argument is ignored
-        and the labels are always only shown when hovering over it.
+    show_labels : bool, default None
+        Whether the graph is annotated with labels next to each node. By default, 
+        the labels are shown when the number of nodes is at most 10. 
+        Otherwise, names are displayed next to the node only when hovering over it. 
+        It is also possible to display labels regardless of the number of nodes, setting 
+        variable as True or hide labels when a variable is False.
```
ChristianZimpelmann, line 66: I guess "when the variable is False" is better

ChristianZimpelmann:
With `sphinx 5.1.1.` which came out 2 weeks ago, readthedocs does not compile anymore due to an error of automodapi. See my comment in this issue: https://github.com/astropy/sphinx-automodapi/issues/148

For now, I fixed sphinx at 5.1.0 (and python at 3.8 in ``.readtehdocs.yaml`` which readthedocs seems to require).

ChristianZimpelmann (approved):
All done now and we can merge.

Thanks for the great work!

PR #394, opened 2022-05-31
393 definition of regelaltersgrenze not correct

TeBackh:
### What problem do you want to solve?

Problem of Issue 393. I fixed the parameters of the piecewise function for the regelaltersgrenze.

### Todo

- [X] Replace cohort thresholds and adjustment rates for 2007 reform
- [X] If it is a relevant contribution, put it in `CHANGES.rst`.
- [ ] check rules for cohorts before 1945.
- [ ] create tests

PR #408, opened 2022-07-30
Tests regelaltersgrenze

LauraGergeleit:
Referring to #394, `ges_rente_regelaltersgrenze` was added to the tests at `test_renten_anspr`. Already exiting testcases were verified or corrected using the official site of the German pension insurance and further tests for the edge cases were added.


ChristianZimpelmann (approved):
Looks great! Thank you very much!

PR #409, opened 2022-08-10
check for near-equality instead of strict equality within groups.

ChristianZimpelmann:

We currently check whether input variables ending on (`_hh`) are constant within households. Since we check for strict equality, for floating point columns, this might raise unexpected errors.

This PR makes sure that we check only for near-equality using `np.allclose()`


hmgaudecker:
I am not sure this is a good idea. Variables should be calculated on the household level and distributed across members, thus strict inequality will always hold. Why should we expect numerical errors?

ChristianZimpelmann:
> I am not sure this is a good idea. Variables should be calculated on the household level and distributed across members, thus strict inequality will always hold. Why should we expect numerical errors?

It seems to happen in the data-preparation when we use hh-level variables to calculate new hh-level variables (deviations of less than e-13)

Also it might be relevant if we convert a data set from R or Stata to python.

What disadvantage do you see? Would those be less relevant if we decrease the tolerance?

hmgaudecker:
I did not even look at the tolerance. I just think we are trying to solve a problem on the GETTSIM side that should not be there in the first place if data preparation is done "correctly".

By the latter I mean "calculate household-level variables on the household level, then merge with an individual-level data file." Or similar operations, of course.

In general, my view is that we need to be **very** careful to introduce ad hoc code. Maybe we don't see a problem with it now and it will bite us a year from now. Who knows. So for these kinds of things, I would say there is a pretty large burden of proof of actual necessity.

ChristianZimpelmann:
> Variables should be calculated on the household level and distributed across members,

And I don't think it is practical to keep all household level variables in a separate file with index hh_id until the very end of the data prep

hmgaudecker:
That's why I wrote "or similar operations".

If you have a good use case why numbers should not be identical, I'll be happy to hear about it.

ChristianZimpelmann:
> That's why I wrote "or similar operations".

Yes, I have sent the message at the same time as yours, so haven' read it before.

For me, it is comparable to the retyping of variables which we allow. And I don't see it as very ad hoc code if we set the tolerance level very low.

But we can of course also do the check beforehand if you have a strong preference!

hmgaudecker:
Sorry, realised too late it was quoted from my first message :smile: 

My guess is there is no need for any check elsewhere, but a case for refactoring "the data-preparation", whatever that referred to. Happy to have a look!

PR #413, opened 2022-08-24
Minor changes tests

ChristianZimpelmann:

Some small changes to test data and tests. Shouldn't be very critical.

- add `jahr_renteneintr` for all observations. We could discuss whether the variable should be empty if person is not yet retired. However, we fill it also in other tests (`zu_verst_eink.csv`) and should be consistent.
- make definition of `INPUT_COLS` etc. consistent in `test_full_taxes_and_transfers.py`
- make precision explicit in `test_grundrente.py`


TeBackh (approved):
Appove! (Really no fundamental changes -  functionality should remain the same).

Issue #395, opened 2022-06-02
rente für (besonders) langjährig Versicherte (inc. rente mit 63) 


TeBackh:
### Current and desired situation

Missing implementation of deduction free retirement of **very** **long** term insured (45+ years) -- "rente mit 63". Feature requested in issue #348. 

Missing implementation of deduction free earlier retirement for **long** term insured (35 years = very many people) ~ Regelaltersgrenze, which is relevant for cohorts pre 1953. --> full retirement age (FRA) < normal retirement age (NRA)

### Proposed implementation

-[ ] Add age thresholds for long term insured in ges_rente.yaml
-[ ] Add age thresholds for very long term insured
-[ ] implement function to determine eligibility
-[ ] change deductions function for eligible persons (zugangsfaktor)

ChristianZimpelmann:
Sounds good!

But I am not sure about

> choose approximation for insurance years

Insurance years should be an input variable such that the user needs to make these approximations outside of GETTSIM.

PR #414, opened 2022-08-25
Visualization with plotly

sofyaakimova:
The problem with the graphs popping out while testing was solved. The visualizations are no longer opened in a browser.


ChristianZimpelmann (approved):
Thank you very much!

I commented the tests back in (were commented out in #395)

PR #415, opened 2022-08-26
Define supported groupings in config.py

ChristianZimpelmann:

Defining supported groupings in config.py and referring to them in interface.py instead of using the hardcoded groups there.

- [x] Define supported groupings in config.py
- [x] Rename two variables in line with name conventions
- [x] Refer in several docstrings explicitely to level on which variable is calculated
- [x] CHANGES.rst


LauraGergeleit (approved):
Looks good to me! :)

PR #417, opened 2022-09-09
Fix parameter dashboard

Eric-Sommer:

When trying to create data for the parameter dashboard, it turned out that one health insurance parameter was not appropriately named.

In addition, I added a comment on what to do if the data are not there


ChristianZimpelmann (approved):
Thanks for the fix!

PR #418, opened 2022-09-19
Calculation level adjustments

LauraGergeleit:

Ensuring all variable names indicate the correct level of calculation, as discussed in #388.
Fix calculation level of Wohngeld, specifically, of considered income of household.


gettsim/taxes/zu_verst_eink/zu_verst_eink.py
```diff
@@ -15,48 +15,67 @@
 """
 
 
-def freibeträge(
+def individual_freibeträge(
```
hmgaudecker, line 18:
```suggestion
def freibeträge_ind(
```
I think we should stick with the convention that the unit of aggregation is appended as suffix, even if it is more a demarcation from other stuff here.

gettsim/taxes/zu_verst_eink/zu_verst_eink.py
```diff
@@ -15,48 +15,67 @@
...
     vorsorgeaufw: float,
-    sonderausgaben: float,
     _eink_st_behinderungsgrad_pauschbetrag: float,
-    alleinerz_freib_tu: float,
     eink_st_altersfreib: float,
+    alleinerz_freib_tu: float,
 ) -> float:
-    """Calculate allowances.
+    """Calculate individual allowances.
```
hmgaudecker, line 24:
```suggestion
    """Sum up all tax-deductible allowances applicable at the individual level.
```

gettsim/transfers/wohngeld.py
```diff
@@ -39,22 +39,25 @@ def wohngeld_m_hh(
...
-    ges_rentenv_beitr_m_tu: float,
-    ges_krankenv_beitr_m_tu: float,
+    ges_rentenv_beitr_m: float,
+    ges_krankenv_beitr_m: float,
     wohngeld_params: dict,
 ) -> float:
-    """Calculate housing benefit subtractions per tax unit.
+    """Calculate housing benefit subtractions on the individual level.
+
+    Note that eink_st_tu is used as an approximation, since Lohnsteuer
```
hmgaudecker, line 50: Not sure about this. Lohnsteuer would apply to employees only, but we'll need to think of self-employed and (eventually) civil servants, too (might well be that a low-ranked police officer in München with a large family is eligible).

gettsim/transfers/wohngeld.py
```diff
@@ -63,68 +66,66 @@ def wohngeld_abzüge_m_tu(
 
     """
     abzug_stufen = (
-        (eink_st_tu > 0) + (ges_rentenv_beitr_m_tu > 0) + (ges_krankenv_beitr_m_tu > 0)
+        (eink_st_tu > 0) + (ges_rentenv_beitr_m > 0) + (ges_krankenv_beitr_m > 0)
```
hmgaudecker, line 69: Seems okay. Just to be sure: Is health insurance also counted if a pensioner is applying for wohngeld?

gettsim/transfers/wohngeld.py
```diff
@@ -244,41 +245,38 @@ def wohngeld_eink_abzüge_m_ab_2016(
     return freib_behinderung_m + freib_kinder_m
 
 
-def wohngeld_eink_m(
+def wohngeld_eink_m_hh(
```
hmgaudecker, line 248: I don't quite understand why we are down to the individual level now. This seems to be all using variables defined at the household level, will be the same for all members? But if I have a variable without the `_hh` suffix, I expect that I can sum it up to get the aggregate, which does not seem the right thing to do here. But maybe I am missing something.

hmgaudecker (commented):
Thanks! A couple of clarifying questions on wohngeld, also see Zulip discussion [here](https://gettsim.zulipchat.com/#narrow/stream/243798-Transfers/topic/Wohngeld/near/297925681) and below.

gettsim/taxes/zu_verst_eink/zu_verst_eink.py
LauraGergeleit: You're right! I will change it.

gettsim/transfers/wohngeld.py
LauraGergeleit: You are right, that does not seem correct, when looking at the discussion [here](https://gettsim.zulipchat.com/#narrow/stream/243798-Transfers/topic/Wohngeld/near/297925681). I will try to rephrase it, but I'm a bit unsure whether I understood it fully

gettsim/transfers/wohngeld.py
LauraGergeleit: That's how I would interpret it ([§ 16 WoGG](https://www.gesetze-im-internet.de/wogg/__16.html)), since pensioners either pay Pflichtbeiträge or pay into a private/voluntary insurance plan that serves the same purpose. (also see [here](https://www.deutsche-rentenversicherung.de/DRV/DE/Rente/In-der-Rente/Kranken-und-Pflegeversicherung-der-Rentner/kranken-und-pflegeversicherung-der-rentner.html))

gettsim/transfers/wohngeld.py
LauraGergeleit: Christian is right, I forgot to change the docstring, I will correct that. The function should be on the household level, using the aggregated (on hh level) income after the deductions for social insurance and taxes but before the allowances (`wohngeld_eink_vor_freib_m`) and the aggregated (on hh level) allowances (`wohngeld_eink_freib_m_hh`).

gettsim/taxes/zu_verst_eink/zu_verst_eink.py
LauraGergeleit: I will change that, thank you!

ChristianZimpelmann (approved):
Great job! Thanks!

PR #420, opened 2022-09-30
minor fix in description of self-employed input variable

ChristianZimpelmann:
Minor fixes in description of input_variables


docs/gettsim_objects/input_variables.rst
```diff
@@ -32,9 +32,9 @@ household.
 +---------------------------+---------------------------------------------+--------------+
 | _`in_ausbildung`          | Dummy: Employment status “in education”     | bool         |
 +---------------------------+---------------------------------------------+--------------+
-| _`selbstständig`          | Dummy: Individual is self-employed          | bool         |
+| _`selbstständig`          | Dummy: Full-time self-employed              | bool         |
```
hmgaudecker, line 35: Why would it have to be full time? I guess it is more about the main activity? Will need to rethink that classification at some point, anyhow.

ChristianZimpelmann: You are right. We need the translation of "hauptberuflich". 

How about `Dummy: Self-employed (main profession)` ? Alternatively, we can just add the German term

hmgaudecker: Just use that, perfectly fine. 

Before a 1.0 release we should definitely go for a categorical variable with the main type of activity (employment subject to social security contributions, civil servant, self-employed, unemployed, out of the labor force, parental leave, pensioner,  ...). Will get hairy with pensioner / employment definition from 2023 onwards, of course... Don't we already have an issue for that?

ChristianZimpelmann: 👍 

I don't remember and haven't found one.

LauraGergeleit:
Looks fine to me!

LauraGergeleit (approved):
Looks fine to me!

PR #422, opened 2022-10-10
raise warning (instead of error) if data for dashboard is missing.

ChristianZimpelmann:

- Pytest raises an error when looking for tests and the input data for the dashboard (`param_dashboard_data.pickle`) is missing.
- Bokeh expects the code for the dashboard in the script without a `if __name__ == '__main__':` condition.

Solution: raise warning if input data is missing instead of raising an error.

### DIFF
--- a/dashboard/app/main.py
+++ b/dashboard/app/main.py
@@ -1,4 +1,5 @@
 import pickle
+import warnings
 
 import pytz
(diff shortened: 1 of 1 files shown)


hmgaudecker (approved):
Thanks!

PR #423, opened 2022-10-10
Gerinfügigkeitsgrenzen

Eric-Sommer:
Models changes in mini / midi job rulings effective since 2022-10-01

- higher Midi-Job threshold
- mini job threshold is now called 'Geringfügigkeitsgrenze' and requires three new parameters
  - `mindestlohn`
  - `geringf_eink_faktor`
  - `geringf_eink_divisor`
- In addition, a policy change needs to introduced in `policy_functions.py`

### DIFF
--- a/gettsim/parameters/soz_vers_beitr.yaml
+++ b/gettsim/parameters/soz_vers_beitr.yaml
@@ -770,6 +770,28 @@ geringfügige_eink_grenzen_m:
     deviation_from: previous
     reference: Art. 4 G. v. 28.11.2018 BGBl. I S. 2016
     midi_job: 1300
+  2022-10-01:
+    midi_job: 1600
+    reference: Art. 7 G. v. 28.06.2022 BGBl. I S. 969
+    note: the former mini_job parameter depends on the statutory minimum wage since 2022-10-01
+
+geringf_eink_faktor:
+  name:
+    de: Faktor, mit dem der Mindestlohn in der Berechnung der Geringfügigkeitsgrenze multipliziert wird.
+  reference: §8 (1) Nr. 1a SGB IV
+  2022-10-01:
+    scalar: 130
+    reference: Art. 7 G. v. 28.06.2022 BGBl. I S. 969
+
+geringf_eink_divisor:
+  name:
+    de: Betrag, durch den der Mindestlohn, multipliziert mit gering_eink_faktor, dividiert wird,
+        um die Geringfügigkeitsgrenze zu erhalten.
+  reference: §8 (1) Nr. 1a SGB IV
+  2022-10-01:
+    scalar: 3
+    reference: Art. 7 G. v. 28.06.2022 BGBl. I S. 969
+
 
 ag_abgaben_geringf:
   name:
@@ -807,6 +829,31 @@ ges_pflegev_zusatz_kinderlos_mindestalter:
   reference: § 55 Abs. 3 SGB XI
   2005-01-01:
     scalar: 23
+
+mindestlohn:
+  name:
+    de: Allgemeiner gesetzlicher Mindestlohn pro Stunde
+    en: Statutory hourly minimum wage
+  unit: Euro
+  reference: §1 (2) Mindestlohngesetz
+  2015-01-01:
+    scalar: 8.5
+  2017-01-01:
+    scalar: 8.84
+  2019-01-01:
+    scalar: 9.19
+  2020-01-01:
+    scalar: 9.35
+  2021-01-01:
+    scalar: 9.5
+  2021-07-01:
+    scalar: 9.6
+  2022-01-01:
+    scalar: 9.82
+  2022-07-01:
+    scalar: 10.45
+  2022-10-01:
+    scalar: 12
 
 rounding:
   midi_job_faktor_f:
--- a/gettsim/social_insurance_contributions/eink_grenzen.py
+++ b/gettsim/social_insurance_contributions/eink_grenzen.py
@@ -2,7 +2,7 @@ from gettsim.shared import add_rounding_spec
 
 
 def mini_job_grenze(
-    wohnort_ost: bool, geringfügigkeitsgrenze_west: int, gerinfügigkeitsgrenze_ost: int
+    wohnort_ost: bool, geringfügigkeitsgrenze_west: int, geringfügigkeitsgrenze_ost: int
 ) -> float:
     """Select the income threshold depending on place of living
 
@@ -16,7 +16,7 @@ def mini_job_grenze(
     -------
 
     """
-    out = gerinfügigkeitsgrenze_ost if wohnort_ost else geringfügigkeitsgrenze_west
+    out = geringfügigkeitsgrenze_ost if wohnort_ost else geringfügigkeitsgrenze_west
     return float(out)
 
 
--- a/gettsim/social_insurance_contributions/eink_grenzen.py
+++ b/gettsim/social_insurance_contributions/eink_grenzen.py
@@ -1,7 +1,9 @@
 from gettsim.shared import add_rounding_spec
 
 
-def mini_job_grenze(wohnort_ost: bool, soz_vers_beitr_params: dict) -> float:
+def mini_job_grenze(
+    wohnort_ost: bool, geringfügigkeitsgrenze_west: int, gerinfügigkeitsgrenze_ost: int
+) -> float:
     """Select the income threshold depending on place of living
 
     Parameters
@@ -14,8 +16,7 @@ def mini_job_grenze(wohnort_ost: bool, soz_vers_beitr_params: dict) -> float:
     -------
 
     """
-    params = soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["mini_job"]
-    out = params["ost"] if wohnort_ost else params["west"]
+    out = gerinfügigkeitsgrenze_ost if wohnort_ost else geringfügigkeitsgrenze_west
     return float(out)
 
 
@@ -128,6 +129,7 @@ def midi_job_bemessungsentgelt_m(
     midi_job_faktor_f: float,
     bruttolohn_m: float,
     soz_vers_beitr_params: dict,
+    geringfügigkeitsgrenze_west: int,
 ) -> float:
     """Income subject to social insurance contributions for midi job.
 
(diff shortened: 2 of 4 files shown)


ChristianZimpelmann (changes requested):
Thanks, a lot! Looks really good and especially the parameters are very nicely documented! I added two comments.

Additionally, the calculation of contributions for midi jobs was changed. Let's discuss on Zulip who is going to implement these changes.

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -152,32 +200,165 @@ def midi_job_bemessungsentgelt_m(
...
+
+    quotient = midijob_grenze / (midijob_grenze - geringfüg_grenze)
+    einkommen_diff = bruttolohn_m - geringfüg_grenze
+
+    out = quotient * einkommen_diff
+
+    return out
+
+
+def geringfügigkeitsgrenze_west_vor_2022(soz_vers_beitr_params: dict) -> int:
```
ChristianZimpelmann, line 307:
```suggestion
def mini_job_grenze_west_vor_2022(soz_vers_beitr_params: dict) -> int:
```

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -152,32 +200,165 @@ def midi_job_bemessungsentgelt_m(
...
+
+    Returns
+    -------
+    Income subject to employee social insurance contributions for midi job.
+    """
+    geringfüg_grenze = soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["mini_job"][
+        "west"
+    ]
+
+    midijob_grenze = soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["midi_job"]
```
ChristianZimpelmann, line 297:
```suggestion
    midi_job_grenze = soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["midi_job"]
```

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -152,32 +200,165 @@ def midi_job_bemessungsentgelt_m(
...
+        See basic input variable :ref:`bruttolohn_m <bruttolohn_m>`.
+    soz_vers_beitr_params
+        See params documentation :ref:`soz_vers_beitr_params <soz_vers_beitr_params>`.
+
+
+    Returns
+    -------
+    Income subject to employee social insurance contributions for midi job.
+    """
+    geringfüg_grenze = soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["mini_job"][
```
ChristianZimpelmann, line 293: Why west? Can we use `mini_job_grenze`?

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -152,32 +200,165 @@ def midi_job_bemessungsentgelt_m(
...
-            - soz_vers_beitr_params["geringfügige_eink_grenzen_m"]["mini_job"]["west"]
+            - geringfügigkeitsgrenze_west
         )
         * midi_job_faktor_f
     )
 
     return mini_job_anteil + lohn_über_mini * gewichtete_midi_job_rate
 
 
+def midi_beitragspfl_einnahme_m(
```
ChristianZimpelmann, line 223:
```suggestion
def midi_beitragspf_einnahme_m(
```

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -152,32 +200,165 @@ def midi_job_bemessungsentgelt_m(
...
+    einkommen_diff = bruttolohn_m - geringfüg_grenze
+
+    faktor1 = midi_job_faktor_f * geringfüg_grenze
+    faktor2 = (quotient1 - quotient2 * midi_job_faktor_f) * einkommen_diff
+    out = faktor1 + faktor2
+
+    return out
+
+
+def midi_sond_beitragspfl_einnahme_m(
```
ChristianZimpelmann, line 268:
```suggestion
def midi_beitragspf_einnahme_arbeitn_m(
```

gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -335,10 +335,39 @@ def _ges_krankenv_midi_job_sum_arbeitn_arbeitg_m(
     return out
 
 
-def _ges_krankenv_midi_job_arbeitg_m(
+def _ges_krankenv_midi_job_sum_arbeitn_arbeitg_m_ab_2022(
```
ChristianZimpelmann, line 338: I very much like that we have separate functions for `arbeitn`, `arbeitg` and `arbeitn_arbeitg` in `ges_krankenv`. Can you do the same for rentenv, arbeitsl_v and pflegev (If it makes sense there)?

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -124,12 +125,59 @@ def midi_job_faktor_f(
...
 
 
+@add_rounding_spec(params_key="soz_vers_beitr")
+def midi_job_faktor_f_ab_2022(
+    soz_vers_beitr_params: dict,
+) -> float:
+    """Faktor F which is needed for the calculation of Bemessungsentgelt
+    (beitragspflichtige Einnahme) of midi jobs. It is calculated as the ratio of 28 %
+    divided by the total social security contribution rate
+    (Gesamtsozialversicherungsbeitragssatz).
```
ChristianZimpelmann, line 135: Refer to time of introduction of this function (since October 2022). Also do so for previous function. Please also check for all other functions that were "split up in time" in this PR.

gettsim/social_insurance_contributions/eink_grenzen.py
LauraGergeleit: Yes, you are right, using `mini_job_grenze` is more elegant, I will change it. I just used west since it is the same for both regions in 2022, but it might change again in the future I guess, so the other solution is better

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -124,12 +125,59 @@ def midi_job_faktor_f(
     return out
 
 
+@add_rounding_spec(params_key="soz_vers_beitr")
+def midi_job_faktor_f_ab_2022(
+    soz_vers_beitr_params: dict,
+) -> float:
+    """Faktor F which is needed for the calculation of Bemessungsentgelt
+    (beitragspflichtige Einnahme) of midi jobs. It is calculated as the ratio of 28 %
```
ChristianZimpelmann, line 133: Describe where 28% comes from (also in previous function). If 28% did not change, we can still mention that number in brackets.

In the docstring or a comment in the code, it should become clear why we change from using `ges_krankenv_beitr_satz` to `soz_vers_beitr_params["beitr_satz"]["ges_krankenv"]["allgemein"]`. Couldnt we stick to `ges_krankenv_beitr_satz` and `_ges_krankenv_beitr_satz_arbeitg`?

gettsim/social_insurance_contributions/eink_grenzen.py
LauraGergeleit: I will adjust the function names together with the change of `vor_2022` to `vor_10_2022`

ChristianZimpelmann (changes requested):
Looks very good! Thank you very much!

I have a few comments, mostly on documentation and some proposals for name changes.

hmgaudecker:
Thanks for all your work! Might well be that we missed it elsewhere, but I think that as per GEP 1:
- it should be `minijob` and `midijob` (seem to be German words...)
- we should not have `mini_[some_concept]` and `midi_[some_concept]` as opposed to `minijob_[some_concept]` and `midijob_[some_concept]`anywhere

gettsim/parameters/soz_vers_beitr.yaml
```diff
@@ -675,101 +675,126 @@ geringfügige_eink_grenzen_m:
...
       west: 450
       ost: 450
-    midi_job: 850
+    midijob: 850
   2019-01-01:
     deviation_from: previous
     reference: Art. 4 G. v. 28.11.2018 BGBl. I S. 2016
-    midi_job: 1300
+    midijob: 1300
+  2022-10-01:
```
Eric-Sommer, line 773: The `minijob` parameter is not specified explicitly in the law anymore and it is not used by `gettsim`. Didn't we agree to only include explicit parameters? Including it here instead of setting it to null may tempt users to simulate reforms altering this parameter, which will not work.

LauraGergeleit: Yes, you're right, it can be deleted. Thanks for the comment! The `minijob` parameter is a leftover from my first attempt, where I thought I needed to explicitly state the limit, however now gettsim uses the function `minijob_grenze` for the calculations.

gettsim/social_insurance_contributions/arbeitsl_v.py
```diff
@@ -34,6 +34,83 @@ def sozialv_beitr_gesamt_m(
...
+    _ges_rentenv_beitr_midijob_sum_arbeitn_arbeitg_m
+        See :func:`_ges_rentenv_beitr_midijob_sum_arbeitn_arbeitg_m`.
+    in_gleitzone
+        See :func:`in_gleitzone`.
+
+    Returns
+    -------
+
+    """
+    if in_gleitzone:
```
ChristianZimpelmann, line 102: Why is it not always the sum of `sozialv_beitr_arbeitn_m` and `sozialv_beitr_arbeitg_m`?

gettsim/social_insurance_contributions/arbeitsl_v.py
```diff
@@ -1,7 +1,7 @@
 """Functions for modeling unemployment and pension insurance."""
 
 
-def sozialv_beitr_gesamt_m(
+def sozialv_beitr_arbeitn_m(
```
ChristianZimpelmann, line 4: In line with the naming of the four arguments, we can drop `_arbeitn` here.

gettsim/social_insurance_contributions/arbeitsl_v.py
```diff
@@ -34,6 +34,83 @@ def sozialv_beitr_gesamt_m(
...
+    out = (
+        ges_pflegev_beitr_arbeitg_m
+        + ges_krankenv_beitr_arbeitg_m
+        + ges_rentenv_beitr_arbeitg_m
+        + arbeitsl_v_beitr_arbeitg_m
+    )
+    return out
+
+
+def sozialv_beitr_gesamt_m(
```
ChristianZimpelmann, line 70: I would go for `_sozialv_beitr_arbeitn_arbeitg_m`

gettsim/social_insurance_contributions/ges_rentenv.py
```diff
@@ -40,6 +40,50 @@ def ges_rentenv_beitr_m(
...
 
+def ges_rentenv_beitr_arbeitg_m(
+    geringfügig_beschäftigt: bool,
+    _ges_rentenv_beitr_midijob_arbeitg_m: float,
+    _ges_rentenv_beitr_bruttolohn_m: float,
+    soz_vers_beitr_params: dict,
+    in_gleitzone: bool,
+    bruttolohn_m: float,
+) -> float:
+    """Contribution for a respective employer to the pension insurance.
```
ChristianZimpelmann, line 51: for a -> of the

(please also adjust for the other similar functions)

ChristianZimpelmann (changes requested):
Thanks, almost done!

Just some small comments.

gettsim/social_insurance_contributions/arbeitsl_v.py
LauraGergeleit: I added this in order to also use and test the functions in `ges_pflegev.py`, `ges_rentenv.py`, `ges_krankenv.py` and `arbeitsl_v.py` that calculate the total social insurance contributions for midijob. Shall I change it so that it is consistent with the other cases? Maybe we will then have some coverage problems again..

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -149,39 +211,172 @@ def midi_job_bemessungsentgelt_m(
 
     Returns
     -------
-    Income subject to social insurance contributions for midi job.
+    Income subject to social insurance contributions for midijob.
```
ChristianZimpelmann, line 214: Only from October 2022 on?

gettsim/social_insurance_contributions/eink_grenzen.py
```diff
@@ -124,23 +126,83 @@ def midi_job_faktor_f(
...
+
+
+def midijob_bemessungsentgelt_m(
+    midijob_faktor_f: float,
     bruttolohn_m: float,
     soz_vers_beitr_params: dict,
+    minijob_grenze_west: float,
 ) -> float:
-    """Income subject to social insurance contributions for midi job.
+    """Income subject to social insurance contributions for midijob until October 2022.
```
ChristianZimpelmann, line 194: Until September 2022?

gettsim/social_insurance_contributions/eink_grenzen.py
ChristianZimpelmann: Couldn't we just still call this `midijob_bemessungsentgelt_m` and add `10_2022` of course?

gettsim/social_insurance_contributions/ges_pflegev.py
```diff
@@ -191,21 +308,97 @@ def _ges_pflegev_beitr_midi_job_m(
     -------
 
     """
-    # Calculate the sum of employee and employer care insurance contribution.
-    gesamtbeitrag_midi_job_m = (
-        midi_job_bemessungsentgelt_m
-        * 2
+    ag_beitr_midijob_m = (
```
ChristianZimpelmann, line 311: If variable is just returned directly afterwards, can be called `out` (to be consistent with similar functions.

ChristianZimpelmann (commented):
Thanks! Only few comments left..

ChristianZimpelmann (approved):

PR #425, opened 2022-10-11
Make gettsim JAX ready

timmens:

Provide functionality to check that library and user-defined functions can be compiled with JAX (and potentially adjust them).


gettsim/jax.py
```diff
@@ -0,0 +1,277 @@
...
+from typing import Dict
+
+import astor
+
+
+DEFAULT_MAPPING = {" and ": " & ", " or ": " | ", " not ": " ~"}
+
+
+def change_if_to_where_wrapper(
+    func: callable, backend: str = "np", mapping: Dict[str, str] = DEFAULT_MAPPING
```
hmgaudecker, line 15:
```suggestion
    func: callable, backend: str, mapping: Dict[str, str] = DEFAULT_MAPPING
```
I would prefer not to have a default backend.

Also, could you explain the rationale for having the DEFAULT_MAPPING as opposed to baking in the functionality? At least in the context of GETTSIM, I think it would be less confusing to just document what the function is doing rather than having a flexible interface.

hmgaudecker (commented):
This looks amazing, thanks! 

Just a couple of broad suggestions for now, I did not manage to go into a lot of detail. Two broad comments:
- Would it be hard to add testcase for how much of the current test suite runs with Jax as backend? Begets the question of how to specify it / the precise interface of GETTSIM which is in flux currently (`gettsim.config` as it is currently assumed does not seem ideal?)
- The names for the script and test should probably change to something like `convert_source_code_for_jax`? Probably too long and not nice, but you get the idea and I'll be happy to discuss!

timmens:
1. The mapping was a bad idea in the first place. I added negations and boolean expressions to the transformer. Now the code is even more robust. Thanks for the push!
2. I don't have a great idea for a name yet. I changed it to `make_vectorizable` for now, but I don't like it. In the end, the name should probably not be tied too closely to Jax, since the code is backend agnostic and can also be used to provide a speedup if one chooses to use NumPy.
3. If I get a list of functions and respective test inputs it should be easy. I'd simply run my code, `jit` the result, and run it on the test input. But I don't know how easy it is to get that list?

hmgaudecker:
Great! As for the last point, some time next week the dags branch should be useful for that and make it very easy to get that list / the test suite.

hmgaudecker:
> Great! As for the last point, some time next week the dags branch should be useful for that and make it very easy to get that list / the test suite.

Of course, this should have read "next month" :innocent: But it is merged now and I just updated the branch.

@ChristianZimpelmann @timmens, can you have a joint look at how to set up the tests?

timmens:
I've added a test that runs the conversion on all gettsim functions. I further added a test that checks if the converted function can be called with `numpy`/`jax.numpy` arrays. For special cases, this already works. For most cases, however, the source code contains other function calls that are not vectorizable, like Python's built-in `max`.

**My proposal:**
> Convert Python's built-in functions that are currently used in gettsim (e.g. `max` -> `jax.numpy.max`)

In this case, I'd need a list of these functions. Adding these cases to the converter is not too much work ---approximately 30min to 1hour per function.

**TODO:**
1. Decide what to do with Python built-ins like `max`
2. Rewrite functions that are currently not convertible because if-else-clause does not follow our special syntax
3. Discuss for what we needed the dags PR. For testing, I got a list of gettsim functions from `gettsim.functions_loader._load_functions`. Did misunderstand something @hmgaudecker ?

ChristianZimpelmann:
Thanks! That sounds great!

> Decide what to do with Python built-ins like max

I feel like the only alternative to rewriting those functions would be to use an internal max-function in all those functions. But that would again increase the difficulty for GETTSIM users.

> In this case, I'd need a list of these functions. 

This should be easy to collect based on the result of the tests, right? I can put it together if you prefer.

> Rewrite functions that are currently not convertible because if-else-clause does not follow our special syntax

This is something, you definitely don't need to do. @LauraGergeleit and me can take care of it.

timmens:
@ChristianZimpelmann 

I just talked to Hans-Martin. We think the best approach would be to use e.g. `np.max` instead of `max`, where `np` is either `numpy` or `jax.numpy` depending on whether `jax` is installed. In this case, someone would need to swap the built-in functions with a numpy equivalent. What do you think? Is this too difficult for non-experts? If we require the special if-else syntax I think we can also require the usage of `np.max` and the like.

If we choose this option I'd provide the code so that you would simply write
you would write
```python
from gettsim.config import np
```

hmgaudecker:
Great, thanks! I think my only issue with what you write is about "depending whether Jax is installed". 

There needs to be a way for the user to explicitly tell GETTSIM which backend to use. We just do not need to be able to switch after import, I think. Not sure how to achieve that but it should be a pretty common problem?

ChristianZimpelmann:
`np.max` sounds good!

timmens:
I implemented the Jax backend. The workflow is now:

```python
from gettsim.config import set_array_backend
from gettsim.interface import compute_taxes_transfers

set_array_backend("jax")
res = compute_taxes_and_transfers(...)
```

where internally gettsim uses `jax.numpy` instead of `np` **if** jax is installed.

---

There are some more problems, unfortunately. I've noticed that `gettsim` uses `numpy.datetime`; this is not (yet) supported by `jax`. Of course, we could simply use standard `numpy` in this case, but then things would get a bit messy in the source code with `numpy` and `np` floating around, where the latter may even represent `jax.numpy`.

hmgaudecker:
I do not think we (will ever) go below the monthly level in the policy functions (as opposed to selecting the point in time etc). Then it would be fine to just use an integer representation, I think.

Does that seem about right, Christian?

ChristianZimpelmann:
- It is true that we only use np.datetime at the moment to calculate `alter_monate`.
- The only reason we might need to be more precise at some point might be if we would like to sort children by their age (including twins or step-children)

hmgaudecker:
Just had a look and after the small fixes to imports most failures are expected an error messages are sensible. The only one I am struggling with is the "too many values to unpack" ting, e.g., here:

```
_______________________________________________________________ test_convertable[arbeitsl_geld_2_m_hh] _______________________________________________________________

func = <function arbeitsl_geld_2_m_hh at 0x7f0e1f1e3e20>

    @pytest.mark.parametrize("func", gettsim_functions.values())
    def test_convertable(func):
>       make_vectorizable(func, backend="numpy")

src/_gettsim_tests/test_jax.py:267: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/_gettsim/jax.py:31: in make_vectorizable
    tree = _make_vectorizable_ast(func, module=module)
src/_gettsim/jax.py:89: in _make_vectorizable_ast
    new_tree = Transformer(module).visit(tree)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:410: in visit
    return visitor(node)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:486: in generic_visit
    value = self.visit(value)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:410: in visit
    return visitor(node)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:486: in generic_visit
    value = self.visit(value)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:410: in visit
    return visitor(node)
src/_gettsim/jax.py:126: in visit_If
    self.generic_visit(node)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:495: in generic_visit
    new_node = self.visit(old_value)
../../../miniconda3/envs/gettsim/lib/python3.11/ast.py:410: in visit
    return visitor(node)
src/_gettsim/jax.py:122: in visit_BoolOp
    binop = _boolop_to_binop(node)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

node = <ast.BoolOp object at 0x7f0e1eee2e00>

    def _boolop_to_binop(node: ast.BoolOp):
        """Transform BoolOp operation to BinOp.
    
        Args:
            node (ast.BoolOp): A BoolOp node in the ast.
    
        Returns:
            ast.BinOp: The BoolOp expression reformatted using '&' instead of 'and', and '|'
                instead of 'or'.
    
        """
>       left, right = node.values
E       ValueError: too many values to unpack (expected 2)

src/_gettsim/jax.py:233: ValueError
```

Function body is:

```py
    if (
        wohngeld_vorrang_hh
        or kinderzuschl_vorrang_hh
        or wohngeld_kinderzuschl_vorrang_hh
        or erwachsene_alle_rentner_hh
    ):
        out = 0.0
    else:
        out = arbeitsl_geld_2_vor_vorrang_m_hh
```

Also by looking elsewhere, e.g., `arbeitsl_geld_berechtigt` with body:

```python
    out = (
        (
            arbeitsl_monate_gesamt
            <= arbeitsl_geld_params["dauer_auszahlung"]["max_dauer"]
        )
        and (alter < arbeitsl_geld_params["altersgrenze"]["alter"])
        and (sum_ges_rente_priv_rente_m == 0)
        and (arbeitsstunden_w < arbeitsl_geld_params["stundengrenze"])
    )
```
my best guess is that this is due to the use of Python's `and` / `or` statements? (using `&` certainly helped this last test to pass, but I did not check exhaustively and I am not sure whether behavior might be impacted).

Is this assessment correct, @timmens? Any chance to improve the error message?

hmgaudecker:
On further investigation joint with @timmens, the issue is that the re-writing code expects a single "and" / "or" operation and it fails on chaining them.

hmgaudecker:
Thanks for the thorough analysis! 

5. + Testcase (?) definitely seem like the best start.

timmens:
I added some additional test cases and quickly rewrote the code so that boolean expressions are translated to function calls `numpy.logical_and` and the like. This seems much more robust than using the bitwise operators. The new translation function uses a recursive strategy which allows for chained expressions.

hmgaudecker:
This is great, @timmens, thanks a lot! 

I converted stuff so that tests should pass. Probably fine to tackle the rest at a later point.

One (likely) last thing before I would be happy to merge would be to add a page to the docs. Bullets are fine, I shall be happy to convert to text. Ideal contents:
- Gist of the restrictions
- Anything that is missing from the tests currently (e.g., making stuff like `test_transfers__elternzeit_elternzeit_anspruch` automatically run via jax and numpy, adding tests of min/max  and other constructs that are not supported and not caught currently -- at least it is my understanding based on conversation above that we will still need to convert `min(x, y)` to `np.min(x, y)` everywhere?)

timmens:
Thank you!

I'll write something up for the docs.

Regarding `max` and the like. It is very easy to translate these function calls to `numpy.max` calls. If we find calls to other non-vectorizable functions with a clear vectorizable representation in numpy/jax, they can be translated easily too. For `max` and `min`, this would, however, require another restriction: We can only allow for either 1 or 2 arguments; 1 argument implies that we call `max` on a sequence, so we need to call `numpy.amax`, and 2 arguments implies we call `max` between two arguments, so we need to call `numpy.maximum`. **Otherwise,** I can also add a section to the docs saying that users need to specify maximum operations using `np.max` and `np.maximum`.

hmgaudecker:
Ah, great, apologies -- I just misinterpreted what you wrote [up there](https://github.com/iza-institute-of-labor-economics/gettsim/pull/425#issuecomment-1326655134). Amazing this works! 

Just make a note of that restriction (1 or 2 arguments), then.

I have to run now, but if you find a quick way of just setting USE_JAX to globally in the tests when Jax is installed, we'd really be done!

timmens:
Problems and Questions:

1. I don't yet get why we need to set `USE_JAX` in the test file.
1. I get an error when running `pytest`:
   ```console
   FAILED src/_gettsim_tests/test_jax.py::test_convertible[unterhaltsvors_m] - AttributeError: 'Raise' object has no attribute 'value'
   ```
   which is caused because an error is raised in the body of an if condition. I am not entirely sure, but I don't think we can support raising errors inside if-conditions. If this is too restrictive, we need to brainstorm again.
1. I think before fixing everything jax-related we should write an end-to-end test for the numpy backend.

TODO:
-  [ ] Write documentation

hmgaudecker:
> Problems and Questions:
> 
>     2. I get an error when running `pytest`:

Sorry, my bad. I just introduced that in 62188ae8ee9, will fix.

>     1. I don't yet get why we need to set `USE_JAX` in the test file.
> 
>     3. I think before fixing everything jax-related we should write an end-to-end test for the numpy backend.

There are a few, I guess we just did not mark them accordingly :innocent: 
- `test_full_taxes_and_transfers.py` has end-to-end tests
- `test_aggregation_functions.py` is actually just waiting for `USE_JAX` to be enabled (note #482 in that respect)

hmgaudecker:
> > Problems and Questions:
> > 
> > 2. I get an error when running `pytest`:
> 
> Sorry, my bad. I just introduced that in [62188ae](https://github.com/iza-institute-of-labor-economics/gettsim/commit/62188ae8ee91cdb0d7f89570f12eff35ee3bdf62), will fix.

Done, it seems.

timmens:
New testing behaviour should be that both backends, `jax` and `numpy`, are tested if `jax` is installed. If not, only the `numpy` backend is tested. I solve this by using pytest's `conftest.py` module, in which I define a function executed before test collection. Ther, I set `USE_JAX` to true.

Additionally:
- I had to switch to `numpy.datetime64` since `jax.numpy` does not support DateTime functions
- I had to fix two if-conditions merged a couple of hours ago

.github/workflows/main.yaml
```diff
@@ -50,29 +50,35 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
-  # run-jax-tests:
-  #   runs-on: ubuntu-latest
-  #   strategy:
-  #     fail-fast: false
-  #     matrix:
-  #       python-version: ["3.9", "3.10", "3.11"]
+  run-jax-tests:
```
hmgaudecker, line 53: I guess this is irrelevant with the new structure and all that is needed is to make
```yaml
          extra-specs: |
            python=${{ matrix.python-version }}
            jax
```
os-dependent?

hmgaudecker:
> New testing behaviour should be that both backends, `jax` and `numpy`, are tested if `jax` is installed. If not, only the `numpy` backend is tested. I solve this by using pytest's `conftest.py` module, in which I define a function executed before test collection. There, I set `USE_JAX` to true.

Very cool! but I don't quite understand how the entire test suite is then being run with the numpy backend if jax is installed? Or was that related to the run-tests / run-jax-tests settings in the workflow?

Ideally, the fixture in conftest would mean that all tests are executed twice (if jax is installed), once with each backend. That way we would check that the numpy backend is used even if jax is installed.

hmgaudecker (commented):
Really neat, I think we are basically good to go conditional on the point I raised re configuring tests and documenting the restrictions on Python code we can have.

Thanks!!!

src/_gettsim/aggregation_numpy.py
```diff
@@ -1,7 +1,7 @@
 from warnings import warn
 
-import numpy as np
 import numpy_groupies as npg
+from _gettsim.config import numpy_or_jax as np
```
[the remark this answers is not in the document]
timmens: Very good point, I will use your suggestions.

.github/workflows/main.yaml
timmens: Yes, that's much better.

timmens:
1. I've added the file `code_restrictions.md` to the docs folder. It is not integrated into the documentation yet, since I did not know where to place it.

2. Testing: I'm a bit confused about the jax testing part. To double check: Do you mean the whole test suite should be run once with `USE_JAX = True` and once with `USE_JAX = False` if jax is installed using a single pytest call? At the moment, I am unsure if this can be easily implemented. However, it would certainly be possible to call pytest twice with a custom argument `--use-jax` or something similar.

Another remark. Currently, we use `USE_JAX` as a global variable. There are other approaches like using environment variables as jax, or defining a dictionary in `config.py`, which is updated. This is probably a good opportunity to ask for advice from Lars Reimann. Of course, this can also be optimized in another PR.

hmgaudecker:
> 1. I've added the file `code_restrictions.md` to the docs folder. It is not integrated into the documentation yet, since I did not know where to place it.

Excellent, thanks! Part of the docs now as of dfc7de9 (we will need to add appropriate pointers from other sections eventually).

>     2. Testing: I'm a bit confused about the jax testing part. To double check: Do you mean the whole test suite should be run once with `USE_JAX = True` and once with `USE_JAX = False` if jax is installed using a single pytest call? At the moment, I am unsure if this can be easily implemented. However, it would certainly be possible to call pytest twice with a custom argument `--use-jax` or something similar.

This last thing sounds like a great solution! In the matrix, leave out numpy on MacOS then. Tests run so slow there and if 
- jax passes on MacOS
- jax and numpy pass on Ubuntu
that should carry over to numpy on MacOS.

> Another remark. Currently, we use `USE_JAX` as a global variable. There are other approaches like using environment variables as jax, or defining a dictionary in `config.py`, which is updated. This is probably a good opportunity to ask for advice from Lars Reimann. Of course, this can also be optimized in another PR.

Sounds great. @lars-reimann, do you have an opinion there? The only thing I want to avoid is dependence on environment variables. This is too confusing for many people and often hard to achieve when you are starting things from different places (e.g., specifying and env variable in .bashrc and that not being respected when running from VSCode...).

lars-reimann:
Since `USE_JAX` affects the entire system, I think having it as a global variable is fine. Otherwise, I could also see it as another optional parameter to `compute_taxes_and_transfers`, similar to the `rounding` and `debug` parameters.

hmgaudecker:
The problem with adding a keyword to `compute_taxes_and_transfers` (or a successor) is that we to know it need at import time. If that was doable without having to specify a decorator for each affected function (which is what we do for rounding & debugging), that would be the perfect solution indeed!

timmens:
1. I've added a flag `--use-jax`, which can be used to set jax as the backend; the default is numpy. To make this work, I had to
    - Move the `conftest.py` file into the `src` directory. pytest would not recognize the added option when the file was placed in `_gettsim_tests`. After doing some research, my best guess is that the conftest.py file only influences the behaviour for tests in the same directory hierarchy or deeper; but since there are doctests located in `_gettsim` the hook is not executed.
    - Adapt the run-tests command in the workflow file to `pytest src`, since otherwise, pytest would not recognize its contents. I haven't found out why pytest doesn't find the file otherwise.
2. The testing behavior now works as intended; however, there are problems with `np.datetime`. I think it would be best if someone who understands the codebase better could take a look. I'm not sure if we can fix this by replacing the arbitrary backend with `numpy`, or if we need to define our own date using ints; @ChristianZimpelmann ?

.github/workflows/main.yaml
```diff
@@ -33,46 +33,35 @@ jobs:
...
+        if: runner.os != 'macOS'
+        run: >-
+          micromamba run -n gettsim pytest src --cov=./ --cov-report=xml
 
-      - name: Run tests
+      - name: Run tests with jax backend on Linux and MacOS
+        if: runner.os != 'Windows'
         run: >-
-          micromamba run -n gettsim pytest --ignore=dashboard/ --cov=./ --cov-report=xml
+          micromamba run -n gettsim pytest src --use-jax --ignore=dashboard/
```
timmens, line 61:
```suggestion
          micromamba run -n gettsim pytest src --use-jax
```

hmgaudecker:
Okay, so the failures seem to be boiling down to some really subtle difference between `numpy.vectorize` and `jax.numpy.vectorize` combined with partialling parameters.

```pytb
$ pytest src/_gettsim_tests/test_arbeitsl_geld.py -x --pdb -v

src/_gettsim_tests/test_arbeitsl_geld.py::test_arbeitsl_geld[2010] FAILED                                                                                                      >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

input_data =     hh_id  tu_id  p_id  bruttolohn_vorj_m  wohnort_ost   kind  ...  alg_ssc  alg_tax  alg_soli  alg_entgelt           ...3.00   383.62     21.10      1412.28  old test data - anwartschaftszeit, arbeitssuch...    none

[11 rows x 24 columns]
year = 2010

    @pytest.mark.parametrize("year", YEARS)
    def test_arbeitsl_geld(
        input_data,
        year,
    ):
        year_data = input_data[input_data["jahr"] == year].reset_index(drop=True)
        df = year_data[INPUT_COLS].copy()
        policy_params, policy_functions = set_up_policy_environment(date=year)
    
>       result = compute_taxes_and_transfers(
            data=df,
            params=policy_params,
            functions=policy_functions,
            targets="arbeitsl_geld_m",
        )

src/_gettsim_tests/test_arbeitsl_geld.py:44: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/_gettsim/interface.py:149: in compute_taxes_and_transfers
    results = tax_transfer_function(**input_data)
/home/hmg/miniconda3/envs/gettsim/lib/python3.11/site-packages/dags/output.py:22: in wrapper_dict_output
    raw = func(*args, **kwargs)
/home/hmg/miniconda3/envs/gettsim/lib/python3.11/site-packages/dags/signature.py:77: in wrapper_with_signature
    return func(*args, **kwargs)
/home/hmg/miniconda3/envs/gettsim/lib/python3.11/site-packages/dags/dag.py:382: in concatenated
    result = info["func"](**kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = ()
kwargs = {'soz_vers_beitr_params': {'ag_abgaben_geringf': {'ges_krankenv': 0.13, 'ges_rentenv': 0.15, 'st': 0.02}, 'beitr_bemes...erlos': 0.0025}, 'ges_rentenv': 0.0995}, 'bezugsgröße_selbst_anteil': 0.75, ...}, 'wohnort_ost': array([False, False])}

    @functools.wraps(func)
    def wrapper_vectorize_func(*args, **kwargs):
>       return func_vec(*args, **kwargs)
E       TypeError: _ges_rentenv_beitr_bemess_grenze_m() got an unexpected keyword argument 'soz_vers_beitr_params'

src/_gettsim/functions_loader.py:572: TypeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>



>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
[15] > /mnt/econ/gettsim/gettsim/src/_gettsim/functions_loader.py(572)wrapper_vectorize_func()
-> return func_vec(*args, **kwargs)
   6 frames hidden (try 'help hidden_frames')

(Pdb++) inspect.signature(func_vec)
<Signature (wohnort_ost: bool, soz_vers_beitr_params: dict) -> float>
```

I checked in `interface._round_and_partial_parameters_to_functions` that signatures of the vectorized function look exactly the same after partialling:

```pdb
(Pdb++) inspect.signature(partial_func)
<Signature (wohnort_ost: bool, *, soz_vers_beitr_params: dict = {'beitr_satz': {'ges_krankenv': {'allgemein': 0.14, 'sonderbeitrag': 0.009, 'ermäßigt': 0.134}, 'ges_pflegev': {'standard': 0.00975, 'zusatz_kinderlos': 0.0025}, 'arbeitsl_v': 0.014, 'ges_rentenv': 0.0995}, 'beitr_bemess_grenze_m': {'ges_krankenv': {'west': 3750, 'ost': 3750}, 'ges_rentenv': {'west': 5500, 'ost': 4650}}, 'bezugsgröße_selbst_m': {'west': 2555, 'ost': 2170}, 'bezugsgröße_selbst_anteil': 0.75, 'geringfügige_eink_grenzen_m': {'minijob': {'west': 400, 'ost': 400}, 'midijob': 800}, 'ag_abgaben_geringf': {'ges_krankenv': 0.13, 'ges_rentenv': 0.15, 'st': 0.02}, 'ges_pflegev_zusatz_kinderlos_mindestalter': 23, 'datum': numpy.datetime64('2010-01-01'), 'rounding': {'midijob_faktor_f': {'base': 0.0001, 'direction': 'nearest'}, 'minijob_grenze_west': {'base': 1, 'direction': 'up'}, 'minijob_grenze_ost': {'base': 1, 'direction': 'up'}}}) -> float>
```

Note that we are vectorizing at a point where nothing is known yet about which arguments are to be vectorized over and which arguments are the same across all functions. My best guess is that `jax.numpy` assumes we want to vectorize over all arguments wheres `numpy` manages to deal with that case. Did you run into something like this before by any chance, @janosg ? 

Anyhow, I think this should be easily fixable once a `data_inputs` / `vectorize_over` / `params_inputs` decorator is implemented as per the [current draft of GEP 6](https://gettsim--471.org.readthedocs.build/en/471/geps/gep-06.html#usage-and-impact). So unless someone sees a smoking gun, my suggestion would be to comment out the `--use-jax` tests and merge this already.

janosg:
Another thing I wonder: Do you even need any of `vmap` or `vectorize`? Since you are auto-replacing if conditions with `where` and similar things, the functions might run out of the box with vectors for most inputs. In that case, numpy vectorize would most likely slow them down quite a bit as it is implemented in pure Python and meant as syntactic sugar.

timmens:
This is the current implementation of gettsims `_vectorize_func` inside the functions_loader.py module:

```python
def _vectorize_func(func):
    signature = inspect.signature(func)

    # Vectorize
    if USE_JAX:

        # ToDo: user jnp.vectorize once all functions are compatible with jax
        func_vec = np.vectorize(func)

    else:
        func_vec = np.vectorize(func)

    @functools.wraps(func)
    def wrapper_vectorize_func(*args, **kwargs):
        return func_vec(*args, **kwargs)

    wrapper_vectorize_func.__signature__ = signature

    return wrapper_vectorize_func
```

Shouldn't this be replaced by

```
from _gettsim.jax import make_vectorizable
from _gettsim.jax import USE_JAX
def _vectorize_func(func):
    backend = "jax" if USE_JAX else "numpy"
    return make_vectorizable(func, backend)
```

?

hmgaudecker:
> Another thing I wonder: Do you even need any of `vmap` or `vectorize`? Since you are auto-replacing if conditions with `where` and similar things, the functions might run out of the box with vectors for most inputs. In that case, numpy vectorize would most likely slow them down quite a bit as it is implemented in pure Python and meant as syntactic sugar.

That's a neat idea -- something to try on another day :smile:

hmgaudecker:
> ```python
> from _gettsim.jax import make_vectorizable
> from _gettsim.config import USE_JAX
> 
> def _vectorize_func(func):
>     backend = "jax" if USE_JAX else "numpy"
>     return make_vectorizable(func, backend)
> ```
> 
> ?


Certainly plausible :laughing: 

That brings up a `TranslateToVectorizableError` that we did not catch before:

```
E           _gettsim.jax.TranslateToVectorizableError: An if statement is performing multiple operations, which is forbidden.
E           Please only perform one operation in the body of an if-elif-else statement.
E           The source code in question is:
E           
E            > if 'heizkostenentlastung_m' in wohngeld_params:
E            >     max_def_hh_größe_heating = numpy.max(i for i in wohngeld_params[
E            >         'heizkostenentlastung_m'] if isinstance(i, int))
E            >     heating_allowance_m = numpy.where(haushaltsgröße_hh <=
E            >         max_def_hh_größe_heating, wohngeld_params['heizkostenentlastung_m']
E            >         [haushaltsgröße_hh], wohngeld_params['heizkostenentlastung_m'][
E            >         max_def_hh_größe_heating] + (berücks_personen -
E            >         max_def_hh_größe_heating) * wohngeld_params[
E            >         'heizkostenentlastung_m']['jede_weitere_person'])
E            > else:
E            >     heating_allowance_m = 0
```

I am wondering two things:
1. Could we add information on which module the error occurs in? E.g., the above occurs after replacing `max` by `numpy.max`, so it is not always trivial to just search the entire codebase.
2. I thought we were testing for all functions to be vectorizable, somehow that was missed. Could you just point me to the respective test quickly? 

Once I used your suggested code and fixed the error above in [e706d3d](https://github.com/iza-institute-of-labor-economics/gettsim/pull/425/commits/e706d3d495ba9f8d2a6171a1b17f9af1aedc1ac1), I am getting a RecursionError that I have no clue of what to make of:

```pytb
/home/hmg/miniconda3/envs/gettsim/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = {'col_offset': 68, 'ctx': <ast.Load object at 0x7fd33e69de10>, 'end_col_offset': 78, 'end_lineno': 50, ...}
memo = {140545211813696: {}, 140545211813952: {}, 140545236296320: {'test': <ast.Compare object at 0x7fd33b6d0dc0>}, 14054523...ol_offset': 7, 'comparators': [<ast.Name object at 0x7fd33b6d0f40>], 'end_col_offset': 50, 'end_lineno': 98, ...}, ...}
deepcopy = <function deepcopy at 0x7fd3dff287c0>

    def _deepcopy_dict(x, memo, deepcopy=deepcopy):
        y = {}
>       memo[id(x)] = y
E       RecursionError: maximum recursion depth exceeded while calling a Python object

/home/hmg/miniconda3/envs/gettsim/lib/python3.11/copy.py:229: RecursionError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```

Any thoughts, anyone?

src/_gettsim_tests/test_jax.py
```diff
@@ -0,0 +1,518 @@
...
+
+gettsim_functions = _load_functions(sources=PATHS_TO_INTERNAL_FUNCTIONS)
+
+not_convertible_yet = {
+    "wohngeld_miete_m_hh_ab_2009",
+    "ges_rente_zugangsfaktor",
+    "_arbeitsl_geld_2_alleinerz_mehrbedarf_m_hh",
+    "eink_st_altersfreib_ab_2005",
+    "vorsorgeaufw_ab_2020",
+}
```
timmens, line 373: I suspect that these functions cause one problem: Makes sense that the end-to-end test does not run if it uses one of these functions. Regarding the recursion error. I think I used `deepcopy` too often, which somehow caused problems with the `ast.NodeTransformer`. Will be fixed with the next commit.

hmgaudecker:
While I am at it, let me re-up this one, @timmens 

> Could we add information on which module the error occurs in? E.g., the above occurs after replacing max by numpy.max, so it is not always trivial to just search the entire codebase.

Or is this information gone after the first couple of transformations?

timmens:
Sorry, I actually thought about this but forgot to post it here.

I think this is very easy to implement. I'll add it during the next 3 hours. Can you describe how you want the error message to look exactly? Suggestion:

```
E           _gettsim.jax.TranslateToVectorizableError: An if statement is performing multiple operations, which is forbidden.
E           Please only perform one operation in the body of an if-elif-else statement.
E           Function: _gettsim.transfers.rente.ges_rente_zugangsfaktor
E           Problematic source code:
E           ...
```

hmgaudecker (approved):

PR #396, opened 2022-06-20
Abschlagsfreie Rente (besonders) langjährig Versicherte ("Rente mit 63") & Rente für Frauen

TeBackh:
### What problem do you want to solve?

Closes #395 

Implement features of the retirement system, for very long term insured (besonders langjährig Versicherte) -- at least 45 years of contributions. These people can retire before the normal retirement age (NRA) (regelaltersgrenze) without deductions, ie reach their full retirement age (FRA) earlier.

The most recent reform of this feature known as "Rente mit 63", starting with cohorts 1951 (introduced in 2014). Yet, before there have been similar pathways to retirement for very long-term insured at age 65.

### Proposed implementation

- [x] Add cohort specific age thresholds in ges_rente.yaml
- [x] implement functions determining thresholds for respective pensions
- [x] implement relevant thresholds in zugangsfaktor function
- [x] implement function to determine eligibility
- [x] implement wartezeiten 5, 15, 35, 45
- [x] implement eligibility criteria in functions
- [x] (implement threshold for earliest possible pension age (women till 1952 60 vs men 63) )

### Todo

- [x] Target the right branch and pick an appropriate title.
- [x] Put `Closes #395` in the first PR comment to auto-close the relevant issue once
  the PR is accepted. This is not applicable if there is no corresponding issue.
- [x] If it is a relevant contribution, put it in `CHANGES.rst`.

PR #428, opened 2022-11-14
Tests and Improvements: Age limits Rente für Frauen / Rente (besonders) langjährig Versicherte

LauraGergeleit:

In #396, functions for the pension for (very) long term insured and the pension for women were implemented, calculating the retirement age. So far, testcases for these functions were missing. Here, an overview of the retirement ages for cohorts born in or after 1930, considering the different pension forms,  is used to add new testcases in `renten_alter.py`.


hmgaudecker:
Mechanics look great to me, but I won't manage to go through individual tests. @TeBackh, could you have a look (maybe together with @LauraGergeleit)? 

Thanks!

TeBackh (changes requested):
Danke Laura. Sieht soweit gut aus

src/_gettsim_tests/test_data/renten_alter.csv
```diff
@@ -0,0 +1,1369 @@
+jahr,p_id,hh_id,tu_id,geburtsjahr,geburtsmonat,weiblich,alter,m_arbeitsunfähig,m_krank_ab_16_bis_24,m_mutterschutz,m_arbeitslos,m_ausbild_suche,m_schul_ausbild,m_alg1_übergang,m_geringf_beschäft,y_pflichtbeitr_ab_40,m_freiw_beitrag,m_ersatzzeit,m_kind_berücks_zeit,m_pfleg_berücks_zeit,m_pflichtbeitrag,_ges_rente_altersgrenze_abschlagsfrei,Comment,,,Quelle,Calculation of age,yob,mob,regsray,regsram,ltsray,ltsram,wosray,wosram
+2010,1,1,1,1930,1,TRUE,80,0,0,0,0,0,0,0,0,10,0,0,0,0,180,60,Altersrente für Frauen,,,Table taken from a paper which calculates the retirement age for each cohort (variables with y indicate years and variables with m indicate months),(variable_y)+(variable_m)*(1/12),1900,1,65,0,63,0,60,0
```
ChristianZimpelmann, line 2: I would replace `a paper` by the names of the authors. Otherwise, the PR looks good and almost ready to merge!

LauraGergeleit: @TeBackh Do you have the names of the author?
Another open question was whether the conversion of the test cases is comprehensible enough, since I made them by hand, or whether we need a function, for example. What do you think @ChristianZimpelmann?

ChristianZimpelmann: I think for comprehensibility reasons it would be indeed good to use the data from the paper as input and transform them automatically, not by hand.

What kind of transformations were necessary? If it is not too complicated we could also just mention the transformations in the notes.

TeBackh: > @TeBackh Do you have the names of the author? Another open question was whether the conversion of the test cases is comprehensible enough, since I made them by hand, or whether we need a function, for example. What do you think @ChristianZimpelmann?

Seibold, A. (2021). Reference points for retirement behavior: Evidence from german pension discontinuities. American Economic Review, 111(4), 1126-65.

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -42,14 +43,61 @@
...
+    out["m_alg1_übergang"] = 0
+    out["m_geringf_beschäft"] = 0
+    out["m_freiw_beitrag"] = 0
+    out["m_ersatzzeit"] = 0
+    out["m_kind_berücks_zeit"] = 0
+    out["m_pfleg_berücks_zeit"] = 0
+    out["_ges_rente_altersgrenze_abschlagsfrei"] = out["Rentenalter_Jahr"] + out[
+        "Rentenalter_Monat"
+    ] * (1 / 12)
+    out.loc[out["m_pflichtbeitrag"] == 180, "y_pflichtbeitr_ab_40"] = 10
```
ChristianZimpelmann, line 92: Can you add a comment before those two lines, what is done and what the assumption is?

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -42,14 +43,61 @@
...
+    }
+    out = original_data[[*var_names]].rename(columns=var_names).copy()
+
+    # Create IDs
+    out["tu_id"] = out["p_id"]
+    out["hh_id"] = out["p_id"]
+
+    # Create variables needed in GETTSIM
+    out["weiblich"] = True
+    out["jahr"] = random.choice(YEARS)
```
ChristianZimpelmann, line 75: Why do we have a random choice here?

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -42,14 +43,61 @@
...
 def input_data():
-    file_name = "renten_alter.csv"
-    out = pd.read_csv(TEST_DATA_DIR / file_name)
+    # Original Data Source: Seibold, A. (2021).
+    # Reference points for retirement behavior:
+    # Evidence from german pension discontinuities.
+    # American Economic Review, 111(4), 1126-65.
+
+    # IDs were added to the original data
+    # m_pflichtbeitrag is added as an identifier for pension kind
```
ChristianZimpelmann, line 52: -> type of pension?

ChristianZimpelmann (commented):
Looks good! Much easiser to see the assumptions and changes now.

Just three very small comments. @TeBackh, would you also like to have another look?

src/_gettsim_tests/test_renten_alter.py
LauraGergeleit: My idea was: Since the original testcases do not have a year, I had to attach a year. However, if all tests have the same year attached to them, then we would only have one testcase. So to ensure that we test for different years I implemented this, but maybe there is a better solution?

hmgaudecker: You can have the fixture taking arguments: https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

ChristianZimpelmann: Well should the tested outcome depend on the year? I am a bit surprised that it doesn't as the respective parameters probably change over time?

TeBackh: They mostly (and in the test cases completely) depend on birth date not the year of the claim. The gettsim structure and the the tests currently always require a payout/taxation year , however.
It actually sound like a good solution to me to randomize -  maybe we could have a range of say 1990-2020 to make sure that we have the relevant tax and transfer structure established for the test cases

hmgaudecker: We'll never want to have randomization in tests, would lead to failures that cannot be reproduced. Say you run it once and 1995 comes up (and fails), next time it does not and tests pass...

My suggestion above would just line up the evaluation year (used to calculate age, I guess) and the policy year, tests are still being run for the entire range of years in the file. Would there be a use case for anything else in the tests? *(as opposed to counterfactual experiments)*

Speaking of years, how come they go up to 2021? Because we are sure the policy environment did not change for anything that happened after Arthur's analysis/data prep stops? Must be way earlier given paper publication and data publication lags...

TeBackh: > are still being run for the entire range of years in the file

thats good to know. So it's enough to use one example year.

> Speaking of years, how come they go up to 2021? Because we are sure the policy environment did not change for anything that happened after Arthur's analysis/data prep stops? Must be way earlier given paper publication and data publication lags..

The Seibold table contains the retirement ages up to cohort 1975. Laura is only using the columns of Regelaltersrente, abschlagsfreie Altersrente für langjährig Versicherte und Altersrente für Frauen. The information in these columns is up to date with all the policy changes currently known to me. 
 (not so the column for "besonders langjährig Versicherte"- thats why we do not use it) . For potential use of all columns we should stop evaluating in 2014.
I'd argue that policy uncertainty about the "Regelaltersgrenze" and "Altersgrenze für langjährig Versicherte" is mostly relevant for cohorts born after 1964 - as the adjustment rates are already set up to this cohort and the "Altersrente für Frauen" probably remains abolished.

src/_gettsim_tests/test_data/renten_alter.csv
```diff
@@ -1,1369 +1,2449 @@
...
-2018,992,992,992,1968,8,FALSE,52,0,0,0,0,0,0,0,0,0,0,0,0,0,60,67,Regelaltersgrenze,,,Table taken from a paper which calculates the retirement age for each cohort (variables with y indicate years and variables with m indicate months),,,,,,,,,
-2018,993,993,993,1968,9,FALSE,52,0,0,0,0,0,0,0,0,0,0,0,0,0,60,67,Regelaltersgrenze,,,Table taken from a paper which calculates the retirement age for each cohort (variables with y indicate years and variables with m indicate months),,,,,,,,,
-2018,994,994,994,1968,10,FALSE,52,0,0,0,0,0,0,0,0,0,0,0,0,0,60,67,Regelaltersgrenze,,,Table taken from a paper which calculates the retirement age for each cohort (variables with y indicate years and variables with m indicate months),,,,,,,,,
-2018,995,995,995,1968,11,FALSE,52,0,0,0,0,0,0,0,0,0,0,0,0,0,60,67,Regelaltersgrenze,,,Table taken from a paper which calculates the retirement age for each cohort (variables with y indicate years and variables with m indicate months),,,,,,,,,
-2018,996,996,996,1968,12,FALSE,52,0,0,0,0,0,0,0,0,0,0,0,0,0,60,67,Regelaltersgrenze,,,Table taken from a paper which c
```
TeBackh, line 1: what happened to ltsray,ltsram,wosray and wosram?
where is the variable m_pflichtbeitrag coming from?

TeBackh (changes requested):

src/_gettsim_tests/test_data/renten_alter.csv
LauraGergeleit: I had store the values of ltsray,ltsram,wosray and wosram underneath the regsray and regsram and I used `m_pflichtbeitrag` to identify the respective type of pension. So that still had to be done manually, I couldn't come up with another solution within GETTSIM. Do you have another idea?

src/_gettsim_tests/test_renten_alter.py
ChristianZimpelmann: > They mostly (and in the test cases completely) depend on birth date not the year of the claim. 

Thanks for the explanation. I think, we can then just add one year (2010?) and run the tests for this year only.. probably best would be if we made a note/comment somewhere for which time frame we expect those policy rules to be in place.

> (not so the column for "besonders langjährig Versicherte"- thats why we do not use it)

Do you mean that we haven't implemented the full policy regime or that the test cases would be only valid until 2014? In the latter case, I think we should keep them and just set the evaluation year to some year before.

hmgaudecker: I think there is value in keeping all years here because we might introduce seemingly innocuous changes that break these things at other points of the policy system.

Agreed that we should test the pre-2014 environment as extensively as possible.

src/_gettsim_tests/test_data/renten_alter.csv
hmgaudecker: This is too cryptic for me to comment on in case you are valuing my opinion :laughing:

src/_gettsim_tests/test_renten_alter.py
ChristianZimpelmann: > I think there is value in keeping all years here because we might introduce seemingly innocuous changes that break these things at other points of the policy system.

Then, how about running all tests for a set of years (if that does not take too long to run)

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -0,0 +1,110 @@
...
+
+YEARS = [2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021]
+
+
+@pytest.fixture(scope="module")
+def input_data():
+    # Original Data Source: Seibold, A. (2021).
+    # Reference points for retirement behavior:
+    # Evidence from german pension discontinuities.
+    # American Economic Review, 111(4), 1126-65.
```
ChristianZimpelmann, line 49: We should mention the exact table here. Is it in the online appendix? WOuld be also interested to respond to the other question.

src/_gettsim_tests/test_data/renten_alter.csv
TeBackh: ltsray,ltsram,wosray and wosram are original column names from the seibold table. (long term insured retirement age year, long term insured retirement age month, womens retiremen age year...)
I do not understand how you identify the type of pension with the pflichtbeitrag_m. (Ususally people wouldnt pay any contributions while being in pension - the exceptions are probably not build in yet) I guess this needs some comment.

src/_gettsim_tests/test_renten_alter.py
hmgaudecker: Fine, I guess that is how I meant "all years".

Just to be sure: Is there really nothing here that changed late in life beyond "Rente mit 63"? I.e., up to some year a birth cohort was expecting something that was changed later?

src/_gettsim_tests/test_data/renten_alter.csv
LauraGergeleit: [ret_testing_2(1).csv](https://github.com/iza-institute-of-labor-economics/gettsim/files/10447412/ret_testing_2.1.csv)
This is how the original file looks like. The retirement ages for the three different pension types are given in separate columns with the year and month of birth in the first two columns. 
Regelaltersgrenze: regsray (year) and regsram (month)
Altersgrenze für langjährig Versicherte: ltsray and ltsram
Altersrente für Frauen: wosray and wosram

I didn't know how to test all three types for each of the cohorts, since only the lowest is considered in `_ges_rente_altersgrenze_abschlagsfrei` in Gettsim. 

So I copied ltsray and ltsram & wosray and wosram, so that all retirement ages are stored as a separate testcase. 
I hope that makes it clearer :)

src/_gettsim_tests/test_renten_alter.py
ChristianZimpelmann: Thank you! Should add that here.

src/_gettsim_tests/test_data/renten_alter.csv
ChristianZimpelmann: Couldn't you test the variables: `ges_rente_regelaltersgrenze`, `ges_rente_frauen_altersgrenze`, `_ges_rente_langj_altersgrenze`?

I think we should definitly stick to the same data format as the original table. Let me know if you need to transform it in some way and get stuck.

hmgaudecker: Apologies, cannot check right now, but don't we have the original variables before comparing and taking the lowest?

LauraGergeleit: Alrighty, I will have another try during the next days

TeBackh: how does this link to the "pflichtbeitrag_m."

LauraGergeleit: In my approach, "pflichtbeitrag_m." ensured that the correct pension type was calculated and tested. But in the new approach, it won't be needed anylonger

src/_gettsim_tests/test_renten_alter.py
TeBackh: well, that this something which is not considered in the functions yet

TeBackh: all rules are established as they are known now

TeBackh: they do work thereafter for the columns we considered. seibold's sample period is up to year 2014. therefore I guess we can be sure that all the retirement ages relevant up to this date are precise, for everything beyond that, they should be checked (especially for reforms that came up later). I checked for the three columns that we use now. 
We might be able to use more columns for all years. some of the remaining columns refer to pension types that -  to my knowledge - are not implemented yet.

ChristianZimpelmann: > We should open an issue

Will do

src/_gettsim_tests/test_data/renten_alter.csv
TeBackh: > Apologies, cannot check right now, but don't we have the original variables before comparing and taking the lowest?

yeah, we have: ``` _ges_rente_langj_altersgrenze ``` , ```ges_rente_frauen_altersgrenze```

LauraGergeleit:
Now, the original data file (also including "Besonders langjährig Versicherte") is added and I adjusted the test, so that the four different pension types are tested separately. However, there is still an error when it comes to "Besonders langjährig Versicherte". In GETTSIM it is implemented that the pension with 63 starts from cohort 1951. In the data from the paper however, it starts from cohort 1947. @TeBackh, do you know what is correct?

TeBackh:
> In GETTSIM it is implemented that the pension with 63 starts from cohort 1951. In the data from the paper however, it starts from cohort 1947. @TeBackh, do you know what is correct?

Well, so the "Rente mit 63" was introduced in July 2014 - such that the first full cohort that could go with 63, was the cohort 1952. Any previous generations who were not in pension at that date could apply for this pension type too, but they were already older than 63 by then. therefore it was only really relevant for cohort 1951, given the alternative option "pension for long term insured".

TeBackh:
I will check again if cohorts before were actually able to claim the rente mit 63, and then adjust the yaml accordingly -  we could save the first piece of the function if just the starting date was relevant.

hmgaudecker:
Hi, just about ready to go for 0.6 -- a couple of things I am not sure about with respect to this PR, most importantly whether it only updates tests or changes behaviour of functions? 

If it only updates tests: Is it worth to wait for it in releasing 0.6?

If it changes behaviour:
- Someone please adjust the PR title appropriately
- Can someone have a look what is going wrong at this moment?
- How long do you expect a fix to take?

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -0,0 +1,145 @@
...
+    out["m_pflichtbeitrag"] = 540
+
+    # assuming the months of compulsory contributions since the years of 40 required
+    # for the pension for women
+    out["y_pflichtbeitr_ab_40"] = 10
+
+    # after the pension for women was abolished
+    out.loc[
+        np.isnan(out["ges_rente_frauen_altersgrenze"]), "ges_rente_frauen_altersgrenze"
+    ] = out["ges_rente_regelaltersgrenze"]
```
TeBackh, line 123: this looks like a strange thing to do here. better woul be to test only in years when it was not abolished.

src/_gettsim_tests/test_renten_alter.py
```diff
@@ -0,0 +1,145 @@
...
+        "vltsram": "Besonders_Langjährig_Monat",
+    }
+    out = original_data[[*var_names]].rename(columns=var_names).copy()
+
+    # Create IDs
+    out["tu_id"] = out["p_id"]
+    out["hh_id"] = out["p_id"]
+
+    # Create variables needed in GETTSIM
+    out["jahr"] = 2010
```
TeBackh, line 82: Not sure if  I am understanding the test logic correctly but if we ware giving it the year 2010 and the very long term insured threshold 63 for cohort 1947 then there should be a bug because back then it was still 65 -- and that is what Gettsim would tell us. If we instead put 2015 - the 63 is valid and should work.

TeBackh (changes requested):

src/_gettsim_tests/test_renten_alter.py
LauraGergeleit: It didn't seem so strange to me, as the piecewise linear function in `altersgrenze_für_frauen_abschlagsfrei:` in the yaml-file `ges_rente`, actually also includes this jump back to the Regelaltersgrenze

src/_gettsim_tests/test_renten_alter.py
LauraGergeleit: I checked whether it works for 2015, however, the same tests keep failing. 
In the original test file there are no values given for cohorts born before 1947. The problem is that GETTSIM calculates 63 for those cohorts when running the tests for years 2015 and later, while when running the tests for the years before 2015 GETTSIM calculates 65. So I don't really know to what value I should set the input, since I don't want to change the original file manually

ChristianZimpelmann: This column should not have an effect. It is also no longer required by GETTSIM recently. So we can delete the line of code. The used policy parameters and functions do only depend on the date given to `set_up_policy_environment`

```suggestion
```

TeBackh: > 
Isn't there a possibility that the tests based on this data column are only run for 2015 onwards?

ChristianZimpelmann: Which data column do you mean exactly? For most of the other test cases the input data contains a column "jahr" and each policy-year is only checked for the rows in the input data with the respective year

src/_gettsim_tests/test_renten_alter.py
hmgaudecker: Referencing #491 to be able to go back to this discussion.

src/_gettsim_tests/test_renten_alter.py
TeBackh: vltsray and vltsram

hmgaudecker: But we could just map that to the year passed to setup_policy_environment and do not need a column in the data, right?

TeBackh: yeah, thats basically what I though of , to avoid manipulating the input data file

hmgaudecker: That is, it is the same for everybody, or am I mistaken?

TeBackh: ?

hmgaudecker:
Okay, so IIUC, the test data is coded by birth year and the thresholds contained in it show whatever was relevant for the respective cohort in 2010. Correct?

So we should only test for the policy year 2010, or am I missing something?

hmgaudecker:
But the test data seem weird for that interpretation. Running only for the year 2010 and checking differences between GETTSIM and the test data, I get:

|     |   p_id |   geburtsjahr |   geburtsmonat |   _ges_rente_besond_langj_altersgrenze |
|----:|-------:|--------------:|---------------:|---------------------------------------:|
| 564 |    565 |          1947 |              1 |                                63      |
| 565 |    566 |          1947 |              2 |                                63      |
| 566 |    567 |          1947 |              3 |                                63      |
| 567 |    568 |          1947 |              4 |                                63      |
| 568 |    569 |          1947 |              5 |                                63      |
| ... | ... | ... | ... | ... |
| 763 |    764 |          1963 |              8 |                                64.8333 |
| 764 |    765 |          1963 |              9 |                                64.8333 |
| 765 |    766 |          1963 |             10 |                                64.8333 |
| 766 |    767 |          1963 |             11 |                                64.8333 |
| 767 |    768 |          1963 |             12 |                                64.8333 |

whereas GETTSIM sets all of them to 65, which seems correct in the year 2010? Plus, 63 for cohort 1947 can't be right, can it?

What are the test data supposed to represent *precisely*?

src/_gettsim_tests/test_renten_alter.py
hmgaudecker: Sorry, our two messages crossed there. We just do not need that variable, I believe. Unless I am misinterpreting what Arthur's dataset is supposed to do, see below. https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#issuecomment-1406520901

Let's continue down there.

TeBackh:
> But the test data seem weird for that interpretation. Running only for the year 2010 and checking differences between GETTSIM and the test data, I get:
> 	p_id 	geburtsjahr 	geburtsmonat 	_ges_rente_besond_langj_altersgrenze
> 564 	565 	1947 	1 	63
> 565 	566 	1947 	2 	63
> 566 	567 	1947 	3 	63
> 567 	568 	1947 	4 	63
> 568 	569 	1947 	5 	63
> ... 	... 	... 	... 	...
> 763 	764 	1963 	8 	64.8333
> 764 	765 	1963 	9 	64.8333
> 765 	766 	1963 	10 	64.8333
> 766 	767 	1963 	11 	64.8333
> 767 	768 	1963 	12 	64.8333
> 
> whereas GETTSIM sets all of them to 65, which seems correct in the year 2010? Plus, 63 for cohort 1947 can't be right, can it?
> 
> What are the test data supposed to represent _precisely_?

The deduction free retirement thresholds by cohort that are valid since 1.7.2014

hmgaudecker:
Cool, thanks. So in my view, it does not make sense to test based on these data before 2014 then. We can leave tests after that as a reality check. 

To check where things are off for 2014 and later, a relevant excerpt is this one:
```py
year_data[[
    "geburtsjahr",
    "geburtsmonat",
    "_ges_rente_besond_langj_altersgrenze",
    "_ges_rente_besond_langj_altersgrenze_berechnet"
]].query(
    "geburtsmonat == 1 & geburtsjahr in [1945, 1946, 1947, 1948, 1949, 1950, 1951, 1952, 1953]"
)
```
which yields:

|     |   geburtsjahr |   geburtsmonat |   _ges_rente_besond_langj_altersgrenze |   _ges_rente_besond_langj_altersgrenze_berechnet |
|----:|--------------:|---------------:|---------------------------------------:|-------------------------------------------------:|
| 540 |          1945 |              1 |                                65      |                                          63      |
| 552 |          1946 |              1 |                                65      |                                          63      |
| 564 |          1947 |              1 |                                63      |                                          63      |
| 576 |          1948 |              1 |                                63      |                                          63      |
| 588 |          1949 |              1 |                                63      |                                          63      |
| 600 |          1950 |              1 |                                63      |                                          63      |
| 612 |          1951 |              1 |                                63      |                                          63      |
| 624 |          1952 |              1 |                                63      |                                          63      |
| 636 |          1953 |              1 |                                63.1667 |                                          63.1667 |


My view: Setting this age to 63 does not make sense for people born 06/1951 and earlier, either in the test data or in our own calculations.

Instead, it should be (pseudocode):
```py
min(("2014-07-01" - "date of birth").year, 65)
```
since before the introduction of the reform, it was 65 and effective on that date, they could apply for their pension.

Thoughts?

TeBackh:
> Cool, thanks. So in my view, it does not make sense to test based on these data before 2014 then. We can leave tests after that as a reality check.
> 
> To check where things are off for 2014 and later, a relevant excerpt is this one:
> 
> ```python
> year_data[[
>     "geburtsjahr",
>     "geburtsmonat",
>     "_ges_rente_besond_langj_altersgrenze",
>     "_ges_rente_besond_langj_altersgrenze_berechnet"
> ]].query(
>     "geburtsmonat == 1 & geburtsjahr in [1945, 1946, 1947, 1948, 1949, 1950, 1951, 1952, 1953]"
> )
> ```
> 
> which yields:
> 	geburtsjahr 	geburtsmonat 	_ges_rente_besond_langj_altersgrenze 	_ges_rente_besond_langj_altersgrenze_berechnet
> 540 	1945 	1 	65 	63
> 552 	1946 	1 	65 	63
> 564 	1947 	1 	63 	63
> 576 	1948 	1 	63 	63
> 588 	1949 	1 	63 	63
> 600 	1950 	1 	63 	63
> 612 	1951 	1 	63 	63
> 624 	1952 	1 	63 	63
> 636 	1953 	1 	63.1667 	63.1667
> 
> My view: Setting this age to 63 does not make sense for people born 06/1951 and earlier, either in the test data or in our own calculations.
> 
> Instead, it should be (pseudocode):
> 
> ```python
> min(("2014-07-01" - "date of birth"), 65.0)
> ```
> 
> with the first bit appropriately converted to years. IIUC, before the introduction of the reform, it was 65 and effective on that date, they could apply for their pension.
> 
> Thoughts?

Well, people born after 1.7.1949 did benefit from the reform since they could retire earlier -  not with 63, since they were already older but with 64.something. We could in principle delete 1947 and 1948  but we did not want to manipulate the input data. In general I do not see the point of excluding cohorts, since the input parameters and the policy parameters in the yaml agree for the years after 2014 for all cohorts. So we just need to make sure that the tests on these input columns run only for year 2015 onwards.

hmgaudecker:
> Well, people born after 1.7.1949 did benefit from the reform since they could retire earlier - not with 63, since they were already older but with 64.something. 

Exactly, hence I think the formula in my previous post should apply.

> We could in principle delete 1947 and 1948 but we did not want to manipulate the input data. 

Well, the input data are not official numbers but stuff used for a (well-published) paper. So a good orientation, but not "the truth". And I think they are wrong for 1947-1948, probably because Arthur did not need to worry about that case in the paper.

> In general I do not see the point of excluding cohorts, since the input parameters and the policy parameters in the yaml agree for the years after 2014 for all cohorts. So we just need to make sure that the tests on these input columns run only for year 2015 onwards.

That is 2014-07-01 onwards, right? Which leaves the discrepancies above, they happen in 2015, too. No differences present for younger cohorts, of course.

TeBackh:
> > Well, people born after 1.7.1949 did benefit from the reform since they could retire earlier - not with 63, since they were already older but with 64.something.
> 
> Exactly, hence I think the formula in my previous post should apply.
> 
> > We could in principle delete 1947 and 1948 but we did not want to manipulate the input data.
> 
> Well, the input data are not official numbers but stuff used for a (well-published) paper. So a good orientation, but not "the truth". And I think they are wrong for 1947 - June 1951, probably because Arthur did not need to worry about that case in the paper.
> 
Ok, I dont have strong opinions on this. The idea was to keep as transparent as possible where the input data comes from by keeping the original input as is, cite the source, and do any required adjustments via the functions in the tests.  

> > In general I do not see the point of excluding cohorts, since the input parameters and the policy parameters in the yaml agree for the years after 2014 for all cohorts. So we just need to make sure that the tests on these input columns run only for year 2015 onwards.
> 
> That is 2014-07-01 onwards, right? Which leaves the discrepancies above, they happen in 2015, too. No differences present for younger cohorts, of course.

Yes. Thought tests are run only on whole years, which translates to 1.1.2014 and 1.1.2015.

hmgaudecker:
Yeah, that is the correct approach in principle -- but for some edge cases it showed where things are off.

The date in these tests is actually set to July 1 of each year:
```py
set_up_policy_environment(date=f"{year}-07-01")
```

hmgaudecker:
So IIUC, a minimal thing of what should be done is to change the test data for the cohorts pre-07/1951 as suggested [above](https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#issuecomment-1406586690)

```python
min(("2014-07-01" - "date of birth"), 65.0)
```

Thanks!

LauraGergeleit:
> So IIUC, a minimal thing of what should be done is to change the test data for the cohorts pre-07/1951 as suggested [above](https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#issuecomment-1406586690)
> 
> ```python
> min(("2014-07-01" - "date of birth"), 65.0)
> ```
> 
> Thanks!

I just tried to implement it that way, so the input data should be adjusted for those cohorts. However, the tests still fail since GETTSIM calculates 63 for `_ges_rente_besond_langj_altersgrenze` for all cohorts born before 1952. So it seems to me that there is still an error in the GETTSIM implementation or do I miss something?

TeBackh:
> > So IIUC, a minimal thing of what should be done is to change the test data for the cohorts pre-07/1951 as suggested [above](https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#issuecomment-1406586690)
> > ```python
> > min(("2014-07-01" - "date of birth"), 65.0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thanks!
> 
> I just tried to implement it that way, so the input data should be adjusted for those cohorts. However, the tests still fail since GETTSIM calculates 63 for `_ges_rente_besond_langj_altersgrenze` for all cohorts born before 1952. So it seems to me that there is still an error in the GETTSIM implementation or do I miss something?



> > So IIUC, a minimal thing of what should be done is to change the test data for the cohorts pre-07/1951 as suggested [above](https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#issuecomment-1406586690)
> > ```python
> > min(("2014-07-01" - "date of birth"), 65.0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thanks!
> 
> I just tried to implement it that way, so the input data should be adjusted for those cohorts. However, the tests still fail since GETTSIM calculates 63 for `_ges_rente_besond_langj_altersgrenze` for all cohorts born before 1952. So it seems to me that there is still an error in the GETTSIM implementation or do I miss something?

Why does it fail then? 63 is also what the Seibold data said.

hmgaudecker:
Yes, but we decided that what those data said is wrong, right? 🙃

TeBackh:
> Yes, but we decided that what those data said is wrong, right? 🙃

Not precisely, actually. after the reform it was 63 by law, just not in practice. We are running in circles with this and we should actually discuss, which kind of outputs we want gettsim to produce. 

What the current implementation can do is to tell you for a given date whether a person with a certain age was eligible for a cetrain pension type or not. and it tells you whether she is eligible for **any** pension. and then it would calculate the pension payment, because the implementation is designed for calculating transfers for a certain period.

What it is not made for, is to output the lowest possible pension age. (If we want to implement such feature , we should write a function that interprets the inputs correctly and translates what these mean in practice for the person.)

So probably we should implement tests on the eligibility or so to be in line with the aim of the functions.

hmgaudecker:
Ahhh, great, thanks for clearing up the misunderstanding!!! 

I was actually thinking along other lines than you all along. Less because I would think people want to know the lowest possible age for some type of pension, but mostly for the following example.

- policy year 2015
- would like to calculate the pension payment for somebody 
  - born 1949
  - X Entgeltpunkte
  - retirement entry age 63
  - 46 eligible contribution years

I think the current implementation would (wrongly?) return pension payments without deductions relative to old age pension as per "Rente mit 63"?

The alternative would be to run GETTSIM for each year starting at that person's retirement and keep the same pension payment. But then we might run into trouble with other reforms, e.g., "Mütterrente", which changed attribution period also for pensioners IIRC?

TeBackh:
> Ahhh, great, thanks for clearing up the misunderstanding!!!
> 
> I was actually thinking along other lines than you all along. Less because I would think people want to know the lowest possible age for some type of pension, but mostly for the following example.
> 
>     * policy year 2015
> 
>     * would like to calculate the pension payment for somebody
>       
>       * born 1949
>       * X Entgeltpunkte
>       * retirement entry age 63
>       * 46 eligible contribution years
> 
> 
> I think the current implementation would (wrongly?) return pension payments without deductions relative to old age pension as per "Rente mit 63"?
> 
ok, i was mislead by the fact that the retirement-entry-year (not age) enters the zugangsfaktor function but inside it runs on age, you are right. This is some logical  mistake that happend in PR #494 - apparently our discussion there was not sufficient back then. So I open an issue to fix the fix.

Wondring whether we should place some warning about this somewhere in the yaml file (allowing for comments would help ;-) )

> The alternative would be to run GETTSIM for each year starting at that person's retirement and keep the same pension payment. But then we might run into trouble with other reforms, e.g., "Mütterrente", which changed attribution period also for pensioners IIRC?

hmgaudecker:
> Wondring whether we should place some warning about this somewhere in the yaml file (allowing for comments would help ;-) )

Just add as part of the description, e.g., there is some of that in the changes of #494 (see [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/494/files), for whatever reason only in the English version, should update for German). Just update as you see fit!

Great that we cleared that up!

TeBackh (approved):

Issue #430, opened 2022-11-16
Make sure test data has the correct dtype


hmgaudecker:
There are currently lots of conversions of the data used during tests. Below is the output from running the test suite on the `use_dags` branch. This makes it hard to find the actual test output.

Solution should be to just convert one or more integers per column (say, a 14) to floats (14.0) in the csv test data.

```
gettsim/tests/test_arbeitsl_geld.py::test_arbeitsl_geld[2010]
gettsim/tests/test_arbeitsl_geld.py::test_arbeitsl_geld[2011]
gettsim/tests/test_arbeitsl_geld.py::test_arbeitsl_geld[2015]
gettsim/tests/test_arbeitsl_geld.py::test_arbeitsl_geld[2019]
  /home/hmg/econ/gettsim/gettsim/gettsim/interface.py:468: UserWarning: The data types of the following input variables have been converted: 
  
   - bruttolohn_vorj_m from int64 to float
   - sum_ges_rente_priv_rente_m from int64 to float
   - arbeitsstunden_w from int64 to float
  
  Note that the automatic conversion of data types is unsafe and that its correctness cannot be guaranteed. The best solution is to convert all columns to the expected data types yourself.
    warnings.warn("\n".join(collected_conversions) + "\n" + "\n" + general_warning)

gettsim/tests/test_arbeitsl_geld_2.py: 72 warnings
  /home/hmg/econ/gettsim/gettsim/gettsim/interface.py:468: UserWarning: The data types of the following input variables have been converted: 
  
   - bruttokaltmiete_m_hh from int64 to float
   - heizkosten_m_hh from int64 to float
   - bruttolohn_m from int64 to float
   - sum_ges_rente_priv_rente_m from int64 to float
   - kapitaleink_brutto_m from int64 to float
   - arbeitsl_geld_m from int64 to float
   - sonstig_eink_m from int64 to float
   - eink_selbst_m from int64 to float
   - eink_vermietung_m from int64 to float
   - eink_st_tu from int64 to float
   - soli_st_tu from int64 to float
   - sozialv_beitr_gesamt_m from int64 to float
   - kindergeld_m_hh from int64 to float
   - unterhaltsvors_m from int64 to float
   - elterngeld_m from int64 to float
   - wohngeld_vor_vermög_check_m_hh from int64 to float
   - vermögen_bedürft_hh from int64 to float
   - arbeitsstunden_w from int64 to float
  
[...]

Issue #282, opened 2021-03-29
Restructure test_data


ChristianZimpelmann:
### Current and desired situation

Input data for tests in `tests/test_data` are currently defined in .csv and .ods files. 

Only one data format, which does not allow for formulas, but can be labelled

### Proposed implementation

yaml

### Considered alternatives

DataFrames in pickle

hmgaudecker:
In particular, think about structured way of including metadata on the source of the test data:

> Specifically, I would like to be able to document what is the exact source for a test case, which variables and values are given in that source and which variables are manually calculated or assumed to bring these external examples into shape for GETTSIM.

ChristianZimpelmann:
@LauraGergeleit and me were trying around how we could represent tests in a `yaml`-file.

One test case (for two individuals in the same household) could look as follows:

```
2017:
-   individuals:
    -   inputs:
            alleinerziehend: false
            alter: 72
            arbeitsl_geld_m: 0
            bewohnt_eigentum_hh: false
            bruttokaltmiete_m_hh: 460.0
            bruttolohn_m: 0
            eink_selbst_m: 0
            eink_st_tu: 0
            elterngeld_m: 0
            grundrentenzeiten: 300
            heizkosten_m_hh: 60.0
            kapital_eink_m: 0
            kind: false
            kindergeld_m_hh: 0
            prv_rente_m: 0
            rentner: true
            schwerbe_ausweis_g: false
            soli_st_tu: 0
            sonstig_eink_m: 0
            sozialv_beitr_m: 0
            staatl_rente_m: 860.6
            unterhaltsvors_m: 0
            vermiet_eink_m: 0
            vermögen_hh: 0
            wohnfläche_hh: 60
        outputs:
            grunds_im_alter_m_hh: 322.0
            regelbedarf_m_grunds_im_alter_vermögens_check_hh: 1256.0
    -   inputs:
            alleinerziehend: false
            alter: 67
            arbeitsl_geld_m: 0
            bewohnt_eigentum_hh: false
            bruttokaltmiete_m_hh: 460.0
            bruttolohn_m: 0
            eink_selbst_m: 0
            eink_st_tu: 0
            elterngeld_m: 0
            grundrentenzeiten: 48
            heizkosten_m_hh: 60.0
            kapital_eink_m: 0
            kind: false
            kindergeld_m_hh: 0
            prv_rente_m: 0
            rentner: true
            schwerbe_ausweis_g: false
[...]

PR #431, opened 2022-11-17
Adjusting test data to the correct data type

LauraGergeleit:

Addressing #430 by adjusting data type of test data in csv-files. 
Note: the respective columns have to be imported as Text-Datatype, when re-importing the csv-file to Excel, to ensure that the format remains the same.


ChristianZimpelmann:
Thanks!

> Note: the respective columns have to be imported as Text-Datatype, when re-importing the csv-file to Excel, to ensure that the format remains the same.

I am afraid future users will forget about this and sooner or later the formatting of numbers as floats or integers respectively will be lost again if we only change the .csv-file.

Two alternatives would be:
- to run `_convert_data_to_correct_types` after importing the .csv in each test. (with a new keyword-argument which disables warnings)
- to filter out these kind of warnings when running tests.

Note, that we won't have this problem anymore once we addressed #62 . So we just need a temporary fix for now. 

@hmgaudecker, what do you think?

hmgaudecker:
We should just tackle #62 then (@LauraGergeleit ?). Make a note in the PR that this will be addressed by the fix, too

ChristianZimpelmann:
Note: I linked an old, already closed issue, but adjusted it now in the messages above: the correct one is #282.

Let's continue the discussion over there.

hmgaudecker:
Looks already much better! Helped me spot a pending deprecation we should tackle: https://github.com/iza-institute-of-labor-economics/gettsim/issues/468

As a New Year's Resolution, let's work on keeping PRs shorter. So just continue with the (few) remaining conversion warnings as you did above and tackle the yaml conversion in a subsequent issue. Having the correct types in the csv should actually be helpful  for that. And let's get started soon enough so that not too many people have a chance to commit wrong data types :wink:

LauraGergeleit:
> As a New Year's Resolution, let's work on keeping PRs shorter. So just continue with the (few) remaining conversion warnings as you did above and tackle the yaml conversion in a subsequent issue. Having the correct types in the csv should actually be helpful for that. And let's get started soon enough so that not too many people have a chance to commit wrong data types 😉

I adjusted the data types for the remaining csv-files and will tackle the yaml conversion in a new PR soon

hmgaudecker:
Beautiful, thanks a lot!!! 

Immediately you start seeing other stuff when the test suite runs through, see #485

PR #442, opened 2022-11-30
Add parameter updates EStG, SolzG1995, BKGG 2022/23/24

mjbloemer:
- [x] Add parameter updates from Steuerentlastungsgesetz 2022 ([G. v. 23.05.2022 BGBl. I S. 749](https://offenegesetze.de/veroeffentlichung/bgbl1/2022/17#page=5)).
- [x] Add Inflationsausgleichsgesetz 2022, ([G. v. 08.12.2022 BGBl. I S. 2230](https://offenegesetze.de/veroeffentlichung/bgbl1/2022/49#page=2))
- [x] [Jahressteuergesetzes 2022](https://dserver.bundestag.de/btd/20/038/2003879.pdf) (Drucksache 20/3879), with another update which can be found in [Beschlussempfehlung und Bericht des Finanzausschusses](https://dserver.bundestag.de/btd/20/047/2004729.pdf) (Drucksache 20/4729)

Not addressed:
- soli `1: upper_threshold` "parameter"?


spreet41 (commented):
m.M.n. sind die upper thresholds im Steuertarif für das Jahr 2023 jeweils um einen 1€ zu hoch. 
Vgl. dazu [Drucksache 20/4378](https://dserver.bundestag.de/btd/20/043/2004378.pdf) auf S.9.
Oder übersehe ich hier etwas?

mjbloemer:
> m.M.n. sind die upper thresholds im Steuertarif für das Jahr 2023 jeweils um einen 1€ zu hoch. 

I looked back at the previous parameters and it seemed to me the staring values for the next interval were used, e.g. "von 10909 Euro bis 15999 Euro".

There is a rounding for zvE, so you don't have to worry about the undefined parts of the tax function :)

gettsim/parameters/kinderzuschl.yaml
```diff
@@ -20,6 +20,14 @@ maximum:
...
     reference: § 6a (2) BKGG
     note: Ab 2021 wird der Betrag aus anderen Parametern berechnet.
+  2023-01-01:
+    scalar: 250
+    reference: Art. 6 Inflationsausgleichsgesetz. Bundesrat Drucksache 576/22. Noch nicht im BGBl. veröffentlicht.
+    note: Regelung für das Kalenderjahr 2023, abweichend von § 6a Absatz 2.
+  2024-01-01:
+    scalar: null
+    reference: § 6a (2) BKGG
+    note: Ab 2021 wird der Betrag aus anderen Parametern berechnet.
```
ChristianZimpelmann, line 30:
```suggestion
    note: Ab 2024 wird der Betrag wieder aus anderen Parametern berechnet.
```

gettsim/policy_environment.py
```diff
@@ -258,6 +258,8 @@ def _parse_kinderzuschl_max(date, params):
     Since 2021, this measure has been derived from subsistence
```
ChristianZimpelmann, line 258: Probably more precise:

`In 2021, 2022, and from 2024 on, this measure ....`

gettsim/policy_environment.py
```diff
@@ -272,7 +274,7 @@ def _parse_kinderzuschl_max(date, params):
 
     """
 
-    if date.year >= 2021:
+    if date.year >= 2024 and 2023 > date.year >= 2021:
```
ChristianZimpelmann, line 277:
```suggestion
    if (date.year >= 2024) or (2023 > date.year >= 2021) :
```

Brackets just for clarity. But the `or` instead of `and` is important.

gettsim/policy_environment.py
```diff
@@ -272,7 +274,7 @@ def _parse_kinderzuschl_max(date, params):
...
         assert {"kinderzuschl", "kindergeld"} <= params.keys()
         params["kinderzuschl"]["maximum"] = (
             params["kinderzuschl"]["existenzminimum"]["regelsatz"]["kinder"]
```
ChristianZimpelmann, line 280: Problem is that we currently don't have the parameters `params["kinderzuschl"]["existenzminimum"]` implemented for now. I assume they do not exist yet? Then we should at least make a `ToDo:` note in the docstring of the function `_parse_kinderzuschl_max`

ChristianZimpelmann (commented):
Great work! Thanks!

A couple of remarks are below. 

Additionally: It would be important to have test cases for those changes eventually -- even better already in this PR. Do you know whether any of the tax calculators etc. are already updated or are you aware of any new examples by ministries?

gettsim/policy_environment.py
mjbloemer: Are those not the parameters in https://github.com/iza-institute-of-labor-economics/gettsim/blob/4379a90b5639e64628a5cb14cd411b9225b12240/gettsim/parameters/kinderzuschl.yaml#L89 ? They might be outdated for 2023 and 2024. I could update them as part of this PR (but would prefer to do so in a separate PR). For KiZ they should only become relevant (again) in 2024 (we will see, probably not).

mjbloemer: I updated these parameters in #446. But I don't think this PR really depends on that being done first.

hmgaudecker:
@LauraGergeleit, this last thing is great! 

Only thing is that I think it merits its own PR and a serious renaming of the parameters (we can use `piecewise_polynomial` internally but should mostly hide it from the parameters file).

My suggestion would be to just add the resulting number in this PR and to open a new one, which will close #444 and where we can discuss the details?

ChristianZimpelmann (approved):
Looks great now. 

Thank you, @mjbloemer and @LauraGergeleit!

Issue #270, opened 2021-02-12
Variable alleinerziehend unclear


ChristianZimpelmann:
### Bug description

The code expects `alleinerziehend == True` for children of single parents. See code in `unterhaltsvors_m`:

```
    conditions = [
        (alter < 6) & alleinerziehend,
        (alter >= 6) & (alter < 12) & alleinerziehend,
        # Older kids get it only if the parent has income > 600€.
        (alter >= 12)
        & (alter < 18)
        & alleinerziehend
        & (
            unterhaltsvorschuss_eink
            > unterhalt_params["unterhaltsvorschuss_mindesteinkommen"]
        ),
    ]
```
However, the variable description "Dummy: Single parent" would imply that alleinerziehend is only true for the single parent.

### Proposed Change

Change variable desciption to "Dummy: Single parent household". And rename variable accordingly?

hmgaudecker:
Phew, good catch. This may require some thought to get right. @Eric-Sommer @mjbloemer @stichnoth: Is `alleinerziehend` always a household-level concept? I would have thought that when a single parent lives in a flat share, it should not apply to all other members.

If that is correct, I would think of it as a characteristic of the parent-child link. The only question is how to include that, will need to work on #147 soonish, anyhow, though.

Eric-Sommer:
My instant thought was that `alleinerziehend` is a personal concept (there is no `hh`), and you would need to check for the maximum value within the tax unit (e.g., for income tax stuff) *or* the household (e.g., for ALG 2). At the very least, we should make sure this is consistently handled as a personal concept across functions.

hmgaudecker:
At least internally, we would need to do the max checking upfront, far too
expensive to do within many functions.

So how does it work for ALG 2 in the example I made up? Or, maybe more
common, a three gen HH with the middle gen a single mother? I would have
thought that she still counts as "alleinerziehend", but grandparents would
not?

Eric-Sommer <notifications@github.com> schrieb am Fr., 12. Feb. 2021, 08:56:

Eric-Sommer:
> Or, maybe more common, a three gen HH with the middle gen a single mother? I would have thought that she still counts as "alleinerziehend", but grandparents would not?

[**§ 21 SGB II**](https://www.buzer.de/gesetz/2602/a37306.htm)
[...]

Issue #402, opened 2022-06-27
Check for Erwachsene in Household


JHermann99:
### Current situation

When there is no adult in a household, ergo only children, a error is produced by _kinderzuschl_eink.py_ --> _kinderzuschl_eink_regel_m_hh_ab_2011_ since this function requires the _anz_erwachsene_hh_ per household to be at least one. 

### Proposed implementation

I recommend introducing a check that there is at least one adult per household, as the absence of adult might also cause unexpected or unwanted results by other functions. It also does not make sense to have a dependant child without a corresponding adult.

ChristianZimpelmann:
I would like to address this issue more generally and add a bunch of tests on input data in a structured way. 

The following cases come to mind:
- restriction of values of individual variables (e.g. non-negative income, categorical values only certain values, mietstufe)
- restrictions on hh-level (e.g. the one mentioned above)
- consistiency checks between variables (e.g. `y_pflichtbeitr_ab_40 < alter`, `if bewohnt_eigentum_hh: bruttokaltmiete_m_hh == 0`)

hmgaudecker:
Cool, I like that. Fail early, fail often.

MImmesberger:
I updated the issue text to reflect the current state of GETTSIM. This one can serve as the parent issue for the other consistency check issues.

PR #448, opened 2022-12-02
Fix some smaller problems with respect to types and `alleinerz`

ChristianZimpelmann:
### Todo

- [x] Set type of `wohnfläche_hh` to ``float``.
- [x] Make sure data created with `synthetic_data` has types expected by GETTSIM.
- [x] Adjust calculation for Unterhaltsvorschuss (and respective tests) such that only the single parent is defined as 'alleinerz' and not the children. (this is related to #270)


gettsim/tests/test_data/unterhalt.csv
```diff
@@ -1,8 +1,8 @@
 hh_id,tu_id,p_id,alleinerz,alter,bruttolohn_m,eink_selbst_m,sonstig_eink_m,arbeitsl_geld_m,sum_ges_rente_priv_rente_m,kapitaleink_brutto_m,eink_vermietung_m,jahr,monat,,,unterhaltsvors_m,,,,,These test cases wer just to check that we return a Series of nans. We exclude this test case for now.,,,,,,,,,,,,,
 5,5,6,TRUE,45,0,0,0,0,0,0,0,2018,1,,,0,,,,,1,1,23,FALSE,40,0,0,0,0,0,0,0,FALSE,2017
-5,5,7,TRUE,10,0,0,0,0,0,0,0,2018,1,,,205,,,,,1,1,25,FALSE,35,0,0,0,0,0,0,0,FALSE,2017
+5,5,7,FALSE,10,0,0,0,0,0,0,0,2018,1,,,205,,,,,1,1,25,FALSE,35,0,0,0,0,0,0,0,FALSE,2017
```
LauraGergeleit, line 3: There should be other tests for which `alleinerz` for children is still evaluated as `True`, for example in `arbeitsl_geld_2.csv`

LauraGergeleit (commented):
Looks good! One comment with respect to the tests...

hmgaudecker:
Random thought while browsing issues: Is it useful to add a check as proposed in #402 here? While  you are looking at those testcases anyhow...

ChristianZimpelmann:
> Random thought while browsing issues: Is it useful to add a check as proposed in https://github.com/iza-institute-of-labor-economics/gettsim/issues/402 here? While you are looking at those testcases anyhow...

Instead of adding it here, I would like to put some deeper thought into how to specify these kind of checks next year (see my new comment in #402)

LauraGergeleit (approved):
Looks good!

PR #450, opened 2022-12-03
Update CI, use modern package structure

hmgaudecker:

Update the outdated CI pipeline and move to the [src-layout](https://github.com/pytask-dev/cookiecutter-pytask-project#faq), which is the preferred way to structure packages.

### Todo

- [x] Check that conda package actually works
- [x] Decide how strict mypy is supposed to be (currently has lots of failures)
- [x] Run updated pre-commit hooks on all files.
- [x] Target the right branch and pick an appropriate title.


hmgaudecker:
@tobiasraabe, if you get a chance to take a quick look at whether the package is structured reasonably (e.g., I was not sure about test data and parameters, whether those would be included when package is installed etc.). Tests do pass locally.

tobiasraabe:
Hi HM, here are some comments.

- Add check-manifest to pre-commit.
- Manage all package data via Manifest.in.
- Test building package in CI with https://github.com/pytask-dev/pytask/blob/main/.github/workflows/publish-to-pypi.yml.
- Convert changes.rst to markdown.
- In tox.ini under pytest/testpaths add src for doctests
- remove setup.py from coverage settings in tox.ini.

hmgaudecker:
> Hi HM, here are some comments.

Thanks! All great, just a couple of background questions...

> * Add check-manifest to pre-commit.
> * Manage all package data via Manifest.in.
> * remove setup.py from coverage settings in tox.ini.

I thought `setup.py` was not required by "modern" (whatever that means precisely) packaging practices and `setup.cfg` was enough. Nor did I find any in some of the most popular OSE / pytask packages. But it seems to be required by the hook. Would you mind providing some background / what should `setup.py` include then?

> * Test building package in CI with https://github.com/pytask-dev/pytask/blob/main/.github/workflows/publish-to-pypi.yml.

Great, thanks! Will stick to conda packaging here. Does this action serve the same purpose / are you aware of an alternative for that?

> * Convert changes.rst to markdown.

Will convert the entire docs to MyST soon, will do as part of that.

> * In tox.ini under pytest/testpaths add src for doctests

:heavy_check_mark: (locally so far)

tobiasraabe:
You are completely right about `setup.py`. Something was missing in `pyproject.toml`. Now it should work.

I don't know about a similar action for conda.

hmgaudecker (commented):
*(went through all files, just a note so that GitHub remembers, too)*

hmgaudecker:
> I would remove the test directory from the package. I do not think it has a use case. I am not sure what errors should be discovered this way, 

I am mainly worried about users changing functions of GETTSIM proper instead of doing it the supported way. E.g. VS Code makes it very easy for you to jump into imported modules. The temptation might be there to just modify the functions, leading to hard-to-debug errors. Issue template might then include "I ran `gettsim.test()` and everything passed.

> and no one uses it anyways.

Well, all my students do for sure because I prepare the notebooks that way. Creating my own demand, I guess :laughing: 

> Distinguishing between a public `gettsim` and a private `_gettsim` package is especially useful for pytask and pytest because it makes it pretty obvious what the publicly supported sphere is for plugin authors. And it avoids some very nasty circular import errors. In your case, I do not think it is necessary. Do you have a specific situation in mind where it is helpful?

I just saw many packages doing it that way. But somewhat for the same reason as above, I think it will be helpful to very explicitly hide certain details from users. In a way, you can think of the grand scheme of GETTSIM designed by certain smart people as a big plugin system :wink:

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,18 @@
+---
+
+name: Submit Question
+about: Ask a general question about GETTSIM
+title: "QST:"
+labels: "question"
+
+---
+
+#### Question about pytask
```
ChristianZimpelmann, line 10:
```suggestion
#### Question about GETTSIM
```

ChristianZimpelmann (approved):
Looks great! Just found a typo.

This will be super helpful in the future as far as I can judge!

PR #455, opened 2022-12-07
Change Documentation from RST to MyST

mj023:

Change all Documentation Files and the README,CHANGES and COC files from RST to MyST.

### Todo

- [x] Convert Documentation
    - [x] Find way to still reference Github Issues and Users in MyST in CHANGES.md
    - [x] Convert inline link targets to MyST
- [x] Convert Github Facing Documents


hmgaudecker:
Cool! Maybe just search for "rst" in the codebase, seems like tests are failing because one of them is looking for an old README.

And while you are at it, could you add the same docs style as in the templates etc.? Thanks!

mj023:
I did not find a MyST solution for keeping the links from the functions docstrings to the inline link targets in the table in input_variables.rst. I can either keep the file in RST or replace the links in the functions with new ones, that all link to the head of the table instead of directly to the referenced variable. Which option would you prefer?

hmgaudecker:
Can we just use generic https://GETTSIM.rtfd.io/{version}/....... links?
Won't work for dev builds but so be it?

Max Jahn ***@***.***> schrieb am Fr., 9. Dez. 2022, 15:27:

mj023:
I will just put a link target above the table then, it should be possible to link to it from the documentation inside the functions. I was just concerned because people will have to sift through a huge table to find the variable, whereas before it would direct them to the part of the table where the variable is.

hmgaudecker:
Sorry, my bad, I had made the wrong assumption about what you meant.

That solution is perfect.

hmgaudecker:
Maybe hold on a bit before you continue there. We are just thinking about some larger change in the architecture and it might be easier to bulk-adjust the links given the old structure. I would not worry much if these are broken as long as we do not publish a release in that state.

hmgaudecker:
Of course I should have checked seriously earlier today but I was in a rush and only on the phone now.

hmgaudecker (approved):
Great work, @mj023 ! 

There were a couple of smaller things left, I fixed the issue with the table by just making headings out of it. But that was all minor. 

Thanks a lot!

PR #457, opened 2022-12-08
Correction of Arbeitslosengeld 1

paulinaschroeder:

Correct calculation of Arbeitslosengeld according to the law.

### Todo

- [x] add input variables anwartschaftszeit, arbeitssuchend, arbeitsl_geld_m_vorj and arbeitsl_geld_m_v2j
- [x] add new functions to arbeitsl_geld.py to calculate Anspruchsdauer of Arbeitslosengeld
- [x] add parameters for Anspruchsdauer of Arbeitslosengeld
- [ ] answer how to calculate `arbeitsl_geld_m`? (see below)

### How to calculate `arbeitsl_geld_m`?

With this correction GETTSIM now calculates the number of months a person is eligible for ALG 1. Since this can be less than 12 months and `arbeitsl_geld_m` is used as an input variable for income calculations of other transfers I propose to calculate `arbeitsl_geld_m` as the average monthly unemployment benefit a person gets for that year but it might be better to just calculate `arbeitsl_geld_m` as the amount a person gets when they receive ALG 1.


hmgaudecker:
> I propose to calculate `arbeitsl_geld_m` as the average monthly unemployment benefit a person gets for that year but it might be better to just calculate `arbeitsl_geld_m` as the amount a person gets when they receive ALG 1.


Thanks! 

Sounds like the way to go is to add the number of months for which someone is eligible (if we do not have that already) and then use the actual monthly benefit.

docs/gettsim_objects/input_variables.rst
```diff
@@ -146,3 +140,15 @@ household.
 || _`y_pflichtbeitr_ab_40`  || Total years of mandat. contributions after || float       |
 ||                          || age 40                                     ||             |
 +---------------------------+---------------------------------------------+--------------+
+|| _`anwartschaftszeit`     || Months of unemployment insurance           || int         |
+||                          || contributions in the last 30 months        ||             |
```
ChristianZimpelmann, line 144: should be more precisely "in the 30 months before claimining unemployment insurance" if I am not mistaken

docs/gettsim_objects/input_variables.rst
```diff
@@ -146,3 +140,15 @@ household.
...
++---------------------------+---------------------------------------------+--------------+
+|| _`arbeitssuchend`        || Dummy: looking for employment              || bool        |
+||                          ||                                            ||             |
++---------------------------+---------------------------------------------+--------------+
+|| _`arbeitsl_geld_m_vorj`  || Amount of months in which Arbeitslosengeld || int         |
+||                          || was received in previous year              ||             |
++---------------------------+---------------------------------------------+--------------+
+|| _`arbeitsl_geld_m_v2j`   || Amount of months in which Arbeitslosengeld || int         |
```
ChristianZimpelmann, line 152: Can we summarize `arbeitsl_geld_m_v2j` and `arbeitsl_geld_m_vorj` in one parameter? Isn't the relevant variable how long the individual already receives unemployment contribution?

gettsim/parameters/arbeitsl_geld.yaml
```diff
@@ -188,3 +164,41 @@ stundengrenze:
   2012-04-01:
     scalar: 15
     reference: Art. 2 G. v. 20.12.2011 BGBl I S. 2854
+
+anwartschaftszeit:
```
ChristianZimpelmann, line 168: I would rename the variable to `anspruchsdauer` and restructure it as follows:

```yaml
    nach_alter:
      0: 12
      50: 15
      55: 18
      ...
    nach_anwartschafszeit:
      12: 6
      18: 8
      ...
      48: 24
    
```

And then use `piecewise_polynomial` as [here](https://github.com/iza-institute-of-labor-economics/gettsim/blob/d1f7b3ff464a74600892d8d42bb7635e3d32c575/gettsim/transfers/wohngeld.py#L128) for both set of parameters and set anspruchsdauer to the minimum of the two values (let me know if it isn't clear enough what I mean).

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -29,81 +30,143 @@ def arbeitsl_geld_m(
...
+        ):
+            out = arbeitsl_geld_params["anwartschaftszeit"][threshold]["anspruch"]
+        else:
+            out = 0
+
+        return out
+    return out
 
 
 def arbeitsl_geld_berechtigt(
```
ChristianZimpelmann, line 133: The changes in this function change the interpretation of `arbeitsl_geld_berechtigt` as subjects who have exhausted their duration of unemployment benefit will be "berechtigt" for unemployment benefit (just for 0 months only). Is there a specific reason for that? Otherwise, I would stick to the old interpretation that those subjects would be counted as not eligible.

ChristianZimpelmann (changes requested):
Thanks, that is very helpful. But a couple of comments in the code..

Additionally, it will be important to also adjust (and expand) the test cases for ALG I.

ChristianZimpelmann:
With respect to the question "How to calculate arbeitsl_geld_m?":

> Sounds like the way to go is to add the number of months for which someone is eligible (if we do not have that already) and then use the actual monthly benefit.

You mean add as additional **outcome** variable?

This would be also my prefered option: monthly benefits + months of elgibility remaining. And then the user can decide how to aggregate those information.

docs/gettsim_objects/input_variables.rst
paulinaschroeder: Yes, if the input variable is how long the person already receives ALG 1 we can combine them.

gettsim/config.py
```diff
@@ -127,6 +124,10 @@
     "m_kind_berücks_zeit": float,
     "m_pfleg_berücks_zeit": float,
     "y_pflichtbeitr_ab_40": float,
+    "anwartschaftszeit": bool,
+    "arbeitssuchend": bool,
+    "arbeitsl_geld_bezug_m": int,
+    "versich_pfl_m": int,
```
hmgaudecker, line 130:
```suggestion
    "soz_vers_pflicht_5j": int,
```
I first thought "Versicherung Pflege" without having seen any context :smile: 

We are already using `soz_vers_` elsewhere, I am not married to the rest of the name except there should be no `_m` ending (reserved for measurements at the monthly level as opposed to the unit being a month)

paulinaschroeder: sounds good 👍

docs/gettsim_objects/input_variables.rst
```diff
@@ -146,3 +140,17 @@ household.
...
 ||                          || age 40                                     ||             |
 +---------------------------+---------------------------------------------+--------------+
+|| _`anwartschaftszeit`     || Dummy: at least 12 months of unemployment  || bool        |
+||                          || contributions in the 30 months before      ||             |
+||                          || claiming unemployment insurance          ||             |
++---------------------------+---------------------------------------------+--------------+
+|| _`arbeitssuchend`        || Dummy: looking for employment              || bool        |
+||                          ||                                            ||             |
++---------------------------+---------------------------------------------+--------------+
+|| _`arbeitsl_geld_bezug_m` || Amount of months the individual already    || int         |
```
ChristianZimpelmann, line 150: - add "uninterruptedly"?

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -42,71 +45,147 @@ def arbeitsl_geld_m(
...
-    arbeitsl_monate_v2j: int,
+def restliche_anspruchsdauer(
+    anwartschaftszeit: int,
+    dauer_nach_versich_pfl: int,
+    dauer_nach_alter: int,
+    arbeitsl_geld_bezug_m: int,
 ) -> int:
-    """Aggregate months of unemployment over the last two years.
+    """Calculate the remaining amount of months a person can receive unemployment
+    benifit this year.
```
ChristianZimpelmann, line 55:
```suggestion
    benefit this year.
```

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -42,71 +45,147 @@ def arbeitsl_geld_m(
...
     )
+
     return out
 
 
+def dauer_nach_alter(
+    alter: int,
+    arbeitsl_geld_params: dict,
+) -> int:
+    """Calculate the lenght of unemployment benifit according to age.
```
ChristianZimpelmann, line 128:
```suggestion
    """Calculate the lenght of unemployment benefit according to age.
```

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -42,71 +45,147 @@ def arbeitsl_geld_m(
     return out
 
 
-def arbeitsl_monate_gesamt(
-    arbeitsl_monate_lfdj: int,
-    arbeitsl_monate_vorj: int,
-    arbeitsl_monate_v2j: int,
+def restliche_anspruchsdauer(
```
ChristianZimpelmann, line 48:
```suggestion
def arbeitl_geld_restliche_anspruchsdauer(
```
For functions we should typically add the name of the transfer.

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -1,13 +1,16 @@
...
+
 from gettsim.piecewise_functions import piecewise_polynomial
 from gettsim.taxes.eink_st import _eink_st_tarif
+from gettsim.transfers.rente import ges_rente_regelaltersgrenze
 
 
 def arbeitsl_geld_m(
     anz_kinder_tu: int,
-    arbeitsl_geld_berechtigt: bool,
+    arbeitsl_geld_berechtigt: int,
```
ChristianZimpelmann, line 13: Why is type `int`?

gettsim/transfers/arbeitsl_geld.py
```diff
@@ -42,71 +45,147 @@ def arbeitsl_geld_m(
...
+        arbeitssuchend
+        and (restliche_anspruchsdauer > 0)
+        and (alter < regelaltersgrenze)
         and (arbeitsstunden_w < arbeitsl_geld_params["stundengrenze"])
     )
+
     return out
 
 
+def dauer_nach_alter(
```
ChristianZimpelmann, line 124: I think we do not need `dauer_nach_alter` and `dauer_nach_versich_pfl` as separate functions. Can put this code to the beginning of `restliche_anspruchsdauer`

gettsim/parameters/arbeitsl_geld.yaml
```diff
@@ -188,3 +164,30 @@ stundengrenze:
...
+      The lenght of ALG 1 eligibility depends on age and on how many months a
+      person was subject to compulsory insurance
+  reference: § 147 Abs. 2 SGB III
+  1997-03-24:
+    nach_alter:
+      0: 12
+      50: 15
+      55: 18
+      58: 24
+    nach_versich_pfl:
```
ChristianZimpelmann, line 186:
```suggestion
    nach_versicherungspflichtige_monate:
```
No character limit on parameters. Can be more verbose.

ChristianZimpelmann (commented):
Structur looks good now, but I made a few more comments.  Thanks for your work!

docs/gettsim_objects/input_variables.md
```diff
@@ -486,3 +462,37 @@ Type: float
 Received Bürgergeld in previous year
 
 Type: bool
+
+(anwartschaftszeit)=
+
+## `anwartschaftszeit`
```
hmgaudecker, line 468:
```suggestion
## `arbeitsl_geld_anwartschaftszeit`
```

Apologies I missed that one. `anwartschaftszeit` is probably more common in pensions context, so we'll definitely need a prefix. Probably need to make it private or ignore errors for the moment.

hmgaudecker: Sure, we just need to be explicit here.

paulinaschroeder: Ok 👍

ChristianZimpelmann: > `anwartschaftszeit` is probably more common in pensions context, so we'll definitely need a prefix. Probably need to ignore errors for the moment.

I checked again and `anwartschaftszeit` is indeed only used for Arbeitslosengeld (pensions use the term Anwartschaft though).

Should we really increase `GEP_01_CHARACTER_LIMIT_USER_FACING_COLUMNS` to 32 (like for all other columns) for this exception? Would be fine with me, but I would prefer going with `anwartschaftszeit` only.

hmgaudecker: Thanks! Just go for it as is, distinction will be irrelevant in the near future, anyhow.

hmgaudecker: And apologies, @paulinaschroeder, for not reading your earlier message carefully. I actually had not realised the difference!

docs/gettsim_objects/input_variables.md
```diff
@@ -429,7 +405,7 @@ Months of schooling (incl college, unifrom age 17, max. 8 years)
 
 Type: float
 
-(m_alg1_übergang)=
+(m*alg1*übergang)=
```
ChristianZimpelmann, line 408: Looks like an auto-formatting error

docs/gettsim_objects/input_variables.md
```diff
@@ -486,3 +462,37 @@ Type: float
...
+
+Dummy: looking for employment
+
+Type: bool
+
+(m_durchg_alg1_bezug)=
+
+## `m_durchg_alg1_bezug`
+
+Amount of months the individual already uninterruptedly receives Arbeitslosengeld
```
ChristianZimpelmann, line 487:
```suggestion
Months the individual already uninterruptedly receives Arbeitslosengeld
```

docs/gettsim_objects/input_variables.rst
```diff
@@ -0,0 +1,161 @@
+.. _input_variables:
```
ChristianZimpelmann, line 1: This file is no longer needed.

src/_gettsim/config.py
```diff
@@ -129,6 +126,10 @@
     "m_kind_berücks_zeit": float,
     "m_pfleg_berücks_zeit": float,
     "y_pflichtbeitr_ab_40": float,
+    "m_anwartschaftszeit": bool,
+    "arbeitssuchend": bool,
+    "m_durchg_alg1_bezug": int,
```
ChristianZimpelmann, line 131: for `m_pfleg_berücks_zeit` etc. we use the type `float`. For consistency, I would also do the same for m_durchg_alg1_bezug and soz_vers_pflicht_5j

ChristianZimpelmann (changes requested):
Thanks! Is almost done now. Just a couple of small comments.

ChristianZimpelmann (approved):
Great! We can merge after tests have run through.

@paulinaschroeder, can you also open a new issue shortly describing the missing implementation aspect related to Lohnsteuer and the problems in the test data (if you still have web-links to "correct" test cases available, you can add them, too). We should address those issues after #150 is merged.

PR #458, opened 2022-12-08
Fix Vorsorgeaufwendungen

### LINKED ISSUE #376: Potential Bug in calculation of "Sonstige Vorsorgeaufwendungen"
### Bug description
In script: vorsorgeaufw.py  "Vorsorgeaufwendungen" are calculated. which are later needed for the calculation of a taxable income.
From my point of view, parts of the code are not correct. Let me discuss the following code snippet.

v. 0.5.0 code is:
```
def vorsorgeaufw_ab_2020(
...
sonstige_vors = 12 * (
            ges_pflegev_beitr_m
            + (1 - eink_st_abzüge_params["vorsorge_kranken_minderung"])
            * ges_krankenv_beitr_m
        )
        limit_below = sonstige_vors + 12 * arbeitsl_v_beitr_m
        limit_above = eink_st_abzüge_params["vorsorge_sonstige_aufw_max"]

        out = min(max(sonstige_vors, limit_below), limit_above)
```

From my point of view, there are two possible mistakes and I have one issue with the naming. Let me briefly explain myself:

1) Consider following line of code:
```
limit_below = sonstige_vors + 12 * arbeitsl_v_beitr_m
```
The scalar(?) **sonstige_vors** is used here. 
In the calculation of **sonstige_vors** only the "Basis GKV" is considered by:
```
 (1 - eink_st_abzüge_params["vorsorge_kranken_minderung"])
            * ges_krankenv_beitr_m
```
which is correct if you are exceeding the limit of 1900€ for a "abhangig beschäftigte" person. 
But it is not correct when it is used to calculate **limit_below**.
Please compare here the following quote.
"Ein Anspruch auf Krankengeld gehört jedoch nicht zum Basisschutz. Deshalb kürzt das Finanzamt bei Pflichtversicherten den Kran­ken­ver­si­che­rungsbeitrag pauschal um 4 Prozent. Dieser unberücksichtigte Betrag und die Beiträge für Wahlleistungen oder eine Aus­lands­kran­ken­ver­si­che­rung kannst Du nur dann als sonstige Vorsorgeaufwendungen geltend machen, wenn der Höchstbetrag von 1.900 Euro noch nicht ausgeschöpft ist." (see e.g. [Source1](https://www.finanztip.de/vorsorgeaufwendungen/), [Source2](https://www.haufe.de/steuern/finanzverwaltung/was-sind-sonstige-vorsorgeaufwendungen_164_417272.html). Of course much better would it be to give reference to EStG, but pretty sure the above mentioned sources are right in that case.)
The second part of the quote states that full payments of GKV should be included in **limit_below**.

2) Consider the following line of code:
```
out = min(max(sonstige_vors, limit_below), limit_above)
```

From my understanding of the code you are restricting **out**, to take no value larger than **limit_above**, which is here 1900€ for 
a "abhangig beschäftigte" person.

Please compare here the following quote.
"Zu den Sonderausgaben, die Du in **voller Höhe** von der Steuer absetzen kannst, gehören auch die Beiträge für Pfle­ge­ver­si­che­rung und Kran­ken­kas­se – aber nur die, die Dich auf sozialhilfegleichem Niveau absichern. Das ist die sogenannte Basiskrankenversicherung. Der von einer Kran­ken­kas­se erhobene Zusatzbeitrag zählt auch dazu" (see e.g. [Source1](https://www.finanztip.de/vorsorgeaufwendungen/), [Source2](https://www.haufe.de/steuern/finanzverwaltung/was-sind-sonstige-vorsorgeaufwendungen_164_417272.html). Of course much better would it be to give reference to EStG, but pretty sure the above mentioned sources are right in that case.)
From my point of view, the quote states that a Basiskrankenversicherung is fully deductible.

I personally don't like the coding of a nested min(max)) combination.

3) I want to discuss my issues with the naming. The naming here is from my point of view, not well aligned with the law termini.

Suggestions would be:
**sonstige_vors** -> **basiskrankenversicherung**

**limit_below** -> **sonstige_vors**

**limit_above** -> **max_sonst_vors**

### Expected behavior
1) leads to calculations of Vorsorgeaufwendungen, which are sligtly too low. Nothing dramatic, but still wrong for certain income intervals. For Single Households i would quess this is relevant for monthly incomes between 1200 and 1800€.

2) leads to calculations of Vorsorgeaufwendungen, which might be too low. For Single Households i would quess this is relevant for monthly incomes above 1800€ approx.

3) naming might help, to read code easier

Combining 1)-3): a suggestion for the code would be:

```
basiskrankenversicherung = 12 * (
            ges_pflegev_beitr_m
            + (1 - eink_st_abzüge_params["vorsorge_kranken_minderung"])
            * ges_krankenv_beitr_m
        )
        sonst_vors_max = eink_st_abzüge_params["vorsorge_sonstige_aufw_max"]
        sonst_vors = min(12 * (arbeitsl_v_beitr_m + ges_pflegev_beitr_m + ges_krankenv_beitr_m), sonst_vors_max)

        out = max(basiskrankenversicherung, sonst_vors)
```

I hope I got it right here...
Open for discussions.

ChristianZimpelmann:

Fix several bugs in Vorsorgeaufwendungen.

Closes #376 

### Todo

- [x] Fix bug based on @michaelhebsaker's suggestion
- [x] Calculate Vorsorgeaufwendungen on tax unit level
- [x] Adjust test cases and add new test cases (especially those with dual earner couples).


ChristianZimpelmann:
I adjusted all existing test cases to the new variable name `vorsorgeaufw` -> `vorsorgeaufw_tu`.

So far, we do not have any test cases in `vorsorgeaufw.csv` for dual-earner couples. @LauraGergeleit, could you have a look for some?

LauraGergeleit:
> So far, we do not have any test cases in `vorsorgeaufw.csv` for dual-earner couples. @LauraGergeleit, could you have a look for some?

I added two testcases for dual-earners. I also found [another calculation example](https://www.haufe.de/finance/haufe-finance-office-premium/vorsorgeaufwendungen-365-abzugsbetrag-bei-ehegattenlebenspartnern-im-fall-der-einzelveranlagung-von-ehegatten_idesk_PI20354_HI9286136.html), however here private health care contributions are paid but this input is so far not considered in GETTSIM when calculating `vorsorgeaufw_tu`. Should I try to add this testcase?

hmgaudecker:
I would vote for the intermediate way (we should not extend the scope of this PR):
- Add the testcase since you just made the investment into how the code and testcases work
- [Mark it as expected failure](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail)
- Add an issue to the list to add private health insurance contributions (in general / to vorsorgeaufwendungen). If there is a corresponding issue already, point to the new test.

Thanks!

src/_gettsim_tests/test_vorsorgeaufw.py
```diff
@@ -41,6 +43,8 @@ def input_data():
 
 
 @pytest.mark.parametrize("year, target", itertools.product(YEARS, OUT_COLS))
+# testcase 19 for 2022 still fails due to missing inputs
+@pytest.mark.xfail
```
ChristianZimpelmann, line 47: I think the way it is implemented, all test cases are expected to fail. Instead, I propose:
add a new column to the test data `expected_to_fail`. And add a separate function `test_vorsorgeaufw_not_yet_implemented` that only uses the test cases for which `input_data["expected_to_fail"] == True` (and the existing function should only use `input_data["expected_to_fail"] == False`

hmgaudecker: No column in the test data, please. We want to see the expected failure right next to the test, not somewhere else.

I agree that pytest syntax for expected failures with parametrize is not exactly pretty (they should really implement @tobiasraabe's syntax from pytast :wink:), but we should use that nevertheless: https://docs.pytest.org/en/latest/how-to/skipping.html#skip-xfail-with-parametrize

LauraGergeleit: Thanks for the comments! I adjusted it now and hope that the implementation is correct

src/_gettsim_tests/test_vorsorgeaufw.py
```diff
@@ -41,7 +39,24 @@ def input_data():
...
+        (2004, "vorsorgeaufw_tu"),
+        (2005, "vorsorgeaufw_tu"),
+        (2010, "vorsorgeaufw_tu"),
+        (2018, "vorsorgeaufw_tu"),
+        (2020, "vorsorgeaufw_tu"),
+        (2021, "vorsorgeaufw_tu"),
+        pytest.param(
+            2022,
+            "vorsorgeaufw_tu",
+            marks=pytest.mark.xfail(reason="missing input variables"),
```
hmgaudecker, line 54:
```suggestion
            marks=pytest.mark.xfail(reason="missing input variables, see #488"),
```

src/_gettsim_tests/test_vorsorgeaufw.py
ChristianZimpelmann: Thanks! Looks much better! If I see correctly, we, however, still do not differentiate between the test cases in 2022 which we expect to fail and those for which we do not expect that.

Hans-Martin, how would you do that if not specified in the test data? Just based on indices?

hmgaudecker: Ah, this test structure confuses me time and again. One more reason to have them at a more granular level based on yaml.

My vote goes to keeping the current solution, i.e., ignoring some tests in 2022 that should not really be ignored.

ChristianZimpelmann: > My vote goes to keeping the current solution, i.e., ignoring some tests in 2022 that should not really be ignore

Then the new test cases are not run at all :/ (if I am not mistaken) Sounds not like a good solution as I am not sure when we get along to address the missing feature.

The options seem to be:
- filter failing tests by new column in data
- filter failing tests by identifiers
- remove failing tests (and copy the respective code to issue #488 )

ChristianZimpelmann: Ok!

Laura, can you then implement the third option?

LauraGergeleit: > Laura, can you then implement the third option?

Done!

ChristianZimpelmann:
Perfect! We can merge then! Thanks

PR #461, opened 2022-12-13
Fix calculation of income for Kinderzuschlag 

ChristianZimpelmann:

Fix some issues with the current calculation of income for Kinderzuschlag.

### Todo

- [x] Make adjustments to calculation of income (in particular, do not consider income of children for `kinderzuschl_eink_eltern_m` and compare maximum income threshold (`kinderzuschl_eink_max_m_tu`) to `kinderzuschl_eink_eltern_m_tu` instead of `kinderzuschl_bruttoeink_eltern_m_tu`.
- [x] Adjust tests and potentially add new test cases


ChristianZimpelmann:
I just adjusted the test cases.

- I threw out househould 5 since it contained three adults in the tax unit.
- For household 1, the calculated Kinderzuschlag changed due to this PR: It does no longer get Kinderzuschlag as gross income of the parents (600) is below the minimum income for couples. @LauraGergeleit, can you double-check whether this is correct? (unfortunately, we do not have sources for this test case)

hmgaudecker:
Open an issue with the fifth testcase, in case we want to support that down
the road?

Christian Zimpelmann ***@***.***> schrieb am Do., 5. Jan.
2023, 18:19:

ChristianZimpelmann:
> Open an issue with the fifth testcase, in case we want to support that down
the road?

Three adults in the same household would be fine of course, but this test case had them in the same tax unit. This clearly doesn't make sense. We currently have up to two adults + their dependent children in a tax unit and should switch to only 1-2 persons, the latter only if they file their taxes together.

So I would not keep it. It will be better to build more test cases based on proper sources (they should exist for Kinderzuschlag) at some point instead of fixing this one.

hmgaudecker:
Ah, my bad, I thought these were proper test cases and only the mapping of actually relevant units to tu/hh got mixed up somehow. Apologies for not digging deeper.

hmgaudecker (approved):
Looks good to me, although there is one thing I'd like to see addressed here or in a very quick follow-up PR:

- Spelling inconsistency of 
  - `arbeitsl_geld_2_brutto_eink_m`
  - `kinderzuschl_bruttoeink_eltern_m_tu`

  should be solved in favour of `bruttoeink` everywhere
- The difference between `[x]_bruttoeink_[y]` and `[x]_eink_[y]` should be documented more clearly. It should be clear from the docstrings what each of them will be used for eventually. Ideally, also mention in the docstring of `[x]` why both are needed (one relevant for min, one for max?).

ChristianZimpelmann:
Thanks, @hmgaudecker. I have done both. Just waiting for Laura to doublecheck my changes. Then we can merge.

LauraGergeleit:
>     * For household 1, the calculated Kinderzuschlag changed due to this PR: It does no longer get Kinderzuschlag as gross income of the parents (600) is below the minimum income for couples. @LauraGergeleit, can you double-check whether this is correct? (unfortunately, we do not have sources for this test case)
Yes, I double-checked it. It is correct that this household does no longer receive Kinderzuschlag, for the reason you mentioned. However, I noticed that the `kinderzuschl_bruttoeink_eltern_m` in the testcases is not consistent with the data given on `bruttolohn_m` for the same tests.

LauraGergeleit (commented):
The changes look fine to me, I only have one comment.

PR #462, opened 2022-12-14
Renaming and restructuring of `piecewise_polynomial` parameters

LauraGergeleit:

Closes ##444. We should think about a renaming of the parameters used for `piecewise_polynomial`, so that we use `piecewise_polynomial` internally but 'hide' it from the parameter files. 
Then, we could also avoid to specify values in the parameter file that are not specifically documented in the law but used in `piecewise_polynomial`, as it is the case for `soli_st` (s. ##444). 

Remaining problem with `1: upper_threshold` in `soli_st`:
While the threshold is overwritten with the correct value, the `intercept_at_lower_threshold` is calculated using the value specified in the parameter file. Thus, the test for `soli_st` still fail, when changing the parameter in the yaml-file.


hmgaudecker:
Thanks ! Apologies, I was not clear enough I think.

What I had in mind, and that is hopefully in line with @mjbloemer's line of thinking, was to 

1. only specify the parameters in the law in `soli_st.yaml`. From the top of my head, I might call them `rate`, `tatsächliche_max_rate`, `freigrenze` *(not married to any of those)*
2. We'll call an internal function that translates these parameters to the parameters for piecewise polynomial we had before. Machinery similar to functions that vary by year.

This way, you only need to look at the law to implement parameters for a new year, everything else is handled behind the scenes.

*Note: The implementation does not need to be super-pretty now, I am just thinking hard about how to best unify the DAG to include functions like this, will be working hard to have that in during the first quarter of 2023, so don't worry too much about that part.*

hmgaudecker:
Oh, and in c83f860 I made those updates. Note that the structure in 1991 was much simpler, the current stuff only started appearing in 1995. Rest would have to be adjusted accordingly and we need a function translating that to the previously-defined parameters for `piecewise_polynomial`

hmgaudecker:
Let's tackle this once we have implemented GEP 6 #471 or similar.

PR #470, opened 2022-12-22
Execute notebooks on rtd. Closes #469

### LINKED ISSUE #469: ENH: Automate execution of notebooks as part of building the documentation

#466 reverted unwanted behavior introduced in #429 that turned up only when manually executing the notebooks for the documentation.

The ideal way would be to only ever commit empty notebooks and to execute things automatically when the documentation is built. Should be possible based on https://docs.jupyter.org/en/latest/running.html#using-a-command-line-interface

To do

- [ ] Figure out how to execute our notebooks from the command line
- [ ] Ideally, continue in case of errors in notebooks 
- [ ] Capture exceptions, continue in case of expected errors (e.g., advanced usage has some) but fail for unexpected ones
- [ ] Include as part of the documentation build
- [ ] Figure out whether there is a pre-commit hook removing the output from notebooks before committing.

Might be relevant for estimagic as well, so pinging @janosg @timmens in case they have ideas / suggestions / faced similar issues.

hmgaudecker:

Automate building the documentation.

- [x] Execute notebooks as part of the rtd build. 
- [x] Add pre-commit hook to commit empty notebooks only
- [x] Use mamba on rtd
- [x] Document PR in docs/changes.rst.


janosg (approved):
Nice!

Just one stupid question: I don't see where you trigger the actual build of the notebook such that it is displayed with output in the online documentation. Is this a readthedocs setting or did I miss something?

hmgaudecker:
> I don't see where you trigger the actual build of the notebook such that it is displayed with output in the online documentation.

This is nbsphinx' default behaviour: https://nbsphinx.readthedocs.io/en/0.8.11/usage.html#nbsphinx_execute

Other useful references:
- https://nbsphinx.readthedocs.io/en/0.8.11/allow-errors.html
- https://nbsphinx.readthedocs.io/en/0.8.11/allow-errors-per-cell.html

PR #472, opened 2022-12-22
Format yaml files

hmgaudecker:
Over the last couple of days, I saw many of cosmetic changes in yaml files causing merge conflicts. Once we tackle #282 it will only make things worse.

Hence, this PR add yamllint and yamlfix to the pre-commit hooks.

It also applies the hooks to all relevant files.


ChristianZimpelmann (approved):
Looks great! 

But it seems like the PR is about to revert some of the changes of recently merged PR's. We should have a look at that before merging.

hmgaudecker:
> But it seems like the PR is about to revert some of the changes of recently merged PR's. We should have a look at that before merging.

Great that you notice! Can you give an example?

hmgaudecker:
@ChristianZimpelmann -- could you have another look at what this PR might revert? 

Note I just switched from None/null values being represented by empty lines in yaml to an explicit null -- this is in line with the respective GEP for values being present etc; it is awkward for, say, English descriptions that are missing. But we should be consistent either way and I would think that being explicit is the better side of the trade-off.

ChristianZimpelmann:
Looks great! will merge now

PR #475, opened 2023-01-04
update schedule for pre-commit autoupdate from weekly to monthly

ChristianZimpelmann:

- To avoid too much noise, update schedule for pre-commit autoupdate from weekly to monthly
- I also realized that `yamlfix` introduces CRLF line endings on windows which lets `yamllint` to fail. When changing the order such that `mixed-line-ending` is run after the yaml pre-commit hooks, the second commit runs through.

### Todo

- [ ] Decide whether monthly schedule is indeed the prefered option (alternatives: quarterly or weekly)
- [ ] Find out whether there is a better fix for the second problem

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,9 +62,9 @@ repos:
         args:
           - --in-place
           - --wrap-summaries
-          - '88'
+          - "88"
           - --wrap-descriptions
-          - '88'
+          - "88"
           - --blank
   - repo: https://github.com/PyCQA/flake8
     rev: 5.0.4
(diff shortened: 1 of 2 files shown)


hmgaudecker:
Cool, thanks! 

Noise is mostly coming from flake8 6.0.0 breaking some of the plugins, but I would be fine with monthly frequency, no strong opinion.

Could you check whether the updated order works for you on Windows? Just introduce some fake change in a yaml file.

ChristianZimpelmann:
Thanks!

> Could you check whether the updated order works for you on Windows? Just introduce some fake change in a yaml file.

`yamlfix` still fails during the first commit and introduces a CRLF line ending, but committing twice works. I did not find a better solution, so I think we can merge it.

hmgaudecker (approved):
Thanks!

Issue #479, opened 2023-01-09
BUG: Parameters for "minimum child alimony" before 2016 are missing in "unterhalt.yaml"


JuergenWiemers:
Because of these missing parameters all simulations with 
```python
policy_params, policy_functions = set_up_policy_environment(2015)
```
and earlier years and at least one means-tested benefit as target fail with
```python
KeyError: 'mindestunterhalt'
```
on the current main branch. 

ToDo:

- [ ] Remove `unterhaltsvors_m` from test fixtures for `full_taxes_and_transfers` pre 2016. Currently there so that no error is raised.

ChristianZimpelmann:
Thanks! This should be indeed fixed.

A difficulty will be to find the correct reference of the parameter as the [Mindestunterhaltsverordnung](https://www.buzer.de/1_Mindestunterhalt-Verordnung.htm#buildInHistory) is only available from 2016 on.

If I understand correctly, the parameter was based on the [Düsseldorfer Tabelle](https://www.olg-duesseldorf.nrw.de/infos/Duesseldorfer_Tabelle/Tabelle-2018/Duesseldorfer-Tabelle-2018.pdf) before. We should check whether it is also mentioned in a BGBl.

hmgaudecker:
Thanks to both of you! 

IIRC, this would not be the only occasion where informal numbers were translated into law only at later date.

It seems easy enough to just get the first row of each year's Düsseldorfer Tabelle and to update [unterhalt.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/src/_gettsim/parameters/unterhalt.yaml) accordingly. 

However, this is unlikely to happen **very** soon (say a month ?) and if someone had numbers floating around for earlier years, a PR would be appreciated. 

If someone tackles this a couple of additional notes:
-  I find the mixing of operators in the keys (< vs <=: "*unter 6, unter 12, bis 17 Jahre*") confusing and would prefer 5 / 11 / 17 or 6 / 12 / 18
- Incidentally, it is 6 / 12 / 18 in `unterhaltsvors_altersgrenzen`, but these do not seem to be used anywhere and can be safely deleted, AFAICT

JuergenWiemers:
> However, this is unlikely to happen **very** soon (say a month ?) and if someone had numbers floating around for earlier years, a PR would be appreciated.

[...]

PR #480, opened 2023-01-10
Nicer error message relating to #479

hmgaudecker:

Produce a nicer error if `mindestunterhalt` is required before 2016, link to #479 in that message.

### DIFF
--- a/src/_gettsim/transfers/unterhalt.py
+++ b/src/_gettsim/transfers/unterhalt.py
@@ -38,6 +38,14 @@ def unterhaltsvors_m(
 
     """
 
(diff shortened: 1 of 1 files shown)


hmgaudecker:
@JuergenWiemers, I briefly tested this and it should do -- do you think it would have been helpful seeing this when you ran into #479?

JuergenWiemers:
The message is definitely a lot more user-friendly than the previous `KeyError: 'mindestunterhalt'` exception. 😅 This should work as a stop-gap until #479 gets a proper fix.

JuergenWiemers (approved):

PR #486, opened 2023-01-12
Show all plots in docs, raise error outside of vectorized function

hmgaudecker:
- Noted yesterday that the advanced usage tutorial did not show all plots
- The nicer error message implemented in #480 broke tests in #425, move it to a different place (fine for now, I think).


ChristianZimpelmann (approved):
Looks good, thanks! Can be merged when tests have passed.

PR #490, opened 2023-01-18
fix b028 error in f-strings.

### LINKED ISSUE #489: BUG: Take care of flake8-bugbear B028 error
### Bug description

In several PR's (e.g. #458), pre-commits fail due to [this](https://github.com/PyCQA/flake8-bugbear/commit/ef39e9a1b5306393ce42ea231d3174449a929c81) change in `flake8-bugbear`.

> **B028**: Consider replacing ``f"'{foo}'"`` with ``f"{foo!r}"`` which is both easier to read and will escape quotes inside ``foo`` if that would appear.

### Two options how to fix

- Make changes in code base as suggested.
- Ignore error **B028**

ChristianZimpelmann:

I also like the suggested notation, so I made the requested changes.

Closes #489

### DIFF
--- a/src/_gettsim/functions_loader.py
+++ b/src/_gettsim/functions_loader.py
@@ -260,7 +260,7 @@ def _format_duplicated_functions(duplicated_functions, functions, source):
     """Format an error message showing duplicated functions and their sources."""
     lines = []
     for name in duplicated_functions:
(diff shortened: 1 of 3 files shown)


hmgaudecker (approved):
Great, thanks!

Issue #493, opened 2023-01-20
BUG: Introductory phase of Rente für besonders langjährig Versicherte is imprecise


TeBackh:
### Bug description

Imprecise parameter input of Rente für besonders langjährig Versicherte. Unnecessary part in the parameter definition

- [x] Set age to 63 for all cohorts before 1953 from 1.7.2014

hmgaudecker:
Thanks! I must admit that I do not quite follow. Can you elaborate a little on which parameter is imprecise and what is the proposed solution?

TeBackh:
Since 1st of July 2014 all cohorts up to cohort 1952 could claim the pension for very long term insured with 63 instead of 65. Since the people born before 1.7.1949 were already 65 by then they were not affected and the people born between 1.7.1949 and 1.7.1951 could not retire with 63 because they had some age between 63 and 65 by the time of introduction. In the previous set up this mechanical "gradual "  reduction  in the retirement age that is caused by the introduction date was explicitely written in the parameter setup, which was unnecessary since the introduction date should do the job alone. and further we only incorporated a decrease from cohort 1950 and not from 1949.5 (which was "inprecise" or basically wrong).
-- and it was my own bug -- I fix it

hmgaudecker:
Cool, thanks! 

Hold on a bit until we have the new structure with time-varying functions in place. The possibility to retire for that reason simply should not be there before 2014-07-01, implementing that will be natural fairly soon. Age should not play a role then (other than being in the window $[63, 65)$ )

TeBackh:
I had done it already, was only fighting with the pre commit hooks. (Laura's tests depend on it , thats why I wanted to fix it quickly)

Issue #491, opened 2023-01-18
ENH: Add information when certain reforms were introduced


ChristianZimpelmann:
#### Is your feature request related to a problem?

Regelaltersrente (NRA) is in `ges_rente.yaml` constant after 1984 (only dependend on birth year).

In reality, that regime seems only be in place from 2007 on.

For the calculation of actual pensions this doesn't matter, but it is relevant when somebody wants to look at expected retirement ages of subjects.

Also: Check pension functions whether `dates_active`-decorators are needed.

hmgaudecker:
Another example would be Vorsorgeaufwendungen, originally set on a path to a new regime 2005-2025. There was a discrete change announced in 2022, that the last three years of the transition would happen more quickly, so the new regime would be in place already in 2023. Implemented in #476.

Not sure how much of that we want to have in here, but for forward-looking agents in 2020 this definitely was not to be expected with 100% certainty...

hmgaudecker:
Another example would be to change the interface of the functions determining age eligibility rules in the pensions case so that discussions like this https://github.com/iza-institute-of-labor-economics/gettsim/pull/428#discussion_r1088075001 don't come up.

PR #494, opened 2023-01-20
fixed parameter definition rente mit 63 plus notes.

TeBackh:

### Todo
- [x] delete unnecessary code lines that lead to imprecision

### DIFF
--- a/src/_gettsim/parameters/ges_rente.yaml
+++ b/src/_gettsim/parameters/ges_rente.yaml
@@ -521,8 +521,8 @@ rentenwert:
     en: Pension value old and new federal states
   description:
     de: >-
-      Der aktuelle Rentenwert drückt den Betrag aus, der der monatlichen Rente für
-      einen Entgeltpunkt entspricht.
+      Der aktuelle Rentenwert drückt den Betrag aus, der der monatlichen Rente für einen 
+      Entgeltpunkt entspricht.
     en: >-
       The current pension value expresses the amount of monthly pension paid for one
       Entgeltpunkt.
@@ -831,7 +831,7 @@ altersgrenze_für_frauen_abschlagsfrei:
       upper_threshold: 1944.916666666
       rate_linear: 1
     2:
-      upper_threshold: 1951.916666666  # constant 65
+      upper_threshold: 1951.916666666 # constant 65
       rate_linear: 0
     3:
       upper_threshold: 1952
@@ -870,7 +870,7 @@ altersgrenze_für_frauen_vorzeitig:
       upper_threshold: 1953
       rate_linear: 7
     2:
-      upper_threshold: inf  # constant 67
+      upper_threshold: inf # constant 67
       rate_linear: 0
 
 altersgrenze_langj_versicherte_vorzeitig:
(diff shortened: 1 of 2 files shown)


src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -875,12 +875,12 @@ altersgrenze_für_frauen_vorzeitig:
...
 altersgrenze_langj_versicherte_vorzeitig:
   name:
-    de: Altersgrenze für vorzeitigen Bezug der Altersrente für langjährig Versicherte.
+    de: Altersgrenze für vorzeitigen Bezug der Altersrente für langjährig Versicherte
     en: Early retirement age for long term insured
   description:
     de: >-
-      Frühesmögliches Alter ab dem die Altersrente für langjährig Versicherte bezogen
-      werden kann (mit Abschlägen).
+      Frühesmögliches Alter ab dem die Altersrente für langjährig Versicherte be-
```
hmgaudecker, line 882: Please don't break words. In general, why all those unrelated changes to the texts?

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -973,23 +976,16 @@ altersgrenze_besonders_langj_versicherte:
       upper_threshold: inf
       rate_linear: 0
       intercept_at_lower_threshold: 65
-  2014-07-01:
+  2014-07-01:  # See comment above *
```
hmgaudecker, line 979: Leave as is, no inline comments, please. 

Are we sure that the results are correct on, say, 1 August 2014? Just wondering about the granularity here since thresholds are based on years but introduction was at the monthly level.

hmgaudecker (commented):
Thanks! 

As mentioned in #493, we should wait until we have the new machinery for time-varying functions in place. `_ges_rente_besond_langj_altersgrenze` should not exist before 2014-07-01.

src/_gettsim/parameters/ges_rente.yaml
TeBackh: To calculate the eligibility for a given date it should work -  and I think this is the main function at the moment.
(For the details we should talk really , there is just too many layers of thinking right know to provide relevant answers in written comment.)

hmgaudecker: Great. Let's see about that once we have merged #495 and potentially other larger changes.

src/_gettsim/parameters/ges_rente.yaml
ChristianZimpelmann: I guess pre-commit hooks complaining about line length. I will ask Laura to run pre-commit hooks over all files to avoid it in the future

ChristianZimpelmann (approved):
I agree with @hmgaudecker's comments. Otherwise, it is ready to merge.

Thanks!

ChristianZimpelmann:
@hmgaudecker, why did you add the `blocked'-label. If I understand correctly, without this PR the new test cases in #428 are failing.

hmgaudecker:
> @hmgaudecker, why did you add the `blocked'-label. If I understand correctly, without this PR the new test cases in #428 are failing.

Because I would like to see a proper solution for the introduction of the function, which will only be reasonable with the machinery from #495. That will not be part of the upcoming release.

If easier, please go ahead and merge, noting that point appropriately (I guess that was your idea behind #491?) so that it will not be forgotten.

ChristianZimpelmann (approved):
I added a note on `_ges_rente_besond_langj_altersgrenze` and addressed the last comment.

Ready to be merged after tests have run through.

Thanks @TeBackh and @hmgaudecker

PR #495, opened 2023-01-20
`@dates_active` decorator

### LINKED ISSUE #334: Improve specification of time dependent functions

The mapping of time dependent functions to their time independent name is currently specified as several if conditions in `load_reforms_for_date`.

To obtain use this mapping in tests, we currently use this code:

```
    time_dependent_functions = {}
    for year in range(1990, 2021):
        year_functions = load_reforms_for_date(datetime.date(year=year, month=1, day=1))
        new_dict = {func.__name__: key for key, func in year_functions.items()}
        time_dependent_functions = {**time_dependent_functions, **new_dict}
```

Which doesn't look nice and also needed to be adjusted once some column changed twice during the same year.
      

### Proposed implementation

Define the mapping of time dependent functions to their time independent name such that the mapping can be directly imported somewhere else.

I could think of the following nested dict structure:
```
{
    "vorsorge": {
        None: vorsorge_bis_2004,
        "01-01-2005": vorsorge_ab_2005_bis_2009,
        "01-01-2010": vorsorge_ab_2010_bis_2019,
        "01-01-2020": vorsorge_ab_2020,
    }
}
```

### Considered alternatives

Specify the mapping in a .yaml file. Problem here is that we could not map the function directly, but only the function name.

lars-reimann:

Closes partially #334

### Todo


lars-reimann:
I only partially replaced the conditionals in `load_functions_for_date` with the `@dates_active` decorator so far to keep the PR small and allow it to be merged quickly. Other functions can be gradually migrated to the new system.

hmgaudecker:
Wow, that looks great!  Just a couple of quick comments / thoughts:

- I think it is a good idea to revert [108ab4f](https://github.com/iza-institute-of-labor-economics/gettsim/pull/495/commits/108ab4f5f1cbac2470939b9223fc2ec30cf3f06f) -- the version we will release with this should only support recent Python (easy to get with conda and gives new Python users much improved error messages), so given that Python can't distinguish between the two formats, the tests should make clear that we support both.
- It would be great to build in an automatic check that time periods do not overlap for functions that end up having the same key in the DAG. Ideally with a nice error message and early (so you notice immediately when coding things up). If that proves to be hard, fail when overwriting a key. But that should not silently work, I think *(I did not check current behavior)*
- @janosg commented on GEP 6 that @tobiasraabe has been using a nicer way to store function attributes than a bunch of dunder-attributes. I briefly checked the other day but only found class attributes via the attrs library, not sure whether they also work for functions?

janosg:
My point was just that instead of several dunder attributes there should be just one (a dict containing all information). It is very easy to lose dunder attributes when applying other decorators. I don't think `functools.wraps` copies everything. If we have just one, each decorator implemented in gettsim can make sure it does not lose that information.

lars-reimann:
> I think it is a good idea to revert [108ab4f](https://github.com/iza-institute-of-labor-economics/gettsim/pull/495/commits/108ab4f5f1cbac2470939b9223fc2ec30cf3f06f) -- the version we will release with this should only support recent Python (easy to get with conda and gives new Python users much improved error messages), so given that Python can't distinguish between the two formats, the tests should make clear that we support both.

The behavior of `datetime.date.fromisoformat` was changed in Python 3.11. Previously, the dash separators were necessary, even though the ISO 8601 format doesn't require them. We could now either

1. Make our own ISO-compliant parser (e.g. using a regex) or use a 3rd party library to support Python 3.9 and 3.10.
2. Only support the format `YYYY-MM-DD`, validate the string input, add highlight this in the documentation.

In my opinion, the second option makes more sense. The dashes improve readability and we ensure consistency in our source code.

ChristianZimpelmann:
> In my opinion, the second option makes more sense. The dashes improve readability and we ensure consistency in our source code.

Sounds like the better solution for me, as well.

hmgaudecker:
Apologies, I should have been more explicit than "recent Python".

My point was that we will not support Python<3.11 for other reasons in the first version of GETTSIM that will include these changes.

lars-reimann:
> * I think it is a good idea to revert [108ab4f](https://github.com/iza-institute-of-labor-economics/gettsim/pull/495/commits/108ab4f5f1cbac2470939b9223fc2ec30cf3f06f) -- the version we will release with this should only support recent Python (easy to get with conda and gives new Python users much improved error messages), so given that Python can't distinguish between the two formats, the tests should make clear that we support both.

I've added a validation step now to ensure only the format `YYYY-MM-DD` is accepted. If we find this to be too restrictive, we could later losen the check.

> * It would be great to build in an automatic check that time periods do not overlap for functions that end up having the same key in the DAG. Ideally with a nice error message and early (so you notice immediately when coding things up). If that proves to be hard, fail when overwriting a key. But that should not silently work, I think _(I did not check current behavior)_

That's done. If two function collide a `ConflictingTimeDependentFunctionsError` is raised.

> * @janosg commented on GEP 6 that @tobiasraabe has been using a nicer way to store function attributes than a bunch of dunder-attributes. I briefly checked the other day but only found class attributes via the attrs library, not sure whether they also work for functions?

The data is now all stored in a dict in a `__gettsim__` attribute for decorated functions.

hmgaudecker:
Finally got around to this. I do not see right away where the changes from the jax PR interact with this one so that tests start failing. @lars-reimann @timmens, could you guys maybe have a look at it together? I'd think that would be most efficient.

> The data is now all stored in a dict in a `__gettsim__` attribute for decorated functions.

I was just about to open another issue to apply this to other metadata as well. But since this requires additional changes anyhow, we might just as well include it here. _(Maybe it is even the root cause because rounding happens to get lost? Wild guess)_

From a quick search, I believe it only applies to `__rounding_params_key__`, which should become part of `__gettsim__` but please double-check.

Btw, is everybody happy with the `__gettsim__` name? (@ChristianZimpelmann @janosg?) I think it's nice for users who want to extend this stuff, not so great for searching / thinking about it within the codebase. No strong opinion.

src/_gettsim/shared.py
```diff
@@ -39,6 +42,123 @@ def inner(func):
...
+            f"The start date {start} must be before the end date {end}."
+        )
+
+
+def _check_for_conflicts(dag_key: str, function_name: str, start: date, end: date):
+    if dag_key not in TIME_DEPENDENT_FUNCTIONS:
+        return
+
+    for f in TIME_DEPENDENT_FUNCTIONS[dag_key]:
+        # While testing, the same function might be added to the registry again,
```
hmgaudecker, line 122: I am unable to follow this comment. Rather add a sentence or two and be very explicit what is happening. I would also suggest:
- changing the name to `_check_for_conflicts_in_time_dependent_functions`
- turning that long comment into the docstring.

src/_gettsim/taxes/zu_verst_eink/eink.py
```diff
@@ -137,6 +138,7 @@ def eink_rente_zu_verst(
     return eink_rente_zu_verst_m * 12
 
 
+@dates_active(start="2009-01-01", change_name="sum_eink")
 def sum_eink_ohne_kapital(
```
hmgaudecker, line 142:
```suggestion
def sum_eink_ohne_kapital_eink(
```

While we are at it...

src/_gettsim/taxes/zu_verst_eink/eink.py
```diff
@@ -196,6 +198,7 @@ def kapitaleink(
     return max(out, 0.0)
 
 
+@dates_active(end="2008-12-31", change_name="sum_eink")
 def sum_eink_mit_kapital(
```
hmgaudecker, line 202:
```suggestion
def sum_eink_mit_kapital_eink(
```

hmgaudecker (commented):
Looks great! Just a couple of minor comments. I think that before merging, I'll try open another PR porting the remaining cases where we'll need it. Hope I'll find the time!

src/_gettsim/policy_environment.py
```diff
@@ -298,21 +293,13 @@ def load_functions_for_date(date):
...
-    else:
-        functions["eink_st_altersfreib"] = eink_st_altersfreib_ab_2005
+    # Using TIME_DEPENDENT_FUNCTIONS here leads to failing tests.
+    functions = {
+        f.__info__["dates_active_dag_key"]: f
+        for f in load_internal_functions().values()
+        if is_time_dependent(f) and is_active_at_date(f, date)
+    }
 
     if year <= 1996:
```
ChristianZimpelmann, line 304: Is there a specific reason why we do not use the `dates_active` decorator for those functions, yet?

ChristianZimpelmann (commented):
Looks great, thank you!

- `load_internal_functions` is a nice addition!
- I like `__info__` more than `__gettsim__` (sorry for not responding before).

src/_gettsim/policy_environment.py
lars-reimann: I generally like to split a new feature and associated refactorings into two PRs for several reasons:

1. The new feature can immediately be used in new code.
2. A long-running branch (new feature) that touches many files (refactorings) inevitably leads to merge conflicts.
3. A new feature requires an in-depth review, while a review of a manual refactoring is more about finding copy & paste errors.

The three functions I changed are just to verify that the decorator works as expected in the existing code.

src/_gettsim/shared.py
lars-reimann: > turning that long comment into the docstring.

My rule is to use docstrings to document the inferface of an API element, i.e. what its users need to know, and comments to explain the implementation, i.e. what people working on the API element need to know.

I wouldn't mind changing it to a docstring but currently, I just clarified the comment.

lars-reimann:
Not sure why readthedocs is failing now but since other PRs are also affected, it seems to be unrelated to the changes in this PR.

janosg:
It's the latest update of pydata-sphinx-theme. 

We pinned it in estimagic for now. See [here](https://github.com/OpenSourceEconomics/estimagic/blob/75d37470061ea7b65fd35aac4ad9b6c157f19b82/docs/rtd_environment.yml#L19)

src/_gettsim/shared.py
```diff
@@ -39,6 +42,127 @@ def inner(func):
...
+                start,
+                end,
+                f.__name__,
+                f.__info__["dates_active_start"],
+                f.__info__["dates_active_end"],
+            )
+
+
+class ConflictingTimeDependentFunctionsError(Exception):
+    """Raised when two time-dependent functions have overlapping time ranges."""
```
hmgaudecker, line 147: Ahh, I am basically missing this sentence in the above comment. Then it becomes obvious of what is going on. Could you just repeat it there?

hmgaudecker (commented):
Basically done! Thanks a lot, great work!

src/_gettsim/policy_environment.py
ChristianZimpelmann: Makes a lot of sense. Thanks for the explanation!

hmgaudecker:
Excellent, thank you! 

Let me get my hands dirty a little on porting further changes, maybe I'll have a comment or two based on that. No PRs on the horizon that will use this machinery AFAIC, else we can merge earlier.

*NB: No need to change this, but I did not mean to push you towards including a docstring in the function, the comment just was not clear to me without knowing the docstring of the Exception. :smile:*

lars-reimann:
> _NB: No need to change this, but I did not mean to push you towards including a docstring in the function, the comment just was not clear to me without knowing the docstring of the Exception. 😄_

No worries, good to have the docstring anyway.

hmgaudecker:
Truly excellent work, it has been a joy to apply it !

PR #496, opened 2023-01-20
Checking yaml-files for too long lines

LauraGergeleit:

The pre-commit hooks were run over all files to identify and fix too long lines in all yaml-files.

### DIFF
--- a/src/_gettsim/parameters/arbeitsl_geld.yaml
+++ b/src/_gettsim/parameters/arbeitsl_geld.yaml
@@ -34,8 +34,8 @@ dauer_auszahlung:
     en: Requirements for ALG 1
   description:
     de: >-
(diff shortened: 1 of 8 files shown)


hmgaudecker:
Please don't merge this. I am pretty sure some setting is not being respected. Lines were 88 characters long already, I had run all that.

Can you be specific where some hook had been complaining first, please?

LauraGergeleit:
@ChristianZimpelmann asked me to check all files and fix those that are marked as too long. `yamlfix` marked all those lines, I corrected here

hmgaudecker:
Where and when precisely did it flag them?

Maximum line length should be 88 and I do not believe any of them was longer beforehand (please correct me if I am wrong). So it seems like on some machines, a setting is not being respected.

PR #500, opened 2023-01-24
Harmonise hooks with estimagic & friends

hmgaudecker:
Following up on #450, others (@janosg @tobiasraabe @timmens) much improved upon what I had been doing in estimagic, pytask, and elsewhere. This PR ports those changes to GETTSIM.

Most importantly, flake8 is replaced by ruff.


hmgaudecker (commented):
Just a reminder that I double-checked everything.

hmgaudecker:
*(failure of pre-commit.ci is due to being slow in updating ruff, just ignore)*

ChristianZimpelmann (approved):
Looks great!

I just don't like that adding lists seem to be flagged now. In my opinion, the alternative makes the code less intuitive (for non-Python experts).

hmgaudecker:
> I just don't like that adding lists seem to be flagged now. In my opinion, the alternative makes the code less intuitive (for non-Python experts).

Fair enough, the good thing is that these fixes happen automatically. So people can write it the way they find it intuitive and then it is applied by the hooks.

PR #501, opened 2023-01-24
Fix imports, update env

### LINKED ISSUE #498: BUG: Import paths became inconsistent
### Bug description

There are many imports now from `_gettsim`, which is nonsense. I introduced that in #450  and it should not become part of a released version.

### Fix

Mirror the directory structure of `_gettsim` in `gettsim` by and add appropriate import statements s.t. imports continue to work as before.

hmgaudecker:

Closes #498 plus some additional maintenance.

### Todo

- [x] Make sure imports work broadly as in 0.5.1
- [x] Make sure one environment.yml file is enough (rtd_env was identical)
- [x] ~Prepare for release (update docs, too)~


hmgaudecker (commented):
Double-checked.

Issue #502, opened 2023-01-25
ENH: Installation via conda-forge


hmgaudecker:
#### Is your feature request related to a problem?

We removed `conda-build` and IIUC, that had also been used to publish the package via the `gettsim` conda channel. We should go for the standard and use the conda build channel. Our docs on maintaining are horribly outdated and should get a small overhaul in the process.

- [ ] Installation via conda-forge
- [ ] Update documentation for maintaining GETTSIM

@timmens, we talked about that yesterday in the context of #501. I think we should have different PRs, though. Please base a PR off the `fix_imports` branch unless that is merged by the time you get around to it. I'll be happy to write the docs then, maybe with a bit of verbal input.

hmgaudecker:
Done via #503 and #511

PR #503, opened 2023-01-25
Packaging

timmens:

Automate PyPI publishing using GitHub actions. This is a preparation step before tackling #502.

### Todo

- [x] Automate PyPI publishing using workflow
   - [x] Set up workflow
   - [x] Create secret token `PYPI_API_TOKEN` (@hmgaudecker, I cannot do this since token creation requires access to the repository settings)
      From [pypi-publish-notes](https://github.com/marketplace/actions/pypi-publish):
      > The secret used in ${{ secrets.PYPI_API_TOKEN }} needs to be created on the settings page of your project on GitHub. See [Creating & using secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).


.gitignore
```diff
@@ -37,7 +39,6 @@ waf-*
 waf3-*
 *latexmk
 *synctex.gz
-.*
 !.environment.osx.yml
```
hmgaudecker, line 42:
```suggestion
.
!.github
```
Please continue to ignore hidden files and whitelist those that should be included under VC.

The whitelist patterns for the environment files are from a very distant past and can all be removed...

hmgaudecker:
So I registered on PyPI and created the token. However, I don't think I can *register the project* ? Seems like that will happen with the upload? Or I do not find the option...

hmgaudecker:
I am pretty sure that `setup.py` is not required anymore; remember having had that discussion with @tobiasraabe, probably in the context of #450, but I cannot check ATM.

timmens:
After discussing with @janosg we came to a conclusion that
```yaml
on: push
```
is preferred to
```yaml
on: release
```
While the latter specifies more directly that we want to publish to PyPI only when creating a new release, we realized that the former also tests that in each action run the package building works.

What remains to be discussed @hmgaudecker

1. I suggest merging this PR and testing it directly by releasing a new patch on main. I've tried the workflow on one of my repositories, and it works.
2. I would also suggest updating the documentation in a different PR. The reason is that I don't know all the details of the steps that have to be taken when releasing a new version on conda-forge yet, since I can only start working on this component once the project has been uploaded to PyPI.

timmens:
I've just tried it out on one of my projects. The workflow is triggered when releasing a new version from a different branch than main, and the package is uploaded to PyPI. So yes, I strongly expect that this works.

hmgaudecker (approved):
Great, let's give it a shot!

Issue #497, opened 2023-01-20
BUG: Precommit yamllint detects too long lines


ChristianZimpelmann:
### Bug description

`yamllint` run on all files returns a bunch of lines that appear to be too long (see output below).

It seems to be an encoding error as all respective lines I checked contain a German special characters. As other people do not get this error, it might be specific to Windows.

#### Output

```
src/_gettsim/parameters/arbeitsl_geld_2.yaml
  10:89     error    line too long (90 > 88 characters)  (line-length)
  311:89    error    line too long (92 > 88 characters)  (line-length)
  335:89    error    line too long (89 > 88 characters)  (line-length)
  484:89    error    line too long (89 > 88 characters)  (line-length)

src/_gettsim/parameters/arbeitsl_geld.yaml
  37:89     error    line too long (90 > 88 characters)  (line-length)

src/_gettsim/parameters/kinderzuschl.yaml
  229:89    error    line too long (89 > 88 characters)  (line-length)
  263:89    error    line too long (90 > 88 characters)  (line-length)

.github/workflows/main.yaml
  9:1       warning  truthy value should be one of [false, true]  (truthy)
  53:3      warning  comment not indented like content  (comments-indentation)

src/_gettsim/parameters/ges_rente.yaml
  524:89    error    line too long (90 > 88 characters)  (line-length)
  801:89    error    line too long (89 > 88 characters)  (line-length)
  856:89    error    line too long (89 > 88 characters)  (line-length)
  878:89    error    line too long (89 > 88 characters)  (line-length)
  949:89    error    line too long (90 > 88 characters)  (line-length)

src/_gettsim/parameters/wohngeld.yaml
  487:89    error    line too long (89 > 88 characters)  (line-length)
  537:89    error    line too long (89 > 88 characters)  (line-length)
  749:89    error    line too long (89 > 88 characters)  (line-length)
  763:89    error    line too long (89 > 88 characters)  (line-length)
  765:89    error    line too long (91 > 88 characters)  (line-length)
  789:89    error    line too long (89 > 88 characters)  (line-length)
  931:89    error    line too long (90 > 88 characters)  (line-length)
  1966:89   error    line too long (91 > 88 characters)  (line-length)
  1988:89   error    line too long (89 > 88 characters)  (line-length)

src/_gettsim/parameters/kindergeld.yaml
  32:89     error    line too long (92 > 88 characters)  (line-length)
[...]

PR #506, opened 2023-01-26
add note on utf-8 encoding under Windows.

ChristianZimpelmann:
See #497


hmgaudecker (approved):
Great, thanks.

I will be watching https://github.com/adrienverge/yamllint/issues/347 to hopefully see this resolved on the yamllint side, this is an annoying thing to have to tell people...

Issue #379, opened 2022-04-12
Revise groups supported by GETTSIM


cpz-research:
### Current situation

Two groups: tax units and households

### Proposed implementation

- rename households (`_hh`) to bedarfsgemeinschaft (`_bg`)
- check for all transfers if they are calculated on the correct level (see e.g. #218 or #335  )

### Open Questions
- Do we need the concept of a core family?
- Do we need special treatment for wohngeld-Gemeinschaft?

ChristianZimpelmann:
I thought a bit more about this issue and think we should for now only do the two points under "Proposed implementation".

- I think core family is not needed at the moment
- Wohngeld-Gemeinschaft might be interesting in the future, but then we would also need to make the Wohngeld vs ALG 2 Vorrang-Check more complex (`wohngeld_vorrang_hh`).

Do you agree, @Eric-Sommer @mjbloemer @hmgaudecker ?

Eric-Sommer:
Isn't this:

> * check for all transfers if they are calculated on the correct level (see e.g. [Reduce calculation of wohngeld to right level #218](https://github.com/iza-institute-of-labor-economics/gettsim/issues/218) or [Calculate kinderzuschl_heizkosten_m on tax unit level #335](https://github.com/iza-institute-of-labor-economics/gettsim/issues/335)  )

inherently the same as this?

> * Do we need special treatment for wohngeld-Gemeinschaft?

hmgaudecker:
Well, of course it does not need to be done in the same PR but it does concern the same issue of relations between rows of data, so we should at least get the conceptual point clear.

>    what about non biological parents or legal guardians? (Suggestion: also include them as elternteil_1)

I guess we do not care about biology at all. Or could there be cases where the, say, biological father pays alimony (or not, for that matter -> `unterhaltsvorschuss`) and the mother is living with a new partner, which is relevant for taxes and transfers? @mjbloemer @stichnoth @JuergenWiemers  (feel free to ping more knowledgeable people)

>  What about parents not living in the same household,

Distinguishing this would be precisely a point of such a structure. The `elternteil` would have a different `tu_id` / `bg_id`  

> or even being unobserved (there might be some underage children that live alone)?

Same here -- not sure what we could do with those and I don't think we can handle them at present, either?

[...]

PR #509, opened 2023-01-27
Discussion of groupings of / relationships among individuals

hmgaudecker:

Discuss and make a clear plan for #379: Which groupings of individuals (Haushaltsgemeinschaften, Bedarfsgemeinschaften, Familien) and relationships amongst them (Ehepartner, Einstandspartner, Eltern, Kinder, andere Verwandte, Verschwägerte) should GETTSIM support and how?

- `docs/gettsim_developer/hh_concepts.md` collects concepts by type of tax / transfer ([rendered version](https://gettsim--509.org.readthedocs.build/en/509/gettsim_developer/hh_concepts.html#relevant-concepts-of-units-of-individuals))
- `docs/gettsim_developer/hh_examples.yaml`  collects test cases. The more, the better -- please add as many as you can and don't worry about the format!

This might eventually become a GEP.


hmgaudecker:
Thanks for the examples and comments, @JuergenWiemers !

Just a couple of quick reactions:

- id for Bedarfsgemeinschaften should indeed be endogenous. See, for example, the first example household that I pushed. This said, if you pass a column `id_bedarfsgemeinschaft` as part of the data / `columns_overriding_functions`, that will be used. Same as for any other column/ function.
- outputs of examples should indeed be what we expect the functions to produce (if I get the idea of what is meant, I can of course help with working out the details).

ChristianZimpelmann:
I just made several changes to the examples. 

- I think we have all important household constallations now. Not in all interactions, but this should not be necessary.
- @JuergenWiemers rightly points out that some of the outputs depend on unspecified inputs (e.g. whether parents receive Kinderzuschlag for 20 year old child depends on `in_ausbildung` and `arbeitsstunden_w`. We should either rename the output variable (`potentielle_kinderfreibeträge`) or add further input variables. I would prefer the former.
- I added `id_tax_unit` as output.
- I added `unterhaltsvors` as output since there are eligibility differences based on the hh constellation (should be probably rather `potentieller_unterhaltsvors`)

Here two further responses to your comments, @JuergenWiemers, which I removed: 

> Sind Kinderfreibeträge hier als halbe Kinderfreibeträge (je Elternteil) zu
> verstehen? Warum dann bei den folgenden beiden Haushaltstypen jeweils nur ein
> halber Kinderfreibetrag trotz zwei Kindern? 

Nein, ganze Kinderfreibeträge der Tax unit. Im ersten Beispiel hat also die Tax unit der beiden Eltern 2
Kinderfreibeträge. Im nächsten Beispiel sind die Eltern getrennt veranlagt und
jeder hat nur einen bzw. zwei halbe Kinderfreibeträge.

> Was soll die "id_bedarfsgemeinschaft" anzeigen? Soll sie anzeigen,
> für welche Personen in einem Haushalt die Anspruchsprüfung gemeinsam durchgeführt
> werden muss? Oder soll sie die Zuordnung der Personen zu (eventuell nicht bedürfigen)
> Bedarfsgemeinschaften _nach_ der Anspruchsprüfung anzeigen? Ich vermute, dass Ersteres zutrifft.

Can you clarify the second option? The first seems correct though.

hmgaudecker:
> @JuergenWiemers rightly points out that some of the outputs depend on unspecified inputs (e.g. whether parents receive Kinderzuschlag for a 20 year old child depends on in_ausbildung and arbeitsstunden_w. We should either rename the output variable (potentielle_kinderfreibeträge) or add further input variables. I would prefer the former.

What do you mean by "add further input variables" ? These are input columns in GETTSIM, no? If they determine the scope of a Bedarfsgemeinschaft, they should be inputs to that function.

ChristianZimpelmann:
> What do you mean by "add further input variables" ? These are input columns in GETTSIM, no? If they determine the scope of a Bedarfsgemeinschaft, they should be inputs to that function.

Yes, they will be inputs of that function. But the question if whether we would like to add all input columns for all outcomes we are considering to that document. This would become quite lengthy if we added e.g. all income variables.

I now added some note at the begining and some more example households and think I found a good solution.

docs/gettsim_developer/hh_examples.yaml
```diff
@@ -0,0 +1,737 @@
...
+        - 0
+        - 1
+        - 0
+    - potentielle_kinderfreibeträge:
+        - 1
+        - 1
+        - 0
+        - 0
+eltern_verheiratet_1_kind_bis24_1_kind_ueber_24:
+# child separate Bedarfsgemeinschaft if over 24
```
ChristianZimpelmann, line 179:
```suggestion
# child in separate Bedarfsgemeinschaft if over 24
```

docs/gettsim_developer/hh_examples.yaml
```diff
@@ -0,0 +1,737 @@
...
+        - 0
+        - 1
+        - 1
+    - potentielle_kinderfreibeträge:
+        - 1
+        - 1
+        - 0
+        - 0
+eltern_verheiratet_1_kind_mit_eigenem_kind:
+# child in separte Bedarfsgemeinschaft if it has children (Bedarfsgemeinschaft never
```
ChristianZimpelmann, line 304:
```suggestion
# child in separte Bedarfsgemeinschaft if he/she has children (Bedarfsgemeinschaft never
```

hmgaudecker:
I just moved the examples you created so they can become actual tests, see 6ddbe34. Thanks a lot for all those, looks great!!! 

In the process, I realised a bit of a misunderstanding we had: There is a lot of information which is present only in the name of the file (used to be outermost key in the old yaml), but which is not present in the example data yet. E.g., we will need kids' income / education status to determine whether they are part of a bedarfsgemeinschaft. This explains why Christian's comment above confused me. I am rather sure we will need many more input variables. E.g., we might have the following function:

```py
def id_bedarfsgemeinschaft(
    id_person,
    id_haushalt,
    id_einstandspartner,
    id_elternteil_1,
    id_elternteil_2,
    alter,
    in_ausbildung,
    kindeseinkommen_in_passender_definition,
    ...
)
```


Some other comments:
- I am not sure I quite understand the business of 1/2 Kinderfreibetrag. I gather that the colloquial use is the Kinderfreibetrag for both parents, but the way it is defined in the law and [used in GETTSIM](https://github.com/iza-institute-of-labor-economics/gettsim/blob/245ee83f9186e2aecdd787e021e4594e52be8e44/src/_gettsim/parameters/eink_st_abzuege.yaml#L8) is the version per parent. Is that what is meant by 0.5? Or is it further divided by two for some reason? In [eltern_verheiratet_2_kinder.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509/files#diff-2d22e8144706d8c713ac72a1d10ddb7a89c4a19d33eaf3ceb13b23d83fd45fcb), usage seems to be consistent with the "per parent" definition.
- I do not think we will need the distinction between grandmother in SGB II and SGB XII, respectively. My understanding was that we will use the same concept of `bedarfsgemeinschaft` for SGB II and SGB XII?
- Can someone point towards the resolution of bedarfsgemeinschaften when kids have their own kids? The case we have ([eltern_verheiratet_1_kind_mit_eigenem_kind.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509/files#diff-af0035772d762cbabe2ce4cad43adcac0b78499a64cd54b1ce550ef224c2aff2)) is with everyone in the same household. What happens if a kid fathers a child and does not live together with the mother? Would he continue to be part of the bedarfsgemeinschaft of his parents or not?
- @ChristianZimpelmann, re your comment: *An important output we are not yet considering here is the status of "Alleinerziehend" (There might even be differences in the definition relevant for taxes, Kinderzuschlag, Wohngeld, and Bürgergeld)*: This seems out of the scope of this discussion, though obviously related. The status of Alleinerziehend 
- Even though the structure is a yaml-file, we won't be able to use anything but scalars for each individual (e.g., `id_eltern` in [eltern_verheiratet_2_kinder.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509/files#diff-2d22e8144706d8c713ac72a1d10ddb7a89c4a19d33eaf3ceb13b23d83fd45fcb). We'll also need to transpose files to actually use them as tests, but maybe it is clearer like this now.
- How do we think about `unterhaltsvorschuss`? Can it be determined by looking at the child / bedarfsgemeinschaft / haushalt alone, or do we somehow need to check references to other individuals? In my view, the answer to this will determine whether we deal with this here or elsewhere. I do not have a good sense of it.
- I think one of the toughest parts will be to deal with cases where we observe only subsets of the relevant persons (see [here](https://gettsim--509.org.readthedocs.build/en/509/gettsim_developer/hh_concepts.html#teile-der-daten-nicht-vorhanden)). Any ideas would be greatly appreciated!

The next steps will  be to come up with functions:

- `id_bedarfsgemeinschaft`
- `id_steuer` *I am not married to the name, but `id_tax_unit` is definitely out of place (English, confusing given former usage including children). Any better suggestions are more than welcome!*
- `anzahl_kinderfreibeträge`
- `unterhaltsvors_berechtigt` *possibly, see above)*

To get there, I would appreciate any input, either,
- explaining relevant variables and how they would enter here (e.g., spell out which income variable is needed in the example for `id_bedarfsgemeinschaft` above)
- adding to the test data, does not need to be conclusive, but any start would be greatly appreciated!

@JuergenWiemers @ChristianZimpelmann @mjbloemer @ritter9 @michaelhebsaker

ChristianZimpelmann:
> I am not sure I quite understand the business of 1/2 Kinderfreibetrag. I gather that the colloquial use is the Kinderfreibetrag for both parents, but the way it is defined in the law and [used in GETTSIM](https://github.com/iza-institute-of-labor-economics/gettsim/blob/245ee83f9186e2aecdd787e021e4594e52be8e44/src/_gettsim/parameters/eink_st_abzuege.yaml#L8) is the version per parent. Is that what is meant by 0.5? Or is it further divided by two for some reason? In [eltern_verheiratet_2_kinder.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509/files#diff-2d22e8144706d8c713ac72a1d10ddb7a89c4a19d33eaf3ceb13b23d83fd45fcb), usage seems to be consistent with the "per parent" definition.

I think they are two aspects:

- It is currently written such that each child is associated with 1 Kinderfreibetrag, typically each parent would get 1/2 Kinderfreibetrag.
- `potentielle_kinderfreibeträge` should actually be names `potentielle_kinderfreibeträge_tu` as it specifies Kinderfreibeträge on `id_steuer`-level. In the example you mention, both parents get 2 * 0.5 = 1 Kinderfreibeträge. The tax unit of both parents gets 2 Kinderfreibeträge

I hope this clears things up. It might make sense to multiply all values of `potentielle_kinderfreibeträge_tu` with 2, not sure what is more common here.

> Can someone point towards the resolution of bedarfsgemeinschaften when kids have their own kids? 

This seems to be relevant: https://www.arbeitsagentur.de/datei/dok_ba015897.pdf

- It seems only about kids living in the same household.
- If I understand correctly, it is however even more complicated: If the child is unable to work, he_she is part of the Bedarfsgemeinschaft of the parents, but the grandchild is not. Not sure whether we would like to model that.

What I just realized: I think what is still missing, is examples for [Kinderwohngeld](https://www.wohngeld.net/kinderwohngeld.html)

hmgaudecker:
> > Can someone point towards the resolution of bedarfsgemeinschaften when kids have their own kids?
> 
> This seems to be relevant: https://www.arbeitsagentur.de/datei/dok_ba015897.pdf
> 
> * It seems only about kids living in the same household.
> 
> * If I understand correctly, it is however even more complicated: If the child is unable to work, he_she is part of the Bedarfsgemeinschaft of the parents, but the grandchild is not. Not sure whether we would like to model that.

Thanks! I don't follow all the way yet. Do you have more specific pointers to which parts of that 99-page document? Seems like you had a closer look already :upside_down_face:

hmgaudecker:
> What I just realized: I think what is still missing, is examples for [Kinderwohngeld](https://www.wohngeld.net/kinderwohngeld.html). Although it would just be a special case of [this case](src/_gettsim_tests/test_data/id_variables/2023/eltern_verheiratet_1_kind_with_income.yaml) if I understand correctly.

The link seems to be broken. But indeed, I do not think (correct me if I am wrong) that there is anything in `kinderwohngeld` that is conceptually relevant here. There will be a formula that determines whether a kid will be taken out of a Bedarfsgemeinschaft and Wohngeld should be an input (eventually). So should become a testcase indeed once we model it (if ever).

hmgaudecker:
> > Thanks! Is there any reason at that stage to bring in the id_steuer - level? To me it is just confusing, we have the individual relationships, we are associating the values with parents, and in the yaml-files we are using the per-parent definition of Kinderfreibeträge...
> 
> I think it is naturally a variable on id_steuer-level and also currently defined like that: `eink_st_kinderfreib_tu`. But I am fine with changing to individual level.

Sure we'll need it at the `id_steuer`-level eventually! However, with the coming structure it seems much simpler to think about it at the individual level and then aggregate using the built-in functionality. Apologies, should have been clearer about this.

hmgaudecker:
@MImmesberger: Please check at some point whether we incorporated all tests that are part of this PR somewhere in `main`. 

From this PR, we may want to merge `hh_concepts.md` eventually, but certainly nothing else. Even for that document, it might be easier to start over.

MImmesberger:
Tests are all part of `main`.

MImmesberger:
Won't be merged but I will make a new PR that merges `hh_concepts.md`.

PR #510, opened 2023-01-27
Update setup.cfg

timmens:

setup.cfg contains faulty entries. They are fixed in this PR.

### Todo

- [x] Fix links and add correct license classifier to setup.cfg


hmgaudecker (approved):
Not much to check ☺️

PR #514, opened 2023-01-30
Rewrite `create_synthetic_data`

ChristianZimpelmann:

`create_synthetic_data` is currently overly complicated and slow (especially when many households are created at the same time). Main problem seems to be that a lot of pandas objects are created and concatenated.

To fix the problems, I rewrote most of the code. Remarks:

- all variables that do not vary across households are now collected in a `dict of lists`: `specs_constant_over_households`
- all variables that should vary across households are now collected in a `dict of lists of lists`: `specs_heterogeneous`
- At the moment, it is no longer possible to create single and couple households or households with varying numbers of children with one function call. I would deliberately drop this feature, but this is up for discussion and could be implemented, as well.

### Todo

- [x] Rewrite code.
- [x] Improve documentation.
- [x] Adjust test cases
- [x] Adjust tutorials

### Collection of ideas

- allow for more than 2 children by either extending official data on Bedarfsgemeinschaften or by dropping use of official data
- allow for easier specification of variables which are constant within households via broadcasting (i.e. allow for `mietstufe: 2` instead of `mietstufe: [2, 2, 2, 2]`


hmgaudecker:
Thanks, sounds great!

+1 for removing that feature

ChristianZimpelmann:
I can have a look at it on Monday.

ChristianZimpelmann:
- I adjusted and expanded test cases and solved the merge conflict.
- I would add the "Collection of ideas" to a new issues
- only open ToDo is the adjustments of tutorials. I would only address this after we fixed #570.

ChristianZimpelmann:
Somehow, the last commit [`4bd9432a4aaaa8538367fef9dbaa3f01b74cb0f9`](https://github.com/iza-institute-of-labor-economics/gettsim/commit/4bd9432a4aaaa8538367fef9dbaa3f01b74cb0f9) does not show up here yet.

@hmgaudecker, please check again before your review.

src/_gettsim/synthetic.py
```diff
@@ -1,436 +1,211 @@
...
 
-    bruttolohn_m, kapitaleink_brutto_m, eink_selbst_m, vermögen_bedürft_hh (int):
-        values for income and wealth, respectively.
-        only valid if heterogenous_vars is empty
+def create_basic_households(
+    n_adults, n_children, specs_constant_over_households, specs_heterogeneous
+):
+    """Create basic variables for all households.
+
+    Basic variables are variables which: - are important to differentiate the
```
hmgaudecker, line 82: Did autoformatting mess something up here?

src/_gettsim/synthetic.py
```diff
@@ -1,436 +1,211 @@
...
-            gen_female,
-            age_children,
-            baujahr,
-            double_earner,
-            policy_year,
-            bruttolohn_m=kwargs.get("bruttolohn_m", 2000.0),
-        )
+    # Identify number of households and individuals per household
+    if len(specs_heterogeneous) > 0:
+        n_households = len(specs_heterogeneous[list(specs_heterogeneous.keys())[0]])
```
hmgaudecker, line 110: Is there any way to make this more readable?

src/_gettsim/synthetic.py
```diff
@@ -1,436 +1,211 @@
...
-        hh_typen = ["single", "couple"]
-    if n_children is None:
-        n_children = [0, 1, 2]
-    if age_adults is None:
-        age_adults = [35, 35]
-    if gen_female is None:
-        gen_female = [False, True]
-    if age_children is None:
-        age_children = [3, 8]
+    hh_typ_string = f"{'single' if n_adults == 1 else 'couple'}_{n_children}_children"
```
hmgaudecker, line 106: Please replace this instance and the same code in line 167 by a function.

src/_gettsim_tests/test_synthetic.py
```diff
@@ -1,50 +1,120 @@
...
+import pytest
 from _gettsim.synthetic import create_synthetic_data
 
-from _gettsim_tests._helpers import cached_set_up_policy_environment
 
-
-def test_synthetic():
-    """Test creation of synthetic data."""
-    # run with defaults
+def test_default():
```
hmgaudecker, line 8:
```suggestion
def test_with_defaults():
```

Otherwise I expect the default to be tested not to change or so; these are rather sanity checks.

src/_gettsim_tests/test_synthetic.py
hmgaudecker, line 1: Why 4+ asserts in one test? If the first one fails, I do not get information about the remaining ones. Would prefer one assert statement per test, probably better to work with fixtures then in order to reduce overhead?

hmgaudecker (commented):
Very nice, much clearer than before! Just a smaller things.

hmgaudecker (approved):
Excellent, thank you!

hmgaudecker:
Only realised ex post that docs are failing because some notebook seems to be using the old interface of `create_synthetic_data`, please fix that before merging, should be trivial.

ChristianZimpelmann:
Thanks!

See my comment above:

> - only open ToDo is the adjustments of tutorials. I would only address this after we fixed #570.

The documentation is currently incomplete. Easiest to fix is probably to revert to the state before the change of pre-commit hooks. So we shouldn't make any changes to the current version on main.

hmgaudecker:
Hope that last commit will do, this brought all print-statements back in and changed the hooks.

ChristianZimpelmann:
Thanks for fixing the tutorials! I updated them now using the new interface of create_synthetic

docs/tutorials/advanced_usage.ipynb
```diff
@@ -235,12 +235,13 @@
...
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
     "#### Error Messages: Missing Inputs\n",
     "\n",
-    "However, if we have failed to add a required column, the function throws an error with a message that specifies which columns are missing. For example, the variable `bruttolohn_m` holds information on monthly gross earnings and is required to compute child benefits. Dropping it from the data triggers the error."
+    "However, if we have failed to add a required column, the function throws an error with a message that specifies which columns are missing. For example, the variable `arbeitsstunden_w` holds information on weakly working hours and is required to compute child benefits. Dropping it from the data triggers the error."
```
hmgaudecker, line 244:
```suggestion
    "However, if we have failed to add a required column, the function throws an error with a message that specifies which columns are missing. For example, the variable `arbeitsstunden_w` holds information on weekly working hours and is required to compute child benefits. Dropping it from the data triggers the error."
```

docs/tutorials/advanced_usage.ipynb
```diff
@@ -325,72 +326,13 @@
...
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "kindergeld.kindergeld_m_hh.value_counts(normalize=True).round(2)"
+    "In this section we will compute the *Kindergeld* for example households. Since households can also claim a tax credit instead of the child benefit (the *Kinderfreibetrag*), we will also compute the income taxes for each household. By default, GETTSIM chooses the financially more favorable option for each case. The results will thus let us inspect how the policy affects different income levels in our data. "
```
hmgaudecker, line 335:
```suggestion
    "In this section we will compute lump-sum child benefits (*Kindergeld*) for example households. Since households can also claim a tax credit (*Kinderfreibetrag*) instead of the child benefit, we will also compute the income taxes for each household. By default, GETTSIM chooses the financially more favorable option for each case. The results will thus let us inspect how the policy affects different income levels in our data. "
```

hmgaudecker (approved):
Excellent, thanks! Just a couple of typos / language details, will implement myself.

PR #517, opened 2023-02-01
A couple of renamings to be consistent with current naming convention.

ChristianZimpelmann:
A couple of renamings to be consistent with current naming convention.


hmgaudecker (approved):
Excellent, thank you!

PR #518, opened 2023-02-02
Support unmarried parents

ChristianZimpelmann:

- Currently, GETTSIM does not support unmarried parents as children are as expected as part of a tax unit.
- This problem will be handled in #509 more generally.
- In the meantime, this PR adds support for unmarried parents by introducing a new grouping "Bedarfsgemeinschaft" which consists of up to 2 adults and their dependent children. Children are now expected in separate tax units, but in the same bedarfsgemeinschaft as their parents.
- It is unclear whether this PR will be merged at some point or just over superseded by #509.

If it is ever merged, the following ToDos should be considered:

- [ ] Calculate ALG 2 on bg-level (currently hh-level)
- [ ] Add function calculating income taxes per person. This variable is used for several transfers. Even better would be the following alternative: Implement an easy way to aggregate from tu-level to bg- or hh-level


hmgaudecker (commented):
Just to raise awareness that before doing something like this in main, we will need to discuss. Probably the naming `eink_abhängig_beschäftigt` is not enough; for the purposes of Einkommensteuer it is well conceivable that Werbungskosten > Einkommen (expensive training programmes, ...)

ChristianZimpelmann:
Just to not forget: Based on this PR, it might be possible to get rid of input variable `kind`.

ChristianZimpelmann:
all changes of this branch are implemented on `main` now.

PR #519, opened 2023-02-02
Improve description of eink_vermietung_m in input_variables.md

ChristianZimpelmann:


docs/gettsim_objects/input_variables.md
```diff
@@ -147,7 +147,7 @@ Type: float
 
 ## `eink_vermietung_m`
 
-Monthly rental income
+Monthly rental income net of maintenance expenses
```
hmgaudecker, line 150:
```suggestion
Monthly rental income net of deductions.
```

Better not get started enumerating all possibilities...

hmgaudecker (changes requested):

docs/gettsim_objects/input_variables.md
ChristianZimpelmann: Sounds good. But the other variable description also do not have a period at the end. Should we change it?

hmgaudecker: No worries, all good! Did not check for consistency.

Issue #499, opened 2023-01-23
BUG: Revise ALG 2 Regelsatz for adult children


ChristianZimpelmann:
### Bug description

Regelsatz for children (`arbeitsl_geld_2_kindersatz_m_hh_ab_2011`) is currently calculated as:

```
    # Sum payments for each age group
    out = (
        arbeitsl_geld_2_params["regelsatz"][6] * anz_kinder_bis_6_hh
        + arbeitsl_geld_2_params["regelsatz"][5] * anz_kinder_ab_7_bis_13_hh
        + arbeitsl_geld_2_params["regelsatz"][4] * anz_kinder_ab_14_bis_24_hh
    )
```
However, Regelbedarfsstufe seem to be only relevant until the age of 18. For older children, Regelsatz 3 is relevant if I am not mistaken. See:

- [§ 20 V SGB II.](https://www.gesetze-im-internet.de/sgb_2/__20.html#:~:text=%C2%A7%2020%20Regelbedarf%20zur%20Sicherung,pers%C3%B6nliche%20Bed%C3%BCrfnisse%20des%20t%C3%A4glichen%20Lebens.)
- [Wikipedia](https://de.wikipedia.org/wiki/Regelbedarf)

@Eric-Sommer, @mjbloemer, can you confirm this?

Eric-Sommer:
I agree 👍 . I have no idea why it's implemented this way.

PR #525, opened 2023-02-09
address bug in Regelsatz for ALG2.

LauraGergeleit:

Addresses issue https://github.com/iza-institute-of-labor-economics/gettsim/issues/499. Currently, for the years since 2011, children aged 18 or older receive Regelsatz 4, even tough it is specified in the law that they should receive Regelsatz (see [here](4. Personenkreis nach § 20 Abs. 2 Satz 2 SGB II, 4. Personenkreis nach § 20 Abs. 2 Satz 2 SGB II).

This bug is now fixed and probably stems from the old regulation before 2010, where the same portion of the Regelsatz was applicable for children aged 14-24.


ChristianZimpelmann (approved):
Looks good. Thanks!

@hmgaudecker, `ruff` seems to have become more restrictive than before. Is this expected?

hmgaudecker:
Yeah, just merge anyhow, will ignore those guys/fix some in a separate PR.

PR #526, opened 2023-02-10
Correcting timing of Bürgergeld reform

LauraGergeleit:

Addressing issue #508. So far, the changes stemming from the Bürgergeld reform were all implemented to be starting from `2023-01-01`, however, the articles come into effect at different times, see https://offenegesetze.de/veroeffentlichung/bgbl1/2022/51#page=59
The changes to `eink_anr_frei` and `eink_anr_frei_kinder` merely start from `2023-07-01`


src/_gettsim_tests/test_arbeitsl_geld_2.py
```diff
@@ -75,7 +75,7 @@
 ]
 
 
-YEARS = [2005, 2006, 2009, 2013, 2018, 2019, 2022, 2023]
+YEARS = [2005, 2006, 2009, 2013, 2018, 2019, 2022, 2024]
```
hmgaudecker, line 78: Given how much this has changed at different points during the year, I believe we should be more fine-grained in the tests. 

I.e., instead of years, go for dates (or months) and have one test for January 2023 and another one for July 2023. 

Who knows what will change in 2024, so the above is not future-proof for sure.

hmgaudecker (commented):
Looks great! 

I do think we need to adjust the tests a little bit, though.

src/_gettsim_tests/test_arbeitsl_geld_2.py
ChristianZimpelmann: Sounds good. It shouldn't be too difficult to change the `year` column in the `.csv` to a `date` column which can contain entries like `2023-07`.

ChristianZimpelmann (approved):
Looks great! Thank you!

PR #527, opened 2023-02-10
Update hooks / files.

hmgaudecker:
In #521, ruff added some stuff that led to pre-commit failures. 

This PR adds patterns to ignore and fixes some code.

Also renamed `unterhaltsv.py` -> `unterhaltsvors.py` in accordance with function / parameter identifier.


lars-reimann:
Since the DAG approach of GETTSIM requires functions to list all values they depend on as parameters, it seems like `PLR0913` (too many arguments) can usually only be fixed by adding `noqa: PLR0913` to the function. Thus, I'd suggest to disable this rule globally.

hmgaudecker:
Yes, I definitely think we should do that down the road, particularly once we blow up interfaces further by taking apart the params dictionaries.

In the short run, however, I would like to keep it and exclude stuff only on a per-file basis; good to be reminded of this every now and then particularly for the functions that are not part of the DAG itself.

ChristianZimpelmann (approved):
I would also prefer deactivating the error globally instead of adding all the `noqa: PLR0913` comments. But feel free to decide otherwise.

hmgaudecker:
Cool, thanks to both of you again! In the end it does not matter much whether we I do the search and replace now or downstream, so I added #531 to not forget about it. Very good point!

PR #530, opened 2023-02-15
Fix income of marginally employed for Wohngeld

ChristianZimpelmann:

- Make sure that income of marginally employed is taken into account for Wohngeld (after deduction of Werbungskosten)
- Add an appropriate test case (based on the deviation with ifo)
- While on it, simplify calculation of `eink_rente_zu_verst` and `eink_rente_zu_verst_m`


hmgaudecker (approved):
Excellent, thank you!

Issue #333, opened 2022-01-19
Treatment of helper functions


ChristianZimpelmann:
### Current and desired situation

Most functions that are defined in the several function modules are written such that they are called using the DAG.

Other functions ("helper functions") are, however, called from within other functions directly (e.g. `_st_tarif` or `_berechne_vorsorge_bis_2004`). 

For those functions, the used arguments are not restricted to names of other functions and input variabels (as it is the case for all other functions). #314 introduces a tests that checks this restriction (`test_all_arguments_inputs_params_function_names`). It currently ignores all functions starting with an underscore. The same PR also made sure that all "helper functions" start with an underscore. This seems a good fix for now, but is not optimal, especially if the number of internal functions for which this restriction should be checked increases.

AFAIK, when building the DAG, we do not differentiate between the typical functions and helper functions.

### Proposed implementation

Move all helper functions to a separate file (or several separate files).

### Considered alternatives

Make some naming convention to identify functions that should not be part of the DAG.

hmgaudecker:
Could we take stock so we know how many we are talking about? The two examples are quite different:

- `st_tarif` is just a tiny wrapper around piecewise polynomial, I'd be happy to move it anywhere (btw, the docstring is still taken directly from IZAΨMOD, I believe, and should be changed / deleted)
- `berechne_vorsorge_bis_2004` just seems to be there in order to implement some complex conditional statements. I would hope that this can just go by suitable refactoring; it definitely should continue to be in the `vorsorge` module.

More generally, I am not sure about the use of `test_all_arguments_inputs_params_function_names`. Is this not automatically checked as part of the DAG generation? That is, does the test not simply add a maintenance burden by trying to replicate the DAG-generation to some extent?

ChristianZimpelmann:
Those two are indeed currently the only ones. But I could imagine that it might sometimes be helpful to have these helper functions to follow the DRY principle. I suggest we leave it as it is right now and discuss it at some later point when we have a bigger discussion about the structure of the repo.

> More generally, I am not sure about the use of test_all_arguments_inputs_params_function_names. Is this not automatically checked as part of the DAG generation?

The thing that this test checks in addition, is that all input variables are documented in `TYPES_INPUT_VARIABLES`. In line with this insight, I changed the name to `test_all_input_vars_documented` and moved it to `test_docs.py`.

> (btw, the docstring is still taken directly from IZAΨMOD, I believe, and should be changed / deleted)

Done

[...]

PR #534, opened 2023-02-16
Store data of `@add_rounding_spec` decorator in `__info__` dict

lars-reimann:

Stores the data of the `@add_rounding_spec` decorator in a dict called `__info__` that gets added to the decorated function. Other decorators like `@dates_active` (#495) or a potential `@helper` (#333) can add their data to this dict. 

This was originally intended to make it easier to copy the data around when the function gets wrapped, e.g. by other decorators (see [this discussion](https://github.com/iza-institute-of-labor-economics/gettsim/pull/495#issuecomment-1399223131)). However, `functools.wraps` copies the `__dict__` of the wrapped function, which includes the custom fields:

```py
from functools import wraps

def f():
    pass

f.test = "test"

@wraps(f)
def g():
    pass

g.test # Output: "test"
```

But it's still a cleaner design to have all decorator data in one place instead of spread over multiple fields, and helps in cases where `functools.wraps` isn't used.

### Todo

- [x] ~~Document PR in CHANGES.md.~~


lars-reimann:
Since this is just a minor refactoring, I omitted the entry to `CHANGES.md`.

src/_gettsim/shared.py
```diff
@@ -24,12 +24,16 @@ def add_rounding_spec(params_key):
     Returns
     -------
     func : function
-        Function with __rounding_params_key__ attribute
+        Function with __gettsim__["rounding_params_key"] attribute
 
     """
 
     def inner(func):
```
hmgaudecker, line 31: So IIUC, there could be an `@wraps(func)` just before this function definition (possibly with appropriate arguments) and we would not need the code below the line `# Remember data from decorator` anymore?

@janosg @timmens, this might be interesting for you guys, too!

hmgaudecker (commented):
Looks great, many thanks! 

Only thing I am not sure about is the identifier for the dict, i.e. currently `__gettsim__`. We had a discussion on this the other day with @janosg and others and I think our preliminary conclusion was something like `__info__` because we may expect some entries to come from upstream packages like dags, but I am not entirely sure anymore and can't find anything written. 

Janoś, can you help? Your memory tends to work better than mine...

src/_gettsim/shared.py
hmgaudecker: Although the example from the [docs](https://docs.python.org/3/library/functools.html#functools.wraps) suggests that using `@wraps` will require (hopefully only) an extra line for the `change_name` keyword of `@dates_active` (#495)?

lars-reimann: > So IIUC, there could be an `@wraps(func)` just before this function definition (possibly with appropriate arguments) and we would not need the code below the line `# Remember data from decorator` anymore?

The wraps doesn't help in this case since 
1. we are potentially creating the `dict` here the first time or adding a new entry,
2. we return `func` itself instead of a new function that calls `func` internally.

src/_gettsim/interface.py
```diff
@@ -540,10 +540,10 @@ def _round_and_partial_parameters_to_functions(functions, params, rounding):
...
-            # Make sure that rounding parameter attribute is transferred to partial
+            # Make sure any GETTSIM metadata is transferred to partial
             # function. Otherwise, this information would get lost.
-            if hasattr(function, "__rounding_params_key__"):
-                partial_func.__rounding_params_key__ = function.__rounding_params_key__
+            if hasattr(function, "__gettsim__"):
+                partial_func.__gettsim__ = function.__gettsim__
 
             processed_functions[name] = partial_func
         else:
```
lars-reimann, line 549: `functools.wraps` could _probably_ be used here, however so we don't need the `hasattr` and manual check, however.

Some test code:
```py
from functools import wraps, partial

def f(x):
    return x

f.test = "test"

g = wraps(f)(partial(f, 1))

print(g.test) # Output: test
print(g()) # Output: 1
```

lars-reimann: I don't find `partial_func = functools.wraps(function)(functools.partial(function, **partial_params))` particularly readable, though.

src/_gettsim/shared.py
janosg: @hmgaudecker are you suggesting to have our own wraps function that creates the `__info__` dict if not present and copies it otherwise?

src/_gettsim/interface.py
hmgaudecker: Definitely not! 

*(I am afraid you'll have to get used to me throwing around some wild ideas without fully understanding what's going on. Occasionally I do make a useful hit...)*

src/_gettsim/shared.py
hmgaudecker: I actually did not fully see through it, but I did think something like this would be helpful. I.e., ideally we'll want to make sure that every function in the DAG has an `__info__` dict so we can just call `func.__info__.get(key)` without additional checks. 

This should also happen for functions the user puts into the DAG.

More of a note, probably for a different PR?

lars-reimann: While we evaluate the decorators we need the `hasattr` check since the decorators can be applied in any order, so `__info__` might not exist yet. I also don't see a good way of storing their data in a place other than a field of the decorated functions.

Afterwards, however, I absolutely agree that ensuring
1. that the dict `__info__` exists,
2. that it has specific keys like `rounding_params_key`
before accessing the values gets annoying. 

I also don't like storing data with a fixed structure in dictionaries since
1. accessing it using strings is brittle (linters don't catch typos),
2. IDEs don't offer robust auto-completion or refactorings,
3. any behavior that uses the data must be defined externally, so the structure of the data becomes part of the interface.

I was thinking about creating a class `PolicyFunction` that contains the actual callable and any additional data. The `function_loader` could then collect the functions (their decorators are applied automatically) and then create `PolicyFunction` objects. Afterwards, the code could work with the created objects.

Could be something along the lines of:

```py
class PolicyFunction:
    def __init__(
        self,
        f: Callable,
        dag_key: str,
        start_date: datetime.date,
        end_date: datetime.date
    ):
        ...

    def is_active_at_date(date: datetime.date) -> bool:
        ... 
```

lars-reimann: But yes, that would be another PR. I could create an issue for it.

hmgaudecker: > But yes, that would be another PR. I could create an issue for it.

Sounds good. We should think about whether to have a base class in dags then, subclassing it here.

janosg: > I also don't like storing data with a fixed structure in dictionaries since
> 
>     1. accessing it using strings is brittle (linters don't catch typos),
> 
>     2. IDEs don't offer robust auto-completion or refactorings,
> 
>     3. any behavior that uses the data must be defined externally, so the structure of the data becomes part of the interface.
> 
> ...
I understand your points. I definitely agree with the first two, but not necessarily with the last one. I think a well designed public data interface might be just the thing we want here!

But I think the idea here is that the `__info__` dict does not have a fixed structure. Any decorator in `gettsim` or `dags` might add something to it. But (by convention) None would delete or overwrite anything that is there.

hmgaudecker: I'll tend to side with Lars here, though I shall be happy to be convinced otherwise.

In general, well-defined interfaces are a good thing. 

If we add a decorator in GETTSIM that needs a new field, we would add that to the class. 

If dags adds an attribute we don't care about, no problem, just ignore (if we already have that, it will be much more obvious we would be overwriting it).

If users want to do magic like this, they'll better do it in their own data structures than to hijack the dict.

Am I missing something?

PR #544, opened 2023-02-27
Add Kindesunterhalt

### LINKED ISSUE #541: ENH: Add Kindesunterhalt

GETTSIM does not consider alimony payments of a parent outside the household -- only Unterhaltsvorschuss which is payed to all single parents.

####  ToDo:

- Add new input variable `kindesunterhalt_m`
- Pay Unterhaltsvorschuss only if no (or not enough?) `kindesunterhalt`
- consider `kindesunterhalt` everywhere where we currently consider `unterhaltsvors_m`

LauraGergeleit:

Closes #541. Adding input variables for child alimony payments as well as alimony payments for divorced spouses. Replacing the use of `unterhaltsvors_m` with the new variable for child alimony payments. 

`unterhaltsvors_m` is only paid to a single parent in case no or not sufficient child alimony is paid, see [source](https://www.biallo.de/soziales/news/unterhaltsvorschuss/#unterhaltsvorschuss-voraussetzungen-berechnung-und-h%C3%B6he).

### Todo

- [x] Add new input variables `kind_unterh_brutto_m` and `kind_unterh_netto_m` 
- [x] Pay Unterhaltsvorschuss only if no or not enough kindesunterhalt
- [x] consider kindesunterhalt everywhere where we currently consider `unterhaltsvors_m`


LauraGergeleit:
I had to change the input variable names for `kindesunterhalt_brutto_m` and `kindesunterhalt_netto_m` as these names exceeded the maximum length of letters.

docs/gettsim_objects/input_variables.md
```diff
@@ -496,3 +496,21 @@ Months of subjection to compulsory insurance in the 5 years before claiming unem
...
+
+## `kind_unterh_netto_m`
+
+Monthly actual child alimony payments
+
+Type: float
+
+## `partner_unterhalt_m`
+
+Monthly alimony payments to divorced spouse
```
hmgaudecker, line 514:
```suggestion
Monthly alimony payments from divorced spouse
```

docs/gettsim_objects/input_variables.md
```diff
@@ -496,3 +496,21 @@ Months of subjection to compulsory insurance in the 5 years before claiming unem
 insurance
 
 Type: float
+
+## `kind_unterh_brutto_m`
+
+Monthly gross child alimony payments
```
hmgaudecker, line 502:
```suggestion
Monthly gross child alimony payments to be received
```

docs/gettsim_objects/input_variables.md
```diff
@@ -496,3 +496,21 @@ Months of subjection to compulsory insurance in the 5 years before claiming unem
...
+
+Type: float
+
+## `kind_unterh_netto_m`
+
+Monthly actual child alimony payments
```
hmgaudecker, line 508:
```suggestion
Monthly actual child alimony payments received
```

docs/gettsim_objects/input_variables.md
```diff
@@ -496,3 +496,21 @@ Months of subjection to compulsory insurance in the 5 years before claiming unem
...
+
+Type: float
+
+## `partner_unterhalt_m`
```
hmgaudecker, line 512: Just occurred to me while updating the description: Do we let this become negative in case we'll look at both payments made and received? Or should we think about more fine-grained cases?

Probably the same with kids above.

hmgaudecker (changes requested):
Super-quick and great work as always! We should clear up a misunderstanding before continuing, though.

docs/gettsim_objects/input_variables.md
ChristianZimpelmann, line 502: If we keep this variable, we need to make clearer what is meant by brutto netto here.

hmgaudecker:
Ohhhhh, I am sorry, somehow my major comment did not make it through. So the "misunderstanding" was completely incomprehensible. Apologies, not sure what happened there.

I think we'll actually need one more variable, though for most applications, `kind_unterh_tats_m` will do.

1. `kind_unterh_brutto_m` -- the amount a court decided that a child is supposed to receive by a parent not living in the same household.
2. `kind_unterh_netto_m` -- Same as `kind_unterh_brutto_m`, after applying deductions (most notably, 1/2 of `kindergeld_m`)
3. `kind_unterh_tats_m` -- The amount actually received, accounting for non-payment.
4. `unterhaltsvors_m` -- whatever is covered by the govt, usually the difference between `mindestunterhalt_m` and `kind_unterh_tats_m`, if positive (other rules may apply)

Importantly, and that is the misunderstanding I was referring to: We will definitely need  `kind_unterh_tats_m` and  `unterhaltsvors_m` everywhere we have `unterhaltsvors_m` in main. They might be treated differentially.

ChristianZimpelmann:
Thanks! What exactly do we need these variables for: `kind_unterh_netto_m`,  `kind_unterh_brutto_m`. I would think for the transfers it is enough to know how much was actually payed.

hmgaudecker:
For the applications with SOEP data, probably.

For example households and analyses of what is going on with taking into account `kindergeld` payments, unlikely.

For family economics applications analysing incentives after divorce, for sure not.

ChristianZimpelmann:
I don't quite understand yet. For sure the variables would be relevant in those analyses, e.g. to calculate the actually payed Unterhalt (pot. using some statistical or behavioral model). But do we need it in GETTSIM, i.e., do we need it as input for any of the potential targets in GETTSIM?

hmgaudecker:
The net version could be a relevant output, for example?

docs/gettsim_objects/input_variables.md
LauraGergeleit: Concerning this discussion: I found that the alimony payments paid might affect `arbeitsl_geld_2` as well as `wohngeld_m_hh` (potentially also `grunds_im_alter`) by reducing the considered income, see [SGBII](https://www.buzer.de/11b_SGB_II.htm) and [WoGG](https://www.wohngeld.org/wohngeldgesetz-wogg/paragraph18/). Thus, it might actually be necessary to have differing input variables depending on whether payments are made or received.

ChristianZimpelmann:
> The net version would be a relevant target in such analyses, for example?

Ok, makes sense. 

But for now and for this PR, I think we should focus on `kind_unterh_tats_m` and should not introduce the brutto, netto variables as new input variables. (Unless they are relevant for any of the transfers, given we have already)

My preference would even be to stick to the name `kindesunterhalt_m` as long as we do not consider the other two variables. It is inconsistent that we use `_tats` for the child variable, but not for the partner variable.

hmgaudecker:
Again, `netto` will almost never be an input and you will not need `brutto` in most applications just as you don't need the inputs for pensions in most applications. Apologies if I did not have that straight further up.

Let's do it for now we are at it; this is not hard. Takes much more time to think through stuff like this when you revisit.

Not sure about what you mean with an inconsistency?

ChristianZimpelmann:
> Again, `netto` will almost never be an input and you will not need `brutto` in most applications just as you don't need the inputs for pensions in most applications.

I just don't see why we should list it as input variable if we do not have any application for which we need it (so far). Or do I misunderstand where/how you would like to add the brutto/netto variables to GETTSIM?

> Not sure about what you mean with an inconsistency?

For the alimony payments by the partner, there might also be a difference between what the partner is required to pay and what he/she actually pays. I don't see a reason why we should consider this for child alimony, but not for partner alimony.

In general, I think this discussion makes the PR much more complicated than it should be.

LauraGergeleit:
> The discussion is complicated, yes, but the implementation seems trivial. So let's get on with it, please.
I already incorporated the changes in my last commit or is there still something missing?

hmgaudecker:
And if I am not mistaken, we do not have anything like `partner_unterhalt`, @ChristianZimpelmann ?

LauraGergeleit:
> And if I am not mistaken, we do not have anything like `partner_unterhalt`, @ChristianZimpelmann ?

Yes, I think so far `partner_unterhalt` is still missing/not implemented, but from [this](https://github.com/iza-institute-of-labor-economics/gettsim/issues/541#issuecomment-1445900785) comment, I understood that it should be added as well but maybe that was a misunderstanding

hmgaudecker:
Oh my memory... Apologies I did not check the code again. How about renaming to `ex_p_unterh_tats_m` ? 

Thinking of it, though, maybe better to open an issue for it because at present, I would think that this will be subsumed in some residual income measure (do we have `sonstiges_eink` or so?)

LauraGergeleit:
Yes, maybe it is better and clearer to leave it out in this PR and open a new issue. And yes, there is an input variable `sonstig_eink_m`, with the description: "Additional income: includes private and public transfers that are not yet implemented in
GETTSIM (e.g., BAföG, Kriegsopferfürsorge)" so it probably included alimony payments thus far

mjbloemer:
Did you consider the term "Zahlbetrag" instead of "netto"? I think this is typically used.

hmgaudecker:
> Did you consider the term "Zahlbetrag" instead of "netto"? I think this is typically used.

Speaking for myself, I was not aware of that term. But I would be worried a little about confusion with `_tats_` ?

src/_gettsim/transfers/unterhaltsvors.py
```diff
@@ -113,3 +125,28 @@ def unterhaltsvorschuss_eink_m_tu(  # noqa: PLR0913
     )
 
     return out
+
+
+def kind_unterh_netto_m(
```
hmgaudecker, line 130: I Just realised that this should go to a separate module `unterhalt.py` (no need to worry about inconsistency `unterh_` vs. `unterhalt` in this PR).

LauraGergeleit: Alright, I will change that. Will we also need separate tests for this function? I think the missing tests are the reason for the decreased coverage

docs/gettsim_objects/input_variables.md
LauraGergeleit: For this discussion on alimony payments paid and not received, I just opened a new issue (#550) as I think this should be handled in a separate PR

src/_gettsim/transfers/unterhaltsvors.py
hmgaudecker: Yeah, would be good, trivial as they may be...

hmgaudecker:
> > Did you consider the term "Zahlbetrag" instead of "netto"? I think this is typically used.
> 
> Speaking for myself, I was not aware of that term. But I would be worried a little about confusion with `_tats_` ?

@mjbloemer, would you mind commenting on the naming once more? We have

1. `kind_unterh_brutto_m` -- the amount a court decided that a child is supposed to receive by a parent not living in the same household. *Potential alternative could be `kind_unterh_anspruch_m` or the like.*
2. `kind_unterh_netto_m` -- Same as `kind_unterh_brutto_m`, after applying deductions (most notably, 1/2 of `kindergeld_m`),  *Potential alternative could be `kind_unterh_zahlbetrag_m`.*
3. `kind_unterh_tats_m` -- The amount actually received, accounting for non-payment.
4. `unterhaltsvors_m` -- whatever is covered by the govt, usually the difference between (`mindestunterhalt_m` - `kindergeld_m`) and `kind_unterh_tats_m`, if positive (other rules may apply).

With `zahlbetrag`, I'd be somewhat worried about confusion between 2. and 3., but if that happens to be the terminology, we could use it. `anspruch` might be nicer, too, for 1.. In favor of `brutto`, I kind of like the symmetry you would have in applications where you see both the parent who has to pay and the kid.

Thoughts would be greatly appreciated, I'd love to merge soon.

mjbloemer:
I would be slightly in favor of `kind_unterh_zahlbetrag_m` as it is often used. Also in "Düsseldorfer Tabelle". @stichnoth : what do you think?

`kind_unterh_anspruch_m` is also better in my opinion.

LauraGergeleit:
Now something got messed up, I will have to look at it again tomorrow, sorry!

hmgaudecker:
> Now something got messed up, I will have to look at it again tomorrow, sorry!

No worries, my fault that all these things were touching the same test data in parallel.

src/_gettsim/parameters/unterhalt.yaml
```diff
@@ -15,73 +15,61 @@ mindestunterhalt:
...
-    1: 6
-    2: 12
-    3: 18
+      Child benefit counts as the child's income, which is why it is taken into
+      account when calculating the child alimony payment. If the parent taking care of
+      the child receives the child benefit for a minor child, the other parent liable
+      for child alimony payments can deduct half of the child benefit when calculating
+      alimony.
+  reference: § 1612b BGB
+  unit: perecent
```
hmgaudecker, line 72:
```suggestion
  unit: share
```
Percent would have to be 50 / 100, I'd hope.

hmgaudecker (approved):

PR #546, opened 2023-03-06
Remove stale code (dashboard and 'dynamic_pension_data'). 

hmgaudecker:
Both are included in https://github.com/iza-institute-of-labor-economics/gettsim-code-for-picking.

Remove code that has not been maintained in a while. We store it publicly in case it turns out to be helpful at some point, but it should not be part of GETTSIM proper (just got in my way of regexing things for dates_active...)


ChristianZimpelmann (approved):
Makes a lot of sense. Looks good!

PR #551, opened 2023-03-14
Double-check wohngeld tests in 2023

### LINKED ISSUE #548: Double-check Tests for Wohngeld
Just stumbled across https://www.bmwsb.bund.de/Webs/BMWSB/DE/themen/stadt-wohnen/wohnraumfoerderung/wohngeld/wohngeldrechner-2023-artikel.html (also official tables there).

The numbers in the tests came from unofficial sources because we were too quick for official ones.

- [x] Double-check the numbers we got from unofficial sources in December 
- [x] Replace sources in the test files with official sources.

LauraGergeleit:

Closes #548. Double-checked the test cases for wohngeld in 2023 using https://www.bmwsb.bund.de/Webs/BMWSB/DE/themen/stadt-wohnen/wohnraumfoerderung/wohngeld/wohngeldrechner-2023-artikel.html. 

I could verify all results except for the testcase with 6 household members (hh_id: 14). There I get a result which is off by around 30€. The same test case with only 5 household members, however, yields the same result in the unofficial and official calculator. 

Additionally, I added rounding as I found that `wohngeld_vor_vermög_check_m_hh` and `wohngeld_nach_vermög_check_m_hh` should always be rounded to the nearest full euro, see [here]([url](https://www.wohngeld.org/wohngeldgesetz-wogg/anlage-3/)).


hmgaudecker (approved):
Excellent, thank you! 

Just in case you are touching yaml files once more, could you please follow these steps:

> Under Windows, yaml-files are by default not loaded with the correct encoding (UTF-8). This leads the pre-commit hook yamllint to erroneously detect too long lines. To fix the problem, set the Windows environment variable PYTHONUTF8 to 1. See [here](https://dev.to/methane/python-use-utf-8-mode-on-windows-212i) for more information.

(by @ChristianZimpelmann, last paragraph above [this link](https://gettsim.readthedocs.io/en/stable/gettsim_developer/how-to-contribute.html#development-workflow))

PR #552, opened 2023-03-19
Fix 2023 kindergeld period

Eric-Sommer:

The recent increase in Kindergeld became valid by 2023-01-01, as stated [here](https://www.buzer.de/7_InflAusG.htm). According to the current implementation, it has been valid since 2022-01-01. This is corrected here.


mjbloemer (approved):

hmgaudecker:
Great catch, thanks!

PR #565, opened 2023-03-29
Autodoc to Autodoc2

mj023:

Replace Autodoc with Autodoc2 to make the docstrings compatible with markdown. Then discuss if the docstring style should be changed to Sphinx style or if we should wait until Autodoc2 supports a different style.

### Todo

- [ ] Replace Autodoc with Autodoc2
- [ ] Change Docstrings to Sphinx Style


hmgaudecker:
When merging `main`, I reverted the changes away from NumPy-style docstrings in `interface.py`.

In https://github.com/sphinx-extensions2/sphinx-autodoc2/issues/33 there is code for getting NumPy-style docstrings to work in autodoc2. Let's hope it will be integrated there soonish, else we should be able to adapt it locally.

hmgaudecker:
Closing this since I autodoc2 seems to be moving along fairly slowly, so I don't think we want to add it as a dependency. We probably won't need too much rst in the vast majority of docstrings anymore, anyhow.

PR #576, opened 2023-04-27
Filter `PolicyTestSet`

lars-reimann:

* Add a function to filter a `PolicyTestSet` by `test_name` and/or `date`.
* Rename `household_name` to `test_name`.

See also: [Zulip discussion
](https://gettsim.zulipchat.com/#narrow/stream/224837-High-Level-Architecture/topic/tests/near/353133153)

### Todo

- [X] Document PR in CHANGES.md. (omitted since changes are not user-facing)

### DIFF
--- a/src/_gettsim_tests/_policy_test_utils.py
+++ b/src/_gettsim_tests/_policy_test_utils.py
@@ -30,8 +30,11 @@ class PolicyTestSet:
         return pd.concat([test.output_df for test in self.test_data], ignore_index=True)
 
     def filter_test_data(
-        self, *, test_name: str | None = None, date: datetime.date | None = None
+        self, *, test_name: str | None = None, date: datetime.date | str | None = None
     ) -> PolicyTestSet:
+        if isinstance(date, str):
+            date = _parse_date(date)
+
         filtered_test_data = [
             test
             for test in self.test_data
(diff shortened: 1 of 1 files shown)


src/_gettsim_tests/_policy_test_utils.py
```diff
@@ -29,22 +29,34 @@ def merged_input_df(self) -> pd.DataFrame:
     def merged_output_df(self) -> pd.DataFrame:
         return pd.concat([test.output_df for test in self.test_data], ignore_index=True)
 
+    def filter_test_data(
+        self, *, test_name: str | None = None, date: datetime.date | None = None
+    ) -> PolicyTestSet:
```
hmgaudecker, line 34: I think a docstring with an example would be very helpful here, many of our core contributors won't be too familiar with type hints, wildcards, etc.

hmgaudecker (approved):
Looks good to me, maybe except for documenting the filter a little.

@ChristianZimpelmann, care to try it out?

ChristianZimpelmann (commented):
Thanks!

This is very helpful.

This call works now fine:

`data.filter_test_data(date=datetime.date(2019, 1, 1)).parametrize_args`

- It is still a bit inconvenient that I need to specify the date as `datetime` object instead of a string. But I could live with it.

src/_gettsim_tests/_policy_test_utils.py
lars-reimann: I've added a docstring.

lars-reimann:
> Thanks!
> 
> This is very helpful.
> 
> This call works now fine:
> 
> `data.filter_test_data(date=datetime.date(2019, 1, 1)).parametrize_args`
> 
> * It is still a bit inconvenient that I need to specify the date as `datetime` object instead of a string. But I could live with it.

You can now pass a string as well.

ChristianZimpelmann (approved):
Great! Thanks a lot!

PR #577, opened 2023-04-28
Add Erwerbsminderungsrente

nafetsk:
Implementation of pension for reduced earning capacity (Erwerbsminderungsrente) as addition to the pension system already existing in gettsim. 

The special feature of the pension for reduced earning capacity are the additional earning points. Average earning points per year are calculated for every individual and attributed for the time between retirement and an allocation age limit (Zurechnungszeitsgrenze). There are also special rules for (Zugangsfaktor) and (Rentenfaktor) both relevant for the calculation of the total pension claim. 

Three new input variables are introduced.

- `voll_erwerbsgemind`: Unable to provide more than 3 hours of market labor per day.
- `teilw_erwerbsgemind`: Able to provide at least 3 but no more than 6 hours of market labor per day.
- XXX


src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,224 @@
...
+    -------
+    Erwerbsminderungsrente pension claim
+
+    """
+
+    if anspruch_erwerbsm_rente:
+        out = (
+            entgeltp_erwerbsm_rente
+            * erwerbsm_rente_zugangsfaktor
+            * rentenwert
```
amageh, line 126: How is the distinction of Rentenwert Ost/West implemented in GETTSIM? I think technically we need two separate variables for pension points that were earned in the West and East respectively as these are each multiplied by a different Rentenwert.

amageh (commented):
Nice work, I left some comments that we can discuss later :-).

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,224 @@
...
+        See params documentation :ref:`ges_rente_params <ges_rente_params>.
+    erwerbsm_rente_params
+        See params documentation :ref:`erwerbsm_rente_params <erwerbsm_rente_params>.
+
+    Returns
+    -------
+    Zugangsfaktor for Erwerbsminderungsrente (pension for reduced earning capacity)
+
+    """
+    zugangsfaktor = (
```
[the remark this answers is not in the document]
nafetsk: I decided for the yearly basis because gettsim is doing it that way for the normal pension and so I could use the same parameter -> ges_rente_params["zugangsfaktor_veränderung_pro_jahr“]. But I think it would be no big deal to change that.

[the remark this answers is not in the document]
hmgaudecker: Should not matter for precision (or is it fixed to integers?) and implementation should be as close as possible to the law / regulations.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,79 @@
...
+    scalar: 780
+zurechnungszeitsgrenze_in_m:
+  name:
+    de: Altersgrenze für die Zurechnungszeit der Erwerbsminderungsrente
+    en: null
+  description:
+    de: >-
+      Anhebung von 2020 bis 2028 um 1 Monat pro Kalendarjahr und ab 2028
+      jeweils 2 Monate bis 67 erreicht ist. Nicht ganz sicher seit wann
+      60(720)Jahre gilt.
```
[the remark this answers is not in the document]
nafetsk: Not sure how to implement that. Making an entry for every month in the yaml file seems wrong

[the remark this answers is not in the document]
hmgaudecker: Have a look at the phasing-in of "Rente für besonders langjährig Versicherte". If in doubt, please ping @TeBackh who implemented that.

Thanks!

src/_gettsim/transfers/erwerbsm_rente.py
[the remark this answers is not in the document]
amageh: That's fine then, thank you for the answers!

hmgaudecker:
This will be a very large PR.

It would be helpful to move things that affect other dimensions than Erwerbsminderungsrente to separate Pars.

E.g. if we need to differentiate the stock of points earned by geographic region, we should do so in two stock variables, not one stock, one share. Doing so will affect all pension payments (I suppose) and issues related to that can be discussed without reference to Erwerbsminderungsrente.

nafetsk:
> This will be a very large PR.
> 
> It would be helpful to move things that affect other dimensions than Erwerbsminderungsrente to separate Pars.
> 
> E.g. if we need to differentiate the stock of points earned by geographic region, we should do so in two stock variables, not one stock, one share. Doing so will affect all pension payments (I suppose) and issues related to that can be discussed without reference to Erwerbsminderungsrente.



> This will be a very large PR.
> 
> It would be helpful to move things that affect other dimensions than Erwerbsminderungsrente to separate Pars.
> 
> E.g. if we need to differentiate the stock of points earned by geographic region, we should do so in two stock variables, not one stock, one share. Doing so will affect all pension payments (I suppose) and issues related to that can be discussed without reference to Erwerbsminderungsrente.

Sorry for the late response. You are right, the differentiation between EPs should be discussed on another PR. I removed the last commit

TeBackh:
I am wondering about the variable "m_beitragszeit". What would be included? can it be composed of other already existing variables?

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -170,37 +170,6 @@ def erwerbsm_rente_vor_grundr_m(
     return out
 
 
-@dates_active(end="2020-12-31")
-def erwerbsm_rente_m(erwerbsm_rente_vor_grundr_m: float) -> float:
-    return erwerbsm_rente_vor_grundr_m
-
-
-@dates_active(start="2021-01-01", change_name="erwerbsm_rente_m")
```
amageh: fiy @TeBackh, we decided to exclude Grundrente from this PR for now as the required changes will be more involved. @nafetsk @MaxBlesch will likely work on it in a new PR in the future.

TeBackh:
Thanks Stefan,
looks good. :) I had some requests and questions and comments -  see review. I will have another go once these are resolved

TeBackh (changes requested):

src/_gettsim/transfers/erwerbsm_rente.py
TeBackh: ok

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+    out = entgeltp + (
+        (((m_zurechnungszeitsgrenze - m_diff_zu_2001) / 12) - (age_of_retirement))
+        * durchschn_entgeltp_y
+    )
+    return out
+
+
+def rentenfaktor_erwerbsm_rente(
```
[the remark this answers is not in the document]
nafetsk: your are right it is "Rentenartfaktor"

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,215 @@
...
+    return out
+
+
+def erwerbsm_rente_zugangsfaktor(
+    erwerbsm_rente_params: dict,
+    ges_rente_params: dict,
+    age_of_retirement: float,
+    jahr_renteneintr: int,
+) -> float:
+    """Calculating Zugangsfaktor for Erwerbsminderungsrente
```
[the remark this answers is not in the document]
nafetsk: "    For each month that a pensioner retires before the age limit, 0.3% of 
    the pension is deducted. The maximum deduction is capped at 10.8%. 
    This deduction is the norm for the pension for reduced earning capacity."

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,215 @@
...
+            "intercepts_at_lower_thresholds"
+        ],
+    )
+    zugangsfaktor = 1 + (age_of_retirement - m_altersgrenze_abschlagsfrei / 12) * (
+        ges_rente_params["zugangsfaktor_veränderung_pro_jahr"][
+            "vorzeitiger_renteneintritt"
+        ]
+    )
+    out = max(zugangsfaktor, erwerbsm_rente_params["min_zugangsfaktor_erwerbsm_rente"])
+
```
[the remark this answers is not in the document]
TeBackh: this seems resolved too

TeBackh (approved):
Thanks, looks good. All my comments have been resolved. I guess it would be Hans-Martins turn now.

amageh:
> Thanks, looks good. All my comments have been resolved. I guess it would be @hmgaudecker 's turn now.

Thank you so much for the thorough review, @TeBackh! @hmgaudecker can you give it a final go before we merge?

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,315 @@
...
+def ges_rente_vorauss_erwerbsm(
+    erwerbsgemindert: bool,
+    m_pflichtbeitrag: float,
+    ges_rente_wartezeit_5: bool,
+) -> bool:
+    """
+    Determine eligibility Erwerbsminderungsrente
+    (pension for reduced earning capacity)
+
+    Requirements are:
```
hmgaudecker, line 56: This section can go. It is mostly documenting other functions, which are documented elsewhere. So better to look them up directly.

The only relevant thing would be the exception in `m_pflichtbeitrag`. But I'd actually be in favor of a separate input variable for that one. If someone is comfortable to take the shortcut that you did here, it is trivial to do so outside of GETTSIM.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,315 @@
...
+        erwerbsgemindert and ges_rente_wartezeit_5 and m_pflichtbeitrag >= 36
+    )
+
+    return anspruch_erwerbsm_rente
+
+
+@dates_active(start="2004-01-01")
+def entgeltp_erwerbsm_rente(
+    entgeltp: float,
+    durchschn_entgelpt: float,
```
hmgaudecker, line 89:
```suggestion
    durchschn_entgeltp: float,
```

(and elsewhere)

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,315 @@
...
+    erwerbsm_rente_params: dict,
+    jahr_renteneintr: int,
+    age_of_retirement: float,
```
hmgaudecker, line 92: Seems like it was not introduced here, but seeing 

```py
    jahr_renteneintr: int,
    age_of_retirement: float,
```

next to each other makes me wonder why we use different languages here?

And why would that not be just current age? Because it will differ over the course of the year? Touches on a discussion we just had with @TeBackh -- I don't think we can try to calculate `erwerbsminderungsrente` in 2010 for someone who took it up in 2003.

hmgaudecker (commented):
Thanks!!! 

Just some high-level comments for now, can't continue for a couple of days or so. I think they should be worth discussing, though.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,128 @@
...
+  description:
+    de: Bis 2012 immer 63 ab 2012 Anhebung bis 65
+    en: null
+  unit: year
+  reference: >-
+    § 77 SGB VI Zugangsfaktor:
+    https://sozialversicherung-kompetent.de/rentenversicherung/
+    leistungsrecht/295-erwerbsminderungsrente-und-rentenabschlaege.html
+  type: piecewise_linear
+  2001-01-01:
```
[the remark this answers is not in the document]
amageh: We chose to start the implementation in 2001 because the Erwerbsminderungsrente was introduced in a reform that year, adding the system before that (Berufsunfähigkeitsrente) would be a task for a new PR in my opinion.

Not sure if we should maybe add a comment about this and add a `dates_active` decorator to the function `erwerbsm_rente_m`?

src/_gettsim/transfers/erwerbsm_rente.py
amageh: This goes back to what we discussed [here](https://github.com/iza-institute-of-labor-economics/gettsim/issues/629#issuecomment-1665713136) I think. The Zurechnungszeit depends on age but also the date of the pension start since different amounts of Zurechungszeit are granted depending on whether the pension started before 2014, between 2014 and 2018, between 2018 and 2019, and after 2019. 

I agree on the naming issue, I guess since these variables are not directly institutional names I would suggest going with English?

nafetsk: 'jahr_renteneintr' is a basic input variable. Since all input variables have german names I think it would be better to switch to german instead.

src/_gettsim/parameters/erwerbsm_rente.yaml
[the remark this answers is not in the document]
hmgaudecker: Yes, please. Otherwise we'll have a wrong policy environment pre-2001

src/_gettsim/transfers/erwerbsm_rente.py
hmgaudecker: Yes, German sounds better.

But I still don't think we'll need that. What the function tries to do, say, is to calculate for the policy year 2023 the disability insurance benefit of somebody who took it up in 2015. We won't be able to do that consistently. In the policy year 2023, we can only reliably calculate the pension of people who retire in 2023.

The way to go about these dynamic settings will be to call GETTSIM repeatedly for different policy years. Say, if first year is 2015:

- 2015 for everyone
- 2016 once for people who took up DI pension in 2015 (pension level fixed); once for everyone else
- 2017 once for people who took up DI pension before 2017 (pension level fixed); once for everyone else
- ...

amageh: I am not sure I understand why we will not be able to do this consistently. If we have information on the retirement date, age at retirement, and pension points, this should be enough to compute entitlements. The only thing that will change over time with the policy environment is the Rentenwert. 

Can you maybe elaborate a bit more?

hmgaudecker: The thing is that we want to get rid of some functions when they become irrelevant. Think of `berufsunfähigkeitsrente` being implemented already. Will be gone from 2001 onwards as per `@dates_active`.

So this would work for the current use case, but it is not future-proof.

amageh: Is it possible to find another workaround? If I understand correctly, this would mean we can only evaluate policies that affect retirees before or at the start of their pension, right? 

I think that would be quite unfortunate. For instance, there is an upcoming reform for 2024 that will increase disability benefits for individuals who didn't benefit from the reforms in the past decade. I am not sure how we would be able to implement that reform without having access to the age and date of retirement.

hmgaudecker: Of course that will be possible! 

In general, there will be a little more work on the user side (=call GETTSIM repeatedly), that's all.

If there is a new reform that requires something like that as a state variable, we will need it only from 2024. Will be interesting to see how "individuals who didn't benefit from the reforms in the past decade" shall be identified...

TeBackh: Maybe we should leave that change in the structure to a new PR, as it seems to affect many other components of the current model (all non em pensions ). 
Also wondering whether for now, where GETTSIM starts only in 1984 for most policies it would be preferrable to have at least a good approximation for pensioneers from earlier years rather than nothing.

hmgaudecker: But those are unrelated changes, except they all are in the context of pensions? No change in structure required elsewhere in this PR. We should tackle those eventually, but not here.

Anything before 2005 needs a serious look, anyhow -- I don't think GETTSIM is of much use currently for that period. 1984 as a starting date should mostly be a placeholder, maybe someone put it into IZAΨMOD because the SOEP started at that time. 

Not sure what you mean with approximations in this context?

TeBackh: > Not sure what you mean with approximations in this context?

I mean that the currently implemented pension thresholds are mostly true also for people who retired before 1984 (as we are testing them for cohorts  from 1900 with this sheet from the seibold paper). So we would be able to calculate pretty precise pensions for people who retired before 1984.

hmgaudecker: Yes, sure, but what speaks against calling GETTSIM repeatedly then? Just use years before 1984, if the only output you request are pensions and all state variables are there, that should be no problem. Put differently, the starting date for 1984 in this case seems arbitrary and it seems like it can be set to an earlier value.

TeBackh: Thats right. We should do this. 

Still somehow wondering whether we want the user to call gettsim 35 times when they have people with retirement entry years across a range of 35 years. but yeah thats how it is then

hmgaudecker: Just a simple loop -- whether you call it twice or 35 times doesn't matter :smile: 

But yes, there is no denying that the origins of GETTSIM are in cross-sectional work. As we discussed in person the other day, given this structure there simply is no way of scaling dynamic stuff up as a one-stop experience.

Put differently, to keep things manageable, we always depict a policy environment as it is active at one particular date. Trying to achieve calculating the pension for someone who retired years ago would change this and it is simply not maintainable in the medium run.

TeBackh: Maybe last note on this (perhaps obvious): We can only take over the zugangfsfaktor and Engeltpunkte from the year of retirement to subsequent years as state vars and we need to "freshly" calc the actual pension for every output year also for people who retired years ago, as the Rentenwert changes every year.

MaxBlesch: We were discussing this among us Berliners and are wondering why we shouldn't implement it in a cross sectional way. Would it be possible to have a Zoom Call next week on this and discuss that? Also I agree with @TeBackh earlier comment, that we should handle this in a different PR.

hmgaudecker: What do you mean by "implementing it in a cross-sectional way"? That's how I'd think of what I am suggesting. 

Again, it works for a simple case. But if functions cease to exist, you need to carry them around for as long as somebody might still be alive even if they retired 70 years ago. No loss in functionality, just a little more work for the user (which will be much less under updated interface).

hmgaudecker: Early Monday morning (finish by 10am) would work for a call, I am blocked with a coauthor for most of the remainder of the week.

MaxBlesch: > What do you mean by "implementing it in a cross-sectional way"? That's how I'd think of what I am suggesting.
> 
> Again, it works for a simple case. But if functions cease to exist, you need to carry them around for as long as somebody might still be alive even if they retired 70 years ago. No loss in functionality, just a little more work for the user (which will be much less under updated interface).

If you call it from 2015 up 2023 to calculate pension for someone who retired 2015, you will need to have the person in the dataset for all these years, i.e. a panel structure.Therefore I called it cross-sectional. But maybe I am just understanding it in the wrong way. 



> Early Monday morning (finish by 10am) would work for a call, I am blocked with a coauthor for most of the remainder of the week.

I am in the US for a research stay now, so that will be too early for me. Maybe we can arrange that in a quieter week and postpone these changes until then and do these in a separate PR.

hmgaudecker: I really do not want to merge something that is due to be changed very soon again. I find the current structure simply too opaque to be able to review it in finite time, I am sorry. 

This is just about simplification, not about any changes that would take serious time.

hmgaudecker: > If you call it from 2015 up 2023 to calculate pension for someone who retired 2015, you will need to have the person in the dataset for all these years, i.e. a panel structure.Therefore I called it cross-sectional. But maybe I am just understanding it in the wrong way.

Oh yeah, I was thinking from a GETTSIM perspective, you were thinking from a data perspective. All good.

Again, that case is supported, just a little bit of extra work. Just call GETTSIM for 2015 with the relevant variables measured *for the year 2015* (you just happen to label the data 2023, but they really are for the year of retirement).

Just to expand from @TeBackh and my conversations on this: In the structure you envision, we just do not see a way to consistently deal with pension credits earned *after* the first time a pension was drawn. You'd have to keep track of the entire history between 2015 and 2023.

CHANGES.md
```diff
@@ -6,13 +6,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
...
 ## Unpublished
 
-- {gh}`577` Implement pension for reduced earning capacity ({ghuser}`nafetsk`)
-- {gh}`590` Add allowance for child income for Kinderzuschlag.
-  ({ghuser}`ChristianZimpelmann`).
-- {gh}`624` Don't create functions for other time units if this leads to a cycle in the
-  graph ({ghuser}`lars-reimann`).
-- {gh}`620` Adds income from self employment to income for Bürgergeld ({ghuser}`amageh`,
-  {ghuser}`lillyfischer`, {ghuser}`TeBackh`)
+- {gh}`577` Implement Erwerbsminderungsrente (public disability insurance)
```
amageh, line 9: I think some changes were accidentally deleted here?

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
 ---
 altersgrenze_abschlagsfrei:
   name:
+    de: Regelaltersgrenze
+    en: Standard retirement age
```
amageh, line 5:
```suggestion
    en: Statutory retirement age
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
-    de: Bis 2012 immer 63 ab 2012 Anhebung bis 65
-    en: null
-  unit: year
-  reference: >-
-    § 77 SGB VI Zugangsfaktor:
-    https://sozialversicherung-kompetent.de/rentenversicherung/
-    leistungsrecht/295-erwerbsminderungsrente-und-rentenabschlaege.html
-  type: piecewise_linear
+      Anspruch nehmen kann. Bis 2012 lag die Regelaltersgrenze bei 63. Ab dann
+      Schrittweise Anhebung bis 65.
```
amageh, line 10:
```suggestion
      schrittweise Anhebung bis 65.
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+    en: >-
+      Age limit up to which one can claim the public disability insurance  without
+      deductions. Up to 2012, the standard retirement age was 63.
+      From then on Gradual increase until 65.
```
amageh, line 14:
```suggestion
      From then on gradual increase until 65.
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
-  unit: year
-  reference: >-
-    § 77 SGB VI Zugangsfaktor:
-    https://sozialversicherung-kompetent.de/rentenversicherung/
-    leistungsrecht/295-erwerbsminderungsrente-und-rentenabschlaege.html
-  type: piecewise_linear
+      Anspruch nehmen kann. Bis 2012 lag die Regelaltersgrenze bei 63. Ab dann
+      Schrittweise Anhebung bis 65.
+    en: >-
+      Age limit up to which one can claim the public disability insurance  without
```
amageh, line 12:
```suggestion
      Age limit up to which one can claim the public disability insurance without
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+  2022-01-01:
+    scalar: 64.666666
+  2023-01-01:
+    scalar: 64.833333
+  2024-01-01:
+    scalar: 65
 altersgrenze_abschlagsfrei_63:
   name:
+    de: Regelaltersgrenze für langjährige Versicherte
+    en: Standard retirement age for long-term insured persons
```
amageh, line 61:
```suggestion
    en: Statutory retirement age for long-term insured persons
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
-      Anspruch nehmen kann wenn man bestimmte zeitliche Bedingungen erfüllt
-    en: null
-  description:
-    de: null
-    en: null
-  unit: year
-  reference: >-
-    § 77 Abs. 4 SGB VI Zugangsfaktor:
+      Anspruch nehmen kann wenn man bestimmte zeitliche Bedingungen
+      erfüllt. Bediungen werden in § 77 Abs. 4 SGB VI definiert.
```
amageh, line 66:
```suggestion
      erfüllt. Bedinungen werden in § 77 Abs. 4 SGB VI definiert.
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+    en: >-
+      age limit up to which one can claim the public disability insurance
```
amageh, line 68:
```suggestion
      Age limit up to which one can claim the public disability insurance
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+      without deductions if you meet certain time conditions.
```
amageh, line 69:
```suggestion
      without deductions under certain time conditions.
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+  unit: Year
+  reference: § 77 Abs. 4 SGB VI Zugangsfaktor
   2001-01-01:
     scalar: 63
 zurechnungszeitsgrenze:
   name:
-    de: Altersgrenze für die Zurechnungszeit der Erwerbsminderungsrente
-    en: null
+    de: Zurechnungszeitsgrenze
+    en: Compensatory time limit
```
amageh, line 78:
```suggestion
    en: Supplementary time limit
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
-  reference: >-
-    https://sozialversicherung-kompetent.de/rentenversicherung/
-    leistungsrecht/295-erwerbsminderungsrente-und-rentenabschlaege.html
-    for the time before 2004 see: https://www.bgbl.de/xaver/bgbl/text.xav?SID=&
-    tf=xaver.component.Text_0&tocf=&qmf=&hlf=xaver.component.Hitlist_0&bk=bgbl&s
-    tart=%2F%2F*%5B%40node_id%3D%27955873%27%5D&skin=pdf&tlevel=-2&nohist=1&sins
-    t=23CF02AC  (Anlage 23 (p.1837))
-  type: piecewise_linear
+    en: >-
+      Age limit for the compensatory period for the public disability insurance.
```
amageh, line 85:
```suggestion
      Age limit for the supplementary period of the public disability insurance.
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -1,128 +1,232 @@
...
+      Increase from 2020 to 2027 by 1 month per year, and from 2028
+      2 months each until 67 is reached.
```
amageh, line 87:
```suggestion
      by 2 months each year until 67 is reached.
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -106,87 +95,24 @@ def entgeltp_erwerbsm_rente(
...
         See params documentation :ref:`erwerbsm_rente_params <erwerbsm_rente_params>.
-    jahr_renteneintr
-        See basic input variable :ref:`jahr_renteneintr <jahr_renteneintr>.
     age_of_retirement
         See :func:`age_of_retirement`.
 
     Returns
     -------
-    Final Entgeltpunkte for Erwerbsminderungsrente
+    Final earning points for Erwerbsminderungsrente (public disability insurance)
```
amageh, line 103:
```suggestion
    Final pension points for Erwerbsminderungsrente (public disability insurance)
```

amageh (commented):
I added some suggested changes for the translation of specific terms. Please ignore my suggestions if there already is a standard within GETTSIM and approve otherwise. Thanks!

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +546,21 @@ Type: int
 Number of own children below the age of 25 (incl. not in hh)
 
 Type: int
+
+## `erwerbsgemindert`
+
+Dummy: Beeing able to work less than 3 hours on the general labor market per day
```
hmgaudecker, line 552:
```suggestion
Unable to provide more than 3 hours of market labor per day.
```

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +546,21 @@ Type: int
...
+
+Type: bool
+
+## `teilw_erwerbsm_rente`
+
+Dummy: Beeing able to work between 3 and 6 hours on the general labor market per day
```
hmgaudecker, line 558:
```suggestion
Able to provide at least 3 but no more than 6 hours of market labor per day.
```

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +546,21 @@ Type: int
...
+
+Type: bool
+
+## `erwerbsm_rente_mit_63`
```
hmgaudecker, line 562: This is opaque to me. The only explanation is the line below; I don't know what the conditions for this would be in the first place?

Also, please use the official term rather than the colloquial one.

src/_gettsim/config.py
```diff
@@ -179,5 +181,8 @@ def set_array_backend(backend: str):
     "kind_unterh_anspr_m": float,
     "kind_unterh_erhalt_m": float,
     "steuerklasse": int,
+    "erwerbsgemindert": bool,
+    "erwerbsm_rente_mit63": bool,
```
hmgaudecker, line 185:
```suggestion
    "erwerbsm_rente_mit_63": bool,
```
Won't flag again and this should change, anyhow. But this is what was used above in the list of input variables.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,232 @@
...
+    scalar: 64.333333
+  2021-01-01:
+    scalar: 64.5
+  2022-01-01:
+    scalar: 64.666666
+  2023-01-01:
+    scalar: 64.833333
+  2024-01-01:
+    scalar: 65
+altersgrenze_abschlagsfrei_63:
```
hmgaudecker, line 58: Please use official term. Who knows for how long this will stay at 63?

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,232 @@
...
+    scalar: 66.333333
+  2028-01-01:
+    scalar: 66.5
+  2029-01-01:
+    scalar: 66.666666
+  2030-01-01:
+    scalar: 66.833333
+  2031-01-01:
+    scalar: 67
+min_zugangsfaktor_erwerbsm_rente:
```
hmgaudecker, line 201:
```suggestion
min_zugangsfaktor:
```
The params object is already called `erwerbsm_rente`, no need to double up.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,232 @@
...
+      sind bei 10.8% gedeckelt.
+    en: Deductions for early retirement are capped at 10.8%.
+  unit: Factor
+  reference: § 77 SGB VI Zugangsfaktor
+  note: >-
+    Rentenreformgesetz 1999 (RRG 1999)
+    "http://www.portal-sozialpolitik.de/index.php?page=rentenversicherung"
+  2000-01-01:
+    scalar: 0.892
+rentenartfaktor_erwerbsm_rente:
```
hmgaudecker, line 217:
```suggestion
rentenartfaktor:
```
The params object is already called `erwerbsm_rente`, no need to double up.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,232 @@
...
+  name:
+    de: Rentenartfaktor für teilweise Erwerbsminderungsrente
+    en: Pension type factor for partial reduction in public disability insurance
+  description:
+    de: null
```
hmgaudecker, line 222: Please add a description. I have no idea what a "Rentenartfaktor" is supposed to be, would have to guess.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+            * erwerbsm_rente_zugangsfaktor
+            * rentenwert
+            * rentenartfaktor_erwerbsm_rente
+        )
+    else:
+        out = 0.0
+    return out
+
+
+def ges_rente_vorauss_erwerbsm(
```
hmgaudecker, line 47:
```suggestion
def ges_rente_anspruch_erwerbsm(
```

? I would find that clearer and you are using it inside the function, too. But once we can write it out as `voraussetzung` I am fine with it (won't be long).

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+    return anspruch_erwerbsm_rente
+
+
+def entgeltp_erwerbsm_rente(
+    entgeltp: float,
+    durchschn_entgeltp: float,
+    erwerbsm_rente_params: dict,
+    age_of_retirement: float,
+) -> float:
+    """Calculating Entgeltpunkte for Erwerbsminderungsrente
```
hmgaudecker, line 83:
```suggestion
    """Entgeltpunkte which Erwerbsminderungsrente is based on
```

Correct?

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+    return out
+
+
+def durchschn_entgeltp(
+    entgeltp: float,
+    age_of_retirement: float,
+) -> float:
+    """Calculation of average earning points as part of the "Grundbewertung".
```
hmgaudecker, line 119:
```suggestion
    """Average earning points as part of the "Grundbewertung".
```

Please go through docstrings and get rid of unnecessary verbs in the beginning. It is obvious that we are doing calculations in each function *(yes, I know lots of old functions still have that...)*

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+        See basic input variable :ref:`entgeltp <entgeltp>
+    age_of_retirement
+        See :func:`age_of_retirement`.
+
+    Returns
+    -------
+    average entgeltp
+    """
+
+    beleg_gesamtzeitr = age_of_retirement - 16
```
hmgaudecker, line 137: Remove hard-coded number.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def rentenartfaktor_erwerbsm_rente(
+    teilw_erwerbsm_rente: bool,
+    erwerbsm_rente_params: dict,
+) -> float:
+    """Checking Rentenartfaktor for Erwerbsminderungsrente
+    (public disability insurance)
+
+    Partial pension - Rentenartfaktor = 0.5
```
hmgaudecker, line 150: Remove. Might change in future.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,228 @@
...
+
+def erwerbsm_rente_zugangsfaktor(
+    ges_rente_params: dict,
+    erwerbsm_rente_params: dict,
+    age_of_retirement: float,
+    erwerbsm_rente_mit_63: bool,
+) -> float:
+    """Calculating Zugangsfaktor for Erwerbsminderungsrente
+    (public disability insurance)
+    For each month that a pensioner retires before the age limit, 0.3% of
```
hmgaudecker, line 185: Please don't hard-code numbers in docstrings. Pattern is enough.

hmgaudecker (commented):
Sorry for taking so long. This is very nice now!

I do have a bunch of things that will make it easier on future maintainers & users. Just wordings, hardly any actual changes to the code.

src/_gettsim/transfers/erwerbsm_rente.py
nafetsk: ges_rente_vorauss_vorzeitig
ges_rente_vorauss_regelrente
ges_rente_vorauss_frauen
ges_rente_vorauss_langj
...
are the names of functions checking the eligibility for other pension types. I wanted to match that pattern

docs/gettsim_objects/input_variables.md
nafetsk: As far as I know, there is no technical term for this. This is an exceptional case in the calculation of the access factor. I have added a small explanatory text. Is that sufficient?

The legal foundation is
§ 77 Abs. 4  SGB VI
> (4) Bei Renten wegen verminderter Erwerbsfähigkeit und bei Hinterbliebenenrenten, deren Berechnung 40 Jahre mit den in § 51 Abs. 3a und 4 und mit den in § 52 Abs. 2 genannten Zeiten zugrunde liegen, sind die Absätze 2 und 3 mit der Maßgabe anzuwenden, dass an die Stelle der Vollendung des 65. Lebensjahres die Vollendung des 63. Lebensjahres und an die Stelle der Vollendung des 62. Lebensjahres die Vollendung des 60. Lebensjahres tritt.

hmgaudecker: How do we call the corresponding parameter in regular pensions then?

nafetsk: I think it is very similar to the rules for `langjährig und besonders langjährig Versicherte`in the regular pension.
But I have not yet seen this term beeing used for the disability pension. But I think we can adopt the name, to at least make it consistent in GETTSIM.

hmgaudecker: Oh, lovely, and it is yet another number (40 instead of 35/45), right? 

But yeah, I'd think `besond_langj` would be best. Seems most similar in spirit. Also note that the text also has the 62->60 mapping, hence the 63 doesn't quite capture it. Btw, is that relevant for us?

nafetsk: Ok I will rename that parameter accordingly.
The jump from 62 to 60 refers to the maximum deduction that is specified elsewhere in the code.
The limit is either 40 years or, in special cases, 35 years. 
> - abschlagsfreier Beginn der Rente mit Vollendung des 63. Lebensjahres, wenn 40 Jahre Wartezeit nach § 51 Absatz 3a SGB VI vorliegen oder
> - abschlagsfreier Beginn nach den Tabellenwerten des § 264d SGB VI, wenn mindestens 35 Jahre Wartezeit nach § 51 Absatz 3a SGB VI vorliegen- diese Regelung gilt aber nur noch für EM-Rente die bis zum 31.12.2023 beginnen (Auslaufregelung).

hmgaudecker: I just don't overlook whether these two points are part of the code right now? If not, it seems like we'll want to include them in this PR, too?

nafetsk: We discussed this and decided that it was too complicated to get these time variables from the data. It needs some kind of approximation which is better done by the user and not by Gettsim.

hmgaudecker: The user would have to do the approximation indeed (unless she had data from the DRV), but GETTSIM needs to know the rule. Same as everywhere?

amageh: Hi, I assumed the variable `erwerbsm_langj_vers` already indicates this approximation/exception set by the user? 

If we want GETTSIM to process different types of time variables/time thresholds that the German pension system recognizes, I think it would make sense to do this in a separate PR, as this will affect other types of pensions as well.

hmgaudecker: Oh shoot, good that you are writing this. I completely missed that. 

This is exactly how, e.g., the `ges_rente_vorauss_besond_langj` works, no? We should be symmetric here, shouldn't be hard -- the whole machinery is in place already.

I am terribly sorry I did not realise this before.

amageh: Thanks! This is my understanding as well. What exactly do you mean by symmetry: the name? In that case I would suggest renaming the variable once more to `erwerbsm_vorauss_besond_langj` in line with `ges_rente_vorauss_besond_langj`. Not sure if the `besond` is necessary, but I agree it is more in line with the spirit of the rules for old age pensions.

hmgaudecker: What I mean by symmetry is (correct me if I am wrong, did not check, might misremeber):

For old age pensions we expect eligibility conditions (contribution periods, age) and calculate eligibility.

We should do the same here (contribution periods, age, disability status), rather than expecting eligibility as an input.

Note that given the DAG-structure, you can trivially pass eligibility. It just should not be the primitive, I think.

docs/gettsim_objects/input_variables.md
```diff
@@ -572,3 +574,23 @@ Type: int
...
 
 Type: int
+
+## `erwerbsgemindert`
+
+Dummy: Unable to provide more than 3 hours of market labor per day.
+
+Type: bool
+
+## `teilw_erwerbsm_rente`
```
hmgaudecker, line 584:
```suggestion
## `teilw_erwerbsgemindert`
```

Should be symmetric to above, I think.

src/_gettsim/config.py
```diff
@@ -183,5 +185,8 @@ def set_array_backend(backend: str):
     "kind_unterh_anspr_m": float,
     "kind_unterh_erhalt_m": float,
     "steuerklasse": int,
+    "erwerbsgemindert": bool,
+    "erwerbsm_langj_vers": bool,
```
hmgaudecker, line 189:
```suggestion
    "teilw_erwerbsgemindert": bool,
```

(swap order)

src/_gettsim/config.py
```diff
@@ -183,5 +185,8 @@ def set_array_backend(backend: str):
...
+    "teilw_erwerbsm_rente": bool,
```
hmgaudecker, line 190:
```suggestion
    "erwerbsm_langj_vers": bool,
```

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,256 @@
...
+  unit: Factor
+  reference: >-
+    § 67 SGB VI Rentenartfaktor:
+  note: >-
+    "https://sozialversicherung-kompetent.de/rentenversicherung/
+    leistungsrecht/1051-rentenartfaktor.html"
+  2001-01-01:
+    teilw: 0.5
+    voll: 1.0
+altersgrenze_grundbew:
```
hmgaudecker, line 239:
```suggestion
altersgrenze_grundbewertung:
```
No reason for abbreviations here.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,256 @@
...
+      erfüllt. Bedingungen werden in § 77 Abs. 4 SGB VI definiert.
+    en: >-
+      Age limit up to which one can claim the public disability insurance
+      without deductions under certain time conditions.
+      These conditions are defined in § 77 4 SGB VI.
+  unit: Year
+  reference: § 77 Abs. 4 SGB VI Zugangsfaktor
+  2001-01-01:
+    scalar: 63
+zurechnungszeitsgrenze:
```
hmgaudecker, line 75:
```suggestion
zurechnungszeitgrenze:
```

Please watch spelling a bit more closely. This really is unnecessary work for reviewers.

src/_gettsim/parameters/erwerbsm_rente.yaml
```diff
@@ -0,0 +1,256 @@
...
+  name:
+    de: Zurechnungszeitsgrenze
```
hmgaudecker, line 77:
```suggestion
    de: Zurechnungszeitgrenze
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+@dates_active(start="2001-01-01")
+def erwerbsm_rente_m(  # noqa: PLR0913
+    erwerbsm_rente_zugangsfaktor: float,
+    entgeltp_west_erwerbsm_rente: float,
+    entgeltp_ost_erwerbsm_rente: float,
+    rentenartfaktor: float,
+    ges_rente_vorauss_erwerbsm: bool,
+    ges_rente_params: dict,
+) -> float:
+    """Erwerbsminderungsrente (public disability insurance claim)
```
hmgaudecker, line 13:
```suggestion
    """Erwerbsminderungsrente (amount paid by public disability insurance if claimed)
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+        See :func:`entgeltp_ost_erwerbsm_rente`.
+    rentenwert
+        See :func:`rentenwert`.
+    rentenartfaktor
+        See :func:`rentenartfaktor`.
+    ges_rente_vorauss_erwerbsm
+        See :func:`ges_rente_vorauss_erwerbsm`.
+    Returns
+    -------
+    Erwerbsminderungsrente (public disability insurance claim)
```
hmgaudecker, line 34:
```suggestion
    Erwerbsminderungsrente (amount paid by public disability insurance if claimed)
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+
+    return anspruch_erwerbsm_rente
+
+
+def entgeltp_west_erwerbsm_rente(
+    entgeltp_west: float,
+    entgeltp_ost: float,
+    entgeltp_zurechnungszeit: float,
+) -> float:
+    """Entgeltpunkte accumulated in western germany which Erwerbsminderungsrente
```
hmgaudecker, line 87:
```suggestion
    """Entgeltpunkte accumulated in Western Germany which Erwerbsminderungsrente
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+    entgeltp_zurechnungszeit
+        See :func:`entgeltp_zurechnungszeit`.
+
+    Returns
+    -------
+    Final pension points for Erwerbsminderungsrente (public disability insurance)
+
+    """
+
+    anteil_entgeltp_ost = entgeltp_ost / (entgeltp_west + entgeltp_ost)
```
hmgaudecker, line 143: If we need this more frequently it would be useful to make a (trivial) function out of it.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+    Returns
+    -------
+    average entgeltp
+    """
+
+    beleg_gesamtzeitr = (
+        age_of_retirement - erwerbsm_rente_params["altersgrenze_grundbew"]
+    )
+
+    entgeltp_gesamt = entgeltp_west + entgeltp_ost
```
hmgaudecker, line 216:
```suggestion
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+
+    durchschn_entgeltp = entgeltp_gesamt / beleg_gesamtzeitr
```
hmgaudecker, line 218:
```suggestion
    durchschn_entgeltp = (entgeltp_west + entgeltp_ost) / beleg_gesamtzeitr
```

Easier to see what is happening this way.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+    m_pflichtbeitrag: float,
+    m_freiw_beitrag: float,
+    ges_rente_anrechnungszeit_45: float,
+    m_ersatzzeit: float,
+    m_kind_berücks_zeit: float,
+    m_pfleg_berücks_zeit: float,
+    ges_rente_params: dict,
+) -> bool:
+    """Whether Wartezeit von 40 Jahren Wartezeit (Vertrauensschutz) has been completed.
+    Aggregates time periods that are relevant for the eligibility of Altersrente für
```
hmgaudecker, line 318: Is this a copy & paste error? E.g., function is called `langj`, but docstring talks about `bes_langj`.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+
+    return out
+
+
+def entgeltp_ost_erwerbsm_rente(
+    entgeltp_west: float,
+    entgeltp_ost: float,
+    entgeltp_zurechnungszeit: float,
+) -> float:
+    """Entgeltpunkte accumulated in eastern germany which Erwerbsminderungsrente
```
hmgaudecker, line 121:
```suggestion
    """Entgeltpunkte accumulated in Eastern Germany which Erwerbsminderungsrente
```

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
+from _gettsim.shared import dates_active
+
+
+@dates_active(start="2001-01-01")
```
hmgaudecker, line 4: I guess all functions in this module should be decorated with `@dates_active` -- possibly with different dates unless none of this has changed since 2001.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,358 @@
...
+        ges_rente_params["zugangsfaktor_veränderung_pro_jahr"][
+            "vorzeitiger_renteneintritt"
+        ]
+    )
+    out = max(zugangsfaktor, erwerbsm_rente_params["min_zugangsfaktor"])
+
+    return out
+
+
+def langj_versicherte_wartezeit(  # noqa: PLR0913
```
hmgaudecker, line 308:
```suggestion
def _erwerbsm_rente_langj_versicherte_wartezeit(  # noqa: PLR0913
```

For now we'll need to be explicit about this.

src/_gettsim/transfers/rente.py
```diff
@@ -1728,7 +1728,7 @@ def ges_rente_anrechnungszeit(  # noqa: PLR0913
         See basic input variable :ref:`m_arbeitsl <m_arbeitsl>`.
     m_ausbild_suche
         See basic input variable :ref:`m_ausbild_suche <m_ausbild_suche>`.
-    m_schul_ausbild
+    #
```
hmgaudecker, line 1731: Seems to be a fluke?

src/_gettsim_tests/test_data/erwerbsm_rente/2001/hh_id_1.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - false
+      - false
+      - true
+      - true
+    teilw_erwerbsm_rente:
+      - false
+      - false
+      - false
+      - false
+      - true
```
hmgaudecker, line 66: This seems weird. From the docstrings I inferred that `erwerbsgemindert` and `teilw_erwerbsgemindert` (new suggestion) are mutually exclusive?

src/_gettsim_tests/test_data/erwerbsm_rente/2001/hh_id_1.yaml
hmgaudecker, line 1: This should become five test cases.

Please give them names that check what is being tested. E.g. `voll_erwerbsgemindert_birthyear_1940`.

*(The hh_id-stuff is only there for legacy reasons, i.e. the bulk conversion of csv-files)*

docs/gettsim_objects/input_variables.md
```diff
@@ -572,3 +574,23 @@ Type: int
 Number of own children below the age of 25 (incl. not in hh)
 
 Type: int
+
+## `erwerbsgemindert`
```
hmgaudecker, line 578:
```suggestion
## `voll_erwerbsgemindert`
```
I think that will make things clearer, but feel free to leave it as is if I am misunderstanding something.

hmgaudecker (commented):
We are getting close!

src/_gettsim_tests/test_data/erwerbsm_rente/2001/hh_id_1.yaml
hmgaudecker: We had one for csv -> yaml, but I am not sure whether it is helpful here. @lars-reimann would know. 

Otherwise, I'd just read them into a DataFrame and write out again row-by-row.

lars-reimann: The conversion script is available [here](https://github.com/iza-institute-of-labor-economics/gettsim-code-for-picking/blob/main/test_data_conversion_scripts/convert_csv_tests_to_yaml.py).

nafetsk:
> > I think all suggestions are implemented now @hmgaudecker . The only problem left is that the function name 'erwerbsm_rente_langj_versicherte_wartezeit' is too long but i dont know how to shorten it without losing precision.
> 
> Even too long if you add an underscore upfront?

I was not aware of that feature. Now it works👍

src/_gettsim_tests/test_data/erwerbsm_rente/2001/hh_id_1.yaml
nafetsk: Thanks a lot. I finally used that script [here.](https://github.com/nafetsk/split_up_yaml/blob/main/split_up_yaml.py) Maybe we can add it to the [gettsim-code-for-picking](https://github.com/iza-institute-of-labor-economics/gettsim-code-for-picking) repo so people can reuse it.

src/_gettsim/transfers/erwerbsm_rente.py
```diff
@@ -0,0 +1,332 @@
...
+
+    else:
+        out = erwerbsm_rente_params["rentenartfaktor"]["voll"]
+
+    return out
+
+
+@dates_active(start="2001-01-01")
+def erwerbsm_rente_zugangsfaktor(
+    ges_rente_params: dict,
```
hmgaudecker, line 223: Please always put all `params` last. Please don't forget to adjust the docstrings along with these changes.

src/_gettsim_tests/test_data/erwerbsm_rente/2001/teilw_erwerbsgemindert_birthyear_1980_claim.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - 44.15686275
+    entgeltp_west_erwerbsm_rente:
+      - 0.0
+    erwerbsm_rente_zugangsfaktor:
+      - 0.892
+    ges_rente_vorauss_erwerbsm:
+      - true
+    rentenartfaktor:
+      - 0.5
+  provided:
```
hmgaudecker, line 17:
```suggestion
```
Just get rid of the `provided` keys everywhere, unless the source is based on some external reference. The idea of `provided` is that this is what is published somewhere (e.g., some calculator of the DRV). Very often, the inputs of some calculator on the web there won't be enough for GETTSIM's functions, hence we add the `assumed` for additional input columns. If we just made up everything ourselves, it should just  be `assumed`

hmgaudecker (approved):
Wonderful, thanks a lot! Just a couple of small changes left, nothing substantial.

PR #580, opened 2023-05-14
Improve docs, fix deps

hmgaudecker:

Small stuff that came up when releasing 0.7


timmens (approved):
Looks great!

PR #583, opened 2023-05-30
feat: conversion between different time units

lars-reimann:

### Todo

### DIFF
--- a/src/_gettsim/time_conversion.py
+++ b/src/_gettsim/time_conversion.py
@@ -1,0 +1,207 @@
+_M_PER_Y = 12
+_W_PER_Y = 365.25 / 7
+_D_PER_Y = 365.25
+
+
+def y_to_m(value: float) -> float:
+    """
+    Converts yearly to monthly values.
+
+    Parameters
+    ----------
+    value
+        Yearly value to be converted to monthly value.
+
+    Returns
+    -------
+    float
+        Monthly value.
+    """
+    return value / _M_PER_Y
+
+
+def y_to_w(value: float) -> float:
+    """
+    Converts yearly to weekly values.
+
+    Parameters
+    ----------
+    value
+        Yearly value to be converted to weekly value.
+
+    Returns
+    -------
+    float
+        Weekly value.
+    """
+    return value / _W_PER_Y
+
+
+def y_to_d(value: float) -> float:
+    """
+    Converts yearly to daily values.
+
+    Parameters
+    ----------
+    value
+        Yearly value to be converted to daily value.
+
+    Returns
+    -------
+    float
+        Daily value.
+    """
+    return value / _D_PER_Y
+
+
+def m_to_y(value: float) -> float:
+    """
+    Converts monthly to yearly values.
+
+    Parameters
+    ----------
+    value
+        Monthly value to be converted to yearly value.
+
+    Returns
+    -------
+    float
+        Yearly value.
+    """
+    return value * _M_PER_Y
+
+
+def m_to_w(value: float) -> float:
+    """
+    Converts monthly to weekly values.
+
+    Parameters
+    ----------
+    value
+        Monthly value to be converted to weekly value.
+
+    Returns
+    -------
+    float
+        Weekly value.
+    """
+    return value * _M_PER_Y / _W_PER_Y
+
+
+def m_to_d(value: float) -> float:
+    """
+    Converts monthly to daily values.
+
+    Parameters
+    ----------
+    value
+        Monthly value to be converted to daily value.
+
+    Returns
+    -------
+    float
+        Daily value.
+    """
+    return value * _M_PER_Y / _D_PER_Y
+
+
+def w_to_y(value: float) -> float:
+    """
+    Converts weekly to yearly values.
+
+    Parameters
+    ----------
+    value
+        Weekly value to be converted to yearly value.
+
+    Returns
+    -------
+    float
+        Yearly value.
+    """
+    return value * _W_PER_Y
+
+
+def w_to_m(value: float) -> float:
+    """
+    Converts weekly to monthly values.
+
+    Parameters
+    ----------
+    value
+        Weekly value to be converted to monthly value.
+
+    Returns
+    -------
+    float
+        Monthly value.
+    """
+    return value * _W_PER_Y / _M_PER_Y
+
+
+def w_to_d(value: float) -> float:
+    """
+    Converts weekly to daily values.
+
+    Parameters
+    ----------
+    value
+        Weekly value to be converted to daily value.
+
+    Returns
+    -------
+    float
+        Daily value.
+    """
+    return value * _W_PER_Y / _D_PER_Y
+
+
+def d_to_y(value: float) -> float:
+    """
+    Converts daily to yearly values.
+
+    Parameters
+    ----------
+    value
+        Daily value to be converted to yearly value.
+
+    Returns
+    -------
+    float
+        Yearly value.
+    """
+    return value * _D_PER_Y
+
+
+def d_to_m(value: float) -> float:
+    """
+    Converts daily to monthly values.
+
+    Parameters
+    ----------
+    value
+        Daily value to be converted to monthly value.
+
+    Returns
+    -------
+    float
+        Monthly value.
+    """
+    return value * _D_PER_Y / _M_PER_Y
+
+
+def d_to_w(value: float) -> float:
+    """
+    Converts daily to weekly values.
+
+    Parameters
+    ----------
+    value
+        Daily value to be converted to weekly value.
+
+    Returns
+    -------
+    float
+        Weekly value.
+    """
+    return value * _D_PER_Y / _W_PER_Y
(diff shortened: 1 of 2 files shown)


src/_gettsim/time_conversion.py
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+
+def _create_functions_for_time_units(  # noqa: PLR0915
+    name: str, func: Callable | None = None
+) -> dict[str, Callable]:
+    result = {}
+    info = getattr(func, "__info__", None)
+
+    if name.endswith("_y"):
```
hmgaudecker, line 268: This block seems very verbose to me. Can't we do the check for the name upfront and then reduce the amount of repetition?

Something like (untested):

```py
time_units = ["y", "m", "w", "d"] # should be some global?
id_groupings = ("", "_hh", "_tu")  # Will change in near future, should be some global?
possible_suffixes = {f"_{unit}{id_g}": f"{unit}" for id_g in id_groupings for unit in time_units}
orig_suffix, orig_unit = [(k, v) for k, v in possible_suffixes.items() if name.endswith(k)][0]
_target_units = [u for u in time_units if not u == orig_unit]
target_suffixes = {orig_suffix.replace(orig_unit, u, 1): u for u in _target_units}
result = {
    _replace_suffix(name, orig_suffix, target_suffix): _create_function_for_time_unit(
        name, info, converter_funcs[f"{orig_unit}_to_{target_unit}"]
    )
    for target_suffix, target_unit in target_suffixes.items()
}
```

where `converter_funcs` is a dict with keys containing string version of the function name. Maybe there is a more elegant solution.

hmgaudecker (commented):
Looks great! One note added where I'd like to see a more algorithmic version to make this future-proof.

Also, we should add tests (and probably add a function to do the sequence of aggregation over groupings and time units) to allow for stuff like:

- existing: `foo_y`
- demanded: `foo_d_hh`

and:

- existing: `foo_y`
- demanded: `foo_d`, `foo_d_hh`

*(just to be pedantic so we don't end up with weird errors because there are two ways now to come up with `foo_d_hh`)*

hmgaudecker (commented):
Thanks! 

My comments are mainly about naming things.

One thing I am not quite clear is how getting from, say, `x_m` to `x_y_hh` works. Maybe we had a misunderstanding there what I meant with my earlier comment that you addressed in [87825d7](https://github.com/iza-institute-of-labor-economics/gettsim/pull/583/commits/87825d7347c12687b50ebba68717f0e60b9d118a)?

Please check out the `lohnst_y_prototype` branch, where I tried things out. 
- There are some failures in `test_docs` that appeared between two weeks ago and now, probably more of an issue of the testing code, already comes up without the most recent commit described next
- In [a8f0a29](https://github.com/iza-institute-of-labor-economics/gettsim/commit/a8f0a29c6c8889e95702442d8eddf462acb3170c), I removed the function `kapitaleink_brutto_y`, which was just required as a stepping stone to get from `kapitaleink_brutto_m` to `kapitaleink_brutto_y_tu`. That leads a bunch of tests to fail. Might be just because we are passing `kapitaleink_brutto_y` as part of the tests, but it should not matter which of the three variables we use as input. Put differently, the aggregation/time conversion machinery should not depend on whether something that is in the graph is a function or a column.

lars-reimann:
> One thing I am not quite clear is how getting from, say, x_m to x_y_hh works. Maybe we had a misunderstanding there what I meant with my earlier comment that you addressed in [87825d7](https://github.com/iza-institute-of-labor-economics/gettsim/pull/583/commits/87825d7347c12687b50ebba68717f0e60b9d118a)?

[_create_derived_functions](https://github.com/iza-institute-of-labor-economics/gettsim/blob/fb43d1d695cbfbbc9afd9490b01be3647bfa4427/src/_gettsim/functions_loader.py#L133) first creates the `time_conversion_functions`, which includes `x_y` (unless it exists already). Then the code to create aggregation functions works as before but on the pool of `time_conversion_functions` and `user_and_internal_functions`.

lars-reimann:
> * There are some failures in `test_docs` that appeared between two weeks ago and now, probably more of an issue of the testing code, already comes up without the most recent commit described next

That really seems to be an issue with the test code: It calls an internal part of the function loading pipeline that does not add derived functions (time conversion/aggregation) to the result.

hmgaudecker:
> > There are some failures in `test_docs` that appeared between two weeks ago and now, probably more of an issue of the testing code, already comes up without the most recent commit described next
> 
> That really seems to be an issue with the test code: It calls an internal part of the function loading pipeline that does not add derived functions (time conversion/aggregation) to the result.

Excellent, that'll need work, anyhow. Currently, the docstrings always link to functions at the level of aggregation that is required, sometimes leading to empty links. That'll become all the more common then.

I'd say we should mark that one as `xfail` as part of this PR.

hmgaudecker:
> > * In [a8f0a29](https://github.com/iza-institute-of-labor-economics/gettsim/commit/a8f0a29c6c8889e95702442d8eddf462acb3170c), I removed the function `kapitaleink_brutto_y`, which was just required as a stepping stone to get from `kapitaleink_brutto_m` to `kapitaleink_brutto_y_tu`. That leads a bunch of tests to fail. Might be just because we are passing `kapitaleink_brutto_y` as part of the tests, but it should not matter which of the three variables we use as input. Put differently, the aggregation/time conversion machinery should not depend on whether something that is in the graph is a function or a column.
> 
> I've had a look at the failures of the `wohngeld` tests:
> 
> * The test data contains the column `kapitaleink_brutto_y`
> 
> * `create_time_conversion_functions` also adds `kapitaleink_brutto_m`, `kapitaleink_brutto_w`, and `kapitaleink_brutto_d`
> 
> * `_create_aggregation_functions` adds the `_tu` and `_hh` functions as needed
> 
> * Note, however, that no function called `kapitaleink_brutto_y` is created
> 
> * This leads to the (IMO misleading) error message shown below since `kapitaleink_brutto_y` is listed in `OVERRIDE_COLS` in the test, but doesn't actually override a function:
> 
> 
> ```
> E           ValueError: You passed the following user column which is unnecessary because no functions
> E           require them as inputs.
> E           
> E           [
> E               "kapitaleink_brutto_y",
> E           ]
> ```
> 
> * The solution in this case is to remove `kapitaleink_brutto_y` from `OVERRIDE_COLS` in the tests

Excellent analysis, thanks!!!

~~Ideally, I would like to see a different approach on the GETTSIM side. From a user perspective, it should not matter whether a function is written out or created as part of the aggregation/time conversion utilities.~~

~~That is, prior to me deleting `kapitaleink_brutto_y`, it was an explicit function doing nothing but the aggregation. We needed to list it in `columns_overriding_functions`. Now it *would* be implicitly created.~~

After writing the prior two paragraphs, I realised this probably won't be possible to do that in a consistent way. Problem is that we may reverse the order of the DAG. That is, we normally expect `kapitaleink_brutto_m` and expect it to be converted to `kapitaleink_brutto_y`. But now this could be just the other way round and I don't think there is a consistent way to predict much of that without hard-coding expected inputs etc.

Do you agree or am I missing something? Would go with your suggestion to drop that line, then.

We'll need to clearly flag this behavior in the documentation of `columns_overriding_functions`, though!

lars-reimann:
> Do you agree or am I missing something? Would go with your suggestion to drop that line, then.

> We'll need to clearly flag this behavior in the documentation of columns_overriding_functions, though!

Is there a use case where the same variable is specified as a data column and as a function and the user still wants to use the function?

hmgaudecker:
IIRC, we decided for the current behaviour after an extensive discussion because it is very easy to do something by accident. That is, what you describe is perfectly intuitive. It is also clear if my dataset has five columns. But with many dozens of input columns, it may quite easily happen that there is a column in the dataset that overrides a function when this is not the intended behaviour.

Hence, we decided for the Python Zen's second commandment: *Explicit is better than implicit.*

That was done well before we had automatic aggregation functions, let alone time conversion functions.

I think what you suggest is very reasonable for the time being; in a redesigned API it should change, anyhow. See #596.

hmgaudecker (approved):
Great work.

Remaining to-dos:
- mark the docs test as xfail as discussed
- open a new PR that will change the behavior of `columns_overriding_functions` in `compute_taxes_and_transfers`

lars-reimann:
> IIRC, we decided for the current behaviour after an extensive discussion because it is very easy to do something by accident. That is, what you describe is perfectly intuitive. It is also clear if my dataset has five columns. But with many dozens of input columns, it may quite easily happen that there is a column in the dataset that overrides a function when this is not the intended behaviour.
> 
> Hence, we decided for the Python Zen's second commandment: _Explicit is better than implicit._
> 
> That was done well before we had automatic aggregation functions, let alone time conversion functions.
> 
> I think what you suggest is very reasonable for the time being; in a redesigned API it should change, anyhow. See #596.

Makes sense. Maybe instead of having the user type out the entire list of overridden functions, it would be enough to allow users to configure whether to raise an error/warn/do nothing if a column overrides a function.

PR #585, opened 2023-06-05
Add Erziehungsgeld

nafetsk:
„Erziehungsgeld“ existed until 2006 and was then replaced by the “Elterngeld”, which is already implemented. 

Only the period from the last change in the "Bundeserziehungsgeldgesetz" in 2004 to the abolition at the end of 2006 is being considered in this PR.


hmgaudecker:
Just a remark: I love the idea of adding Erziehungsgeld! But in order to use it, serious looks at Elterngeld and the period before 2015 or so in general will be necessary. I am afraid that we have not been doing particularly well in making sure that time period works, let alone correctly.

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+        See :See basic input variable :ref:`hat_kinder <hat_kinder>`.
+    erziehungsgeld_anspruch_kind_tu
+        See :func:`erziehungsgeld_anspruch_kind_tu`.
+
+    Returns
+    -------
+    eligibility of parental leave benefit (erziehungsgeld) as a bool
+
+    """
+    out = erziehungsgeld_anspruch_kind_tu and (arbeitsstunden_w <= 30) and hat_kinder
```
MaxBlesch, line 209: Please remove hardcoded threshold.

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+    if erziehungsgeld_anspruch_kind and alleinerz_tu:
+        out = (
+            bruttolohn_vorj_m_tu * 12
+            - eink_st_abzuege_params["werbungskostenpauschale"]
+        ) * erziehungsgeld_params["pauschal_abzug_auf_einkommen"]
+    elif erziehungsgeld_anspruch_kind and (not alleinerz_tu):
+        out = (
+            bruttolohn_vorj_m_tu * 12
+            - eink_st_abzuege_params["werbungskostenpauschale"]
+            * 2  # is that already hard coded?
```
MaxBlesch, line 257: What do you mean by that?

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+    ----------
+    erziehungsgeld_params
+        See params documentation :ref:`erziehungsgeld_params <erziehungsgeld_params>`.
+    alleinerz_tu
+        See :func:`alleinerz_tu`.
+    alter_monate
+        See :func:`alter_monate`.
+    anz_kinder_mit_kindergeld_tu
+        See :func:`anz_kinder_mit_kindergeld_tu`.
+    budget_erz_geld
```
MaxBlesch, line 287: Is that newly introduced? If so, we would need a description in the documentation?

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+    elif (
+        erziehungsgeld_anspruch_kind
+        and (7 <= alter_monate <= 12)
+        and erziehungsgeld_eink_relev_kind < erziehungsgeld_einkgrenze_kind
+        and (not budget_erz_geld)
+    ):
+        out = erziehungsgeld_params["erziehungsgeld_satz"]["regelsatz"]
+    elif (
+        erziehungsgeld_anspruch_kind
+        and (7 <= alter_monate <= 12)
```
MaxBlesch, line 141: Please remove hardcoded threshold.

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Returns
+    -------
+    income threshold for parental leave benefit (erziehungsgeld)
+    """
+    # There are different income threshold depending on the age of the child,
+    # the fact if a person is a single parent, and if regelsatz or budgetsatz is applied
+
+    einkommensgrenze = {
+        (True, True, False): erziehungsgeld_params["einkommensgrenze_bis_6m"][
```
MaxBlesch, line 300: Please split that up in nested functions. Where you select in each function based on one bool. Lets zoom about it!

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -0,0 +1,330 @@
...
+        (False, True, True): erziehungsgeld_params["einkommensgrenze_ab_7m"][
+            "alleinerz"
+        ],
+        (False, False, True): erziehungsgeld_params["einkommensgrenze_ab_7m"]["paar"],
+    }
+    key = (alter_monate < 7, alleinerz_tu, budget_erz_geld)
+    out = einkommensgrenze[key]
+
+    # For every additional child the income threshold is increased by a fixed amount
+    out += (anz_kinder_mit_kindergeld_tu - 1) * erziehungsgeld_params[
```
MaxBlesch, line 325: I think this should be more prominent, i.e. an extra function.

src/_gettsim_tests/test_data/erziehungsgeld/2005/hh_id_6.yaml
```diff
@@ -0,0 +1,108 @@
+---
+info:
+  note: ''
+  source: own calculations
```
MaxBlesch, line 4: Just to make sure. There is probably no official calculation?

src/_gettsim_tests/test_data/erziehungsgeld/2005/hh_id_4.yaml
```diff
@@ -0,0 +1,88 @@
...
+    geburtsjahr:
+      - 1965
+      - 1965
+      - 2004
+    budget_erz_geld:
+      - false
+      - false
+      - true
+    inanspruchn_erzgeld:
+      - true
```
MaxBlesch, line 68: We shouldn't have inanspr true and then there is none. This doesn't make sense

src/_gettsim_tests/test_data/erziehungsgeld/2005/hh_id_8.yaml
```diff
@@ -0,0 +1,108 @@
...
+    - 0
+    - 19640
+    - 33140
+  erziehungsgeld_kind_m:
+    - 0.0
+    - 0.0
+    - 333.55
+    - 300.0
+  erziehungsgeld_m:
+    - 633.55
```
MaxBlesch, line 105: We probably don't handle the case where one parent claims for one child erziehungsgeld and the other parent for the other kid? I think we shouldn't care about that corner case, but just wanted to know?

MaxBlesch (changes requested):
Thanks so much for the PR. It is looking already really good. Lets talk in the coming week about the comments!

src/_gettsim_tests/test_data/erziehungsgeld/2005/hh_id_4.yaml
nafetsk: That's true, however, I then wonder why we even check for eligibility for erziehungsgeld if "inanspruchn_erzgeld" is only true if money is also paid out.

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +546,15 @@ Type: int
...
 
 Type: int
+
+## `budget_erz_geld`
+
+Dummy: applied for "Budgetsatz" of parental leave benefit
+
+Type: bool
+
+## `inanspruchn_erzgeld`
```
MaxBlesch, line 556: Maybe we need a general discussion about these claim variables. If those are true and then gettsim calculates that individuals are not eligible. How do we handle that?

hmgaudecker: The interpretation can only ever be conditional on GETTSIM deciding that someone is eligible. Anything else would be measurement error or GETTSIM needs refinement.

hmgaudecker: However, thinking of it, I think we should get rid of it for now. People get the transfer if they qualify, period. Else you'll need to remove erziehungsgeld from the targets. More general handling of take up is in the making.

Btw, please write out: `erziehungsgeld`.  The two abbreviations here are inconsistent with each other already.

MImmesberger:
Is there a reason why you don't follow the implementation of Kindergeld (in GETTSIM Kindergeld is being ascribed to the child)? This would both simplify the code and solve the take-up issue (as we can be agnostic about which parent receives the Erziehungsgeld). I think we can get rid of the `inanspruchn_erziehungsgeld` variable this way.

MImmesberger:
In the last commit, I changed the grouping from `tu` to `fg`. Two questions:

1. Is that correct? Erziehungsgeld should be payed out to the children's parents only.
2. Given `fg` is correct, we should use `anz_erwachsene_fg` and `anz_kinder_mit_kindergeld_fg` as function inputs instead of their `tu` version. These targets do not exist yet, should we keep the `tu` version for now or implement the `fg` version right here?

MaxBlesch:
Great work @MImmesberger. Regarding your comments:

- I would keep the takeup. As soon we have a better solution to deal with gendered transfers, we should delete it.

- It is good that you deleted it. Totally agree!

MImmesberger:
Okay, great!

@hmgaudecker Do you still want to take a look at it? Because you asked about that: For the basic input variables I kept the abbreviation `erzgeld` because of the character limit.

src/_gettsim_tests/test_data/erziehungsgeld/2005/alleinerz_one_child_budget.yaml
hmgaudecker, line 1: rename test to `budgetsatz`

Also, I am not quite sure from the name what the test is supposed to test?

docs/gettsim_objects/input_variables.md
```diff
@@ -572,3 +572,15 @@ Type: int
 Number of own children below the age of 25 (incl. not in hh)
 
 Type: int
+
+## `budget_erzgeld`
```
hmgaudecker, line 576:
```suggestion
## `budgetsatz_erzieh`
```
Makes searching slightly easier, at least. And `budget` by itself is too ambiguous (if not familiar with the context).

Will change before being released, anyhow.

docs/gettsim_objects/input_variables.md
```diff
@@ -572,3 +572,15 @@ Type: int
...
+
+Dummy: applied for "Budgetsatz" of parental leave benefit
+
+Type: bool
+
+## `inanspruchn_erzgeld`
```
hmgaudecker, line 582:
```suggestion
## `inanspruchn_erzieh`
```

hmgaudecker (approved):
Just a couple of tiny renamings!

hmgaudecker:
I was just about to hit "merge", but then stumbled across this sentence in the PR description:

> Only the period from the last change in the "Bundeserziehungsgeldgesetz" in 2004 to the abolition at the end of 2006 is being considered in this PR.

Is this correct, @nafetsk @MImmesberger? If so, please add `@dates_active(start="2004-01-01", ...)` to all functions in `erziehungsgeld.py`. Additionally, add another option with `@dates_active(start="1985-01-01", end="2003-12-31", change_name="erziehungsgeld_m")` that yields an error. We should never produce wrong results.

MImmesberger:
In 2004 there was a Neubekanntmachung of the BErzGG. There were some changes regarding the Regelsatz and the maximum age of the children since the 80's that aren't considered in this PR and before 1998, the budgetierte Erziehungsgeld didn't exist.

I will set the decorators accordingly and add the parameters after 1998.

@nafetsk Are you aware of any other changes?

hmgaudecker:
Thanks! I was just going to suggest buzer.de as a greate source for diffing versions of laws, but their archive does not go back that far. @mjbloemer, @JuergenWiemers: would one of you happen to know resources for doing that?

MImmesberger:
After reading some more material, I'm not sure about the changes that happened between 1998 and 2004 (e.g. [this source](https://www.dji.de/fileadmin/user_upload/bibs/218/ErlerBeitragDJI-Kompendium.dt.pdf) suggests that there might be changes regarding the eligibility criteria). So, I set the start date to 2004.

MaxBlesch:
At the beginning of this implementation @nafetsk and I had a discussion if we should include the pre 2004 regime, but wanted to do 2004-2006 first. For us it is most important to have mirrored the policy change from Elterngeld to Erziehungsgeld. So, I think @MImmesberger solution is good. 

When checking the code right now, I realized that Elterngeld does not have the `@dates_active` decorators. Maybe we can do that in #619 or I can file a issue!

PR #589, opened 2023-06-14
make sure `arbeitsl_geld_2_eink_anr_frei_m` uses the correct variable

ChristianZimpelmann:

- make sure `arbeitsl_geld_2_eink_anr_frei_m` uses the correct variable
- improve documentation


hmgaudecker (approved):
Excellent, thank you very much! Good catch.

Issue #568, opened 2023-04-17
BUG: Consider `eink_selbst_m` in `arbeitsl_geld_2_eink_anr_frei_m`


ChristianZimpelmann:
### Bug description

- The function `arbeitsl_geld_2_eink_anr_frei_m` should be based on Erwerbseinkommen.
- It only considers `bruttolohn_m` at the moment.
- We should add `eink_selbst_m`

### ToDo

- check whether reasoning above is correct
- implement test case for self-employed income and Bürgergeld
- implement proposed change
- add reference to law

PR #590, opened 2023-06-15
Fix child income deduction for Kinderzuschlag

ChristianZimpelmann:
### Todo

- [x] Add allowance for child income for Kinderzuschlag
- [x] Adjust test cases (the failing tests are not based on an external source). Maybe [this calculator](https://www.kinderzuschlag.org/rechner/) can be used.

### Note
Two further bugs which might or might not be fixed within this PR:

- Find out whether child income of other sources than `bruttolohn_m` should be deducted from Kinderzuschlag (we currently only consider bruttolohn. 
- After addressing the point above, #568 is also relevant for Kinderzuschlag.


ChristianZimpelmann:
Thanks!

PR #593, opened 2023-06-16
Implement reform of gesetzliche Pflegepflegeversicherung effective as of 2023-07-01

### LINKED ISSUE #595: ENH: new Pflegeversicherung regime from 2023-07-01
The contribution rate for elderly care is going to increase on 2023-07-01 ([draft](https://dserver.bundestag.de/btd/20/069/2006983.pdf))

Moreover, the top-up is going to be changed. So far, every adult above 23 without children has to pay a top-up *unless* he/she receives *Bürgergeld*  (not sure we have that covered already). Now, the top-up rate is reduced by the number of children, starting with the second child. More importantly, children are only considered until they are 25. See also [here](https://www.iwkoeln.de/fileadmin/user_upload/Studien/Kurzberichte/PDF/2023/IW-Kurzbericht_2023-Pflegeversicherungsbeitrag.pdf).

- [ ] check whether we exempt receivers of Bürgergeld
- [ ] implement new top-up regulation

paulinaschroeder:

Adapt calculation of `ges_pflegev_beitr_m` to new reform (Pflegeunterstützungs- und -entlastungsgesetz - PUEG)

Closes #595 

### Todo

- [x] Increase contribution rates
- [x] Introduce deduction of contribution rate for people with two or more children
- [x] Add new test cases

### DIFF
--- a/docs/gettsim_objects/input_variables.md
+++ b/docs/gettsim_objects/input_variables.md
@@ -519,6 +519,6 @@ Type: int
 
 ## `anz_eig_kind_bis_24`
 
-Amount of own children below the age of 25 (incl. not in hh)
+Number of own children below the age of 25 (incl. not in hh)
 
 Type: int
--- a/docs/gettsim_objects/input_variables.md
+++ b/docs/gettsim_objects/input_variables.md
@@ -517,7 +517,7 @@ Tax Bracket (1 to 5) for withholding tax
 
 Type: int
 
-## `anz_kinder_u_25`
+## `eigene_kind_bis_24`
 
 Amount of children below tha age of 25
 
--- a/docs/gettsim_objects/input_variables.md
+++ b/docs/gettsim_objects/input_variables.md
@@ -519,6 +519,6 @@ Type: int
 
 ## `eigene_kind_bis_24`
 
-Amount of children below tha age of 25
+Amount of own children below the age of 25 (incl. not in hh)
 
 Type: int
--- a/docs/gettsim_objects/input_variables.md
+++ b/docs/gettsim_objects/input_variables.md
@@ -516,3 +516,9 @@ Type: float
 Tax Bracket (1 to 5) for withholding tax
 
 Type: int
+
+## `anz_kinder_u_25`
+
+Amount of children below tha age of 25
+
+Type: int
--- a/docs/gettsim_objects/input_variables.md
+++ b/docs/gettsim_objects/input_variables.md
@@ -517,7 +517,7 @@ Tax Bracket (1 to 5) for withholding tax
 
 Type: int
 
-## `eigene_kind_bis_24`
+## `anz_eig_kind_bis_24`
 
 Amount of own children below the age of 25 (incl. not in hh)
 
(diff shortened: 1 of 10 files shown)


docs/gettsim_objects/input_variables.md
```diff
@@ -516,3 +516,9 @@ Type: float
 Tax Bracket (1 to 5) for withholding tax
 
 Type: int
+
+## `anz_kinder_u_25`
+
+Amount of children below tha age of 25
```
ChristianZimpelmann, line 522:
```suggestion
Amount of own children below the age of 25 (incl. not in hh)
```

I assume not in hh count as, well? Otherwise, we won't need a new input variable.

docs/gettsim_objects/input_variables.md
```diff
@@ -516,3 +516,9 @@ Type: float
 Tax Bracket (1 to 5) for withholding tax
 
 Type: int
+
+## `anz_kinder_u_25`
```
ChristianZimpelmann, line 520:
```suggestion
## `eigene_kind_bis_24`
```

src/_gettsim/social_insurance_contributions/ges_pflegev.py
```diff
@@ -29,14 +29,71 @@ def ges_pflegev_zusatz_kinderlos(
     return out
 
 
+def ges_pflegev_beitr_regulär_besch_m(
```
ChristianZimpelmann, line 32: We should use two functions for the situation before 2023-07 and the situation after and make use of the `@dates_active` decorator (see the code base for several examples). 

Then, we don't need to define the parameter `abschlag_kinder` for the time before that.

src/_gettsim/social_insurance_contributions/ges_pflegev.py
```diff
@@ -29,14 +29,71 @@ def ges_pflegev_zusatz_kinderlos(
...
+    # Add additional contribution for childless individuals
+    if ges_pflegev_zusatz_kinderlos:
+        beitr_regulär_beschäftigt_m += (
+            _ges_krankenv_bruttolohn_m
+            * sozialv_beitr_params["beitr_satz"]["ges_pflegev"]["zusatz_kinderlos"]
+        )
+
+    # Substract contribution for individuals with two or more children under 25
+    for i in range(2, 5):
+        if anz_kinder_u_25 == i:
```
ChristianZimpelmann, line 75: I think we can get ride fo the loop and the second part of the code by something like:

```
if anz_kinder_u_25 >= 2:
    .... * min(anz_kinder_u_25 - 1, 4)

**Review by ChristianZimpelmann — CHANGES_REQUESTED** [review]
Thanks, that is a great starting point! 

I added a larger comment and a few renaming suggestions.

Before merging, we should also add a few new test cases.

File: docs/gettsim_objects/input_variables.md [hunk 5]
```diff
@@ -516,3 +516,9 @@ Type: float
 Tax Bracket (1 to 5) for withholding tax
 
 Type: int
+
+## `anz_eig_kind_bis_24`
+
+Amount of own children below the age of 25 (incl. not in hh)
```
ChristianZimpelmann, line 522:
```suggestion
Number of own children below the age of 25 (incl. not in hh)
```

src/_gettsim/parameters/sozialv_beitr.yaml
```diff
@@ -274,6 +274,15 @@ beitr_satz:
     reference: >-
       (arbeitsl_v:) Art. 2 Nr. 15 G. v. 18.12.2018 BGBl. I S. 2651
       (ges_krankenv mean_zusatzbeitrag:) BAnz AT 31.10.2022 B5
+  2023-07-01:
```
ChristianZimpelmann, line 277: I think we need `deviation_from: previous` here such that parameters for the other social insurances

ChristianZimpelmann (approved):
Great job!

PR #601, opened 2023-07-07
Add groupings `sn_id`, `fg_id`, and `bg_id`

### LINKED ISSUE #582: ENH: Minimal example for endogenous groupings

In #509, we have been discussing various ways to aggregate individuals into relevant units based on relations between individuals.
- For example, a *Bedarfsgemeinschaft* in SGB II / XII is based on the parent-child relation (`bg_id`)
- A successor for `tax_unit` (Steuersubjekt, `st_id` as tentative names)
- Eligibility for child exemptions in the income tax (`n_kinderfreibeträge`)

The main change is that these units are endogenous now, whereas the only way to set them was as input variables so far.  The functions will be part of the DAG, but they do not operate on a single row, but need to perform merge-like operations. 

Another downstream complication may be that when we do aggregation by `bg_id` or `st_id`, we cannot assume the dataset to be ordered.

#### Minimal test case

This is not a correct depiction of the rules, but should capture the gist of what we need. I use `-1` to indicate missing variables, this is just a placeholder right now.

```yaml
---
info:
  note: >-
    Parents with three kids, assume that one is eligible to be in the same 
    Bedarfsgemeinschaft. One is not because it is living outside the household, another
    is not because it is 25 years of age. Both parents should be a single steuersubject,
    each kid is its own.
  source: ''
inputs:
  provided:
    p_id:
      - 0
      - 1
      - 2
      - 3
      - 4
    hh_id:
      - 0
      - 0
      - 1
      - 0
      - 0
    alter:
      - 50
      - 48
      - 18
      - 20
      - 25
    p_id_ehepartner:
      - 1
      - 0
      - -1
      - -1
      - -1
    p_id_elternteil_1:
      - -1
      - -1
      - 0
      - 0
      - 1
    p_id_elternteil_2:
      - -1
      - -1
      - 1
      - 1
      - 0
outputs:
  bg_id:
    - 0
    - 0
    - 1
    - 0
    - 2
  st_id:
    - 0
    - 0
    - 1
    - 2
    - 3
```

lars-reimann:

Closes #582

Add groupings
* `bg`
* `fg`
* `sn`


lars-reimann:
@hmgaudecker So far this is a simple proof-of-concept with some dummy code for `st_id`. Some questions about this:

1. How should `st_id` be defined?
2. How should `bg_id` be defined?

For the latter, the note in the minimal test case provides some context. However, it's for example not quite clear, whether the age limit is 25 year or something lower.

src/_gettsim/functions_loader.py
```diff
@@ -104,15 +108,17 @@ def load_and_check_functions(
             user_functions,
             aggregation_functions,
             time_conversion_functions,
+            endogenous_groupings,
         ],
-        ["internal", "user", "aggregation", "time_conversion"],
+        ["internal", "user", "aggregation", "time_conversion", "endogenous_groupings"],
```
hmgaudecker, line 113:
```suggestion
        ["internal", "user", "aggregation", "time_conversion", "groupings"],
```

The `endogenous` is just for our understanding during the dev process.

hmgaudecker:
Great start, thanks! 

> @hmgaudecker So far this is a simple proof-of-concept with some dummy code for `st_id`. Some questions about this:
> 
>     1. How should `st_id` be defined?

```py
def st_id_numpy(
    p_id: numpy.ndarray,
    p_id_ehepartner: numpy.ndarray,
    gemeinsam_veranlagt: numpy.ndarray[bool]
):
```
Just add a check whether `gemeinsam_veranlagt` evaluates to `True` (ideally for both partners and [raise an error if not the same for both partners](https://github.com/iza-institute-of-labor-economics/gettsim/blob/3a9998ab5dc2cd9d1c9374c944bbdd665284235d/src/_gettsim_tests/test_data/id_variables/2023/partner_widerspruechliche_veranlagungs_info.yaml)). Will need to worry about speed eventually, though.

>     2. How should `bg_id` be defined?
> 
> However, it's for example not quite clear, whether the age limit is 25 year or something lower.

The age limit is `<= 24` indeed. 

If that age limit is met, additional qualifications are that the kid's income [is not too high](https://github.com/iza-institute-of-labor-economics/gettsim/blob/3a9998ab5dc2cd9d1c9374c944bbdd665284235d/src/_gettsim_tests/test_data/id_variables/2023/eltern_verheiratet_1_kind_with_income.yaml) and that the kid [does not have own children](https://github.com/iza-institute-of-labor-economics/gettsim/blob/3a9998ab5dc2cd9d1c9374c944bbdd665284235d/src/_gettsim_tests/test_data/id_variables/2023/eltern_verheiratet_1_kind_mit_eigenem_kind.yaml). There might be more eventually, but that should do for a start!

lars-reimann:
> Just add a check whether gemeinsam_veranlagt evaluates to True (ideally for both partners and [raise an error if not the same for both partners](https://github.com/iza-institute-of-labor-economics/gettsim/blob/3a9998ab5dc2cd9d1c9374c944bbdd665284235d/src/_gettsim_tests/test_data/id_variables/2023/partner_widerspruechliche_veranlagungs_info.yaml)). Will need to worry about speed eventually, though.

I've updated `st_id` accordingly.

-----

Regarding `bg_id`: 

* Are these the correct inputs?
  ```py
  def bg_id_numpy(
      p_id: numpy.ndarray,
      hh_id: numpy.ndarray,
      alter: numpy.ndarray,
      p_id_ehepartner: numpy.ndarray,
      p_id_elternteil_1: numpy.ndarray,
      p_id_elternteil_2: numpy.ndarray,
      enough_income: numpy.ndarray[bool],
  ):
      pass
  ```

src/_gettsim/groupings.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def bg_id_numpy(  # noqa: PLR0913
+    p_id: numpy.ndarray,
+    hh_id: numpy.ndarray,
+    alter: numpy.ndarray,
+    p_id_ehepartner: numpy.ndarray,
+    p_id_elternteil_1: numpy.ndarray,
+    p_id_elternteil_2: numpy.ndarray,
+    enough_income: numpy.ndarray[bool],
```
hmgaudecker, line 17:
```suggestion
    eigener_bedarf_gedeckt: numpy.ndarray[bool],
```

src/_gettsim/groupings.py
```diff
@@ -0,0 +1,125 @@
...
+
+def create_groupings() -> dict[str, Callable]:
+    return {"bg_id": bg_id_numpy, "st_id": st_id_numpy}
+
+
+def bg_id_numpy(  # noqa: PLR0913
+    p_id: numpy.ndarray,
+    hh_id: numpy.ndarray,
+    alter: numpy.ndarray,
+    p_id_ehepartner: numpy.ndarray,
```
hmgaudecker, line 14:
```suggestion
    p_id_einstandspartner: numpy.ndarray,
```
Being married is not required, but relationship has to be close to it ("eheähnliche Gemeinschaft")

hmgaudecker (commented):
Wonderful! Many thanks.

Essentially just a few naming suggestions.

Maybe a good next step would be to take the [test data from #509](https://github.com/iza-institute-of-labor-economics/gettsim/pull/509/files) and move them over here?

I think the structure from there (just one directory `id_variables` instead of `bg_id`, `st_id`, ...; better name will be `groupings`) will minimize code duplication

Eventually we will have to worry about speed. Do you have suggestions on how to best set up performance benchmarks?

lars-reimann:
> Eventually we will have to worry about speed. Do you have suggestions on how to best set up performance benchmarks?

I did not use it yet, but [`pytest-benchmark`](https://pypi.org/project/pytest-benchmark/) looks promising.

lars-reimann:
The additional validation is now implemented, without the option to treat non-existing `p_ids` as missing.

hmgaudecker:
Sorry for being slow to get back to this one. We had very fruitful discussions at the workshop, the gist of which is included in GEP-01 [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/601/commits/38272c1415a0a094a9f70bcf9dc2a1c5630d3f62)

I do think we are pretty clear now on where this is headed. I suggest the following for the current PR:
- rename `st_id` to `sn_id` (problem is that `st_id` looks like Steuer-ID, which is the federal one attached to a single person. We are looking for the Steuernummer at the local tax office, which may refer to couples, too)
- rename `hh_id` to `vg_id` (not sure this is relevant at all right now, probably we won't have a function for that but assume it is a root node, so just a matter of renaming once we use the new ids)
- add the intermediate step for `fg_id`, which checks the family relations as is done at present, but ignores `eigener_bedarf_gedeckt`. Think of this as the potential Bedarfsgemeinschaft.
- I am not perfectly sure whether we'll add the additional variable `eigener_bedarf_gedeckt` or make it a function of parameters in the end of the day. I'd leave that up to you whether you think it's better to include the step from `fg_id` to `bg_id` right here as it is at present, or push off to another day, assuming `bg_id` == `fg_id` for the moment.

@mjbloemer, please chime in in case something from this list runs counter to what we discussed *(I do know `fg_id` will be anachronism)*

Apologies for these specification changes!!! All previous effort was not in vain at all, however: Working through this PR helped clarify lots of things! 

Finally, while you are at it, please add tests for the lines flagged by codecov:

image

lars-reimann:
Apart from the change to `hh_id` the specification changes are implemented now. I'll do this in a separate PR since it touches countless files.

src/_gettsim/config.py
```diff
@@ -67,13 +67,36 @@ def set_array_backend(backend: str):
...
 SUPPORTED_GROUPINGS = {
     "hh": {
-        "name": "household",
-        "description": "all individuals living in the same household.",
+        "name": "Haushalt",
+        "description": "The relevant unit for Wohngeld. Encompasses more people than"
+        " the Bedarfsgemeinschaft (e.g., possibly more than 2 generations). Relevant"
+        " unit for Wohngeld. `vg` derives from Verantwortungs- und"
+        " Einstehensgemeinschaft, though this might be a misnomer.",
+        "is_endogenous": False,
```
hmgaudecker, line 75:
```suggestion
        "potentially_endogenous": False,
```

We can always override the own calculation by passing columns, so `is_` is a bit misleading.

Put differently, I will never be able to derive these columns from other variables.

src/_gettsim/interface.py
```diff
@@ -430,8 +436,13 @@ def _fail_if_group_variables_not_constant_within_groups(data):
         Dictionary containing a series for each column.
 
     """
+    hard_coded_groupings = [
+        key
+        for key, value in SUPPORTED_GROUPINGS.items()
```
hmgaudecker, line 441: I think this check is conceptually slightly off. It should check also check those columns that I pass. E.g., if I pass `bg_id` instead of having GETTSIM calculate it, it should flag wrong values.

src/_gettsim/synthetic.py
```diff
@@ -150,7 +150,12 @@ def create_basic_households(
         }
     )
 
-    group_ids = [f"{g}_id" for g in SUPPORTED_GROUPINGS]
+    hard_coded_groupings = [
```
hmgaudecker, line 153:
```suggestion
    exogenous_groupings = [
```

hmgaudecker (commented):
Excellent work! 

A couple of high-level comments left.

Also, we'll need to think about benchmarking infrastructure. Won't be able to do in this repo, let's discuss at some point. Not crucial *right* now.

src/_gettsim/config.py
```diff
@@ -69,11 +69,33 @@ def set_array_backend(backend: str):
     "hh": {
```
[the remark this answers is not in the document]
lars-reimann: #662 only changed functions, not their parameters. Thus, the grouping "hh" is currently still used. Removing this entry leads to failing tests.

src/_gettsim/groupings.py
```diff
@@ -0,0 +1,155 @@
...
+                message = (
+                    f"{current_p_id_einstandspartner} and {current_p_id} are "
+                    "married, but have different values for "
+                    "gemeinsam_veranlagt."
+                )
+                raise ValueError(message)
+
+            if current_gemeinsam_veranlagt:
+                result.append(p_id_to_st_id[current_p_id_einstandspartner])
+                continue
```
hmgaudecker, line 147: Last statement in the loop; how can it do anything?

src/_gettsim/groupings.py
```diff
@@ -0,0 +1,155 @@
...
+
+def sn_id_numpy(
+    p_id: numpy.ndarray,
+    p_id_ehepartner: numpy.ndarray,
+    gemeinsam_veranlagt: numpy.ndarray[bool],
+):
+    """
+    Compute a Steuernummer (ID) for each person / couple.
+    """
+    p_id_to_st_id = {}
```
hmgaudecker, line 120:
```suggestion
    p_id_to_sn_id = {}
```
All occurences of `st_id` in this function body.

hmgaudecker (commented):
Looks great! I added a couple of to-dos directly, was easier than describing what was needed.

The only thing I am wondering now is whether we should add a few test cases where things (i.e., `hh_ids` we may require, but parents / `vg_id`s / `bg_id`s) may come up in somewhat different places. IIUC, the tests all relate to a single household, are we sure it will work for multiple households with "strange" order?

src/_gettsim/groupings.py
lars-reimann: The next 4 statements are also part of the loop

src/_gettsim/groupings.py
```diff
@@ -0,0 +1,155 @@
...
+        if current_p_id_elternteil_2 >= 0:
+            if current_p_id_elternteil_2 not in p_id_to_p_ids_children:
+                p_id_to_p_ids_children[current_p_id_elternteil_2] = []
+            p_id_to_p_ids_children[current_p_id_elternteil_2].append(current_p_id)
+
+    p_id_to_fg_id = {}
+    next_fg_id = 0
+
+    for index, current_p_id in enumerate(p_id):
+        # Already assigned a fg
```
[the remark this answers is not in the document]
hmgaudecker: Ah, now I get it. Maybe we can be a bit more explicit in the comment above: `Already assigned a vg_id to this p_id via einstandspartner / parent.` Or so.

src/_gettsim/groupings.py
hmgaudecker: Sorry for the noise!

lars-reimann:
> The only thing I am wondering now is whether we should add a few test cases where things (i.e., hh_ids we may require, but parents / vg_ids / bg_ids) may come up in somewhat different places. IIUC, the tests all relate to a single household, are we sure it will work for multiple households with "strange" order?

I've added some tests for this.

hmgaudecker (approved):
Great! 

Rest are cosmetics / explanations for why we are skipping stuff so that does not get lost.

src/_gettsim_tests/test_data/groupings/2023/skip_enkelkind_und_sgb_ii_grossmutter.yaml
```diff
@@ -0,0 +1,31 @@
+---
+info:
+  note: Skipping test for now, because data is missing.
```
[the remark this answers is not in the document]
lars-reimann: The tests are taken from #509. For those tests, input columns are missing, and I cannot infer from the name, what should be tested exactly.

hmgaudecker (approved):
Excellent, thanks a lot!

src/_gettsim_tests/test_data/groupings/2023/skip_enkelkind_und_sgb_ii_grossmutter.yaml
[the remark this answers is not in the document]
ChristianZimpelmann: This is a child living together with their grandmother. If I remember correctly, they should not be put in the same bedarfsgemeinschaft as those never span more than 2 generations.

PR #602, opened 2023-07-12
Correct midijob_faktor_f

### LINKED ISSUE #597: BUG: Midijob Faktor F only changes once a year
### Bug description

Faktor F (used in the calculation of Bemessungsentgelt for Midijobs) is only adjusted at the beginning of each year.
`midijob_faktor_f` currently uses the social insurance contribution rates and therefore it also updates if the contribution rates change within a year.
This is a problem for the implementation of the gesetzliche Pflegeversicherung reform (#593) since the rates change in July 2023 this would currently also change the factor for the second half of 2023.

paulinaschroeder:

Use only parameters from the beginning of a year to calculate `midijob_faktor_f`.

Closes #597 

### Todo

- [x] fix `midijob_faktor_f`
- [x] Document in CHANGES.md.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ versioning](https://semver.org/) and all releases are available on
 
 ## Unpublished
 
+- {gh}`602` Correct `midijob_faktor_f` ({ghuser}`paulinaschroeder`).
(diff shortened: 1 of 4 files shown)


ChristianZimpelmann (approved):
Well done. Thanks!

Can be merged. (Feel free to add your contribution to CHANGES.md before)

PR #603, opened 2023-07-17
add anz_eig_kind_bis_24 to synthetic.

ChristianZimpelmann:

#595 added a new input variable, but we forgot to have the input variable created in `synthetic.py`.

### Todo

- [x] Add `anz_eig_kind_bis_24` to synthetic
- [x] Make sure GETTSIM with default targets can be run over data set created by synthetic (Should we maybe implement this as an automated test. @hmgaudecker, what do you think?)
- [x] While working on it, I also made some wording in the documentation with respect to time-varying policy functions more precise.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ versioning](https://semver.org/) and all releases are available on
 
 ## Unpublished
 
+- {gh}`603` Add anz_eig_kind_bis_24 to synthetic ({ghuser}`ChristianZimpelmann`).
 - {gh}`593` Implement reform of gesetzliche Pflegepflegeversicherung effective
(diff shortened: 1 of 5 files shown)


hmgaudecker:
>  Make sure GETTSIM with default targets can be run over data set created by synthetic (Should we maybe implement this as an automated test. @hmgaudecker, what do you think?)

Excellent idea! 

Eventually, we'll actually need to differentiate the default input variables by date. Assuming that `hat_kinder` was only relevant for Pflegeversicherung, this would be a nice case in point. But that's just a side thought, for now it would be great to have the test around.

ChristianZimpelmann:
> Eventually, we'll actually need to differentiate the default input variables by date.

Yes, good point.

PR #618, opened 2023-07-24
Apply dates_active decorator on abgelt_st.

hmgaudecker:

Abgeltungssteuer did not exist before 2008; make sure it is active in GETTSIM only after that date.

### DIFF
--- a/src/_gettsim/parameters/sozialv_beitr.yaml
+++ b/src/_gettsim/parameters/sozialv_beitr.yaml
@@ -67,19 +67,19 @@ beitr_satz:
   1988-01-01:
     deviation_from: previous
     ges_krankenv:
-      allgemein: 0.1288
+      mean_allgemein: 0.1288
   1989-01-01:
     deviation_from: previous
     ges_krankenv:
-      allgemein: 0.129
+      mean_allgemein: 0.129
   1990-01-01:
     deviation_from: previous
     ges_krankenv:
-      allgemein: 0.125
+      mean_allgemein: 0.125
   1991-01-01:
     deviation_from: previous
     ges_krankenv:
-      allgemein: .128
+      mean_allgemein: .128
   1992-01-01:
     deviation_from: previous
     arbeitsl_v: 0.0315
--- a/src/_gettsim/parameters/sozialv_beitr.yaml
+++ b/src/_gettsim/parameters/sozialv_beitr.yaml
@@ -165,7 +165,7 @@ beitr_satz:
     ges_rentenv: 0.0995
   2008-01-01:
     ges_krankenv:
-      allgemein: 0.14
+      mean_allgemein: 0.14
       sonderbeitrag: 0.009
     ges_pflegev:
       standard: 0.00975
--- a/src/_gettsim_tests/test_full_taxes_and_transfers.py
+++ b/src/_gettsim_tests/test_full_taxes_and_transfers.py
@@ -115,7 +115,7 @@ def test_data_types(
                 internal_type = year_functions[column_name].__annotations__["return"]
             else:
                 # TODO (@hmgaudecker): Implement easy way to find out expected type of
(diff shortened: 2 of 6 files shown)


src/_gettsim_tests/test_full_taxes_and_transfers.py
```diff
@@ -92,8 +114,8 @@ def test_data_types(
             elif column_name in year_functions:
                 internal_type = year_functions[column_name].__annotations__["return"]
             else:
-                # TODO (@hmgaudecker): Implement easy way to find out expected type of
-                #     aggregated functions
+                # TODO(@hmgaudecker): Implement easy way to find out
```
hmgaudecker, line 117: Seems unrelated, revert? 
```suggestion
                # TODO (@hmgaudecker): Implement easy way to find out expected type of
```

src/_gettsim/parameters/sozialv_beitr.yaml
```diff
@@ -165,7 +165,7 @@ beitr_satz:
     ges_rentenv: 0.0995
   2008-01-01:
     ges_krankenv:
-      allgemein: 0.14
+      mean_allgemein: 0.14
```
ChristianZimpelmann, line 168: We made the differentiation between `allgemein` and `mean_allgemein` on purpose. The latter refers to a period in which health insurance contribution rates differed over health insurance providers and we report the mean of them.

Let's discuss the best interface here.

src/_gettsim_tests/test_full_taxes_and_transfers.py
```diff
@@ -93,7 +115,7 @@ def test_data_types(
                 internal_type = year_functions[column_name].__annotations__["return"]
             else:
                 # TODO (@hmgaudecker): Implement easy way to find out expected type of
-                #     aggregated functions
+                #     expected type of aggregated functions
```
hmgaudecker, line 118:
```suggestion
                #     aggregated functions
```

src/_gettsim_tests/test_full_taxes_and_transfers.py
```diff
@@ -44,12 +44,23 @@ def test_full_taxes_and_transfers(
     policy_params, policy_functions = cached_set_up_policy_environment(
         date=test_data.date
     )
+    # TODO(@hmgaudecker): Remove again once unterhaltsvors_m is implemented
+    #     for more years.
+    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/479
+    if test_data.date.year < 2016:
```
ChristianZimpelmann, line 50: Seems a bit hacky, but should be fine for now.

ChristianZimpelmann (dismissed):
Great addition, thanks!

I think we should also add a test for the first year of existance of Abgeltungssteuer (2009).

Otherwise, just one comment we should discuss.

ChristianZimpelmann:
> Should be fine now, @mjbloemer and me did check that until 2008, contribution rates differed across health insurers in all years.

Great! That simplifies things.

I just encountered another problem, see #626.

Since `abgelt_st_y_tu` is actually part of `DEFAULT_TARGETS` (which is not yet time dependend), with this PR and #626 being fixed, we can no longer run GETTSIM with default arguments before 2009. 

Possible solution:

- change this line `targets = DEFAULT_TARGETS if targets is None else targets` such that DEFAULT_TARGETS without respective function to calculate are ignored.

hmgaudecker:
Great catch! 

> change this line `targets = DEFAULT_TARGETS if targets is None else targets` such that DEFAULT_TARGETS without respective function to calculate them are ignored.

I'd be fine with throwing an error for the time being. I don't think we should make a release before the interface is updated, anyhow. #627 can easily be incorporated in the process, then.

ChristianZimpelmann (approved):
Alright! Then we can merge.

But we should tackle #626 soon.

Issue #613, opened 2023-07-23
BUG: Consider income from self-employment for Elterngeld


ChristianZimpelmann:
### Bug description

During the calculation of Elterngeld, only `bruttolohn_m` is considered. We should also take care of subjects who have been self-employed before the birth.

PR #619, opened 2023-07-25
Consider self-employed income for Elterngeld

JonasSchernich:

Fixing #613 by considering `eink_selbst_m` for Elterngeld (not fully done, _elterngeld_proxy_eink_vorj_elterngeld_m yet unchanged)

### DIFF
--- a/src/_gettsim/transfers/elterngeld.py
+++ b/src/_gettsim/transfers/elterngeld.py
@@ -265,8 +265,9 @@ def _elterngeld_anz_mehrlinge_anspruch(
     return out
 
 
-def elterngeld_nettolohn_m(
+def elterngeld_nettoeink_m(  # noqa: PLR0913
     bruttolohn_m: float,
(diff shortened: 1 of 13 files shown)


hmgaudecker:
Great start! 

The way I understood the bug report, it was supposed to be about the prior year's income, which is unchanged. I honestly think we should get rid of that function and let the user calculate it from a prior run of GETTSIM / some educated guess / whatever. 

The change seems to be about current income, which leads to reductions of Elterngeld. 

On an unrelated note: I find this module extremely confusing. E.g., what is `erlass` supposed to abbreviate? `relev` is also a bit unspecific. Improved docstrings would help for a start... 

Thanks for tackling this!

hmgaudecker:
Closing this for now. We should eventually tackle this, but it should come as a serious shot at reasonable support for Elterngeld, which will be much more complicated.

Issue #558, opened 2023-03-21
ENH: Apply `dates_active` to more functions


hmgaudecker:
#495 added the `dates_active` decorator; #557 applies it to many cases. There are many more where application is more cumbersome because additional interfaces will require changes. This issue serves to collect them.

### Arbeitslosengeld II

- [x] Set start date

### Minijobs

- [x] Get rid of ost/west distinction starting in 2000

### Midijobs

- [x] Set start date for midijobs: Remove before 2003-04-01
- [x] Remove "midijob" entries in soz_vers_beitr.yaml
- [x] Will need to fix interfaces of functions before that date

### Kinderzuschlag

- [x] Set start date

hmgaudecker:
If you come across anything, please add to the above list.

PR #623, opened 2023-07-25
Apply dates_active decorator to midijob and ges_pflegev

hmgaudecker:

Midijobs were introduced in 2004. There used to be a hacky solution setting the gleitzone to have length zero. With the `dates_active` machinery, we can make this nicer.

Similar comments apply to `ges_pflegev`, which was introduced in 1995 and changed its rules in 2005.


hmgaudecker:
Implements a bunch of things from #558.

src/_gettsim/social_insurance_contributions/ges_pflegev.py
```diff
@@ -247,13 +372,45 @@ def ges_pflegev_beitr_selbst_m(
...
+        See :func:`_ges_krankenv_bemessungsgrundlage_rente_m`.
+    ges_pflegev_beitr_satz
+        See :func:`ges_pflegev_beitr_satz`.
+
+    Returns
+    -------
+    Monthly health insurance contributions for pension income.
+
+    """
+    out = _ges_krankenv_bemessungsgrundlage_rente_m * (ges_pflegev_beitr_satz * 2)
```
MImmesberger, line 398: I read [here](https://www.spiegel.de/wirtschaft/urteil-rentner-muessen-pflegeversicherung-allein-bezahlen-a-451421.html#) that until March 2004, the public pension fund did pay half of the Pflegeversicherungsbeitrag. I think this is not implemented here.

src/_gettsim/social_insurance_contributions/ges_krankenv.py
```diff
@@ -1,7 +1,48 @@
 from _gettsim.shared import dates_active
 
 
-def ges_krankenv_beitr_m(  # noqa: PLR0913
+@dates_active(end="2003-03-31", change_name="ges_krankenv_beitr_m")
+def ges_krankenv_beitr_m_vor_midijob(
```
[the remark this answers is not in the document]
hmgaudecker: Done.

src/_gettsim/social_insurance_contributions/ges_pflegev.py
hmgaudecker: Excellent catch! Would prefer to do that separately, though. Can you open an issue and maybe a PR?

MImmesberger (approved):

PR #624, opened 2023-07-25
fix: don't create functions for other time units if this leads to a cycle in the graph

### LINKED ISSUE #621: Misleading error message when a root node with possible time conversion is missing
### Bug description
The following error came up after changing the calculation of Elterngeld by taking selbst_eink into account:

**E           ValueError: The DAG contains one or more cycles:
E           
E           [
E               "['eink_selbst_m', 'eink_selbst_y']",
E           ]**

The error could be fixed by adding the selbst_eink_m into the tests, however to me this solution does not appear to be appropriate for that error message.

---

#### Code Sample, a copy-pastable example
example code from file hh_id_2.yaml(same thing for other eltergeld testf iles):
code generating the issue:
```python
    bruttolohn_m:
      - 1800.0
    bruttolohn_vorj_m:
      - 1800.0
```
code not generating it:
```python
    bruttolohn_m:
      - 1800.0
     eink_selbst_m:
      - 0.0
    bruttolohn_vorj_m:
      - 1800.0
```

#### Problem description

I don't see how fixing the test files can solve the issue specified above, since no loop was stopped (as far as I can tell)

lars-reimann:

Closes #621

No longer create functions for other time units if this leads to a loop. This leads to better error messages, e.g. for the tests from #619:

**Before:** 

```txt
E           ValueError: The DAG contains one or more cycles:
E           
E           [
E               "['eink_selbst_y', 'eink_selbst_m']",
E           ]
```

**After:**

```txt
E           ValueError: The following data columns are missing.
E           
E           [
E               "eink_selbst_m",
E           ]
```

### Todo

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ versioning](https://semver.org/) and all releases are available on
 
 ## Unpublished
 
+- {gh}`624` Don't create functions for other time units if this leads to a loop ({ghuser}`lars-reimann`).
 - {gh}`603` Add anz_eig_kind_bis_24 to synthetic ({ghuser}`ChristianZimpelmann`).
 - {gh}`593` Implement reform of gesetzliche Pflegepflegeversicherung effective
   as of 2023-07-01 ({ghuser}`paulinaschroeder`).
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ versioning](https://semver.org/) and all releases are available on
 
 ## Unpublished
 
-- {gh}`624` Don't create functions for other time units if this leads to a loop ({ghuser}`lars-reimann`).
+- {gh}`624` Don't create functions for other time units if this leads to a cycle in the graph ({ghuser}`lars-reimann`).
 - {gh}`603` Add anz_eig_kind_bis_24 to synthetic ({ghuser}`ChristianZimpelmann`).
 - {gh}`593` Implement reform of gesetzliche Pflegepflegeversicherung effective
   as of 2023-07-01 ({ghuser}`paulinaschroeder`).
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -7,6 +7,7 @@ versioning](https://semver.org/) and all releases are available on
 
 ## Unpublished
 
+- {gh}`624` Don't create functions for other time units if this leads to a loop ({ghuser}`lars-reimann`).
```
hmgaudecker, line 10:
```suggestion
- {gh}`624` Don't create functions for other time units if this leads to a cycle in the graph ({ghuser}`lars-reimann`).
```

This is the correct terminology, with "loop" I always think "for" :-)

src/_gettsim_tests/test_time_conversion.py
```diff
@@ -250,3 +236,12 @@ def test_should_apply_converter(self):
         function = _create_function_for_time_unit("test", None, d_to_w)
 
         assert function(1) == 7
+
+
+# https://github.com/iza-institute-of-labor-economics/gettsim/issues/621
+def test_should_not_create_loops():
```
hmgaudecker, line 242:
```suggestion
def test_should_not_create_cycle():
```

src/_gettsim/time_conversion.py
```diff
@@ -304,9 +306,11 @@ def _create_time_conversion_functions(
 
         missing_time_units = [unit for unit in all_time_units if unit != time_unit]
         for missing_time_unit in missing_time_units:
-            result[
-                f"{base_name}{missing_time_unit}{aggregation}"
-            ] = _create_function_for_time_unit(
+            new_name = f"{base_name}{missing_time_unit}{aggregation}"
```
hmgaudecker, line 309: Could you add a comment on how this achieves the desired behavior? I can't wrap my head around it in 3min and this will only get worse once I have not thought about this problem the day before :laughing:

hmgaudecker (commented):
Excellent! Just a bit of terminology and a request to comment the code !

hmgaudecker (approved):

hmgaudecker:
Thanks a lot!

PR #630, opened 2023-08-01
Fix `piecewise_polynomial`, fixes #629

### LINKED ISSUE #629: BUG: `piecewise_polynomial` ignores intercepts except for the first one
### Bug description

The function [check_intercepts](https://github.com/iza-institute-of-labor-economics/gettsim/blob/f1ff0c0122c81da29d5b24e145321981a31460ea/src/_gettsim/piecewise_functions.py#L245C1-L290C22) responsible for preparing the intercepts for piecewise functions is not working properly in case intercepts are supplied for every part of the function. In that case the whole array is overwritten with the first intercept. We were confronted with that issue while working on #577  but apparently other piecewise functions should have encountered the same problem. 
For example in [eink_st.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/blob/f1ff0c0122c81da29d5b24e145321981a31460ea/src/_gettsim/parameters/eink_st.yaml#L229C1-L266C38)

#### Proposed solution
The problem can be fixed by adding a condition checking for that case and preventing the overwriting. 

```python
def check_intercepts(
    parameter_dict, parameter, lower_thresholds, upper_thresholds, rates, keys
):
    """Check and transfer raw intercepte data. If necessary create intercepts.

    Transfer and check raw rates data, which needs to be specified in a
    piecewise_polynomial layout in the yaml file.

    Parameters
    ----------
    parameter_dict
    parameter
    lower_thresholds
    upper_thresholds
    rates
    keys

    Returns
    -------

    """
    intercepts = numpy.zeros(len(keys))
    count_intercepts_supplied = 1

    if "intercept_at_lower_threshold" not in parameter_dict[0]:
        raise ValueError(f"The first piece of {parameter} needs an intercept.")
    else:
        intercepts[0] = parameter_dict[0]["intercept_at_lower_threshold"]
        # Check if all intercepts are supplied.
        for interval in keys[1:]:
            if "intercept_at_lower_threshold" in parameter_dict[interval]:
                count_intercepts_supplied += 1
                intercepts[interval] = parameter_dict[interval][
                    "intercept_at_lower_threshold"
                ]
        if (count_intercepts_supplied > 1) & (count_intercepts_supplied != len(keys)):
            raise ValueError(
                "More than one, but not all intercepts are supplied. "
                "The dictionaries should contain either only the lowest intercept "
                "or all intercepts."
            )
        elif count_intercepts_supplied == len(keys):
            pass

        else:
            intercepts = create_intercepts(
                lower_thresholds, upper_thresholds, rates, intercepts[0]
            )
    return intercepts
```

nafetsk:

Fixing [this bug](https://github.com/iza-institute-of-labor-economics/gettsim/issues/629)

### Todo

- [ ] Check if other parts of the code are effected by this

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 - {gh}`630` Fixes bug in `piecewise_polynomial` that ignored jumps in intercepts
   ({ghuser}`nafetsk`).
+- {gh}`639` Infer `columns_overriding_functions` for `compute_taxes_and_transfers` from
+  functions and data ({ghuser}`lars-reimann`).
 - {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
   ({ghuser}`lars-reimann`).
 - {gh}`618`, {gh}`623` Apply `@dates_active` decorator to Abgeltungssteuer, Midi- and
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 - {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
   ({ghuser}`lars-reimann`).
-- {gh}`630` Fixes bug in `piecewise_polynomial` that ignored jumps in intercepts ({ghuser}`nafetsk`).
+- {gh}`630` Fixes bug in `piecewise_polynomial` that ignored jumps in intercepts
+  ({ghuser}`nafetsk`).
 - {gh}`618`, {gh}`623` Apply `@dates_active` decorator to Abgeltungssteuer, Midi- and
   Minijobs, Pflegeversicherung. ({ghuser}`hmgaudecker`).
 - {gh}`624` Don't create functions for other time units if this leads to a cycle in the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,10 +6,10 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
-- {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
-  ({ghuser}`lars-reimann`).
 - {gh}`630` Fixes bug in `piecewise_polynomial` that ignored jumps in intercepts
   ({ghuser}`nafetsk`).
+- {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
+  ({ghuser}`lars-reimann`).
 - {gh}`618`, {gh}`623` Apply `@dates_active` decorator to Abgeltungssteuer, Midi- and
   Minijobs, Pflegeversicherung. ({ghuser}`hmgaudecker`).
 - {gh}`624` Don't create functions for other time units if this leads to a cycle in the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,16 +1,20 @@
 # Changes
 
-This is a record of all past `gettsim` releases and what went into them
-in reverse chronological order. We follow [semantic
-versioning](https://semver.org/) and all releases are available on
-[Anaconda.org](https://anaconda.org/gettsim/gettsim).
+This is a record of all past `gettsim` releases and what went into them in reverse
+chronological order. We follow [semantic versioning](https://semver.org/) and all
+releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsim).
 
 ## Unpublished
 
-- {gh}`624` Don't create functions for other time units if this leads to a cycle in the graph ({ghuser}`lars-reimann`).
+- {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
+  ({ghuser}`lars-reimann`).
+- {gh}`618`, {gh}`623` Apply `@dates_active` decorator to Abgeltungssteuer, Midi- and
+  Minijobs, Pflegeversicherung. ({ghuser}`hmgaudecker`).
+- {gh}`624` Don't create functions for other time units if this leads to a cycle in the
+  graph ({ghuser}`lars-reimann`).
 - {gh}`603` Add anz_eig_kind_bis_24 to synthetic ({ghuser}`ChristianZimpelmann`).
-- {gh}`593` Implement reform of gesetzliche Pflegepflegeversicherung effective
-  as of 2023-07-01 ({ghuser}`paulinaschroeder`).
+- {gh}`593` Implement reform of gesetzliche Pflegepflegeversicherung effective as of
+  2023-07-01 ({ghuser}`paulinaschroeder`).
 - {gh}`602` Correct `midijob_faktor_f` ({ghuser}`paulinaschroeder`).
 - {gh}`600` Extend the `access_different_date` functionality for `jahresanfang`.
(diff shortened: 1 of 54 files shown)


hmgaudecker:
Thanks, looks reasonable at first sight!  We will need a testcase for this, which fails on main and pins down precisely the change in behaviour.

I do agree, though, that this was probably just an omission in setting up the function back in the day. We probably would not have needed that much code if the only goal was to only ever use the first intercept :see_no_evil: 

Tiny feedback: I had a good look at the code example in #629 before you opened up this PR. If you or anyone else is about to open a PR alongside, it would be great to add a note in the bug report.

hmgaudecker:
Btw, it should not matter for `eink_st` because that function does not have jumps.

nafetsk:
> Btw, it should not matter for `eink_st` because that function does not have jumps.

I think my message was a little confusing. I was not referring to the `eink_st` itself but to the parameter `rente_ertragsanteil` which is located in [eink_st.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/blob/f1ff0c0122c81da29d5b24e145321981a31460ea/src/_gettsim/parameters/eink_st.yaml#L229C1-L266C38). This parameter has jumps between each section and should therefore struggle with the same bug as `zurechnungszeitsgrenze_in_m`

hmgaudecker:
> to the parameter `rente_ertragsanteil` which is located in [eink_st.yaml](https://github.com/iza-institute-of-labor-economics/gettsim/blob/f1ff0c0122c81da29d5b24e145321981a31460ea/src/_gettsim/parameters/eink_st.yaml#L229C1-L266C38).

As mentioned over at [#629](https://github.com/iza-institute-of-labor-economics/gettsim/issues/629#issuecomment-1665481655), this might actually be a place where we do not want to use piecewise polynomial at all. That should really only be used when we need it at the individual level, but IIUC, for `rente_ertragsanteil` it only picks a value, which is the same for everyone in a given year?

hmgaudecker:
I'll be happy to review, but as mentioned above:

>  We will need a testcase for this, which fails on main and pins down precisely the change in behaviour.

nafetsk:
Hey, I added a test which checks if the intercepts are loaded properly. It fails in main and is successful with the bug fix in this PR. The test is quite simple and just checks for the ‘rente_ertragsanteil’ case mentioned above. If that’s not enough I am happy to get feedback on how to do it differently.

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,16 @@
+"""
+Test for BUG-fix #629
+https://github.com/iza-institute-of-labor-economics/gettsim/issues/629
+"""
+import numpy as np
+
+from _gettsim_tests._helpers import cached_set_up_policy_environment
```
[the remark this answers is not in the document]
nafetsk: Ok thanks for the feedback. Since the function I changed `check_intercepts` is only used in `get_piecewise_parameters` and not in `piecewise-polynomial` itself, i will test `get_piecewise_parameter` for a more precise outcome.

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,49 @@
+"""
+Tests for `piecewise_polynomial`
+"""
+import numpy as np
+from _gettsim.piecewise_functions import (
+    get_piecewise_parameters,
+)
+
+
+def test_get_intercepts_for_piecewise_polynomial():
```
hmgaudecker, line 10:
```suggestion
def test_get_piecewise_parameters_all_intercepts_supplied():
```

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,49 @@
...
+    params = {
+        "type": "piecewise_quadratic",
```
hmgaudecker, line 12:
```suggestion
        "type": "piecewise_linear",
```

Or did I miss something?

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,49 @@
...
+        },
+        3: {
+            "lower_threshold": 2041,
+            "upper_threshold": "inf",
+            "rate_linear": 0,
+            "intercept_at_lower_threshold": 1,
+        },
+    }
+
+    piecewise_parameters = get_piecewise_parameters(
```
hmgaudecker, line 40:
```suggestion
    expected = np.array([0.27, 0.5, 0.8, 1]
    actual = get_piecewise_parameters(
```

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,49 @@
...
+        params,
+        "eink_st_tarif",
+        "linear",
+    )
+
+    assert (
```
hmgaudecker, line 46:
```suggestion
    np.testing.assert_almost_equal(actual, expected, decimal=10)
```

You never know when floating point comparisons may fail...

hmgaudecker (commented):
Very nice, mostly stylistic stuff left!

amageh:
Great job, @nafetsk! 

Thank you for the review and excellent feedback @hmgaudecker. I think all comments have been addressed now. We will merge the PR into 'main' if you approve.

CHANGES.md
```diff
@@ -8,6 +8,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 - {gh}`638` Don't use functions in `compute_taxes_and_transfers` that are not active
   ({ghuser}`lars-reimann`).
+- {gh}`630` Fixes bug in `piecewise_polynomial` that ignored jumps in intercepts
```
hmgaudecker, line 11: Should always go on top, timing of commit to main matters, not id.

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,45 @@
...
+        },
+        3: {
+            "lower_threshold": 2041,
+            "upper_threshold": "inf",
+            "rate_linear": 0,
+            "intercept_at_lower_threshold": 1,
+        },
+    }
+    expected = np.array([0.27, 0.5, 0.8, 1])
+    actual = get_piecewise_parameters(
```
hmgaudecker, line 40: Better call by name, not by value -- I'd have to guess what the second and third arguments do. Also, `eink_st_tarif` seems misleading?

src/_gettsim_tests/test_piecewise_polynomial.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+from _gettsim.piecewise_functions import (
+    get_piecewise_parameters,
+)
+
+
+def test_get_piecewise_parameters_all_intercepts_supplied():
+    params = {
+        "type": "piecewise_linear",
+        "progressionsfaktor": True,
```
hmgaudecker, line 13: Is this required? What does it do?

hmgaudecker (approved):
Looks great, many thanks! Just a few more tiny things I'd like to see changed / clarified.

CHANGES.md
amageh: Thanks, I was unsure about this.

src/_gettsim_tests/test_piecewise_polynomial.py
nafetsk: not in our case. I deleted it for simplicity reasons

PR #632, opened 2023-08-07
Monthly precision in Zugangsfaktor

### LINKED ISSUE #612: ENH: Taking the month of retirement into account for pension benefit calculation

Currently, the `ges_rente_zugangsfaktor` function only considers the year of retirement and compares it to the year of birth. Hence, in the current implementation, individuals that retire early or late are treated as if they retired 0, 12, 24,... months before or after the statutory retirement age threshold.

The calculation of the Zugangsfaktor should be sensitive to the month of retirement. The calculation of pension benefits depends on how many months early or late the person retires.

The current implementation of the statutory retirement age thresholds is already based on the month of birth. The only missing part is probably the correct calculation of the Zugangsfaktor.

TeBackh:

Closes #612: Introducing monthly precision in Zugangsfaktor calc

### Todo

- [x] Introducing new variable month of retirement
- [x] implementing function of age of retirement incorporating specifics of law
- [x] inserting age of retirement in zugangsfaktor calculation
- [x] adding input variable in test cases

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,7 +73,11 @@ repos:
     rev: 1.7.0
     hooks:
       - id: nbqa-black
+        additional_dependencies:
+          - black==23.7.0
       - id: nbqa-ruff
+        additional_dependencies:
+          - ruff==v0.0.281
         args:
           - --ignore=B018,T201
   - repo: https://github.com/executablebooks/mdformat
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,9 +6,9 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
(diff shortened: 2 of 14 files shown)


timmens (approved):
👍  (I am pretty sure that `float("Nan")` won't work with JAX, but this is not a problem from my side. I can tackle that problem when we finish the JAX compatibility.)

MImmesberger (approved):

MImmesberger:
@timmens It seems like the codecov check doesn't run because there is no coverage data for the pull request base. (I probably made a mistake when merging `main` into the PR?) After some googling, I have no idea how to fix it. Any advice?

hmgaudecker:
No worries, sometimes it gets messed up. Had the same today in one of Lars' PRs. Maybe main changed too much in the meantime?

Issue #378, opened 2022-04-12
Re-design of interface / API


ChristianZimpelmann:
### Current situation

Running GETTSIM requires two function calls: `set_up_policy_environment` and `compute_taxes_and_transfers`

### Proposed implementation

- Allow for running GETTSIM in one function call: `compute_taxes_and_transfers`
- Allow for running GETTSIM in three steps: `set_up_policy_environment`, `set_up_execution_order` (sets up dag), `execute_calculation_of_taxes_and_transfers` (executes dag)

### Considered alternatives

- Different names for functions (I first thought about `set_up_dag` and `execute_dag`, but that might be too much details about the implementation?)
- Split up in more steps

### Note

While working on the PR, we could also think about the following:
- Add new argument `evaluation_date` (This would be a prerequiste for working on #211 and  #317 )
- Let `set_up_policy_environment` return all internal functions (instead of just the time dependent ones)
- allowing for longer variable names

hmgaudecker:
Just a quick shot, working backwards

> Add new argument `evaluation_date` (This would be a prerequiste for working on [Age / time / cohort effects: GETTSIM version #211](https://github.com/iza-institute-of-labor-economics/gettsim/issues/211) and  [Define age as a function. #317](https://github.com/iza-institute-of-labor-economics/gettsim/pull/317) )

Definitely do that. If for the moment we restrict `policy_date` to `evaluation_date`, no problem, but we should define the interface right away.

> Let `set_up_policy_environment` return all internal functions (instead of just the time dependent ones)

Definitely. The distinction is just arbitrary. 

For the internal steps, I would do:
1. `read_params` (only gets raw yaml files)
2. `convert_params(raw_params, policy_year, evaluation_date, ...)` (does stuff like applying transformation for `piecewise_polynomial`)
3. `get_raw_policy_functions` (returns all functions that will form input for the DAG in raw form, before renamings)
4. `select_policy_functions(raw_policy_functions, params)` (applies DAG and selects time-varying functions, partials params)
5. `execute_calculation(policy_functions, data)`

Steps 2-4 will depend a bit on how we will handle the time-varying functions stuff. I will need to write the GEP first, but cannot possibly find room for that in the next two weeks.

ChristianZimpelmann:
Alright, let's put this issue on hold until then.

hmgaudecker:
*(Based on discussion we are having right now in a Zoom call re eligibility ages in #396)*

[...]

PR #638, opened 2023-08-14
fix: don't use functions in `compute_taxes_and_transfers` that are not active

### LINKED ISSUE #626: BUG: Deactivated functions are used by `compute_taxes_and_transfers`
### Bug description

In #618, the functions `abgelt_st_y_tu` is made available only from 2009 on. However, in 2008 you can still request `abgelt_st_y_tu` without recieving an error. While `abgelt_st_y_tu` is not part of `policy_functions`, when calling `compute_taxes_and_transfers`, GETTSIM seems to collect the function from `PATHS_TO_INTERNAL_FUNCTIONS`

Interestingly, I get an error as expected if the `change_name` attribute is used in the decorator:

```python
@dates_active(start="2009-01-01", change_name="abgelt_st_y_tu")
def abgelt_st_y_foo_tu(zu_verst_kapitaleink_y_tu: float, abgelt_st_params: dict) -> float:
    pass
```

### Proposed solution

Is there any reason functions from `PATHS_TO_INTERNAL_FUNCTIONS` are still collected in `compute_taxes_and_transfers` instead of just having all active functions in `policy_params`?

lars-reimann:

Closes #626
Closes partially #378

* Don't load any internal functions implicitly anymore.
* Also load functions that are not time-dependent into the policy environment.

### Todo

### DIFF
--- a/src/_gettsim/functions_loader.py
+++ b/src/_gettsim/functions_loader.py
@@ -37,7 +37,7 @@ if TYPE_CHECKING:
 
 
 def load_and_check_functions(
-    user_functions_raw,
+    functions_raw,
     columns_overriding_functions,
     targets,
     data_cols,
@@ -45,8 +45,7 @@ def load_and_check_functions(
 ):
     """Create the dict with all functions that may become part of the DAG by:
 
-    - merging user and internal functions
-    - vectorize all functions
+    - vectorizing all functions
     - adding time conversion functions, aggregation functions, and combinations
 
     Check that:
@@ -55,8 +54,8 @@ def load_and_check_functions(
 
     Parameters
     ----------
-    user_functions_raw : dict
-        A dictionary mapping column names to policy functions by the user.
+    functions_raw : dict
+        A dictionary mapping column names to policy functions.
     columns_overriding_functions : str list of str
         Names of columns in the data which are preferred over function defined in the
         tax and transfer system.
@@ -81,17 +80,17 @@ def load_and_check_functions(
     """
 
     # Load user and functions.
-    user_functions_raw = [] if user_functions_raw is None else user_functions_raw
-    user_functions = _load_functions(user_functions_raw)
+    functions_raw = [] if functions_raw is None else functions_raw
+    functions = _load_functions(functions_raw)
 
     # Vectorize functions.
-    vectorized_user_functions_functions = {
-        fn: _vectorize_func(f) for fn, f in user_functions.items()
+    vectorized_functions = {
+        fn: _vectorize_func(f) for fn, f in functions.items()
     }
 
     # Create derived functions
     time_conversion_functions, aggregation_functions = _create_derived_functions(
-        vectorized_user_functions_functions, targets, data_cols, aggregation_specs
+        vectorized_functions, targets, data_cols, aggregation_specs
     )
 
     # Check for implicit overlap of functions and data columns.
@@ -100,7 +99,7 @@ def load_and_check_functions(
     ]
     for funcs, name in zip(
         [
-            user_functions,
+            functions,
             aggregation_functions,
             time_conversion_functions,
         ],
@@ -110,7 +109,7 @@ def load_and_check_functions(
 
     all_functions = {
         **time_conversion_functions,
-        **vectorized_user_functions_functions,
+        **vectorized_functions,
         **aggregation_functions,
     }
 
--- a/src/_gettsim/functions_loader.py
+++ b/src/_gettsim/functions_loader.py
@@ -80,7 +80,7 @@ def load_and_check_functions(
 
     """
 
-    # Load user and internal functions.
+    # Load user and functions.
     user_functions_raw = [] if user_functions_raw is None else user_functions_raw
     user_functions = _load_functions(user_functions_raw)
 
--- a/src/_gettsim/functions_loader.py
+++ b/src/_gettsim/functions_loader.py
@@ -86,8 +86,7 @@ def load_and_check_functions(
 
     # Vectorize functions.
     vectorized_user_functions_functions = {
-        fn: _vectorize_func(f)
-        for fn, f in user_functions.items()
+        fn: _vectorize_func(f) for fn, f in user_functions.items()
     }
 
     # Create derived functions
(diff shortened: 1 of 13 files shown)


hmgaudecker:
Just checked — does the latter test fail because we are requesting a function in a year where it is inactive? In that case it should fail! Just change the behavior of the test then so that it expected the error that occurs, would actually be a good check.

lars-reimann:
> Just checked — does the latter test fail because we are requesting a function in a year where it is inactive? In that case it should fail! Just change the behavior of the test then so that it expected the error that occurs, would actually be a good check.

The plot test fails now because
* Previously the `functions` passed to `plot_dag` were treated as functions that were added to the implicitly loaded internal functions, so even though `functions` was empty, `"erwachsene_alle_rentner_hh"` was available.
* Now the `functions` passed to `plot_dag` are treated as the complete list of functions, and nothing is added implicitly anymore, so `"erwachsene_alle_rentner_hh"` is no longer available.

I think expecting this test to fail would be best solution.

hmgaudecker:
> * Previously the `functions` passed to `plot_dag` were treated as functions that were added to the implicitly loaded internal functions, so even though `functions` was empty, `"erwachsene_alle_rentner_hh"` was available.

Ah, now I understand, so `plot_dag` does not reference a specific policy environment. Yes, then we should require passing all functions. Just document that clearly now and make sure the examples from the docs still work. 

I think there was an issue collecting things for the API redesign, please make a note there, too!

hmgaudecker:
Btw, shouldn't the title of this PR and the entry in the changelog be the exact opposite?

lars-reimann:
> Btw, shouldn't the title of this PR and the entry in the changelog be the exact opposite?

It should be correct: Previously, the policy environment that was created by `create_policy_environment` only contained time-dependent functions (the ones with `@dates_active` decorator). This was also the behavior before that decorator was introduced. Now the policy environment contains all functions that are active at the given date, including the ones that are perpetually active.

lars-reimann:
> > * Previously the `functions` passed to `plot_dag` were treated as functions that were added to the implicitly loaded internal functions, so even though `functions` was empty, `"erwachsene_alle_rentner_hh"` was available.
> 
> Ah, now I understand, so `plot_dag` does not reference a specific policy environment. Yes, then we should require passing all functions. Just document that clearly now and make sure the examples from the docs still work.

The [plots ](https://gettsim--638.org.readthedocs.build/en/638/how_to_guides/visualizing_the_system.html) still look good.

lars-reimann:
> I think there was an issue collecting things for the API redesign, please make a note there, too!

I've like to the related issue in the first comment now.

src/_gettsim/functions_loader.py
hmgaudecker, line 1: I think I generally like what you did, but I am a little lost. Before we had user-provided functions and internal ones, but that distinction is gone or more blurry now? I don't quite know what to make of names like `vectorized_user_functions_functions`. 

Again, I believe that what the code does is likely to be fine, but currently the identifiers don't quite match what is happening. Or do I better look at #639 first?

hmgaudecker: Thanks for the explainer! I suppose the behaviour changed a bit with the introduction of `@dates_active` and from that point on, current behavior really had stopped making any sense.

src/_gettsim/functions_loader.py
```diff
@@ -103,18 +97,17 @@ def load_and_check_functions(
...
     for funcs, name in zip(
         [
-            internal_functions,
-            user_functions,
+            functions,
             aggregation_functions,
             time_conversion_functions,
         ],
-        ["internal", "user", "aggregation", "time_conversion"],
+        ["user", "aggregation", "time_conversion"],
```
hmgaudecker, line 104:
```suggestion
        ["hard-coded", "aggregation", "time_conversion"],
```

src/_gettsim/functions_loader.py
lars-reimann: > Or do I better look at https://github.com/iza-institute-of-labor-economics/gettsim/pull/639 first?

No, I think #639 should have priority.

src/_gettsim_tests/test_full_taxes_and_transfers.py
```diff
@@ -75,6 +77,8 @@ def test_data_types(  # noqa: PLR0912
     imports = _convert_paths_to_import_strings(PATHS_TO_INTERNAL_FUNCTIONS)
     functions = _load_functions(imports)
 
+    out = OUT_COLS if test_data.date.year <= 2008 else [*OUT_COLS, "abgelt_st_y_tu"]
```
hmgaudecker, line 80: Same

hmgaudecker (approved):
Looks great, just some tiny cosmetic changes.

src/_gettsim/functions_loader.py
hmgaudecker: ?

hmgaudecker: > Sorry, I had too many tabs open. #638 (this one) should have priority.

I better not send a screenshot of my tabs...

src/_gettsim_tests/test_data/full_taxes_and_transfers/2008/hh_id_1.yaml
```diff
@@ -344,5 +344,15 @@ inputs:
       - 0.0
       - 0.0
       - 0.0
+    sum_eink_ohne_kapital_eink_y:
```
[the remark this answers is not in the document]
lars-reimann: Without these additions, the tests fail, since `sum_eink_ohne_kapital_eink_y` and `abgelt_st_y_tu` are not defined in 2008, but other functions seem to rely on them (which may be a bug).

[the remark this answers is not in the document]
hmgaudecker: That is a bug indeed. Fixed it below.

And we should improve the error message. It should list the functions that require some node. Could you open an issue for that, please?

hmgaudecker:
Cool, I think this can be merged then -- please don't forget that issue, though!

src/_gettsim_tests/test_data/full_taxes_and_transfers/2008/hh_id_1.yaml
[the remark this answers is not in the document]
lars-reimann: -> #640

PR #639, opened 2023-08-14
feat: infer `columns_overriding_functions` from functions and data

lars-reimann:

Relates to #378.

Previously, users had to explicitly pass the `columns_overriding_functions` to `compute_taxes_and_transfers`. They had to follow some rules, however:

1. All columns that were in the `data` and in `functions` (or one of the derived functions for other time units #583 or groupings #601) must be included.
2. Only columns that were in the `data` must be included.
3. Only columns that were in `functions` (or one of the derived functions for other time units #583 or groupings #601) must be included.

The result was that the user had to specify exactly the columns in the data that clash with any function; otherwise an error would be raised. Once that user had succeeded with this guessing game, which became more difficult with the introduction of derived functions, the result would be that the values in the column would be used, and the corresponding function would not be executed. 

It was also difficult to specify `columns_overriding_functions` for our tests that target many different years. During some periods, a function might not exist yet, so we needed to manually create different `columns_overriding_functions` lists for those different periods (e.g. https://github.com/iza-institute-of-labor-economics/gettsim/blob/91b36477c861bb9c4fe3ede4a5924f24dcc20ef2/src/_gettsim_tests/test_full_taxes_and_transfers.py#L90)

This PR removes the `columns_overriding_functions` parameter from `compute_taxes_and_transfers` and instead derives it automatically using the rules defined above.

See also: https://github.com/iza-institute-of-labor-economics/gettsim/issues/378#issuecomment-1613164091

### Todo

### DIFF
--- a/src/_gettsim/interface.py
+++ b/src/_gettsim/interface.py
@@ -89,7 +89,13 @@ def compute_taxes_and_transfers(  # noqa: PLR0913
     )
     data = _convert_data_to_correct_types(data, functions_overridden)
     columns_overriding_functions = set(functions_overridden)
-    _warn_if_functions_and_columns_overlap(columns_overriding_functions)
+
+    # Warn if columns override functions.
+    if columns_overriding_functions:
+        warnings.warn(
+            FunctionsAndColumnsOverlapWarning(columns_overriding_functions),
+            stacklevel=2,
+        )
 
     # Select necessary nodes by creating a preliminary DAG.
     nodes = set_up_dag(
@@ -359,17 +365,17 @@ def _create_input_data(
     return input_data
 
 
-def _warn_if_functions_and_columns_overlap(
-    columns_overriding_functions: set[str],
-) -> None:
-    """Warn if functions which compute columns overlap with existing columns.
+class FunctionsAndColumnsOverlapWarning(UserWarning):
+    """
+    Warning that functions which compute columns overlap with existing columns.
 
     Parameters
     ----------
     columns_overriding_functions : set[str]
         Names of columns in the data that override hard-coded functions.
     """
-    if columns_overriding_functions:
+
+    def __init__(self, columns_overriding_functions: set[str]) -> None:
         n_cols = len(columns_overriding_functions)
         first_part = format_errors_and_warnings(
             f"Your data provides the column{'' if n_cols == 1 else 's'}:"
@@ -381,10 +387,11 @@ def _warn_if_functions_and_columns_overlap(
             hard-coded functions of the taxes and transfers system.
             If you want {'this' if n_cols == 1 else 'a'} data column to be used
             instead of calculating it within GETTSIM you need not do anything.
-            If you want {'this' if n_cols == 1 else 'a'} data column to be calculated
-            by hard-coded functions, remove {'it' if n_cols == 1 else 'them'} from the
-            *data* you pass to GETTSIM. {'' if n_cols == 1 else '''You need to pick one
-            option for each column that appears in the list above.'''}
+            If you want {'this' if n_cols == 1 else 'a'} data column to be
+            calculated by hard-coded functions, remove
+            {'it' if n_cols == 1 else 'them'} from the *data* you pass to GETTSIM.
+            {'' if n_cols == 1 else '''You need to pick one option for each column
+            that appears in the list above.'''}
             """
         )
         how_to_ignore = format_errors_and_warnings(
@@ -399,15 +406,7 @@ def _warn_if_functions_and_columns_overlap(
                 )
             """
         )
-        warnings.warn(
-            f"{first_part}\n{formatted}\n{second_part}\n{how_to_ignore}",
-            category=FunctionsAndColumnsOverlapWarning,
-            stacklevel=2,
-        )
-
-
-class FunctionsAndColumnsOverlapWarning(UserWarning):
-    pass
+        super().__init__(f"{first_part}\n{formatted}\n{second_part}\n{how_to_ignore}")
 
 
 def _fail_if_duplicates_in_columns(data):
--- a/src/_gettsim/interface.py
+++ b/src/_gettsim/interface.py
@@ -387,7 +387,27 @@ def _warn_if_functions_and_columns_overlap(
             option for each column that appears in the list above.'''}
             """
         )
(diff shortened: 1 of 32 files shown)


lars-reimann:
Issuing a warning makes sense, I'll add that later.

src/_gettsim_tests/test_full_taxes_and_transfers.py
```diff
@@ -49,23 +48,11 @@ def test_full_taxes_and_transfers(
     if test_data.date.year <= 2008:
         out.remove("abgelt_st_y_tu")
 
-    # TODO(@hmgaudecker): Remove again once unterhaltsvors_m is implemented
```
hmgaudecker, line 52: I moved this to-do to #479 and #585. Would be great to ping the owner of a to-do before deleting it -- better safe than sorry!

src/_gettsim_tests/test_interface.py
```diff
@@ -46,49 +42,20 @@ def func_before_partial(arg_1, arbeitsl_geld_2_params):
 
 
 @pytest.mark.parametrize(
```
hmgaudecker, line 44: Do we still need the parametrization here? I find it rather hard to see through what is happening; seems like there is a single case left? But see my request for another test above, maybe that'll bring the usefulness back.

src/_gettsim_tests/test_interface.py
```diff
@@ -443,8 +375,8 @@ def arbeitsl_geld_2_m_double(arbeitsl_geld_2_m):
         data,
         {},
         functions=[arbeitsl_geld_2_m_double],
-        targets="arbeitsl_geld_2_m_double_hh",
         aggregation_specs=aggregation_specs,
+        targets="arbeitsl_geld_2_m_double_hh",
```
hmgaudecker, line 379:
```suggestion
        targets="arbeitsl_geld_2_double_m_hh",
```

Unrelated, but let's use the occasion to stick to our own naming conventions.

src/_gettsim_tests/test_interface.py
```diff
@@ -474,8 +406,8 @@ def test_aggregation_specs_missing_group_sufix():
             data,
             {},
             functions=[],
-            targets="arbeitsl_geld_2_m_agg",
             aggregation_specs=aggregation_specs,
+            targets="arbeitsl_geld_2_m_agg",
```
hmgaudecker, line 410:
```suggestion
            targets="arbeitsl_geld_2_agg_m",
```
Same.

hmgaudecker (commented):
Very nice! Some smaller stuff directly in the code. 

Now that we added the warning, the test outputs show a very long list of warnings, which makes it hard to find the actual results. We should catch these warnings throughout in the tests, I think -- maybe except for `test_full_taxes_and_transfers` ?

src/_gettsim/interface.py
```diff
@@ -373,6 +359,57 @@ def _create_input_data(
...
+            *data* you pass to GETTSIM. {'' if n_cols == 1 else '''You need to pick one
+            option for each column that appears in the list above.'''}
+            """
+        )
+        how_to_ignore = format_errors_and_warnings(
+            """
+            If you want to ignore this warning, add the following code to your script
+            before calling GETTSIM:
+
+                import warnings
```
hmgaudecker, line 395:
```suggestion
                import warnings
                from gettsim import FunctionsAndColumnsOverlapWarning
```

hmgaudecker: I guess this is a thing to put into the main namespace as many people will want to use it?

src/_gettsim/interface.py
```diff
@@ -373,6 +359,57 @@ def _create_input_data(
...
+            """
+        )
+        warnings.warn(
+            f"{first_part}\n{formatted}\n{second_part}\n{how_to_ignore}",
+            category=FunctionsAndColumnsOverlapWarning,
+            stacklevel=2,
+        )
+
+
+class FunctionsAndColumnsOverlapWarning(UserWarning):
```
hmgaudecker, line 409: Apologies for being pedantic, but can't we include the string with the warning in the class itself now? I would find that much more natural.

Actually, we might not need the function `_warn_if_functions_and_columns_overlap` at all anymore. That is, the calling code could be (trying the walrus without code checking here, very adventurous!):

```py
if columns_overriding_functions := set(functions_overridden):
    warnings.warn(
        FunctionsAndColumnsOverlapWarning(columns_overriding_functions),
        stacklevel=2,
    )
```

and the warning class something like:


```py
class FunctionsAndColumnsOverlapWarning(UserWarning):
    def __init__(self, columns_overriding_functions):
        super().__init__(self.message(columns_overriding_functions))

    def message(self, columns_overriding_functions):
        return f"(repeat previous): {columns_overriding_functions}"
```

hmgaudecker (commented):
Excellent! Just an import statement + idea for a simplification left.

src/_gettsim/interface.py
```diff
@@ -373,6 +365,51 @@ def _create_input_data(
...
+            that appears in the list above.'''}
+            """
+        )
+        how_to_ignore = format_errors_and_warnings(
+            """
+            If you want to ignore this warning, add the following code to your script
+            before calling GETTSIM:
+
+                import warnings
+                from gettsim import FunctionsAndColumnsOverlapWarning
```
hmgaudecker, line 403:
```suggestion
                from gettsim import FunctionsAndColumnsOverlapWarning

```

src/_gettsim_tests/test_interface.py
```diff
@@ -45,50 +41,40 @@ def func_before_partial(arg_1, arbeitsl_geld_2_params):
...
-    with expectation:
-        _fail_if_columns_overriding_functions_are_not_in_functions(
-            columns_overriding_functions, functions
+def test_dont_warn_if_functions_and_columns_dont_overlap():
+    with warnings.catch_warnings():
+        warnings.filterwarnings("error", category=FunctionsAndColumnsOverlapWarning)
+        compute_taxes_and_transfers(
+            data=pd.DataFrame({"p_id": [0]}),
+            params={},
+            functions={"dupl": lambda x: x},
```
hmgaudecker, line 60:
```suggestion
            functions={"some_func": lambda x: x},
```

hmgaudecker (approved):
Wonderful, many thanks!!!

PR #645, opened 2023-08-23
Vertrauensschutz in Rente für Arbeitslose

TeBackh:
Special age thresholds for people with legitimate expectations during the phase-in of tightened eligibility requirements. Lots of additional improvements that popped up along the way.


src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,109 @@ abolishment_cohort_rente_für_arbeitsl:
...
+  name:
+    de: >-
+      Altersgrenze für Altersrente wegen Arbeitslosigkeit mit  Vertrauensschutz
+      (abschlagsfrei).
+    en: >-
+      Age threshold for pension for unemployed (without deductions)
+      with Vertrauensschutz
+  description:
+    de: >-
+      Frühesmögliches Alter ab dem die Altersrente für Arbeitslose, für die der
```
hmgaudecker, line 989:
```suggestion
      Frühestmögliches Alter ab dem die Altersrente für Arbeitslose, für die der
```

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,109 @@ abolishment_cohort_rente_für_arbeitsl:
   note: https://www.sozialgesetzbuch-sgb.de/sgbvi/237.html
   1980-01-01:
     scalar: 1952
+altergrenze_arbeitsl_abschlagsfrei_vertrauensschutz:
+  name:
+    de: >-
+      Altersgrenze für Altersrente wegen Arbeitslosigkeit mit  Vertrauensschutz
```
hmgaudecker, line 982:
```suggestion
      Altersgrenze für Altersrente wegen Arbeitslosigkeit mit Vertrauensschutz
```

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,109 @@ abolishment_cohort_rente_für_arbeitsl:
...
+      (abschlagsfrei).
+    en: >-
+      Age threshold for pension for unemployed (without deductions)
+      with Vertrauensschutz
+  description:
+    de: >-
+      Frühesmögliches Alter ab dem die Altersrente für Arbeitslose, für die der
+      Vertrauensschutz gilt abschlagsfrei bezogen werden kann.
+    en: >-
+      Earliest possible age to receive pension for unemployed without  deductions
```
hmgaudecker, line 992:
```suggestion
      Earliest possible age to receive pension for unemployed without deductions
```

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -926,9 +926,10 @@ altersgrenze_arbeitsl_vorzeitig:
   description:
     de: >-
       Frühesmögliches Alter ab dem die Altersrente für Arbeitslose
```
hmgaudecker, line 928:
```suggestion
      Frühestmögliches Alter ab dem die Altersrente für Arbeitslose
```

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,109 @@ abolishment_cohort_rente_für_arbeitsl:
...
+      upper_threshold: 1943.9166666
+      rate_linear: 0.0
+    19:
+      upper_threshold: 1944.0
+      rate_linear: 1
+    20:
+      upper_threshold: 1944.08333333
+      rate_linear: 0.0
+    21:
+      upper_threshold: 1944.16666667 # 25  #  jump to 65 (normal unemployed)
```
hmgaudecker, line 1065: I neither get the comments nor the very high rate here.

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,109 @@ abolishment_cohort_rente_für_arbeitsl:
...
+      Frühesmögliches Alter ab dem die Altersrente für Arbeitslose, für die der
+      Vertrauensschutz gilt abschlagsfrei bezogen werden kann.
+    en: >-
+      Earliest possible age to receive pension for unemployed without  deductions
+      under legitimate expectations.
+  reference: § 237 SGB VI
+  note: >-
+    https://www.gesetze-im-internet.de/sgb_6/__237.html
+  type: piecewise_linear
+  1980-01-01:
```
hmgaudecker, line 998: Two comments:
1. Can we use the correct starting date, i.e., when the relevant section of § 237 SGB VI was added?
2. IIUC, this basically implements this table?

   image

   If so, there should be a simpler way of implementing this than piecewise linear, I think. Happy to discuss.

hmgaudecker (commented):
Thanks! I did not get very far because I realised that I don't get the big picture at all. § 237 SGB VI seems to cover more than Vertrauensschutz. Can you be a bit more specific, what exactly is being modelled?

src/_gettsim/parameters/ges_rente.yaml
TeBackh: For people born after Feb 1944 Vertrauensschutz is no longer available, hence the regular thresholds for the deduction free pension for unemployed apply, which is 65 for this cohort.

src/_gettsim/parameters/ges_rente.yaml
TeBackh: changed to 2005.
Yes, lets discuss

src/_gettsim/parameters/ges_rente.yaml
hmgaudecker: Yeah, that's exactly what is not intuitive then. If this is called Vertrauensschutz and only relevant for a couple of cohorts, others should not even show up here.

TeBackh: yep, thats due to the "exhaustiveness requirement" of the function (needs inputs from -inf to inf)-  had this problem before. maybe the function should be changed then.

TeBackh: I changed the parameter input now. it required a littlebit of hardcoding numbers in the function, as the threshold increase lost its regularity shortly before abolishment. hope thats fine, otherwise can change it

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,60 @@ abolishment_cohort_rente_für_arbeitsl:
...
+      2: 60.416666666
+      3: 60.5
+    1943:
+      1: 60.583333333
+      2: 60.666666666
+      3: 60.75
+    1944:
+      1: 60.833333
+      2: 65.0
+cohort_range_arbeitsl_vertrauenss:
```
hmgaudecker, line 1018: This should go, it just repeats the keys from the previous entry. More below.

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -975,6 +976,60 @@ abolishment_cohort_rente_für_arbeitsl:
...
+      Inanspruchnahme nicht betroffen bzw. es erfolgt eine geringere Anhebung.
+      (Z.B. wenn bereits eine Arbeitslosigkeit vorgelegen hat, als das Gesetz
+      beschlossen wurde.)
+    en: >-
+      Earliest possible age to receive pension for unemployed without deductions
+      under legitimate expectations.
+  reference: § 237 SGB VI Abs. 4
+  note: >-
+    https://www.gesetze-im-internet.de/sgb_6/__237.html
+  2005-12-31:
```
hmgaudecker, line 1002: These should become months 1-12 instead of 1, 2, 3

Why is the starting date 2005-01-01 but 2005-12-31 for the next entry? Should be only one, I  suppose?

src/_gettsim/transfers/rente.py
```diff
@@ -606,16 +606,65 @@ def ges_rente_frauen_altersgrenze(
...
+    ges_rente_params
+        See params documentation
+        :ref:`ges_rente_params <ges_rente_params>`.
+
+    Returns
+    -------
+    lowest full retirement age for unemployed.
+
+    """
+    if (
```
hmgaudecker, line 635: This stops at the end of 1996, so I don't think any of that was law at that time?

src/_gettsim/transfers/rente.py
```diff
@@ -606,16 +606,65 @@ def ges_rente_frauen_altersgrenze(
...
+                "altersgrenze_arbeitsl_abschlagsfrei"
+            ]["intercepts_at_lower_thresholds"],
+        )
+    else:
+        out = ges_rente_regelaltersgrenze
+
+    return out
+
+
+@dates_active(start="1997-01-01", change_name="_ges_rente_arbeitsl_altersgrenze")
```
hmgaudecker, line 655: I do not understand the difference of this and the function `_ges_rente_arbeitsl_altersgrenze_vertrauenss` on line 715ff

src/_gettsim/transfers/rente.py
```diff
@@ -625,13 +674,28 @@ def _ges_rente_arbeitsl_altersgrenze(
...
 
     Returns
     -------
     lowest full retirement age for unemployed.
 
     """
-    if birthdate_decimal < ges_rente_params["abolishment_cohort_rente_für_arbeitsl"]:
+    if (
+        int(birthdate_decimal)
+        <= ges_rente_params["cohort_range_arbeitsl_vertrauenss"]["last"]
```
hmgaudecker, line 689: this should become 

```py
if birthyear in ges_rente_params["altersgrenze_arbeitsl_abschlagsfrei_vertrauensschutz"]:
```
or so.

The if-statements below can be replace by indexing into months (use `birthmonth`, too)

src/_gettsim/transfers/rente.py
```diff
@@ -648,6 +712,61 @@ def _ges_rente_arbeitsl_altersgrenze(
...
+
+    Returns
+    -------
+    lowest full retirement age for unemployed with Vertrauensschutz.
+
+    """
+    third = int((geburtsmonat - 1) / 4) + 1
+
+    if (
+        geburtsjahr < ges_rente_params["cohort_range_arbeitsl_vertrauenss"]["last"]
```
hmgaudecker, line 748: See above. Entry should go, simplify by using birth months.

hmgaudecker (commented):
Very cool. I made some things consistent but did not change the code. 

I think this still can be simplified substantially. Some tiny things, but most importantly, the "functions should refer to the state of law at a particular point in time" - idea is not quite implemented all the way yet.

MImmesberger:
When working on the Vertrauensschutz, I noticed that some rules regarding the statutory retirement ages of Arbeitslose are not implemented yet.

Currently, we mostly use piecewise linear functions. Now I have a reform where a piecewise linear function would have more than 40 segments, so it is both cumbersome to write and difficult to understand in the end. Currently I use the implementation below. But this has the downside that we need a new policy function even though only parameters changed. Any advice? I am leaning towards piecewise linear function. 

```text
  1992-01-01:
    reference: Rentenreformgesetz 1992. BGBl. I S. 2261 1989 § 41
    note: Increase of full retirement age from 60 to 65 for birth cohort 1941-1952.
    [.....]
    1951:
      0: 64.083333
      2: 64.166667
      4: 64.25
      6: 64.333333
      8: 64.416667
      10: 64.5
    1952:
      0: 64.583333
      2: 64.666667
      4: 64.75
      6: 64.833333
      8: 64.916667
      10: 65
    1953:
      scalar: 65
  1997-01-01:
    reference: Wachstums- und Beschäftigungsförderungsgesetz 1996. BGBl. I S. 1461 1996
    note: Increase of full retirement age from 60 to 65 for birth cohort 1937-1941.
    0:
      lower_threshold: -inf
      upper_threshold: 1936.916666666
      rate_linear: 0
      intercept_at_lower_threshold: 60
    1:
      upper_threshold: 1941.916666666
      rate_linear: 1
    2:
      upper_threshold: inf
      rate_linear: 0
```

hmgaudecker:
I don't quite follow. Why do we need a new policy function? In general, writing out the parameters is much better IMHO, even if it becomes lengthy. `piecewise_linear` is really made for individual-level stuff like the tax code, not for converting aggregate parameters. Of course it can be used for that purpose, too -- but it typically becomes very hard to read.
 
If the answer to _Why do we need a new policy function?_ is _Because of the second part for the 1937-1941 cohorts_, better convert that to direct parameters, too.

MImmesberger:
> If the answer to Why do we need a new policy function? is Because of the second part for the 1937-1941 cohorts, better convert that to direct parameters, too.

Yes, exactly!

MImmesberger:
In the last commit I rewrote the functions and parameter inputs for the Rente wg. Arbeitslosigkeit and implemented the Vertrauensschutz rules for 1997 and 2006.

When working on it, I noticed that the former implementation was not quite correct because increases in statutory retirement ages after the abolishment of the Rente wg. Arbeitslosigkeit were also applied there. I guess this is an artifact of an older version of GETTSIM.

This is why tests are currently failing. For example, in `test_data/renten_alter/2010/hh_id_769.yaml` a retirement age of 67 for the Rente wg. Arbeitslosigkeit is the expected output. AFAIK, the full retirement age was 65 when the Rente wg. Arbeitslosigkeit was abolished.

Long story short: We probably have to update the test data.

MImmesberger:
ToDo:

- [ ] Fix the wrong retirement age tests
- [ ] Set an end date for functions that use the Vertrauensschutz variable as an input. End date = Year in which there can be no additional retirement entries through this Vertrauensschutz rule.

src/_gettsim/transfers/rente.py
```diff
@@ -606,44 +606,151 @@ def ges_rente_frauen_altersgrenze(
     return out
 
 
-def _ges_rente_arbeitsl_altersgrenze(
-    birthdate_decimal: float,
+@dates_active(end="1996-12-31", change_name="_ges_rente_arbeitsl_altersgrenze")
```
hmgaudecker, line 609:
```suggestion
@dates_active(end="1991-12-31", change_name="_ges_rente_arbeitsl_altersgrenze")
```
Cohorts affected by the law knew from this point onwards what would happen.

Probably means we'll have to introduce another function for 1992-1996?

Same issue below for 2010, or I do not understand the logic here.

MImmesberger: Yes you are right.
That's actually something I was thinking about this morning: With the current decorator, it is not possible to "reintroduce" a policy, i.e. letting it be active in two different period, right? This applies here because we want to test for Vertrauensschutz, e.g. between 1997 and 2009, but before and after the policy is the same.

hmgaudecker: No, that's not possible given the interface of the decorator. And the case is sufficiently rare as to warrant a (necessarily) more cluttered interface.

In this case it is a one-liner, so the following would be making things more complex, don't use. But you could have a common function that is called under the hood (it would not be visible in the DAG). Would make it clear that it is the same regime.

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -973,15 +1268,110 @@ abolishment_cohort_rente_für_arbeitsl:
       Cohort from which pension for unemployed abolished
   reference: § 237 SGB VI
   note: https://www.sozialgesetzbuch-sgb.de/sgbvi/237.html
-  1980-01-01:
+  2012-01-01:
```
hmgaudecker, line 1271: By the logic just discussed elsewhere, this should be 2007 or so.

src/_gettsim/transfers/rente.py
```diff
@@ -895,22 +1161,54 @@ def ges_rente_vorauss_frauen(
...
+    Eligibility as bool.
+
+    """
+
+    out = arbeitsl_1y_past_585 and ges_rente_wartezeit_15 and pflichtbeitr_8_in_10
+
+    return out
+
+
+@dates_active(start="2012-01-01", change_name="_ges_rente_vorauss_arbeitsl")
```
hmgaudecker, line 1201: See elsewhere, reform was in 2007, forward-looking agents should have worked with that?

`dates_active` should have an end date -- 1951 + x where x is the relevant retirement age (65 or whatever)

src/_gettsim_tests/test_data/grundrente_proxy_rente/2021-07-01/hh_id_10.yaml
```diff
@@ -67,4 +71,4 @@ inputs:
   assumed: {}
 outputs:
   rente_vorj_vor_grundr_proxy_m:
-    - 1025.7
+    - 0.0
```
MImmesberger, line 74: I changed this because in the policy environment of 2021 retirement was not possible
with age 65 (because after the requested changes, GETTSIM "forgets" about the Rente wg.
Arbeitslosigkeit after 2017).

MImmesberger:
In case you are confused by the recent changes: Apparently there were a few months in 1996 where the increase in the FRA was accelerated (in July 1996) while the ERA increased as before (in 1992). Then a few months later (September 1996), they abolished the increase in the ERA but kept the acceleration in the FRA from July 1996.

MImmesberger:
I added some tests for the Rente wg. Arbeitslosigkeit. Those tests are made up using the texts of the relevant laws.

I have one problem that I'm not able to solve: Some tests fail because of a pandas error: `ValueError: If using all scalar values, you must pass an index`.

All of the failed tests seem to address functions that just return a parameter from the params dictionary, i.e.

```python
def func(params):
  return params["some_param"]
```

Other than that, the PR is ready to be reviewed.

hmgaudecker:
Thanks!!! 

> I have one problem that I'm not able to solve: Some tests fail because of a pandas error: `ValueError: If using all scalar values, you must pass an index`.
> 
> All of the failed tests seem to address functions that just return a parameter from the params dictionary, i.e.
> 
> ```python
> def func(params):
>   return params["some_param"]
> ```

So the problem is that for whatever reason, this implies that instead of something like this:

```py
{'entgeltp_update': array([31.0625])}
```
the following arrives in line 738 of interface.py:

```py
{'ges_rente_arbeitsl_vorzeitig': array(60.)}
```

I don't quite see why that happens as the outputs of functions look the same (just a scalar). @lars-reimann, would you mind having a quick look?

In any case, we'll just need some workaround and truly fix it when looking seriously at the interface.

lars-reimann:
Hmm, the tests fail for functions that only depend on `ges_rente_params`:

```py
@dates_active(end="1991-12-31", change_name="ges_rente_arbeitsl_vorzeitig")
def _ges_rente_arbeitsl_vorzeitig_ohne_staffelung(
    ges_rente_params: dict
):
   ...
```

Adding any other dependency resolves the issue:

```py
@dates_active(end="1991-12-31", change_name="ges_rente_arbeitsl_vorzeitig")
def _ges_rente_arbeitsl_vorzeitig_ohne_staffelung(
        ges_rente_params: dict,
        geburtsjahr: int,
):
   ...
```

I didn't figure out why that happens yet, though.

hmgaudecker:
Ahh, sure, the params are partialed in beforehand, so something weird must be happening when doing the calculations via the DAG.

Just add a fake dependency, corresponding `# noqa: XXX` statements, and an issue so as not to forget about it when re-doing the interface.

MImmesberger:
Ah, thanks @lars-reimann @hmgaudecker. It worked!

src/_gettsim/transfers/rente.py
hmgaudecker: What you did works perfectly well and is even better than what I had in mind!

hmgaudecker:
> I added some tests for the Rente wg. Arbeitslosigkeit.

Cool. Only thing is that for new tests, can you please call the file for what is being tested instead of `test_hh_id_[x].yaml`? That was a leftover of the bulk conversion of csv-formatted tests to individuals / households.

E.g., I am just looking at the 2007 change. So there could be tests (making up the year here):
- `test_renten_alter` / `2015` / `cohort_1951_wegen_arbeitsl.yaml`
- `test_renten_alter` / `2015` / `cohort_1952.yaml`

src/_gettsim/transfers/rente.py
```diff
@@ -491,19 +500,68 @@ def _ges_rente_altersgrenze_abschlagsfrei_mit_besond_langj(  # noqa: PLR0913
...
+
+    """
+
+    out = float("inf")
+    if ges_rente_vorauss_regelrente:
+        out = ges_rente_regelaltersgrenze
+    if ges_rente_vorauss_langj:
+        out = min([out, _ges_rente_langj_altersgrenze])
+    if ges_rente_vorauss_besond_langj:
+        out = min([out, _ges_rente_besond_langj_altersgrenze])
```
hmgaudecker, line 545: Just my usual check whether this is being tested for sure? codecov flags it, but that might well be false.

src/_gettsim/transfers/rente.py
```diff
@@ -625,48 +711,183 @@ def ges_rente_frauen_altersgrenze(
...
+def _ges_rente_arbeitsl_altersgrenze_ohne_staffelung(
+    ges_rente_params: dict,
+    geburtsjahr: int,  # noqa: ARG001
+) -> float:
+    """Full retirement age for unemployed.
+
+    Before the WFG (Gesetz für Wachstum und Beschäftigung) was implemented in 1997 the
+    full retirement age was the same for every birth cohort.
+
+    ToDo: Remove fake dependency (geburtsjahr).
```
hmgaudecker, line 774:
```suggestion
    # TODO(@MImmesberger): Remove fake dependency (geburtsjahr).
    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/666
```

Please move this and others outside of docstrings, then it will be flagged by ruff's [flake-8-todo-td](https://docs.astral.sh/ruff/rules/#flake8-todos-td) codes, too, if if does not have this form.

I had converted all to-do comments a while ago. I only realised now that  this did not catch those inside docstrings. Converting all those (some can just be deleted) would be a useful task in the near future. Some are horribly outdated, that is just a distraction.

src/_gettsim/transfers/rente.py
```diff
@@ -829,15 +1042,261 @@ def _ges_rente_altersgrenze_vorzeitig(  # noqa: PLR0913
...
+    """
+
+    return ges_rente_params["altersgrenze_arbeitsl_vorzeitig"]
+
+
+@dates_active(
+    start="2004-07-26", end="2017-12-31", change_name="ges_rente_arbeitsl_vorzeitig"
+)
+def ges_rente_arbeitsl_vorzeitig_mit_vertrauenss_ab_2006(
+    ges_rente_params: dict,
```
hmgaudecker, line 1250: Please make sure that params always appear last in the list of inputs.

src/_gettsim/transfers/rente.py
```diff
@@ -829,15 +1042,261 @@ def _ges_rente_altersgrenze_vorzeitig(  # noqa: PLR0913
...
+    vertra_arbeitsl_2006: bool,
+    _ges_rente_arbeitsl_vorzeitig_ohne_vertrauenss: float,
+) -> float:
+    """Early retirement age of pension for unemployed.
+
+    Includes Vertrauensschutz rules implemented in 2006. Policy becomes inactive in 2018
+    because then all potential beneficiaries have reached the normal retirement age.
+
+    Parameters
```
hmgaudecker, line 1259: Docstring paramets do not match parameters.

Feel free to ignore for the moment. But we should really enable the "D" codes soon in `pyproject.toml` (maybe excluding some sub-codes), then it will be flagged automatically.

src/_gettsim/transfers/rente.py
```diff
@@ -881,15 +1362,20 @@ def ges_rente_vorauss_regelrente(ges_rente_wartezeit_5: bool) -> bool:
...
     y_pflichtbeitr_ab_40: float,
     geburtsjahr: int,
     ges_rente_params: dict,
 ) -> bool:
-    """Determining the eligibility for Altersrente für Frauen (pension for
-    women) Wartezeit 15 years, contributions 10 years after age 40, being a women.
+    """Eligibility for Altersrente für Frauen (pension for
+    women).
+
+    Wartezeit 15 years, contributions for 10 years after age 40, being a women. Policy
```
hmgaudecker, line 1376:
```suggestion
    Wartezeit 15 years, contributions for 10 years after age 40, being a woman. Policy
```

src/_gettsim/transfers/rente.py
```diff
@@ -914,22 +1400,62 @@ def ges_rente_vorauss_frauen(
...
+@dates_active(end="2006-12-31", change_name="_ges_rente_vorauss_arbeitsl")
+def _ges_rente_vorauss_arbeitsl_ohne_2007_reform(
+    arbeitsl_1y_past_585: bool,
+    ges_rente_wartezeit_15: bool,
+    pflichtbeitr_8_in_10: bool,
+) -> bool:
+    """Eligibility for Altersrente für Arbeitslose (pension
+    for unemployed).
+
+    Wartezeit 15 years, 8 contribution years past 10 years, being unemployed for at
```
hmgaudecker, line 1418: Just wondering: Is there really no requirement of being *currently* unemployed? So I could be unemployed at age 59 for a year and claim at age 63 (or whatever) despite being currently employed?

If there is such a requirement, we might always think of the current implementation as an approximation (can always quit / get myself fired at any time), but then we should document that better.

src/_gettsim/transfers/rente.py
```diff
@@ -914,22 +1400,62 @@ def ges_rente_vorauss_frauen(
...
     arbeitsl_1y_past_585: bool,
     ges_rente_wartezeit_15: bool,
     pflichtbeitr_8_in_10: bool,
     birthdate_decimal: float,
     ges_rente_params: dict,
 ) -> bool:
-    """Determining the eligibility for Altersrente für Arbeitslose (pension
-    for unemployed. Wartezeit 15 years, 8 contributionyears past 10 years, being
-    at least 1 year unemployed after age 58 and 6 months.
+    """Eligibility for Altersrente für Arbeitslose (pension
```
hmgaudecker, line 1452:
```suggestion
    """Eligibility for Altersrente für Arbeitslose (pension for unemployed).
```

src/_gettsim/transfers/rente.py
```diff
@@ -914,22 +1400,62 @@ def ges_rente_vorauss_frauen(
...
+    for unemployed).
```
hmgaudecker, line 1453:
```suggestion
```

src/_gettsim/transfers/rente.py
```diff
@@ -914,22 +1400,62 @@ def ges_rente_vorauss_frauen(
...
+
+    Wartezeit 15 years, 8 contributionyears past 10 years, being
```
hmgaudecker, line 1455: Can you watch the line breaks a bit better, please? Alt + Q does it for me in VS Code, happy to look into the settings.

hmgaudecker (approved):
Excellent, thanks a lot! Mostly very general stylistic changes left, no show-stoppers.

src/_gettsim/transfers/rente.py
MImmesberger: According to law, the function should look something like this:

```python
def _ges_rente_vorauss_arbeitsl(
    arbeitsl_1y_past_585: bool,
    ges_rente_wartezeit_15: bool,
    pflichtbeitr_8_in_10: bool,
    arbeitssuchend: bool,
    m_altersteilzeitarb_52: bool
) -> bool:

    return (arbeitssuchend and arbeitsl_1y_past_585 and ges_rente_wartezeit_15 and arbeitsl_1y_past_585) or (m_altersteilzeitarb_52 and arbeitsl_1y_past_585 and ges_rente_wartezeit_15 and arbeitsl_1y_past_585)

def m_altersteilzeitarb_52(
    m_altersteilzeitarb: bool
) -> bool:
     return m_altersteilzeitarb>=52
```
(of course with numbers in params files, etc)

This would require a new input variable: m_altersteilzeitarb (Months of Altersteilzeitarbeit).

But I agree, that `arbeitssuchend` is something that the potential retiree can easily influence. A quick search regarding voluntary unemployment didn't yield any restrictions. (But, of course, this doesn't guarantee that there aren't any.)
I added a few sentences in the docstring:
```text
The person is also required to be unemployed at the time of claiming the pension. As there are no restrictions regarding voluntary unemployment this requirement may be viewed as always satisfied and is therefore not included when checking for eligibility.
``

File: src/_gettsim/parameters/ges_rente.yaml [hunk 26]
```diff
@@ -937,32 +1365,118 @@ altersgrenze_arbeitsl_vorzeitig:
...
+      4: 62.333333
+      5: 62.416667
+      6: 62.5
+      7: 62.583333
+      8: 62.666667
+      9: 62.75
+      10: 62.833333
+      11: 62.916667
+      12: 63.0
+first_birthyear_without_rente_für_arbeitsl:
```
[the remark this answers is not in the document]
MImmesberger: I agree. 

We should do this with the Vertrauensschutz rules as well. That way, we don't carry those in our params dict forever. I think I will edit those now (including `first_birthyear_without_rente_für_arbeitsl `)and then we can change the Altersrente für Frauen using this PR as a template.

src/_gettsim/transfers/rente.py
MImmesberger: Ah sorry, I thought that codecov check passed => Test coverage of newly added / changed lines is 100%. I will have that in mind in the future.

I added some tests that cover these lines but only for one birth cohort (where the tests are particularly easy). I think it makes sense to properly test the pathways into retirement once we are rewriting them (in particular, the Rente für Frauen).

MImmesberger: Do we want to test for `out = float("Nan")` in `referenz_alter_abschlag`?

I think not because i) coding "no eligibility for early retirement" with NaNs as the early retirement age should change in the future and ii) as far as I understood, these tests check against the numbers implied by the law (and not such technical details).

If we want to test for it, I don't know how to include `float("NaN")` as expected output in the yaml file.

MImmesberger:
Thank you for the comments! Will keep those in mind for future PRs. I implemented the changes or replied to your concerns.

hmgaudecker (approved):
Excellent, thanks a lot! I did some renamings directly, please don't use underscores excessively.

PR #649, opened 2023-09-07
allow for up to 10 children in create_synthetic_data.

ChristianZimpelmann:
allow for up to 10 children in create_synthetic_data.

### DIFF
--- a/src/_gettsim/synthetic.py
+++ b/src/_gettsim/synthetic.py
@@ -43,12 +43,12 @@ def create_synthetic_data(
     # Check inputs
     if n_adults not in [1, 2]:
         raise ValueError("'n_adults' must be either 1 or 2")
-    if n_children not in [0, 1, 2]:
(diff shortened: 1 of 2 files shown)


src/_gettsim/synthetic.py
```diff
@@ -172,7 +172,12 @@ def create_constant_across_households_variables(df, n_adults, n_children, policy
         datetime.date(policy_year, 1, 1),
         RESOURCE_DIR / "synthetic_data" / "bedarfsgemeinschaften",
     )
-    hh_typ_string = create_hh_typ_string(n_adults, n_children)
+
+    # Use data for 2 children if there are more than 2 children in the household.
+    hh_typ_string = (
```
MImmesberger, line 177: Is there a reason why you don't use the `create_hh_typ_string` function here?

ChristianZimpelmann: Good point! I adjusted it in the last commit.

MImmesberger (approved):

PR #654, opened 2023-09-14
refactor: rename `hh_id` to `vg_id`

lars-reimann:

As discussed in https://github.com/iza-institute-of-labor-economics/gettsim/pull/601#issuecomment-1677783059, this PR 
* renames `hh_id` to `vg_id`,
* replaces the aggregation indicator `_hh` with `_vg`.

Since this PR touches thousands of files, I've split this from #601.

### Todo


hmgaudecker:
Thanks! Will need to alert some people before merging, but looks great at a glance! 

Small things that I saw right away:
- Please search for `hh` in the entire codebase -- mostly GEPs / docs / tutorials, where it still shows up
- This will turn up `haushaltsgröße_hhn` in `wohngeld.yaml`. That key go completely, it is calculated as `max_definierte_vg_größe`. 
- We should document somewhere the origin of `vg`. Probably GEP-01 ?

lars-reimann:
> We should document somewhere the origin of vg. Probably GEP-01 ?

What should be written there?

hmgaudecker:
> > We should document somewhere the origin of vg. Probably GEP-01 ?
> 
> What should be written there?

Now that you are saying it, it is actually written already :see_no_evil:, right [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/601/commits/38272c1415a0a094a9f70bcf9dc2a1c5630d3f62#diff-00413ff99705e46e6efc409af72baf15c3e34f7f72af01792d857a76475eaf1c)... Apologies.

Reading through that makes me wonder whether this PR is actually helpful or not in its current form. That is, current `hh` will become either `vg`, or `fg`, or `bg`. 

It might be more useful to use the correct naming already, leave all three of {`vg`, `fg`, `bg`} the same (=same behaviour as current `hh`) and then introduce the distinctions made in #601 ?

If that makes sense, it would be
- `vg` in `wohngeld`
- `fg` in `elterngeld`
- `bg` in `arbeitsl_geld_2`, `kinderzuschl`

Does that make sense, @mjbloemer @ChristianZimpelmann @JuergenWiemers ?

ChristianZimpelmann:
> Does that make sense,

I would say so!

ChristianZimpelmann:
> Does that make sense,

I would say so!

Not really sure about `grundrente`. Probably `fg`? It could be the case even that we need another grouping there as I am not sure whether there is ever more than one generation considered.

A tricky part will be the Günstigerprüfung between Wohngeld and ALG II if those are referring to different groupings. But this isn't relevant in this PR.

lars-reimann:
How should the following functions be renamed?

* `grunds_im_alter_m_hh` in `grunds_im_alter.py`
* `grunds_im_alter_vermög_freib_hh` in `grunds_im_alter.py`
* `wohngeld_kinderzuschl_vorrang_hh` in `benefit_checks.py`

hmgaudecker:
- `grunds_im_alter_m_vg` in `grunds_im_alter.py`
- `grunds_im_alter_vermög_freib_vg` in `grunds_im_alter.py`
- `wohngeld_kinderzuschl_vorrang_vg` in `benefit_checks.py`

The last might actually pose a problem because wohngeld operates on different grouping than `arbeitsl_geld_2`, `kinderzuschl`, which we won't be able to solve before #601 and maybe a follow-up. 

Just so you are not surprised.

lars-reimann:
Closing since `main` has already diverged too far from this.

PR #659, opened 2023-10-13
Rounding of Taxes and Transfers

### LINKED ISSUE #88: Rounding is missing for several taxes / transfers

MImmesberger:
Closes #88 

### Todo

- [ ] Wohngeld in steps of 5€.
- [x] Income deduction from Kinderzuschlag in steps of 10€ until 07/2019.
- [x] Vorsorgeaufwendungen are rounded up.
- [x] Income tax is rounded down.
- [x] Unterhaltsvorschuss is rounded up.


MImmesberger:
The PR is ready for review.
1. I couldn't find the legal basis for Wohngeld being payed out in 5€ steps, so I did not implement it. See discussion in #88 
2. The rounding of zu versteuerndes Einkommen includes an additive part (e.g. round down to a number divisible by 54 and then add 18€). A summary is [here](https://www.bmf-steuerrechner.de/ekst/rundungsvorschrift.xhtml#:~:text=Für%20die%20Einkommensteuerformel%202002%2F2003,um%2018%20€%20zu%20erhöhen.). Because this wasn't possible with the current machinery, I added an "add" key to the rounding specification that adds a float after the rounding step.

hmgaudecker:
> I couldn't find the legal basis for Wohngeld being payed out in 5€ steps, so I did not implement it. See discussion in

Yes, let's leave it at the nearest Euro!

src/_gettsim/interface.py
```diff
@@ -604,22 +604,23 @@ def _add_rounding_to_functions(functions, params):
...
                 base=rounding_spec["base"],
                 direction=rounding_spec["direction"],
+                add=rounding_spec["add"] if "add" in rounding_spec else None,
             )(func)
 
     return functions_new
 
 
-def _add_rounding_to_one_function(base, direction):
+def _add_rounding_to_one_function(base, direction, add):
```
hmgaudecker, line 613: Do we have tests for this or is codecov's complaint a fluke?

And please add typing, something like

```python
base: float,
direction: Literal['up', 'down', 'nearest'']
add: float
```

Not sure about what you mean by "optional"? Surely, the function expects to be called with `add` supplied. Do you mean that the typing should be `add: float|None` ?

However, given all the checks below, I'd think it'll be easier to just specify a 0 everywhere.

And we should be a bit more expressive in the name, maybe `to_add_after_rounding` ?

src/_gettsim/parameters/eink_st.yaml
```diff
@@ -276,3 +276,21 @@ rounding:
       base: 1
       direction: down
       reference: § 32a Abs. 1 S. 6 EStG
+  zu_verst_eink_y_tu:
```
hmgaudecker, line 279: Can you add an explanation of what the "add" does here? Seems a bit weird to me. Also, annual income was only recorded in steps of 54 Euros? Not 100 DM / 1.95583 = 51.13 EUR ?

src/_gettsim/parameters/unterhaltsvors.yaml
```diff
@@ -49,3 +49,11 @@ altersgrenzen:
     1: 6
     2: 12
     3: 18
+rounding:
+  unterhaltsvors_m:
+    note:
+      en: Starting date unclear
```
hmgaudecker, line 55: Why?

src/_gettsim_tests/test_rounding.py
```diff
@@ -57,10 +63,15 @@ def test_func():
 
 
 @pytest.mark.parametrize(
-    "base, direction",
-    [(1, "upper"), ("0.1", "down"), (5, "closest")],
+    "base, direction, add",
+    [
+        (1, "upper", None),
```
hmgaudecker, line 68:
```suggestion
        (1, "upper", 0),
```
See above, I don't think this should work.

src/_gettsim_tests/test_rounding.py
```diff
@@ -131,6 +144,40 @@ def test_func(income):
     np.array_equal(calc_result["test_func"].values, np.array(input_values_exp_output))
 
 
+@pytest.mark.parametrize(
+    "base, direction, add, input_values, exp_output",
+    rounding_specs_and_exp_results_with_additive_part,
+)
+def test_rounding_with_additive_part(base, direction, add, input_values, exp_output):
```
hmgaudecker, line 151: I don't think we need a separate test from `test_rounding` anymore, unless we leave the code as is.

hmgaudecker (commented):
Very sweet! I think I would get rid of the case distinction for the `add` parameter (and rename it), just requiring that it is zero if not used.

src/_gettsim/interface.py
MImmesberger: > Do we have tests for this or is codecov's complaint a fluke?

If you are referring to the codecov report from above: this was a mistake on the codecov side. If not, I don't know what you are referring to.

> However, given all the checks below, I'd think it'll be easier to just specify a 0 everywhere.

I agree, I changed the default value from `None` to `0`:

```python
to_add_after_rounding=rounding_spec["to_add_after_rounding"]
                if "to_add_after_rounding" in rounding_spec
                else 0,
            )(func)
``

src/_gettsim/parameters/eink_st.yaml
MImmesberger: Ah that was a typo, should have been 36, not 54. I didn't convert currencies, but just took the rules from the EStG/ the Website of the ministry of finance:

> Für die Einkommensteuerformel 2002/2003 ist zur Ermittlung des abgerundeten zu versteuernden Einkommens (zvE) das zu versteuernde Einkommen auf den nächsten, durch 36 € ohne Rest teilbaren €-Betrag abzurunden, wenn es nicht bereits durch 36 ohne Rest teilbar ist und um 18 € zu erhöhen.

src/_gettsim/parameters/unterhaltsvors.yaml
MImmesberger: Looked into it and realized that the rounding rules were already there when Unterhaltsvorschuss was introduced in 1980. Updated the file accordingly.

src/_gettsim_tests/test_rounding.py
MImmesberger: I removed the `test_rounding_with_additive_part` and included the test in the existing machinery. I wanted to explicitly test the case where `to_add_after_rounding` is not specified in the dict, so I edited the part where this is relevant in the original test.

MImmesberger:
Thank you for the comments! I hope I addressed all your concerns.

MImmesberger:
Sorry, I still have trouble interpreting codecov outputs:

From the output I figured that `_add_rounding_to_one_function` is not properly covered by a test (only implicitly because it is used in the wrapper `_add_rounding_to_functions` which is tested). But testing directly via `test_rounding_callable` doesn't help.

Maybe @lars-reimann knows whats going on?

hmgaudecker:
Oh, codecov happens to be off every now and then. If you are sure it is being run, that's fine!

src/_gettsim/parameters/eink_st.yaml
hmgaudecker: Makes sense for Euro-times. But the 54 before that?

hmgaudecker (approved):
Looks great, just one tiny comment left.

src/_gettsim/parameters/eink_st.yaml
MImmesberger: This is a rule in the EStG as well:

> Für Einkommensteuerformeln 1981 bis 2001 ist zur Ermittlung des abgerundeten zu versteuernden Einkommens (zvE) das zu versteuernde Einkommen auf den nächsten, durch 54 ohne Rest teilbaren DM-Betrag abzurunden, wenn es nicht bereits durch 54 ohne Rest teilbar ist. In 2001 ist dieser ermittelte Wert um 27 DM zu erhöhen.

hmgaudecker: Oh, I am afraid you will have to convert that, then. We were thinking to have a possibility to specify unit=DM, but IIRC we never got around to that, yet?

MImmesberger: Ah, I thought we calculate everything in DM before 2002. 

I converted the DM values to EUR.

hmgaudecker: That would be the goal eventually, but so far, we did not get around to it!

PR #660, opened 2023-10-15
Entgeltpunkte Ost/West

nafetsk:

Distinction between East and West earning points in the calculation of the pension. Differentiation possible via a new input variable called "anteil_entgeltp_ost", which is equivalent to the OPXAZ variable of the SOEP-RV.

### Todo


hmgaudecker:
We might have discussed that before, but it would seem more natural to me to have separate variables `entgeltpunkte_ost` and `entgeltpuntkte_west` rather than the total and fractions. Any specific reasons why you picked the current version?

Also, how do you plan to implement the accumulation of points? Based on place of residence?

Would be good to make a detailed list in the PR first on what you plan to do (or discuss on Zulip), so one can have an early discussion. Some things one will only see when in code, but many things can be cleared up ex ante.

nafetsk:
> We might have discussed that before, but it would seem more natural to me to have separate variables `entgeltpunkte_ost` and `entgeltpuntkte_west` rather than the total and fractions. Any specific reasons why you picked the current version?
> 
> Also, how do you plan to implement the accumulation of points? Based on place of residence?
> 
> Would be good to make a detailed list in the PR first on what you plan to do (or discuss on Zulip), so one can have an early discussion. Some things one will only see when in code, but many things can be cleared up ex ante.

We thought a share variable would make sense, as there is one in SOEP-RV (OPXAZ). 

In the first draft, I had implemented a function that adjusts the share variable depending on the place of residence and the EGPT earned. 
However, we then took it out again because we thought that the accumulation of EGPT over time was hardly implemented and that this should be a separate PR if at all.

And it's true that we should announce PRs earlier and put them up for discussion

hmgaudecker:
Thanks for the explanation! 

The SOEP is irrelevant to the extent that it should not drive implementation details. It's trivial to calculate two levels from total + share outside of GETTSIM. 

>  However, we then took it out again because we thought that the accumulation of EGPT over time was hardly implemented and that this should be a separate PR if at all.

Not sure why you are saying that? There is `entgeltp_update`, which of course is missing lots of things, but the latter is irrelevant to the question at hand.

amageh:
Hi, thanks for the feedback! I agree that using stocks vs. shares east shouldn't make much of a difference. We went for shares since this is what is most widely available in the datasets we used so far and it's the most straightforward way to add EGPT from Zurechnungszeit. Thinking about an updating functionality though, using stocks is probably more simple.

Regarding the updating function: I argued for not adding it for now to keep this change is minimal as possible, since I though more "dynamic" functionalities of GETTSIM will be part of larger future PRs. I am completely fine with adding it again though, if this makes more sense!

hmgaudecker:
But the function is there? With two stocks, it will be obvious it'll need updating.

nafetsk:
Hey, got rid of the share variabel and introduced `entgeltp_west` and `entgeltp_ost` as new input variables. In order to update them I replaced `entgeltp_update` with `entgeltp_ost_update` and `entgeltp_west_update`.

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +538,19 @@ Type: int
 Number of own children below the age of 25 (incl. not in hh)
 
 Type: int
+
```
hmgaudecker, line 541: Please leave above `grundr_zeiten`, no reason to move around.

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +538,19 @@ Type: int
...
+(entgeltp_west)=
+
+## `entgeltp_west`
+
+West earnings points for pension claim
```
hmgaudecker, line 546:
```suggestion
Earnings points for pension claim accumulated in western states
```

docs/gettsim_objects/input_variables.md
```diff
@@ -546,3 +538,19 @@ Type: int
...
+
+Type: float
+
+(entgeltp_ost)=
+
+## `entgeltp_ost`
+
+East earnings points for pension claim
```
hmgaudecker, line 554:
```suggestion
Earnings points for pension claim accumulated in eastern states
```

src/_gettsim/config.py
```diff
@@ -131,7 +131,6 @@ def set_array_backend(backend: str):
     "geburtstag": int,
     "geburtsmonat": int,
     "mietstufe": int,
-    "entgeltp": float,
     "kind": bool,
```
hmgaudecker, line 134: Again, leave up here. No reason to move around.

src/_gettsim/transfers/rente.py
```diff
@@ -23,8 +23,9 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
 @add_rounding_spec(params_key="ges_rente")
 def ges_rente_vor_grundr_m(
     ges_rente_zugangsfaktor: float,
-    entgeltp_update: float,
-    rentenwert: float,
+    entgeltp_ost_update: float,
+    entgeltp_west_update: float,
+    ges_rente_params: dict,
```
hmgaudecker, line 28: params should always be the last argument(s), put after `rentner`

src/_gettsim/transfers/rente.py
```diff
@@ -43,10 +44,12 @@ def ges_rente_vor_grundr_m(
...
         See :func:`ges_rente_zugangsfaktor`.
-    entgeltp_update
-        See :func:`entgeltp_update`.
-    rentenwert
-        See :func:`rentenwert`.
+    entgeltp_ost_update
+        See :func:`entgeltp_ost_update`.
+    entgeltp_west_update
+        See :func:`entgeltp_west_update`.
+    ges_rente_params
```
hmgaudecker, line 51: Same.

src/_gettsim/transfers/rente.py
```diff
@@ -160,7 +170,43 @@ def entgeltp_update(entgeltp: float, entgeltp_update_lohn: float) -> float:
...
+    entgeltp_ost
+        See basic input variable :ref:`entgeltp_ost <entgeltp_ost>`.
+    entgeltp_update_lohn
+        See :func:`entgeltp_update_lohn`.
+
+    Returns
+    -------
+
+    """
+    # Note: We might need some interaction between the two
```
hmgaudecker, line 203: I don't follow this comment. 

In general it is better to open an issue than to leave comments in the code; the latter just gets lost.

src/_gettsim/transfers/grundrente.py
```diff
@@ -302,13 +302,14 @@ def grundr_zuschlag_bonus_entgeltp(
 @add_rounding_spec(params_key="ges_rente")
 def rente_vorj_vor_grundr_proxy_m(  # noqa: PLR0913
     rentner: bool,
-    rentenwert_vorjahr: float,
```
hmgaudecker, line 305: Pls remove from docstring, too.

src/_gettsim/transfers/rente.py
hmgaudecker, line 1: Remove the function `rentenwert_vorjahr` entirely.

hmgaudecker (commented):
Very nice! Just stylistic issues / trivialities left. 

One more relevant thing: As trivial as they may be, it would be nice to add a couple of tests on the two entgeltp_* - functions directly.

src/_gettsim/transfers/rente.py
nafetsk: Me neither to be honest. It was already there in the original version of `entgeltp_update`. According to git blame it was written by @MaxBlesch

nafetsk:
> Very nice! Just stylistic issues / trivialities left.
> 
> One more relevant thing: As trivial as they may be, it would be nice to add a couple of tests on the two entgeltp_* - functions directly.

The testcases in `test_renten_anspr` are checking both update functions. I could add another case that checks for a mixed earning point biography. Is that enough or should I rather create a new test for that?

src/_gettsim/transfers/rente.py
hmgaudecker: Ah, I guess that's what I meant with "It just gets lost". :laughing:

Just delete then, @MaxBlesch can always open an issue in case it was some deeper insight :wink:

hmgaudecker:
> The testcases in `test_renten_anspr` are checking both update functions. I could add another case that checks for a mixed earning point biography. Is that enough or should I rather create a new test for that?

Ah, I did not realise that. As long as both east and west are checked, all is fine.

src/_gettsim/transfers/grundrente.py
```diff
@@ -327,8 +326,10 @@ def rente_vorj_vor_grundr_proxy_m(  # noqa: PLR0913
...
     alter
         See basic input variable :ref:`alter <alter>`.
-    entgeltp
-        See basic input variable :ref:`entgeltp <entgeltp>`.
+    entgeltp_west
+        See basic input variable :ref:`entgeltp_west <entgeltp_west>`.
+    entgeltp_ost
+        See basic input variable :ref:`entgeltp_ost <entgeltp_ost>`.
     ges_rente_zugangsfaktor
         See :func:`ges_rente_zugangsfaktor`.
```
hmgaudecker, line 335: ges_rente_params

nafetsk:
> > The testcases in `test_renten_anspr` are checking both update functions. I could add another case that checks for a mixed earning point biography. Is that enough or should I rather create a new test for that?
> 
> Ah, I did not realise that. As long as both east and west are checked, all is fine.

I added another test case just to make sure it works for a mixed (east/west) biography. Tests should be sufficient now.

hmgaudecker (approved):
Excellent, thanks a lot!

PR #675, opened 2023-12-06
Switch from vg to hh.

### LINKED ISSUE #674: BUG: Change grouping from `vg_id` back to `hh_id`
### Bug description

In #662 we changed the grouping unit for Wohngeld from `hh_id` to `vg_id`. It turned out that this is not correct because having the same `vg_id` (being in one Verantwortungs- und Einstehensgemeinschaft) is a subset of the relevant grouping criteria for Wohngeld. The correct grouping is `hh_id`.

MImmesberger:

Closes #674 

### Todo

### DIFF
--- a/src/_gettsim/groupings.py
+++ b/src/_gettsim/groupings.py
@@ -76,7 +76,7 @@ def fg_id_numpy(  # noqa: PLR0913
 
         p_id_to_fg_id[current_p_id] = next_fg_id
 
-        current_vg_id = hh_id[index]
+        current_hh_id = hh_id[index]
         current_p_id_einstandspartner = p_id_einstandspartner[index]
         current_p_id_children = p_id_to_p_ids_children.get(current_p_id, [])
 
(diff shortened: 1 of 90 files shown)


src/_gettsim/groupings.py
```diff
@@ -76,7 +76,7 @@ def fg_id_numpy(  # noqa: PLR0913
 
         p_id_to_fg_id[current_p_id] = next_fg_id
 
-        current_vg_id = hh_id[index]
+        current_hh_id = hh_id[index]
```
MImmesberger, line 79: I replaced `vg` with `hh` in the groupings function as well. That's correct, right?

lars-reimann (approved):
✔️ All occurrences of `vg` are replaced.
✔️ Changelog is updated.

This looks great to me.

src/_gettsim/groupings.py
hmgaudecker: Correct. VG should go completely, including any reference to it in the docs etc.

hmgaudecker (approved):
Excellent work, thanks!

PR #676, opened 2023-12-09
Linking children's transfers to their parents.

MImmesberger:
This PR is supposed to link transfers on the children level directly to their parents based on `p_id_elternteil_1` and `p_id_elternteil_2`.

This affects the calculation of Kinderfreibetrag, Kindergeld, Erziehungsgeld and Elterngeld.

Additionally, take-up issues are relevant regarding the Kindergeld, Erziehungsgeld and Elterngeld (does `p_id_elternteil_1` or `p_id_elternteil_2` take up the transfer?).


MImmesberger:
I thought about the implementation for some time and realised that having parents-children links would require larger changes to the grouping and aggregation code (at least `_create_aggregation_functions` and `_create_derived_functions`) as groups would be overlapping (because children belong to an "elternteil_1" and "elternteil_2" group and because children may have children themselves). 

An alternative would be to handle `p_id`-specific operations across observations separately from the currently existing aggregation steps.

Have we already discussed how we want to implement his? @hmgaudecker @lars-reimann

hmgaudecker:
Good points. My thinking has been somewhat sloppy indeed.

- We may need a differentiation between variables 
  - Who is the reason for a claim (i.e., the kid)
  - Who actually makes the claim / to whom it is paid out (Kindergeld) / who benefits from it (Kinderfreibetrag) (typically one or both parents, but for kids aged 18-24, it could actually be the kid itself, I believe) 
- We probably need three steps
  1. Determining whether someone is reason for a claim (rules in [BKGG §1](https://www.buzer.de/gesetz/1166/a16512.htm), [§2](https://www.buzer.de/gesetz/1166/a16513.htm), store an indicator with the kid
  2. Distribute Kindergeld across parents (not sure about institutions here, i.e., whether it makes sense to set half a claim for both and have a separate variable for payout, which is always to one person only; or to go with one payout only)
  3. Distribute Kinderfreibeträge across parents (typically one each, but I think this can be overridden by choice)
- For 2. and 3., the initial variables will be stored with the kid, then summed up by the each parent's `p_id`. Just need to make sure this is merged by p_id again, so the resulting vector has the correct length and order.

MImmesberger:
How we implemented the required changes depends on how strong we want to adhere to the current `p_id_elternteil_x` design. We need to the discuss how the user inputs look like.

Taking Kindergeld as an example we could do it this way:

1. GETTSIM calculates the transfer the person in its own right is eligible for (on the individual level). We already do this for Kindergeld and Erziehungsgeld I think. The target could be something like `kindergeld_anspruch` (float). We do this for every person in the dataset (so also for rows that are not eligible for Kindergeld).
2. We use user-specified pointers to the `p_id` of the one who actually claims (and receives) the benefit. The basic input column could be here `p_id_kindergeld_auszahlung`. This is specified on the child level. We can also add defaults and checks whether the user input is correct (i.e. the person that claims the benefit can actually do so).
3. The benefit is aggregated on the basis of `p_id_kindergeld_auszahlung` and summarized in a variable `kindergeld_m`. 

An alternative would be the following:

We utilise the existing `p_id_elternteil_x` columns. The procedure is the same as above, but in step 2, the user specifies a variable `kindergeld_elternteil_auszahlung` that determines whether parent `1` or `2` (or the child) receives the transfer. The downside is that we cannot represent transfers that should be payed out to someone else (e.g. think of grandparents receiving Kindergeld, but also Unterhalt from the parents).

hmgaudecker:
I am afraid I am unable to fully follow. The basic structure seems the same, only whether we add an explicit pointer for  `p_id_kindergeldempfänger` (that seems to be the "right" name to me)? 

If that is the case, I think we should go for the explicit one.

`kinderfreibetrag` is likely the more intricate case. Typically each parent receives one, but there are cases where one parent gets both. So we might need something like `p_id_kinderfreibetragempfänger_1` and  `p_id_kinderfreibetragempfänger_2`?

hmgaudecker:
In general, we always prefer the user specifying things explicitly (even if it creates a slightly higher burden on her) rather than assuming things which cover 90% of the use cases and are wrong for the remainder.

MImmesberger:
I agree, that is what I had in mind as well. 

Then we should add tests of the user input at some point (e.g. if both parents are specified via `p_id_kinderfreibetragempfänger_1` and `p_id_kinderfreibetragempfänger_2`, one of those two persons (or the child itself) must receive the Kindergeld and no one else).

hmgaudecker:
Yes, I am afraid that there will be plenty of data consistency checks in these. I do think, however, that we can add them as we go along. No need to research all of them before they crop up.

MImmesberger:
I added some test files and renamed some functions to illustrate the changes necessary. I propose the following convention:

Take some tax or Freibetrag called `x`:
- `x_kind_m` yields the transfer or Freibetrag the individual is eligible for in its own right. I changed the functions for Kindergeld, Erziehungsgeld and Kinderfreibetrag accordingly.
- `x_eltern_m` aggregates the transfers that some person gets via a new input colum: `p_id_x_empfänger`. One exception: Kinderfreibeträge are distributed via `p_id_kinderfreibetrag_empfänger_1` and `p_id_kinderfreibetrag_empfänger_1`. This is the step that needs to be implemented.
- We rename `x_eltern_m` to `x_m` via a simple function. This is necessary because for some `x` we need to make checks on the parent level before the transfer can be distributed (e.g. Erziehungsgeld, check whether parents don't work too much).

In case you have different naming suggestions let me know (I'm not too happy with `x_eltern_m` but I can't think of a better name that is short and applies to all children related transfers).

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -335,27 +335,50 @@ def sonderausgaben_betreuung_y_tu(
...
+        See :func:`kindergeld_anspruch`.
+
+    Returns
+    -------
+
+    """
+    return kindergeld_anspruch
+
+
+def eink_st_kinderfreib_kind_y(
```
hmgaudecker, line 355: While I like the general idea of what you are suggesting, here it is confusing. There simply is no notion of Kinderfreibetrag at the child level, it is always linked to the taxpayers.

I suggest we aggregate `kindergeld_anspruch` at the parent level and only then add up the payment values here.

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -6,9 +6,8 @@
 
 @dates_active(start="2004-01-01", end="2008-12-31")
 def erziehungsgeld_m(
-    erziehungsgeld_kind_m_fg: int,
+    erziehungsgeld_eltern_m: int,
```
hmgaudecker, line 9: So this would be just the amount before checking eligibility? We'll need some naming convention to make that clear eventually, but should only tackle that after implementing the hierarchical naming.

src/_gettsim_tests/test_data/parent_child_links/2007/eltern_getrennt_1_kind.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - 3
+    p_id_kinderfreibetrag_empfänger_2:
+      - -1
+      - -1
+      - 3
+    p_id_erziehungsgeld_empfänger:
+      - -1
+      - -1
+      - 3
+    kindergeld_anspruch:
```
hmgaudecker, line 77: Shouldn't this be endogenous?

src/_gettsim_tests/test_data/parent_child_links/2007/eltern_getrennt_1_kind.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - false
+    eigener_bedarf_gedeckt:
+      - false
+      - false
+      - false
+    p_id_kindergeld_empfänger:
+      - -1
+      - -1
+      - 3
+    p_id_kinderfreibetrag_empfänger_1:
```
hmgaudecker, line 65: Shouldn't this be endogenous?

src/_gettsim_tests/test_data/parent_child_links/2007/eltern_getrennt_1_kind.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - -1
+      - -1
+      - 3
+    p_id_kinderfreibetrag_empfänger_2:
```
hmgaudecker, line 69: Shouldn't this be endogenous?

hmgaudecker (commented):
> I added some test files and renamed some functions to illustrate the changes necessary. 

Thanks! 

> I propose the following convention:
> 
> Take some tax or Freibetrag called `x`:
> 
> * `x_kind_m` yields the transfer or Freibetrag the individual is eligible for in its own right. I changed the functions for Kindergeld, Erziehungsgeld and Kinderfreibetrag accordingly.
> 
>  * `x_eltern_m` aggregates the transfers that some person gets via a new input colum: `p_id_x_empfänger`. One exception: Kinderfreibeträge are distributed via `p_id_kinderfreibetrag_empfänger_1` and `p_id_kinderfreibetrag_empfänger_2`. This is the step that needs to be implemented.

Great in general! However, not for Kinderfreibetrag, see inline comment.

> * We rename `x_eltern_m` to `x_m` via a simple function. This is necessary because for some `x` we need to make checks on the parent level before the transfer can be payed out (e.g. Erziehungsgeld, check whether parents don't work too much).

Fine for now, will need clear naming convention in the future.

> In case you have different naming suggestions let me know (I'm not too happy with `x_eltern_m` but I can't think of a better name that is short and applies to all children related transfers).

The general problem there has less to do with parents/children than with eligibility checks in general, I think. Same for `bürgergeld`, you have a max payment amount and then look for tapering due to other income. Nothing that needs to be solved in this PR. Will be much easier to tackle once we have the hierarchical naming convention, too.

src/_gettsim/transfers/erziehungsgeld.py
MImmesberger: Yes, `erziehungsgeld_eltern_m` is before eligibility checks on the parental level. Better naming would be something like `erziehungsgeld_vor_eltern_check_kind_m` and `erziehungsgeld_vor_eltern_check_eltern_m` but this would be super long and, as you said, easier to solve later.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
MImmesberger: I agree, but keeping something like

```python
def eink_st_kinderfreib_anspruch_kind(
    kindergeld_anspruch: bool,
) -> bool:
    return kindergeld_anspruch
```

and summarizing this internally based on `p_id_elternteil_x` to `eink_st_kinderfreib_anspruch_eltern: int` makes it easier to use the same infrastructure for both transfers and Freibeträge.

src/_gettsim_tests/test_data/parent_child_links/2007/eltern_getrennt_1_kind.yaml
MImmesberger: Yes, that can be based on `p_id_elternteil_x`.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
hmgaudecker: See my previous point. We can do all that with `kindergeld_anspruch`, really no need to create a synonym.

MImmesberger: My point is that for transfers we would aggregate `x_kind` to `x_eltern` but with your suggested change, Freibeträge would be special because we aggregate `y_kind` to `x_eltern`. I wanted to build on the logic of the aggregation suffixes that we already have. But of course, we can handle this special case internally.

MImmesberger: True, that's a good approach.

MImmesberger:
To summarize the discussion so far, there are two steps:

**Creating `p_id_kinderfreibetrag_1` and `p_id_kinderfreibetrag_2`**
- In general, this just mirrors `p_id_elternteil_1` and `p_id_elternteil_2`. 
- If there are no parents, the user can set `p_id_kinderfreibetrag_1` and `p_id_kinderfreibetrag_2` herself, but we set the usual `-1` default value.
- If only one parent is specified via `p_id_elternteil_1`, `p_id_kinderfreibetrag_1` points to that parent, but not `p_id_kinderfreibetrag_2`.

(A side note: Having one parent that receives both Kinderfreibeträge is only possible if [the other parent doesn't satisfy alimony requirements to some extent](https://www.vlh.de/wissen-service/steuer-abc/wie-funktioniert-das-mit-dem-kinderfreibetrag.html#:~:text=Im%20Falle%20einer%20Trennung%20oder,oder%20bei%20der%20Mutter%20leben.). I would prefer to leave that out for now.)

**Aggregating the children-related transfers / the children-induced entitlements for Freibeträge to their parents.**
- This is based on the pointers that are endogenously created (for the case of Kinderfreibetrag) or specified by the user (all transfers).
- For $x\in\{\text{eink_st_kinderfreib}, \text{erziehungsgeld}, \text{kindergeld}\}$ we need to aggregate individual-level columns marked via `x_kind` (plus the usual aggregation suffixes) to a new column `x_eltern` based on `p_id_x_empfänger` (transfers) or `p_id_x_empfänger_1`/`p_id_x_empfänger_2` (Freibeträge).

The tests that I wrote should cover all the new features that are discussed here.

Does that make sense to everyone? @hmgaudecker @lars-reimann

lars-reimann:
From a design point of view, this seems to be a reasonable solution.

hmgaudecker:
Sounds great! Just one little clarification question:

> (A side note: Having one parent that receives both Kinderfreibeträge is only possible if [the other parent doesn't satisfy alimony requirements to some extent](https://www.vlh.de/wissen-service/steuer-abc/wie-funktioniert-das-mit-dem-kinderfreibetrag.html#:~:text=Im%20Falle%20einer%20Trennung%20oder,oder%20bei%20der%20Mutter%20leben.). I would prefer to leave that out for now.)

Not sure what you mean by "leave out":
1. Disallow the possibility that `p_id_kinderfreibetrag_1` and `p_id_kinderfreibetrag_2` point to the same person?
2. Do not attempt to model the rules that lead to `p_id_kinderfreibetrag_1` and `p_id_kinderfreibetrag_2` pointing to the same person when the user does not specify these two `p_id`-variables explicitly?

I'd be all for 2. (probably not even possible as this seems to be decided on a case-by-case basis), but would prefer to allow for 1.. So users should be able to make it happen, but the typical use case will not cover it.

MImmesberger:
Yes, should have been more clear about this. That's exactly what I had in mind. We generally allow for double Freibeträge but we don't create them endogenously -- they have to be enforced by the user.

MImmesberger:
I realized that parent-child links in the case of Kindergeld are more complex than Freibeträge and Erziehungsgeld because in the past Kindergeld depended on the number of children.

This implies that the notion of "Kindergeld belongs to the child but is payed out to the parent" is wrong in this case; we cannot assign the amount of Kindergeld that the child is eligible for without knowing i) which parent claims Kindergeld and ii) for how many children this parent claims Kindergeld (think of couples with children from different partners).

This sounds like we need to treat Kindergeld very similar to Kinderfreibeträge: It doesn't exist on the child level, but it may be payed out to the child instead of the parent in special cases. However, now I'm not sure whether we want to even cover those special cases?

hmgaudecker:
1. Yes, I think it's fine to generally think in two steps — binary eligibility at the child level and do the amount calculation only where it is required, typically parents.
2. Special cases: nothing to be done in this PR. Should be possible to just point at the kids itself some day, if necessary.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -335,27 +335,60 @@ def sonderausgaben_betreuung_y_tu(
...
     """
-    kinderfreib_total = sum(eink_st_abzuege_params["kinderfreib"].values())
-    out = kinderfreib_total * anz_kinder_mit_kindergeld_tu * anz_erwachsene_tu
 
-    return float(out)
+    out = (
+        sum(eink_st_abzuege_params["kinderfreib"].values())
+        * eink_st_kinderfreib_anz_ansprüche
+    )
+    return out
```
lars-reimann, line 360: I don't see much benefit in the creation of a single-use local variable before immediately returning it. But it's a pattern used frequently in the code base. Maybe @hmgaudecker knows why this is being done.

src/_gettsim/aggregation_jax.py
```diff
@@ -84,3 +84,17 @@ def grouped_all(column, group_id):
     out_on_hh = segment_min(column, group_id)
     out = out_on_hh[group_id]
     return out
+
+
+def sum_values_by_index(column, id_col, p_id):
+    fail_if_dtype_not_numeric_or_boolean(column, agg_func="sum_values_by_index")
+    if column.dtype in ["bool"]:
```
lars-reimann, line 91: Could be replaced with an equality check.

src/_gettsim/aggregation_numpy.py
```diff
@@ -127,6 +127,18 @@ def grouped_cumsum(column, group_id):
...
 
+def sum_values_by_index(column, id_col, p_id):
+    fail_if_dtype_not_numeric_or_boolean(column, agg_func="sum_values_by_index")
+    if column.dtype in ["bool"]:
+        column = column.astype(int)
+    out = numpy.zeros_like(p_id, dtype=column.dtype)
+
+    for position, id_receiver in enumerate(id_col):
+        if id_receiver >= 0:
+            out[numpy.where(p_id == id_receiver)] += column[position]
```
lars-reimann, line 138: Performance can be improved here: Currently, you have an outer loop (for..in) going over the $n$ elements of `id_col` and an inner loop (vectorized `==`) going over the $n$ elements of `p_id`. Thus, you currently have quadratic complexity, i.e. doubling $n$ quadruples the runtime.

You are better off first creating a `dict` mapping `p_id`s to positions (linear complexity) and then using this dictionary to look up the position of the person with `p_id == id_receiver` for each entry in `id_col` (linear complexity). This overall still has linear complexity instead of quadratic.

lars-reimann:
> @lars-reimann I have issues with the tests. On my machine, everything works fine (even when I set `USE_JAX` to True), but here they fail and I have no idea why. Could you have a look?

* The tests work on my system, too.
* After updating the `conda` environment with `conda env update --file environment.yml --prune`, the tests still work.
* The expected/actual values are relatively close to each other. Could be different rounding?
* Dependencies are not pinned, so different versions might be used locally and during CI. This might be the cause for the rounding differences.

@MImmesberger You could try making any change to the `environment.yml` file and pushing that, so a new environment gets created in CI. Currently, it uses a cached one.

lars-reimann:
Hmm, the expected values are all floating point numbers (e.g. `[right]: [9312.0, 4656.0, 0.0, 0.0]`), while the returned numbers are ints (e.g. `[left]:  [9540, 4770, 0, 0]`). Is there some unintended conversion happening along the way?

src/_gettsim/aggregation.py
```diff
@@ -68,3 +70,10 @@ def grouped_all(column, group_id):
 def grouped_cumsum(column, group_id):
     # Not yet implemented for jax
     return grouped_cumsum_numpy(column, group_id)
+
+
+def sum_values_by_index(column, id_col, p_id):
```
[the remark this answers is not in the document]
MImmesberger: Thanks! How would the type hints look like in this case? At this point, we don't know whether the functions returns a numpy.ndarray or a jax.numpy.ndarray.

src/_gettsim/aggregation_numpy.py
MImmesberger: Makes sense, thanks!

src/_gettsim/aggregation.py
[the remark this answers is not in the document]
lars-reimann: Several options:

1. If the `jax.numpy.ndarray` has the same interface as a `numpy.ndarray` (which the name implies), you can just use `numpy.ndarray` as the return type. A caller need not know the actual implementation of the array, only what they can do with it.
2. If the interfaces are not compatible, you can specify a union type `numpy.ndarray | jax.numpy.ndarray`.

MImmesberger:
> Hmm, the expected values are all floating point numbers

The error occurs because I used the wrong parameters. What is more interesting is why pytest didn't complain on my (and Lars') local machine. It complained after I changed the relevant parameters from floats to ints, so I suppose one weak spot in my implementation is that it matters whether a parameter is used in int or float format. 

I don't know what the annotations exactly do, but could it be this line? @lars-reimann 
```python
annotations["returns"] = int if annotations["source_col"] in (int, bool) else float
```

Else, I don't know what could create such a behavior.

hmgaudecker (commented):
Our implementation should not rely on this. The int version is much more readable!

MImmesberger:
Ah, I figured it out. I didn't know that the tests are running on the main branch. This PR wasn't up to date at that time. Sorry for the mess.

hmgaudecker:
> Ah, I figured it out. I didn't know that the tests are running on the main branch. This PR wasn't up to date at that time. Sorry for the mess.

Not sure what you mean? They are not running on the main branch but on branch associated with the PR. And locally on whatever branch you have checked out.

MImmesberger:
> Not sure what you mean? They are not running on the main branch but on the branch associated with the PR. And locally on whatever branch you have checked out.

Then I misinterpreted at article, thanks!

MImmesberger:
I'm done. Parent-child links are specified as dicts in the respective module, just as the aggregation specifications. The current implementation allows for specifications provided by the user (as with the aggregation functions). 

Apparently, there is no codecov base report. We hit 82% of diff, but this is largely due to lines that weren't touched in this PR.

src/_gettsim/aggregation_numpy.py
```diff
@@ -120,13 +121,33 @@ def grouped_cumsum(column, group_id):
...
-    if column.dtype in ["bool"]:
+    if column.dtype == bool:
         column = column.astype(int)
     out = npg.aggregate(group_id, column, func="cumsum", fill_value=0)
 
     return out
 
 
+def sum_values_by_index(
+    column: numpy.ndarray,
```
lars-reimann, line 132: For consistency, we could always use the type `npt.NDArray` instead of `numpy.ndarray`.

lars-reimann (approved):
I did not verify whether the test data is correct. The quality of the implementation is good, though.

src/_gettsim/functions_loader.py
```diff
@@ -291,7 +321,7 @@ def _format_duplicated_functions(duplicated_functions, functions, source):
     return "\n".join(lines)
 
 
-def _load_aggregation_combined_dict_from_strings(sources):
+def _load_aggregation_and_link_combined_dict_from_strings(sources):
```
hmgaudecker, line 324:
```suggestion
def _load_aggregation_and_link_dicts_from_modules(sources):
```

I think this makes clearer what is going on. With the two things, I was thinking of combining aggregations and links...

src/_gettsim/functions_loader.py
```diff
@@ -310,7 +340,8 @@ def _load_aggregation_combined_dict_from_strings(sources):
             aggregation_dicts_defined_in_module = [
```
hmgaudecker, line 340:
```suggestion
            relevant_dicts_defined_in_module = [
```

But maybe the better option is to just do two dictionaries? I find the lambda functions below pretty confusing. 

Or add an argument to the function whether we want aggregation or interpersonal links? Probably best, unless 
1. it is too expensive to do the imports twice or 
2. we need both dicts at the same time for the consistency checks.

Seems to me it can't be 1. as we are calling it multiple times and discarding one of the outputs.

src/_gettsim/functions_loader.py
```diff
@@ -310,7 +340,8 @@ def _load_aggregation_combined_dict_from_strings(sources):
...
                 obj
                 for name, obj in inspect.getmembers(out)
-                if isinstance(obj, dict) and name.startswith("aggregation_")
+                if isinstance(obj, dict)
+                and name.startswith(("aggregation_", "interpersonal_links_"))
                 # if _is_function_defined_in_module(func, out.__name__)
```
hmgaudecker, line 345:
```suggestion
```

Can this line go?

ValueError below has two typos in `aggregation_ dictionarys`

src/_gettsim/functions_loader.py
```diff
@@ -571,6 +606,129 @@ def aggregation_func(source_col, group_id):
...
+            data_cols,
+        )
+        for link_col, link_spec in link_dict.items()
+        if link_spec["source_col"] in user_and_internal_functions
+    }
+
+    return link_functions
+
+
+def _create_one_link_func(
```
hmgaudecker, line 637:
```suggestion
def _create_one_interpersonal_link_func(
```

Note: Multiple references to parents & children, which won't be caught by search and replace in the function body.

src/_gettsim/functions_loader.py
```diff
@@ -571,6 +606,129 @@ def aggregation_func(source_col, group_id):
     return aggregation_func
 
 
+def _create_parent_child_link_functions(
```
hmgaudecker, line 609: So we are testing all the stuff that is flagged by codecov here? I seem to remember a comment on that, but thought I'd double-check.

src/_gettsim/functions_loader.py
```diff
@@ -571,6 +606,129 @@ def aggregation_func(source_col, group_id):
...
+) -> Callable:
+    """Create a function that links parent and child variables."""
+
+    _fail_if_source_col_not_in_functions(
+        link_spec, user_and_internal_functions, data_cols
+    )
+
+    annotations = {}
+
+    if (
```
hmgaudecker, line 650: Can you explain what this is doing?

src/_gettsim/functions_loader.py
```diff
@@ -571,6 +606,129 @@ def aggregation_func(source_col, group_id):
...
+    else:
+        raise NotImplementedError(
+            """Parent-child links based on more
+                                  than two ID columns currently not supported."""
+        )
+
+    return link_func
+
+
+def sum_by_parent(
```
hmgaudecker, line 713: Misnomer. Also, why do we need this function? Seems to be exactly what `sum_values_by_index` does?

src/_gettsim/functions_loader.py
```diff
@@ -571,6 +606,129 @@ def aggregation_func(source_col, group_id):
...
+    source_col: numpy.ndarray,
+    id_col: numpy.ndarray,
+    p_id: numpy.ndarray,
+) -> numpy.ndarray:
+    return sum_values_by_index(source_col, id_col, p_id)
+
+
+def sum_by_parent_multiple_targets(
```
hmgaudecker, line 721: See above, though here the interface changes, of course. Might still be clearer to do it just in the calling code directly, rather than adding a function.

src/_gettsim/synthetic.py
```diff
@@ -157,12 +157,41 @@ def create_basic_households(
     ]
     group_ids = [f"{g}_id" for g in exogenous_groupings]
     df["p_id"] = df.index
+
+    # Create Elternteil IDs
```
hmgaudecker, line 161: We'll probably want to change these eventually. Would be better to do it directly rather than setting things up old-style and then doing research. However, this whole function can be updated once the new structure is used everywhere.

src/_gettsim/transfers/erziehungsgeld.py
```diff
@@ -2,12 +2,18 @@
 Elterngeld."""
 from _gettsim.shared import add_rounding_spec, dates_active
 
+interpersonal_links_erziehungsgeld = {
+    "erziehungsgeld_eltern_m": {
+        "id_col": "p_id_erziehgeld_empf",
```
hmgaudecker, line 7:
```suggestion
        "id_col": "p_id_erziehungsgeld_empf",
```
We should never abbreviate within a word.

src/_gettsim/transfers/kindergeld.py
```diff
@@ -11,20 +11,24 @@
     },
 }
```
hmgaudecker, line 12: Why do we still need the aggregation above? This would be a good chance to get rid of the `grouped_cumsum` altogether. But see comment on proof of concept vs implementation above.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -332,27 +342,59 @@ def sonderausgaben_betreuung_y_tu(
     return float(out)
 
 
-def eink_st_kinderfreib_y_tu(
-    anz_kinder_mit_kindergeld_tu: float,
-    anz_erwachsene_tu: int,
+def eink_st_kinderfreib_y(
```
hmgaudecker, line 345: How do you think about this PR?

1. Proof of concept?
2. Implementation of things doing it the new way?

If 1., we are probably done. If 2., we should strive to get rid of things like `anz_kinder_mit_kindergeld_tu` altogether.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
```diff
@@ -332,27 +342,59 @@ def sonderausgaben_betreuung_y_tu(
...
 
     Returns
     -------
 
     """
-    kinderfreib_total = sum(eink_st_abzuege_params["kinderfreib"].values())
-    out = kinderfreib_total * anz_kinder_mit_kindergeld_tu * anz_erwachsene_tu
 
-    return float(out)
+    return (
```
hmgaudecker, line 363:
```suggestion
    return float(
```

I thought the constituents might be ints?

src/_gettsim/transfers/kindergeld.py
```diff
@@ -11,20 +11,24 @@
...
+
 
 def kindergeld_m(
-    kindergeld_anspruch: bool,
-    kumulativer_kindergeld_anspruch_tu: int,
+    kindergeld_anz_anprüche: bool,
     kindergeld_params: dict,
 ) -> float:
-    """Calculate kindergeld for an individual child.
+    """Kindergeld given number of Kindergeld claims.
```
hmgaudecker, line 26:
```suggestion
    """Sum of Kindergeld for eligible children.
```

hmgaudecker (commented):
Great work! 

Have to stop now -- did not manage to look through the tests yet.

src/_gettsim/functions_loader.py
lars-reimann: We can ignore the cases where the function header is marked but nothing in the function body. Flagged lines in a function body are truly not covered.

src/_gettsim/taxes/zu_verst_eink/freibetraege.py
MImmesberger: I planned to implement the infrastructure in this PR and adjust the existing policy functions in a later PR (that gets rid of `tu` occurrences) such that this PR doesn't get too large. But I'm open to do here too.

src/_gettsim/transfers/erziehungsgeld.py
MImmesberger: I think we had this discussion before. This is due to the character limit (it's a basic input variable).
- `p_id_erziehgeld_empf` has exactly 20 characters
- `p_id_erziehungsgeld_empf` has 24 characters

One solution would be to use `p_id_erziehungsg_emp` but I have the feeling that `p_id_erziehgeld_empf` is easier to understand.

MImmesberger:
GEPs and documentation are updated! @hmgaudecker after you reviewed it, we can merge. I would vote for getting rid of the other `tu`s and correctly specifying the Kindergeld functions in another PR.

src/_gettsim/functions_loader.py
```diff
@@ -420,7 +463,65 @@ def wrapper_rename_arguments(*args, **kwargs):
         return decorator_rename_arguments
 
 
-def _create_one_aggregation_func(  # noqa: PLR0912
+def _check_agg_specs_validity(agg_specs, agg_col):
+    if "aggr" not in agg_specs:
+        raise KeyError(
+            f"No aggr keyword is specified for aggregation column {agg_col}."
```
hmgaudecker, line 469:
```suggestion
            f"`aggr` key is missing for aggregation column {agg_col}."
```

src/_gettsim/functions_loader.py
```diff
@@ -420,7 +463,65 @@ def wrapper_rename_arguments(*args, **kwargs):
...
+        )
+
+    if agg_specs["aggr"] != "count":
+        if "source_col" not in agg_specs:
+            raise KeyError(
+                f"Source_col is not specified for aggregation column {agg_col}."
```
hmgaudecker, line 475:
```suggestion
                f"`source_col` key is missing for aggregation column {agg_col}."
```

src/_gettsim/functions_loader.py
```diff
@@ -431,31 +532,20 @@ def _create_one_aggregation_func(  # noqa: PLR0912
...
-        ) from e
 
-    if aggr != "count":
-        try:
-            source_col = agg_specs["source_col"]
-        except KeyError as e:
-            raise KeyError(
-                f"Source_col is not specified for aggregation column {agg_col}."
-            ) from e
+    _check_agg_specs_validity(agg_specs, agg_col)
```
hmgaudecker, line 548: This should go in a comprehension one level further up, symmetric to `aggregate_by_p_id`.

src/_gettsim/transfers/kindergeld.py
```diff
@@ -11,20 +11,25 @@
...
+aggregate_by_p_id_kindergeld = {
+    "kindergeld_anz_ansprüche": {
+        "p_id_to_aggregate_by": "p_id_kindergeld_empf",
+        "source_col": "kindergeld_anspruch",
+        "aggr": "sum",
+    },
+}
+
 
 def kindergeld_m(
```
hmgaudecker, line 23: Should be simplified to

```python
kindergeld_params["kindergeld"] * kindergeld_anz_ansprüche
```

starting in 2023. Either do here or make an issue.

src/_gettsim_tests/test_aggregation_functions.py
```diff
@@ -11,6 +11,7 @@
     grouped_mean,
     grouped_min,
     grouped_sum,
+    sum_by_p_id,
```
hmgaudecker, line 14: Pls make an issue reminding us to add and test the other `aggr_by_p_id` functions.

hmgaudecker (approved):
Excellent!!! Just some cosmetics left, no show-stoppers.

docs/geps/gep-04.md
```diff
@@ -250,30 +250,75 @@ Using a different reduction function than the sum is as easy as explicitly speci
...
+```
+
+This dict creates a target function `kindergeld_anz_ansprüche` which gives the amount of
+claims that a person has on Kindergeld, based on the `kindergeld_anspruch` function
+which returns bools that show whether a person is a reason for a Kindergeld claim.
+
+The output type will be the same as the input type. Exceptions:
+
+- Input type `bool` and aggregation `sum` leads to output type `int`.
+- Input type `int` and aggregation {math}`\in \{` `any`, `all` {math}`\}` leads to
```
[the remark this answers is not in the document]
MImmesberger: Yes, I don't see a reason not to support float inputs.

MImmesberger:
Codecov complains because i) we didn't test (and implement) the new aggregation functions and ii) internal handling of aggregation_by_p_id is not tested (as discussed yesterday). Would vote for ignoring it.

PR #687, opened 2024-02-05
Use max() instead of explicit if

schra:
max() and min() are frequently used in the code base, so it would be more uniform to do the same here.


MImmesberger (approved):
Thanks!

PR #691, opened 2024-02-29
minor fixes to doc

git-girl:

- there was a dead link in the doc
- there was an unnecessary cell from my understanding of the docs
   - While the cell does show that kinderfreibetraege are used in wohngeld
the the rest of the documentation doesnt go into wohngeld at all.
imo this isn't helpful in understanding kindergeld and more importantly
distracting from the example.

also cool thanks for your work and having this be open source!

### DIFF
--- a/docs/tutorials/advanced_usage.ipynb
+++ b/docs/tutorials/advanced_usage.ipynb
@@ -54,15 +54,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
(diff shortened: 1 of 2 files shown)


MImmesberger:
Yes, you're right. Thanks for spotting the mistakes and correcting them!

@lars-reimann Can you have a look where the `CondaVerificationError` in the `readthedocs` check comes from?

MImmesberger:
@git-girl I included your changes in #695 to fix the issue with the `readthedocs` check. Thanks again!

PR #700, opened 2024-03-07
Data columns overwrite functions regardless of time unit

### LINKED ISSUE #699: BUG: Time conversion fails for user inputs
### Bug description

When playing around with the tests for Elterngeld in #694, I noticed that `eink_st_m_sn` is not recognised as a the time converted counterpart of `eink_st_y_sn`. The same happens for `soli_st_m_sn`. Changing the income tax input in any of the Elterngeld tests reproduces the error.

Is it intentional that we don't convert data inputs or is that a bug?

On `main` the Elterngeld test takes care of this via 
```python
df["soli_st_y_tu"] = df["soli_st_m"].groupby(df["tu_id"]).transform("sum") * 12
```

lars-reimann:

Closes #699

Data columns now always overwrite policy functions regardless of their time unit. If, for example, the value `eink_st_y_sn` is needed, and a data column `eink_st_m_sn` and policy function `eink_st_y_sn` are provided, the converted data column is used.

### Todo

### DIFF
--- a/src/_gettsim_tests/test_time_conversion.py
+++ b/src/_gettsim_tests/test_time_conversion.py
@@ -204,7 +204,7 @@ class TestCreateFunctionsForTimeUnits:
         for expected_name in expected:
             assert expected_name in time_conversion_functions
 
-    def test_should_not_create_functions_that_exist_already(self) -> None:
+    def test_should_not_create_functions_automatically_that_exist_already(self) -> None:
         time_conversion_functions = create_time_conversion_functions(
             {"test1_d": lambda: 1}, ["test2_y"]
         )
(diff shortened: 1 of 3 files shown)


src/_gettsim_tests/test_time_conversion.py
```diff
@@ -205,12 +205,19 @@ def test_should_create_functions_for_other_time_units_for_data_cols(
             assert expected_name in time_conversion_functions
 
     def test_should_not_create_functions_that_exist_already(self) -> None:
```
hmgaudecker, line 207:
```suggestion
    def test_should_not_create_functions_automatically_that_exist_already(self) -> None:
```

src/_gettsim_tests/test_time_conversion.py
```diff
@@ -205,12 +205,19 @@ def test_should_create_functions_for_other_time_units_for_data_cols(
...
+        time_conversion_functions = create_time_conversion_functions(
+            {"test1_d": lambda: 1}, ["test2_y"]
+        )
+
+        assert "test1_d" not in time_conversion_functions
+        assert "test2_y" not in time_conversion_functions
+
+    def test_should_overwrite_functions_with_data_cols(self) -> None:
```
hmgaudecker, line 215:
```suggestion
    def test_should_overwrite_functions_with_data_cols_that_only_differ_in_time_period(self) -> None:
```

hmgaudecker (approved):
Excellent, thanks! I'd just make the names more explicit at the expense of a terrible length. But I had to think for too long previously to make out why results *should* differ.

PR #715, opened 2024-03-18
Fix time conversion after rounding

### LINKED ISSUE #712: BUG: Time conversion after rounding
### Bug description

When using `eink_st_m_sn` as a target, there is an error that the rounding specification is missing in the parameter yaml. This happens because the source function (`eink_st_y_sn`) is rounded and an entry for `eink_st_m_sn` is missing in the parameter yaml.

#### Expected Output

The source function should be rounded as specified. Then, the converted function should be created.

lars-reimann:

Closes #712

Previously, deriving a time-converted function from a rounded function lead to the following error:

```txt
KeyError: Rounding specifications for function test_func_y are expected in the parameter dictionary 
at ['params_key_test']['rounding']['test_func_y']. These nested keys do not exist. 
If this function should not be rounded, remove the respective decorator.
```

This PR fixes this.

### Todo

### DIFF
--- a/src/_gettsim/time_conversion.py
+++ b/src/_gettsim/time_conversion.py
@@ -351,6 +351,11 @@ def _create_function_for_time_unit(
         return converter(x)
 
     if info is not None:
-        func.__info__ = info
+        # The derived function should not be rounded (see #712)
+        func.__info__ = {
+            key: value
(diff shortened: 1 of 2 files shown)


hmgaudecker (approved):
Great, thanks!

Issue #664, opened 2023-11-21
BUG: Altersrente für Frauen Vertrauensschutz.


MImmesberger:
### Bug description

The Altersrente für Frauen has Vertrauensschutz rules that are very similar to the ones of Altersrente wg. Arbeitslosigkeit (implemented in #645).

A good place to start should be the [RRG 1992](https://www.bgbl.de/xaver/bgbl/text.xav?SID=&tf=xaver.component.Text_0&tocf=&qmf=&hlf=xaver.component.Hitlist_0&bk=bgbl&start=%2F%2F*%5B%40node_id%3D%27964463%27%5D&skin=pdf&tlevel=-2&nohist=1&sinst=3DC4811B), [WFG 1997](https://www.bgbl.de/xaver/bgbl/text.xav?SID=&tf=xaver.component.Text_0&tocf=&qmf=&hlf=xaver.component.Hitlist_0&bk=bgbl&start=%2F%2F*%5B%40node_id%3D%27959179%27%5D&skin=pdf&tlevel=-2&nohist=1&sinst=984B3210), [Ruhestandsübergangsgesetz 1996](https://www.bgbl.de/xaver/bgbl/text.xav?SID=&tf=xaver.component.Text_0&tocf=&qmf=&hlf=xaver.component.Hitlist_0&bk=bgbl&start=%2F%2F*%5B%40node_id%3D%27959278%27%5D&skin=pdf&tlevel=-2&nohist=1&sinst=DDEDA4BF) and the [RV-Nachhaltigkeitsgesetz 2004](https://www.bgbl.de/xaver/bgbl/text.xav?SID=&tf=xaver.component.Text_0&tocf=&qmf=&hlf=xaver.component.Hitlist_0&bk=bgbl&start=%2F%2F*%5B%40node_id%3D%27952876%27%5D&skin=pdf&tlevel=-2&nohist=1&sinst=974224E0).

MImmesberger:
This duplicates #718

Issue #663, opened 2023-11-21
BUG: Retirement age tests


MImmesberger:
### Bug description

The `renten_alter` tests test the policy environment of 2010 only. In some cases, the tests expect wrong results (e.g. Altersrente für Frauen; sometimes, a full retirement age above 65 is expected even though this pathway into retirement was abolished with a full retirement age of 65).

Additionally, it is not clear what purpose the `jahr` key serves in the test parameter files. 

### Suggested fix

Rewrite the test files to adhere to the test standards used in other modules.

MImmesberger:
Closed via #717

PR #717, opened 2024-03-29
Adjust statutory retirement ages

### LINKED ISSUE #491: ENH: Add information when certain reforms were introduced

### LINKED ISSUE #665: BUG: Statutory retirement ages
### Bug description

Currently, statutory retirement ages are often implemented via piecewise linear functions. This implementation makes sense for monetary units, but not for retirement ages, especially in the presence of staggered implementation of higher statutory retirement ages. The new implementation should look something like the solution used for the Altersrente wg. Arbeitslosigkeit in #645 

Note: Additionally, the current piecewise functions are not correct for pathways into retirement that were abolished. The current implementation uses the normal retirement age for cohorts that are not eligible for this pathway. This is not necessary because GETTSIM deals with this eligibility concerns elsewhere.

MImmesberger:
Closes #665, #664
Closes #491 partially
Related to #663 

This PR solves two problems (as discussed in #665):
1. Statutory retirement ages (SRAs) are wrong in the parameter files. This is a relict of the missing `dates_active` feature. Back then, we dealt with abolished pathways into retirement by setting the relevant SRA of the abolished pathway to the SRA of a pathway that continued to exist (e.g. setting the full retirement age for Altersrente for Frauen to the Regelaltersgrenze after the 1952 birth cohort). This logic fails once we want to depict changes of the policy environment over time.
2. SRAs are currently implemented via piecewise linear functions. We started to replace this with explicit birth-cohort-age dicts in #645 

Along these lines, this PR changes the test files that currently only test the pension functions for policy year 2010. Due to point 1, there are probably a lot of tests that expect wrong SRAs.

### Todo

- [x] Parameters + Functions of Regelaltersgrenze / NRA
- [x] Parameters + Functions of Altersrente f. Frauen (abolished since birth cohort 1952)
- [x] Parameters + Functions of Altersrente f. langj. Versicherte
- [x] Parameters + Functionf of Altersrente f. bes. langj. Versicherte (since 2012)


src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -1836,42 +2199,36 @@ altersgrenze_besond_langj_versicherte:
     en: null
   type: piecewise_linear
   2012-01-01:
```
MImmesberger, line 2201: Should this be the date of announcement or the date when this came into effect? I think we already discussed this in #645 and went for date of announcement (because difference didn't matter since affected cohorts were sufficiently far away from retirement age). Here, the difference matters, because the reform applies to all cohorts and was announced 5 years before it came into effect (2007 vs 2012).

Imo in this case we should go for the date when it came into effect, but then I'm worried regarding inconsistencies concerning how we set the other dates in this parameter file.

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -1727,75 +1962,203 @@ altersgrenze_langj_versicherte_vorzeitig:
...
+    note: ERA back to 63.
+    scalar: 63.0
+  1997-12-16:
+    reference: Rentenreformgesetz 1999. BGBl. I S. 2998 1997
+    note: ERA back to 62.
+    scalar: 62.0
+  2007-04-20:
+    reference: RV-Nachhaltigkeitsgesetz 2004. BGBl. I S. 1791 2004
+    note: ERA back to 63.
+    scalar: 63.0
```
MImmesberger, line 1983: Note (more to myself than to others, but feel free to chime in if you know more): The ERA reforms for the Rente f. langj. Versicherte seem very strange as it jumps back and forth from 63 to 62.

src/_gettsim/transfers/rente.py
```diff
@@ -288,7 +284,7 @@ def ges_rente_zugangsfaktor(  # noqa: PLR0913
 
     """
 
-    if rentner and ges_rente_vorauss_regelrente:
```
MImmesberger, line 291: Because `ges_rente_m` returns non-zero pensions only if `rentner` is `True`, I think it is not necessary to set the Zugangsfaktor to 0 as well.

src/_gettsim/transfers/rente.py
```diff
@@ -359,34 +355,24 @@ def age_of_retirement(
...
-    rentner
-        See basic input variable :ref:`rentner <rentner>`.
-
 
     Returns
     -------
     Age at retirement.
 
     """
-    if rentner:
```
MImmesberger, line 371: Because `jahr_renteneintr` and `monat_renteneintr` are required to build the DAG independent of the value of `rentner` the distinction doesn't make much sense here.

src/_gettsim/transfers/rente.py
```diff
@@ -463,37 +444,33 @@ def _ges_rente_altersgrenze_abschlagsfrei_mit_besond_langj(  # noqa: PLR0913
...
+    ges_rente_vorauss_arbeitsl
+        See :func:`ges_rente_vorauss_arbeitsl`.
 
     Returns
     -------
     Full retirement age.
 
     """
 
-    out = float("inf")
```
MImmesberger, line 491: The default value of `inf` doesn't make much sense to me (but maybe I'm missing something here?). The default retirement age without deductions should be the Regelaltersgrenze (NRA).

src/_gettsim/transfers/rente.py
```diff
@@ -591,43 +566,46 @@ def _referenzalter_abschlag_mit_rente_arbeitsl_frauen(
...
     elif ges_rente_vorauss_langj:
         out = _ges_rente_langj_altersgrenze
     elif ges_rente_vorauss_frauen:
-        out = ges_rente_frauen_altersgrenze
-    elif _ges_rente_vorauss_arbeitsl:
+        out = _ges_rente_frauen_altersgrenze
+    elif ges_rente_vorauss_arbeitsl:
         out = _ges_rente_arbeitsl_altersgrenze
     else:
-        out = float("NaN")
```
MImmesberger, line 614: Similar to above, the reference age for deductions should be the normal retirement age if the individual is not eligible for any other pathway. This is a bit stupid because those individuals cannot claim their pension before that age (and hence there cannot be any deductions) but with this change, the target is much more useful in applications I think.

MImmesberger:
I mostly changed parameters and adjusted how the functions access them accordingly. 

There are some occasions where I changed default values of targets. Those changes make sense to me but maybe I'm missing something (especially as I wasn't around when the original defaults were introduced). I highlighted them in the comments above to make the review easier.

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -1834,44 +2333,26 @@ altersgrenze_besond_langj_versicherte:
...
-      Cohort from which pension for very long term insured was fully implemented.
-  reference: § 237 SGB VI
-  note: https://www.sozialgesetzbuch-sgb.de/sgbvi/237.html
   2012-01-01:
-    scalar: 1952
+    reference: RV-Altersgrenzenanpassungsgesetz 2007. BGBl. I S. 554 2007
+    scalar: 65.0
+  2014-06-23:
+    reference: RV-Leistungsverbesserungsgesetz 2014. BGBl. I S. 787 2014
+    min_birthyear_new_regime: 1964
```
hmgaudecker, line 2341: Switch order: `entry_age_old_regime`, `entry_age_new_regime`, `max_birthyear_old_regime`, `min_birthyear_new_regime`.

src/_gettsim/parameters/ges_rente.yaml
hmgaudecker: Announcement is fine, won't ever be used before 2012 due to the setting of dates_active. But mention in the notes here so people are not confused.

src/_gettsim/transfers/arbeitsl_geld.py
```diff
@@ -144,16 +142,15 @@ def arbeitsl_geld_berechtigt(  # noqa: PLR0913
         See basic input variable :ref:`arbeitsstunden_w <arbeitsstunden_w>`.
     arbeitsl_geld_params
         See params documentation :ref:`arbeitsl_geld_params <arbeitsl_geld_params>`.
-    geburtsjahr
-        See basic input variable :ref:`geburtsjahr <geburtsjahr>`.
-    ges_rente_params
-        See params documentation :ref:`ges_rente_params <ges_rente_params>`.
+    ges_rente_regelaltersgrenze
+        See basic input variable :ref:`ges_rente_regelaltersgrenze
```
hmgaudecker, line 146:
```suggestion
        See :func:`ges_rente_regelaltersgrenze`.
```

src/_gettsim/transfers/arbeitsl_geld.py
```diff
@@ -144,16 +142,15 @@ def arbeitsl_geld_berechtigt(  # noqa: PLR0913
...
+            <ges_rente_regelaltersgrenze>`.
```
hmgaudecker, line 147:
```suggestion
```

src/_gettsim/transfers/rente.py
hmgaudecker: Then we are missing a test. Seems to me that this is the only way to get a `np.nan` in there.

src/_gettsim/transfers/rente.py
```diff
@@ -398,32 +384,28 @@ def _ges_rente_altersgrenze_abschlagsfrei_ohne_besond_langj(
...
-        See :func:`_ges_rente_vorauss_arbeitsl`.
+    ges_rente_vorauss_arbeitsl:
+        See :func:`ges_rente_vorauss_arbeitsl`.
     Returns
     -------
     Full retirement age.
 
     """
 
-    out = float("NaN")
```
hmgaudecker, line 421: Again, what is behaviour if all of these evaluate to False?

src/_gettsim/transfers/rente.py
hmgaudecker: But this here already checks whether people qualify, right? Tbh, I am not sure what the purpose is in the end -- I would think we'd check this as part of the `voraussetzungen`? Or am I misinterpreting the abbreviation?

src/_gettsim/transfers/rente.py
hmgaudecker: See above, I'll need a bit more explanation first.

src/_gettsim/transfers/rente.py
```diff
@@ -1374,8 +1779,53 @@ def ges_rente_vorauss_regelrente(ges_rente_wartezeit_5: bool) -> bool:
     return ges_rente_wartezeit_5
 
 
-@policy_info(end_date="2017-12-31")
-def ges_rente_vorauss_frauen(
+@policy_info(end_date="1997-12-15", name_in_dag="ges_rente_vorauss_frauen")
+def ges_rente_vorauss_frauen_ohne_geburtsjahr_prüfung(
```
hmgaudecker, line 1783:
```suggestion
def ges_rente_vorauss_frauen_ohne_prüfung_geburtsjahr(
```

src/_gettsim/transfers/rente.py
```diff
@@ -1384,6 +1834,8 @@ def ges_rente_vorauss_frauen(
 ) -> bool:
     """Eligibility for Altersrente für Frauen (pension for women).
 
+    Only individuals below a certain birth year are eligible.
```
hmgaudecker, line 1837:
```suggestion
    Only individuals born before a certain year are eligible.
```

hmgaudecker (commented):
Had a glance, looks mostly sensible to me! Just a couple of clarifications necessary, have not looked at this for too long.

src/_gettsim/parameters/ges_rente.yaml
MImmesberger: That's true, we could just let the function and parameter dates differ. Didn't think about that.

src/_gettsim/transfers/rente.py
MImmesberger: The function is supposed to yield the earliest age at which the person can claim benefits without any deductions. The `float("inf")` is just the value that is returned if the person is not eligible for any pathway into retirement. 

I'm arguing, that the function should return the NRA in that case, because it feels a bit stupid to set all statutory retirement ages to `inf` if the person is not eligible for retirement (in that case, we do the eligibility check downstream). I stumbled on this when I used these targets on real data, seems very unnatural to me.

hmgaudecker: One can argue for most stuff. From my point of view, "eligible with current state variables" is the  only important thing, we can't project into the future much, anyhow. So I am not fully sure why we need this at all...

Put differently, I think the cleanest thing is inf in case we are talking about a future age, but almost anything that is consistent and well documented is fine with me.

hmgaudecker: Put differently, I don't know what "eligible **at all**" would encompass. Anything but age? Some other selection?

src/_gettsim/transfers/rente.py
hmgaudecker: Okay, if zero is consistent, that is fine with me. But it should be documented and fixed using a test.

src/_gettsim/transfers/rente.py
MImmesberger: We are talking about current age.

This functions and all the others where I changed the default outputs (if all eligibility checks fail) are inputs of the Zugangsfaktor calculation. To make the changes that I favour explicit this would be the logic of the Zugangsfaktor calculation:

1. Calculate for each pathway into retirement the statutory retirement age (without any eligibility checks, but conditional on birth year).
2. Compute the earliest possible retirement age `_ges_rente_altersgrenze_vorzeitig`. This checks for eligibility of the pathways that allow for early retirement, but not for elgibility for Regelaltersrente. Hence, if the individual is not eligible for early retirement, it returns the NRA. **I didn't change anything here**.
3. Compute the reference age for deductions `_ges_rente_altersgrenze_abschlagsfrei`. Here I want to use the same logic as in 2: Compute the reference age, checking for eligibility for all pathways that allow for early retirement, but use the NRA if the individual is not eligible for any of those pathways. No eligibility check for Regelaltersrente.
4. Compute the Zugangsfaktor. Return 0 if not `rentner` or if not eligible for Regelaltersrente.

hmgaudecker: Cool. Would just be good if that logic became very apparent somewhere!

MImmesberger:
I think I addressed all the requested changes. The unresolved conversations are all related to the discussion we had [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/717#discussion_r1616003122). To make the logic more apparent for future users, I adjusted the docstrings accordingly.

hmgaudecker (approved):

PR #720, opened 2024-04-10
Combined decorator for policy information

lars-reimann:

Combine the decorators `add_rounding_spec` and `dates_active` into a single decorator `policy_info`.

### Todo

### DIFF
--- a/src/_gettsim/policy_environment.py
+++ b/src/_gettsim/policy_environment.py
@@ -242,8 +242,7 @@ def load_functions_for_date(date):
     functions = {}
     for f in load_internal_functions().values():
         if not is_time_dependent(f) or is_active_at_date(f, date):
-            info = f.__info__ if hasattr(f, "__info__") else {}
-            name = info.get("name_in_dag", f.__name__)
+            name = f.__info__["name_in_dag"] if hasattr(f, "__info__") else f.__name__
             functions[name] = f
 
     return functions
--- a/src/_gettsim/policy_environment.py
+++ b/src/_gettsim/policy_environment.py
@@ -254,7 +254,7 @@ def is_time_dependent(f: Callable) -> bool:
 
 
 def is_active_at_date(f: Callable, date: datetime.date) -> bool:
-    return f.__info__["start_date"] <= date <= f.__info__["dates_active_end"]
+    return f.__info__["start_date"] <= date <= f.__info__["end_date"]
 
 
 def _load_parameter_group_from_yaml(
--- a/src/_gettsim/policy_environment.py
+++ b/src/_gettsim/policy_environment.py
@@ -254,7 +254,7 @@ def is_time_dependent(f: Callable) -> bool:
 
 
 def is_active_at_date(f: Callable, date: datetime.date) -> bool:
-    return f.__info__["dates_active_start"] <= date <= f.__info__["dates_active_end"]
+    return f.__info__["start_date"] <= date <= f.__info__["dates_active_end"]
 
 
 def _load_parameter_group_from_yaml(
--- a/src/_gettsim/policy_environment.py
+++ b/src/_gettsim/policy_environment.py
@@ -243,14 +243,14 @@ def load_functions_for_date(date):
     for f in load_internal_functions().values():
         if not is_time_dependent(f) or is_active_at_date(f, date):
             info = f.__info__ if hasattr(f, "__info__") else {}
-            name = info.get("dates_active_dag_key", f.__name__)
+            name = info.get("name_in_dag", f.__name__)
             functions[name] = f
 
     return functions
 
 
 def is_time_dependent(f: Callable) -> bool:
-    return hasattr(f, "__info__") and "dates_active_dag_key" in f.__info__
+    return hasattr(f, "__info__") and "name_in_dag" in f.__info__
 
 
 def is_active_at_date(f: Callable, date: datetime.date) -> bool:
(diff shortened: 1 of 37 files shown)


lars-reimann:
I've deliberately kept `test_dates_active` unchanged since it's more the mechanism than the name of the decorator.

src/_gettsim/shared.py
```diff
@@ -16,43 +16,21 @@ def __repr__(self):
...
-
 TIME_DEPENDENT_FUNCTIONS: dict[str, list[Callable]] = {}
 
 
-def dates_active(
+def policy_info(
+    *,
     start: str = "0001-01-01",
     end: str = "9999-12-31",
     change_name: str | None = None,
```
hmgaudecker, line 26:
```suggestion
    name_in_dag: str | None = None,
```
I think this makes things clearer, especially given the new context but would have been good right away!

src/_gettsim/shared.py
```diff
@@ -61,6 +39,10 @@ def dates_active(
     ensure that the function name is unique in the file where it is defined. Otherwise,
     the function will be overwritten by the last function with the same name.
```
hmgaudecker, line 40:
```suggestion
    the function would be overwritten by the last function with the same name.
```

src/_gettsim/shared.py
```diff
@@ -98,6 +85,8 @@ def inner(func: Callable) -> Callable:
         func.__info__["dates_active_start"] = start_date
```
hmgaudecker, line 85:
```suggestion
        func.__info__["start_date"] = start_date
```
I think we can simplify a lot here now! The keys in the `__info__` dict should generally match the names of inputs to the decorator, I think. Unless heavily processed, but parsing dates seems below that bar to me.

src/_gettsim/shared.py
```diff
@@ -98,6 +85,8 @@ def inner(func: Callable) -> Callable:
...
         func.__info__["dates_active_end"] = end_date
```
hmgaudecker, line 86:
```suggestion
        func.__info__["end_date"] = end_date
```

src/_gettsim/shared.py
```diff
@@ -16,43 +16,21 @@ def __repr__(self):
...
-    return inner
-
-
 TIME_DEPENDENT_FUNCTIONS: dict[str, list[Callable]] = {}
 
 
-def dates_active(
+def policy_info(
+    *,
     start: str = "0001-01-01",
```
hmgaudecker, line 24:
```suggestion
    start_date: str = "0001-01-01",
```

src/_gettsim/shared.py
```diff
@@ -16,43 +16,21 @@ def __repr__(self):
...
     end: str = "9999-12-31",
```
hmgaudecker, line 25:
```suggestion
    end_date: str = "9999-12-31",
```

src/_gettsim/shared.py
```diff
@@ -16,43 +16,21 @@ def __repr__(self):
...
     change_name: str | None = None,
+    rounding_key: str | None = None,
```
hmgaudecker, line 27:
```suggestion
    params_key_for_rounding: str | None = None,
```
Also for the output. Long but clear. This will go again once we have hierarchical functions and the DAG extends to the params, so that should be the right trade-off for the moment.

hmgaudecker (commented):
Looks great, thanks! Just a couple of renaming suggestions.

src/_gettsim/policy_environment.py
```diff
@@ -243,18 +243,18 @@ def load_functions_for_date(date):
     for f in load_internal_functions().values():
         if not is_time_dependent(f) or is_active_at_date(f, date):
             info = f.__info__ if hasattr(f, "__info__") else {}
-            name = info.get("dates_active_dag_key", f.__name__)
+            name = info.get("name_in_dag", f.__name__)
```
hmgaudecker, line 246:
```suggestion
```

src/_gettsim/policy_environment.py
```diff
@@ -243,18 +243,18 @@ def load_functions_for_date(date):
     for f in load_internal_functions().values():
         if not is_time_dependent(f) or is_active_at_date(f, date):
             info = f.__info__ if hasattr(f, "__info__") else {}
```
hmgaudecker, line 245:
```suggestion
            name = f.__info__["name_in_dag"] if hasattr(f, "__info__") else f.__name__
```

src/_gettsim_tests/test_naming_conventions.py
```diff
@@ -34,8 +34,8 @@ def time_indep_function_names(all_functions: dict[str, callable]) -> list[str]:
     for function_name, function in all_functions.items():
         if hasattr(function, "__info__"):
             info = function.__info__
-            if "dates_active_dag_key" in info:
-                time_indep_function_names.add(info["dates_active_dag_key"])
+            if "name_in_dag" in info:
```
hmgaudecker, line 37: This check should not be necessary anymore?

hmgaudecker (approved):
Excellent, thanks!

src/_gettsim_tests/test_naming_conventions.py
lars-reimann: Assuming that users don't manually add `__info__` but use the decorator, then no. Seems reasonable to me, since `__info__` should not become part of the public interface.

Issue #708, opened 2024-03-09
ENH: Use pointers to make target columns available at source level


MImmesberger:
#### Is your feature request related to a problem?

Currently, pointers allow us to aggregate variables from source individuals (e.g. children) to target individuals (e.g. parents). Sometimes it is necessary to make target columns available on the source level (e.g. in the case of Unterhaltsvorschuss).

For more details, see the discussion we had [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/694#discussion_r1514514289).

lars-reimann:
Should this be another addition to the complex name convention for parameters, like <foreign_key>_<target_column> or how do you picture a general solution for this?

hmgaudecker:
In principle this is a simple join, pinky that we don't have dataframes but numpy/Jax arrays.

The naming convention you propose sounds great!

lars-reimann:
> The naming convention you propose sounds great!

Ok, this would also need

1. A distinct separator between foreign_key_column and target_column (not an underscore),
2. A distinct keyword to indicate the desired interpretation of this parameter.

Both requirements could be fulfilled at once by having a unique infix, like ` <foreign_key_column>_to_<target_column>`. Though we should probably pick something less likely to occur in other parameter names than `_to_`.

That said, I find the current name convention already quite hard to work with due to all the "magic" it encodes (time conversion, aggregation, renaming via `dates_active`, ...). Because of this, I'd rather suggest adding a utility function that takes three inputs:

1. Array for the foreign keys.
2. Array for the primary keys.
3. Array for the targets.

This function can then return another array with the targets that belong to the respective foreign keys. Where needed, this function can be called explicitly. If caching of the result is desired, it can be wrapped into another function like `_unterhaltsvors_empf_einkommen_m`.

hmgaudecker:
I don't follow all the way yet, I am afraid. How would that work with the DAG? How would somebody who wanted to extend GETTSIM call that function?

I'd also be fine with doing something provisional right now and then re-consider when re-designing the interface after the move to a hierarchical scheme.

lars-reimann:
Let's say we go with some name convention and add a parameter `p_id_unterhaltsvors_empf_to_einkommen_m` to some policy function:

1. This would implicitly add dependencies to the columns `p_id_unterhaltsvors` (foreign key), `p_id` (primary key), and `einkommen_m` (target). This does not seem obvious to users and needs special treatment on the implementation side.
[...]

PR #721, opened 2024-04-18
Optionally skip vectorization of a policy function

lars-reimann:

Related to #708

Add a new parameter `is_vectorized` to the `policy_info` decorator to indicate that a function is already vectorized. In that case, it is not vectorized again.

### Todo

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
+- {gh}`721` Optionally skip vectorization of a policy function ({ghuser}`lars-reimann`).
 - {gh}`720` Combined decorator for policy information ({ghuser}`lars-reimann`).
 - {gh}`700` Data columns overwrite functions regardless of time unit
   ({ghuser}`lars-reimann`).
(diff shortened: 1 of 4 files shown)


hmgaudecker (commented):
Looks good to me, thanks! 

Just not sure about the naming: `is_vectorized` sounds a bit like this would be modified after applying the vectorization step. But I don't think this is what will be happening, right?

Would `skip_vectorization` be clearer? 

I'd also value @MImmesberger's opinion!

MImmesberger:
I think `skip_vectorization` is much better than `is_vectorized`.

However, given that #708 is supposed to add a new (core) functionality to GETTSIM, I wonder whether the purpose of the decorator is sufficiently easy to understand for non-experts. It took me a minute to understand the necessity of the decorator, now I'm wondering how long it would take someone who has never heard of vectorization (e.g. Stata users?).

From this point of view it might make sense to use a term that is less precise from the programmers point of view, but more understandable for the user. I don't have a good name right now, but given that the not vectorized functions return columns instead of scalars `function_output=column` with default `scalar` might do the trick. Given the specified function output, we can infer whether the function should be vectorized or not.

hmgaudecker:
Thanks! 

This will only be used in case of things like joins etc. -- so very confined. We will document it there, so it is easy to find when need arises.

I'd say users who have not seen the term "vectorisation" have no business doing anything other than implementing scalar-level functions :wink:. There is a lot of value in technical precision.

Hence, `skip_vectorization` it shall be! Thanks both!

Issue #704, opened 2024-03-08
ENH: Get rid of input variable `kind`


hmgaudecker:
#### Is your feature request related to a problem?

With the structure from #690, the input variable really stops making sense. But the PR is bloated already, so not implementing there.

Furthermore, it is not clear in which sense it is a child. Bundeskindergeldgesetz? Under 18? ... ? Probably mostly follows old IZA $\Psi$ MOD / SOEP conventions.

#### Describe the solution you would like to see

Just calculate endogenously (age, part of household, parent ids, ...), potentially depending on the context.

MImmesberger:
For ALG2 Regelbedarf (`arbeitsl_geld_2_regelsatz_m_bg` in particular), "erwachsen" refers to the age threshold of 18 years. So there it doesn't relate to BKGG.

hmgaudecker:
Cool, then the check should just be on age there (it might be the case already).

MImmesberger:
Reminder to keep [this comment](https://github.com/iza-institute-of-labor-economics/gettsim/pull/694#discussion_r1525043131) in mind when redesigning the `kind` column (if this is not solved via #708)

hmgaudecker:
In case this is still a problem, we should get rid of it before a release.

PR #722, opened 2024-04-18
Join columns to make target columns available at source level

### LINKED ISSUE #708: ENH: Use pointers to make target columns available at source level

lars-reimann:

Closes #708

Add a `join_numpy` function to make target columns available at the source level. 

The function gets
* Foreign keys,
* Primary keys,
* Targets in the same order as the primary key.

It then maps each foreign key to the associated target. Primary keys must be unique and foreign keys must point to some primary key.

### Todo

### DIFF
--- a/src/_gettsim/shared.py
+++ b/src/_gettsim/shared.py
@@ -3,6 +3,9 @@ import re
 import textwrap
 from collections.abc import Callable
 from datetime import date
+from typing import TypeVar
+
+import numpy as np
 
 from _gettsim.config import SUPPORTED_GROUPINGS
 
@@ -260,3 +263,42 @@ def remove_group_suffix(col):
         out = out.removesuffix(f"_{g}")
 
     return out
+
+
+Key: TypeVar = TypeVar("Key")
+Out: TypeVar = TypeVar("Out")
+
+
+def join_numpy(
+    foreign_key: np.ndarray[Key], primary_key: np.ndarray[Key], target: np.ndarray[Out]
+) -> np.ndarray[Out]:
+    """
+    Given a foreign key, find the corresponding primary key, and return the target at
+    the same index as the primary key.
+
+    Parameters
+    ----------
+    foreign_key : np.ndarray[Key]
+        The foreign keys.
+    primary_key : np.ndarray[Key]
+        The primary keys.
+    target : np.ndarray[Out]
+        The targets in the same order as the primary keys.
+
+    Returns
+    -------
+    np.ndarray[Out]
+        The joined array.
+    """
+    if len(np.unique(primary_key)) != len(primary_key):
+        keys, counts = np.unique(primary_key, return_counts=True)
+        duplicate_foreign_keys = keys[counts > 1]
+        raise ValueError(f"Duplicate primary keys: {duplicate_foreign_keys}")
+
+    try:
+        sorter = np.argsort(primary_key)
+        idx = np.searchsorted(primary_key, foreign_key, sorter=sorter)
+        return target[idx]
+    except IndexError as e:
+        invalid_foreign_keys = foreign_key[~np.isin(foreign_key, primary_key)]
+        raise ValueError(f"Invalid foreign keys: {invalid_foreign_keys}") from e
--- a/src/_gettsim/shared.py
+++ b/src/_gettsim/shared.py
@@ -292,8 +292,8 @@ def join_numpy(
     """
     if len(np.unique(primary_key)) != len(primary_key):
         keys, counts = np.unique(primary_key, return_counts=True)
-        duplicate_foreign_keys = keys[counts > 1]
-        raise ValueError(f"Duplicate primary keys: {duplicate_foreign_keys}")
+        duplicate_primary_keys = keys[counts > 1]
+        raise ValueError(f"Duplicate primary keys: {duplicate_primary_keys}")
 
     try:
         sorter = np.argsort(primary_key)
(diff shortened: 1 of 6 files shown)


src/_gettsim/shared.py
```diff
@@ -260,3 +263,42 @@ def remove_group_suffix(col):
...
+        The targets in the same order as the primary keys.
+
+    Returns
+    -------
+    np.ndarray[Out]
+        The joined array.
+    """
+    if len(np.unique(primary_key)) != len(primary_key):
+        keys, counts = np.unique(primary_key, return_counts=True)
+        duplicate_foreign_keys = keys[counts > 1]
```
hmgaudecker, line 295:
```suggestion
        duplicate_primary_keys = keys[counts > 1]
```
Or am I missing something?

src/_gettsim_tests/test_join.py
```diff
@@ -0,0 +1,45 @@
...
+    foreign_key: np.ndarray[int],
+    primary_key: np.ndarray[int],
+    target: np.ndarray[str],
+    expected: np.ndarray[str],
+):
+    assert np.array_equal(join_numpy(foreign_key, primary_key, target), expected)
+
+
+def test_join_numpy_raises_duplicate_primary_key():
+    with pytest.raises(ValueError):
```
hmgaudecker, line 39: Would be good to check the error message, also to differentiate from subsequent test.

src/_gettsim_tests/test_join.py
```diff
@@ -0,0 +1,45 @@
...
+        join_numpy(np.array([1, 1, 1]), np.array([1, 1, 1]), np.array(["a", "b", "c"]))
+
+
+def test_join_numpy_raises_invalid_foreign_key():
+    with pytest.raises(ValueError):
```
hmgaudecker, line 44: Same same.

src/_gettsim/shared.py
```diff
@@ -3,6 +3,9 @@
 import textwrap
 from collections.abc import Callable
 from datetime import date
+from typing import TypeVar
+
+import numpy as np
```
hmgaudecker, line 8: Unless I am missing something, we can use our internal jax.numpy / numpy version. Code should be the same.

src/_gettsim/shared.py
```diff
@@ -260,3 +263,42 @@ def remove_group_suffix(col):
         out = out.removesuffix(f"_{g}")
 
     return out
+
+
+Key: TypeVar = TypeVar("Key")
+Out: TypeVar = TypeVar("Out")
+
+
+def join_numpy(
```
hmgaudecker, line 272:
```suggestion
def join(
```
In case we can be agnostic about whether we use numpy or jax.numpy

hmgaudecker (commented):
Looks great, just some tiny stuff.

Before merging, however, I'd love to see it in action. Can you implement a solution to (parts of) #704 either as part of this PR or in another one using this branch as the basis? Probably together with @MImmesberger or he can give a hint what would be a good use case.

MImmesberger:
I think the most straightforward application is in function `_unterhaltsvors_anspruch_pro_kind_m`. There, we need to replace `_unterhaltsvorschuss_eink_above_income_threshold_fg`. Instead, we should have a bool based on the parent's (linked via `p_id_kindergeld_empf`) return value of `_unterhaltsvorschuss_eink_above_income_threshold`.

In Lars' terminology:
- target: `_unterhaltsvorschuss_eink_above_income_threshold`
- primary key: `p_id`
- foreign key: `p_id_kindergeld_empf`

src/_gettsim/shared.py
lars-reimann: Doesn't seem to work without changes. If I keep the code as is and use JAX, I get this error:

```txt
>           idx = backend.searchsorted(primary_key, foreign_key, sorter=sorter)
E           ValueError: Non-hashable static arguments are not supported. An error occurred while trying to hash an object of type <class 'jaxlib.xla_extension.ArrayImpl'>, [0 1 2]. The error was:
E           TypeError: unhashable type: 'jaxlib.xla_extension.ArrayImpl'
```

lars-reimann:
> I think the most straightforward application is in function _unterhaltsvors_anspruch_pro_kind_m. There, we need to replace _unterhaltsvorschuss_eink_above_income_threshold_fg. Instead, we should have a bool based on the parent's (linked via p_id_kindergeld_empf) return value of _unterhaltsvorschuss_eink_above_income_threshold.

Thanks, the demo is now included.

lars-reimann:
I'll look into updating the type hint check later.

MImmesberger:
> Thanks, the demo is now included.

Thanks! I just adjusted the column name and removed the obsolete aggregation dict at the top of the file. 

I'm happy with the implementation but I'll let @hmgaudecker approve.

src/_gettsim/shared.py
hmgaudecker: Okay, we should keep that in mind then. We should do some benchmarks, but now that Jax is available on Windows, too, we may want to switch completely.

hmgaudecker (approved):
Looks great to me, thanks!

Issue #724, opened 2024-04-24
BUG: input variable `vermögen_bedürft_bg` uses endogeneous grouping


ChristianZimpelmann:
### Bug description

- we request the input variable `vermögen_bedürft_bg` however this grouping is endogeneous. Also, it is used for `hh`-wide transfers like Wohngeld which seems incorrect if `bg` and `hh` differ. Probably should be `vermögen_bedürft_hh` (which we somehow need to distribute on Bedarfsgemeinschaften) or we request individual-level wealth?

See also #412

hmgaudecker:
Great catch, thanks a lot! 

Despite my comment in #412, I think individual-level is the way to go now. It is ultimately what matters in many cases and anything else would be introducing lots of complexity on the GETTSIM side. 

At some point, we need a data preparation guide... There we could explain and make suggestions like trying different ways for distributing wealth measured at some group level.

Issue #666, opened 2023-11-24
BUG: Scalar instead of list returned if function inputs are parameters only


MImmesberger:
### Bug description

The bug was discovered and discussed in #645 

GETTSIM returns something like

```python
{'ges_rente_arbeitsl_vorzeitig': array(60.)}
```

instead of
```python
{'ges_rente_arbeitsl_vorzeitig': array([60.])}
```

resulting in a pandas `ValueError: If using all scalar values, you must pass an index`. This occurred in #645 when functions had a parameter dict as an input (and nothing more) and is resolved once any other input is added.

Once this issue is resolved, we should remove the fake dependencies marked with `ToDo: Remove fake dependency (geburtsjahr).` in the docstrings.

MImmesberger:
Issue also occurs when testing for the Minijob threshold in #680.

hmgaudecker:
Could it be enough to just work with `skip_vectorization`?

MImmesberger:
No, same error message. I didn't investigate further.

MImmesberger:
This is not solved, we just got rid of the error message.

`ValueError: If using all scalar values, you must pass an index` shows up when we create a DataFrame out of GETTSIMs results tree. We got rid of the error because we switched from column-by-column comparison to comparing multiple columns at once (then, pandas can infer what is a column and what is not). We still get a scalar instead of an array in the functions with the fake inputs, pandas just broadcasts them.

I think it would be confusing to users to return arrays alongside with scalars.

(Context: I changed how we created the DataFrame because the way we did it didn't work when we only have a single expected output column. But then, I noticed that broadcasting now fails...)

hmgaudecker:
Ah, you beat me to it! 

No bug, just a matter of delivering output, in my view.

hmgaudecker:
> Also, the fake input fix works only if we use vectorization_strategy=loop.

We should actually set `vectorization_strategy='not_required'` in all these cases!

PR #729, opened 2024-04-25
Make sure GETTSIM runs over simple synthetic data

### LINKED ISSUE #727: BUG: Make sure GETTSIM runs over simple synthetic data
### Bug description

```
data = create_synthetic_data(2, 2)
policy_params, policy_functions = set_up_policy_environment("2024")
compute_taxes_and_transfers(data, policy_params, policy_functions)
```

throws a few errors.

(will open a PR to fix it in a sec)

ChristianZimpelmann:

Closes #727 

Note:
- `vermögen_bedürft_bg` needs to be adjusted again once we addressed #724


src/_gettsim/transfers/grunds_im_alter.py
```diff
@@ -10,7 +10,7 @@ def grunds_im_alter_m_eg(  # noqa: PLR0913
     unterhaltsvors_m_eg: float,
     grunds_im_alter_eink_m_eg: float,
     erwachsene_alle_rentner_hh: bool,
-    vermögen_bedürft_eg: float,
+    vermögen_bedürft_bg: float,
```
MImmesberger, line 13: Why is `bg` the correct grouping if the function is on `eg` level? Am I missing something?

src/_gettsim/transfers/rente.py
```diff
@@ -1817,8 +1817,10 @@ def anteil_entgeltp_ost(
     Proportion of Entgeltpunkte accumulated in East Germany
 
     """
-
-    out = entgeltp_ost / (entgeltp_west + entgeltp_ost)
+    if entgeltp_west + entgeltp_ost == 0:
+        out = 0
```
MImmesberger, line 1821:
```suggestion
        out = 0.0
```

src/_gettsim/transfers/unterhaltsvors.py
```diff
@@ -64,12 +64,50 @@ def unterhaltsvors_m(
     return out
 
 
+@policy_info(start_date="2023-01-01", name_in_dag="_kindergeld_erstes_kind_m")
+def _kindergeld_erstes_kind_ohne_staffelung_m(
+    kindergeld_params: dict,
+) -> float:
+    """Kindergeld for first child with no heterogeneity by number of children.
```
MImmesberger, line 71:
```suggestion
    """Kindergeld for first child when Kindergeld does not depend on number of children.
```

src/_gettsim/transfers/unterhaltsvors.py
```diff
@@ -64,12 +64,50 @@ def unterhaltsvors_m(
...
+
+    """
+    return kindergeld_params["kindergeld"]
+
+
+@policy_info(end_date="2022-12-31", name_in_dag="_kindergeld_erstes_kind_m")
+def _kindergeld_erstes_kind_gestaffelt_m(
+    kindergeld_params: dict,
+) -> float:
+    """Kindergeld for first child with heterogeneity by number of children.
```
MImmesberger, line 90:
```suggestion
    """Kindergeld for first child when Kindergeld does not depend on number of children.
```

MImmesberger (commented):
Thanks, Christian!

MImmesberger:
FYI: I added fake dependencies to your new functions because they depend only on the params yaml. We discovered problems with this (Issue #666).

src/_gettsim/transfers/rente.py
```diff
@@ -1817,8 +1817,10 @@ def anteil_entgeltp_ost(
     Proportion of Entgeltpunkte accumulated in East Germany
 
     """
-
-    out = entgeltp_ost / (entgeltp_west + entgeltp_ost)
+    if entgeltp_west + entgeltp_ost == 0:
```
hmgaudecker, line 1820:
```suggestion
    if entgeltp_west == entgeltp_ost == 0.0:
```
Makes clearer what is meant.

src/_gettsim/transfers/grunds_im_alter.py
MImmesberger: Ah sorry just saw your note. In this case, why don't we keep `vermögen_bedürft_eg` here and set `vermögen` as an individual-level input var and aggregate it to `eg`?

ChristianZimpelmann: You are right! That seems the better option and shouldn't be too much hassle. Will do

hmgaudecker: Yeah, I think this will just fix #724 along the way. Thanks! Just to be sure, the individual-level input should be `vermögen_bedürft`, at least for now (at some point we could think of putting it together from individual pieces).

ChristianZimpelmann: I just opened #732 which introduces that change.

MImmesberger:
Just added some tests that cover the new changes. From my POV, we can merge this one.

hmgaudecker:
Great, go ahead!

hmgaudecker (approved):

PR #730, opened 2024-04-25
Add argument to create_synthetic_data on whether adults are married.

ChristianZimpelmann:

Add argument to create_synthetic_data on whether adults are married.

### DIFF
--- a/src/_gettsim/synthetic.py
+++ b/src/_gettsim/synthetic.py
@@ -11,9 +11,10 @@ from _gettsim.policy_environment import _load_parameter_group_from_yaml
 current_year = datetime.datetime.today().year
 
 
-def create_synthetic_data(
+def create_synthetic_data(  # noqa: PLR0913
     n_adults=1,
     n_children=0,
+    adults_married=True,
     specs_constant_over_households=None,
     specs_heterogeneous=None,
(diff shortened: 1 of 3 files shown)


src/_gettsim/synthetic.py
```diff
@@ -287,8 +306,6 @@ def create_constant_across_households_variables(df, n_adults, n_children, policy
             bg_daten["bruttokaltmiete"][hh_typ_string_lookup]
         ),
         "heizkosten_m_hh": float(bg_daten["heizkosten"][hh_typ_string_lookup]),
-        "kind_unterh_anspr_m": 0.0,
```
MImmesberger, line 290: Why did you remove this and the next line?

MImmesberger (commented):
Thanks! Just one question.

src/_gettsim/synthetic.py
ChristianZimpelmann: all float input variables which are not defined in this dictionnary are set to 0.0 anyway below. So no need to include those two here.

MImmesberger: Ah thanks, didn't know that!

MImmesberger (approved):

PR #732, opened 2024-04-25
Adjust wealth input to individual level.

### LINKED ISSUE #724: BUG: input variable `vermögen_bedürft_bg` uses endogeneous grouping

ChristianZimpelmann:
Closes #724


docs/gettsim_objects/means_testing.md
```diff
@@ -4,14 +4,19 @@
...
 
-'vermögen_bedürft_bg'. Despite small differences over the transfers, we decided, for
-now, to require only one wealth variable as input and use it for all transfers.
+'vermögen_bedürft'. Despite small differences over the transfers, we decided, for now,
+to require only one wealth variable as input and use it for all transfers.
+
+Note that GETTSIM requests wealth on the individual level as input. This variable is
+then aggregated on the respective grouping level for each transfer. In most data sets
+wealth is only measured on the household level and, hence, needs to be split up on the
+individual level before feeding it into GETTSIM.
```
hmgaudecker, line 19: Great. Please add that this is due to the fact that transfers are calculated at the sub-household level (Bedarfsgemeinschaften), which is endogenous. And that if in doubt, robustness checks regarding different distributions of wealth across household members might be in order.

src/_gettsim_tests/test_data/benefit_checks/2006/hh_id_4.yaml
hmgaudecker, line 1: I think I would prefer to revert the changes in `benefit_checks`. These are not about getting the bedarfsgemeinschaft right, purely about testing what happens once we have wealth at that level. We do use "intermediate" variables all the time for that and should continue to do so here IMO.

hmgaudecker (commented):
Excellent, thanks!!! Quick reactions after a glance, won't have time until the afternoon and thus leaving it to @MImmesberger to approve & merge!

hmgaudecker:
Closes #412

src/_gettsim_tests/test_data/benefit_checks/2006/hh_id_4.yaml
ChristianZimpelmann: As from this PR on, Wohngeld and ALG 2 use different wealth variables (on `hh` and `eg` level) we need to adjust the files anyhow. The alternative would be to include both `vermögen_bedürft_eg` and `vermögen_bedürft_hh`. Additionally, Kinderzuschlag will most likely require wealth on the individual level anyhow in the future (see #412).

I would rather stick to the implemented changes.

hmgaudecker: Fine, I did not realise the tests were doing all checks at the same time. Thanks!

MImmesberger (approved):
Looks good to me, thanks!

PR #739, opened 2024-04-30
Unterhaltsvorschuss calculation on child level.

### LINKED ISSUE #735: BUG: Unterhaltsvorschuss and Unterhalt needed on child level
### Bug description

- `unterhaltsvors_m` used to be calculated on the child level (for all parents the variable was zero) (it wasn't well documented that way I guess)
- It was changed to be calculated on the parent level (the variable is now zero for all children)
- This leads to errors in the calculation of transfers (Wohngeld and Kinderzuschlag) in which Unterhaltsvorschuss is deducted on the child-level as child income (which is handled differently to parent income)
- On a similar note, it is currently not clearly described whether the input variable `kind_unterh_erhalt_m` is expected to be specified on the child or parent level. The Unterhaltsvorschuss module seems to imply that we expect it on the parent level which again doesn't fit to how it is used for the calculation of transfers

Was there an important reason for the change? 

- If yes, we need to rewrite the transfer modules or rather create additional child-level variables which we can create there.
- Otherwise, I would argue in favor of reverting it back to child level.
- Either way, we most likely need to know which child received how much Unterhalt, the parent level is not enough.

Note:
- Slightly related to #357

MImmesberger:
Closes #735

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
+- {gh}`722` Join columns to make target columns available at source level
+  ({ghuser}`lars-reimann`).
 - {gh}`732` Change wealth input variable to individual-level (`vermögen_bedürft`)
   ({ghuser}`ChristianZimpelmann`).
 - {gh}`730` Add argument to create_synthetic_data on whether adults are married.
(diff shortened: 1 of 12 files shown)


MImmesberger:
Tests fail due to #740.

ChristianZimpelmann (approved):
Looks good! Thanks!

Can be merged with failing tests being skipped or we can also just wait until #740 is fixed.

PR #741, opened 2024-04-30
fix: fall back to default value if foreign key of `join_numpy` is unresolved

### LINKED ISSUE #740: BUG: `join_numpy` return if foreign key is < 0
In the Unterhaltsvorschuss module, we use the `join_numpy` function to determine whether the parent that receives Kindergeld for a specific child is a single parent:

```python
def parent_alleinerz(
    p_id_kindergeld_empf: np.ndarray[int],
    p_id: np.ndarray[int],
    alleinerz: np.ndarray[bool],
):
    return join_numpy(p_id_kindergeld_empf, p_id, alleinerz)
```

This returns `True` if `p_id_kindergeld_empf` is not determined, i.e. set to `-1`. `True` might be a bad default in this case.

#### Proposed solution

Allow for a default in the `join_numpy` function that is returned if the foreign key is not determined.

lars-reimann:

Closes #740

`join_numpy` now takes a value that is used if a foreign key cannot be resolved, e.g. because it is explicitly set to `-1`.

### Todo

- [ ] ~~Document PR in CHANGES.md.~~

### DIFF
--- a/src/_gettsim/shared.py
+++ b/src/_gettsim/shared.py
@@ -303,7 +303,7 @@ def join_numpy(
     invalid_foreign_keys = foreign_key[
         (foreign_key >= 0) & (~np.isin(foreign_key, primary_key))
     ]
-    print(invalid_foreign_keys)
+
     if len(invalid_foreign_keys) > 0:
         raise ValueError(f"Invalid foreign keys: {invalid_foreign_keys}")
 
--- a/src/_gettsim/shared.py
+++ b/src/_gettsim/shared.py
@@ -300,6 +300,13 @@ def join_numpy(
         duplicate_primary_keys = keys[counts > 1]
         raise ValueError(f"Duplicate primary keys: {duplicate_primary_keys}")
 
+    invalid_foreign_keys = foreign_key[
+        (foreign_key >= 0) & (~np.isin(foreign_key, primary_key))
+    ]
+    print(invalid_foreign_keys)
+    if len(invalid_foreign_keys) > 0:
+        raise ValueError(f"Invalid foreign keys: {invalid_foreign_keys}")
+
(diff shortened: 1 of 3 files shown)


lars-reimann:
Currently, a default value must always be specified. Would that suffice for now?

src/_gettsim/shared.py
```diff
@@ -295,10 +300,29 @@ def join_numpy(
...
-        sorter = np.argsort(primary_key)
-        idx = np.searchsorted(primary_key, foreign_key, sorter=sorter)
-        return target[idx]
-    except IndexError as e:
-        invalid_foreign_keys = foreign_key[~np.isin(foreign_key, primary_key)]
-        raise ValueError(f"Invalid foreign keys: {invalid_foreign_keys}") from e
+    invalid_foreign_keys = foreign_key[
+        (foreign_key >= 0) & (~np.isin(foreign_key, primary_key))
+    ]
+    print(invalid_foreign_keys)
```
MImmesberger, line 306: That's probably not intentional, right?

MImmesberger (commented):
Thanks, that was fast!
I tried it out on the branch of #739 and it works as expected. Regarding the naming: I think `value_if_foreign_key_is_missing` makes it clearer what is happening for the average user than `value_for_unresolved_foreign_key`.

src/_gettsim/shared.py
hmgaudecker: For sure not. But I hope pre-commit will complain, anyhow ;-)

hmgaudecker (approved):
Great, thanks! Approval conditional on fixed tests, ofc. (feel free to merge if they'll be fixed immediately in another PR).

PR #744, opened 2024-05-02
Kindergeldübertrag

MImmesberger:
### What problem do you want to solve?

Closes #742 

The implementation will still rely on manually removing children from their Bedarfsgemeinschaft (via `eigenbedarf_gedeckt`) until the corresponding issue #622 has been addressed.

PR #725, opened 2024-04-25
Several small fixes related to groupings

ChristianZimpelmann:
### What problem do you want to solve?

Closes #745 

Several small fixes related to groupings.

Leave open for now in case more comes up.

### Todo

- [x] simplify use of `bruttokaltmiete_m_bg` 
- [x] fix bug in docstring
- [x] add `unterhaltsvors_m` to default targets instead of `unterhaltsvors_m_hh` (let the user decide on which unit to aggregate)
- [x] shorten very long name of test which contains duplicate and is too long for Windows
- [x] Document PR in CHANGES.md.

Note:
- for now, adjusted how Unterhaltsvorschuss is deducted from Wohngeld (see the discussion in #357). However, I am still unclear what is correct here.

PR #746, opened 2024-05-02
Split Mietkosten up by Bedarfsgemeinschaft

### LINKED ISSUE #745: BUG: Split up Mietkosten across different Bedarfsgemeinschaften
### Bug description

To compute the Bürgergeld Regelsatz for households with multiple Bedarfsgemeinschaften, we need to split up Mietkosten (warm/kalt) as they are defined on the household, not Bedarfsgemeinschaften level.

### Proposed implementation

Mietkosten are split up according to the relative Bedarf of the Bedarfsgemeinschaft in the ExMin-Bericht (don't have a source for this).

### Alternative solution

Mietkosten are split up according to the relative size of the Bedarfsgemeinschaft (e.g. if 60% of people in the household are part of the BG, 60% of Miete are borne by the BG). This is implemented by #694 but seems to be wrong.

MImmesberger:
Closes #745 

This PR is necessary to finish #744.

### DIFF
--- a/src/_gettsim/demographic_vars.py
+++ b/src/_gettsim/demographic_vars.py
@@ -5,10 +5,12 @@ These information are used throughout modules of gettsim.
 """
 
 import datetime
+from collections import Counter
 
 import numpy
 
 from _gettsim.config import SUPPORTED_GROUPINGS
+from _gettsim.shared import join_numpy, policy_info
 
 
 def _add_grouping_suffixes_to_keys(group_dict: dict[str, dict]) -> dict[str, dict]:
@@ -330,3 +332,30 @@ def birthdate_decimal(
     out = geburtsjahr + (geburtsmonat - 1) / 12
 
     return out
+
+
+@policy_info(skip_vectorization=True)
+def anz_paare_hh(
+    p_id_einstandspartner: int,
+    hh_id: int,
+    p_id: int,
+):
+    """Calculate the number of couples in a household."""
+    hh_id_einstandspartner = join_numpy(
+        p_id_einstandspartner, p_id, hh_id, value_if_foreign_key_is_missing=-1
+    )
+
+    # Create bools for whether the hh_id and the einstandspartner's hh_id match
+    matching_hh_id = [
+        foreign_id == primary_id
+        for foreign_id, primary_id in zip(hh_id_einstandspartner, hh_id)
+    ]
+
+    # Zip the ids and values together
+    zipped_data = zip(hh_id, matching_hh_id)
+
+    # Create a Counter object to count occurrences of each ID with a True value
+    counter = Counter(id_ for id_, value in zipped_data if value)
+
+    # Generate the output list by looking up counts for each id
+    return [counter[id_] // 2 for id_ in hh_id]
--- a/src/_gettsim/transfers/kinderzuschl/kost_unterk.py
+++ b/src/_gettsim/transfers/kinderzuschl/kost_unterk.py
@@ -1,8 +1,7 @@
 def kinderzuschl_kost_unterk_m_bg(
     _kinderzuschl_wohnbedarf_eltern_anteil_bg: float,
-    bruttokaltmiete_m_hh: float,
-    heizkosten_m_hh: float,
-    _anteil_personen_in_haushalt_bg: float,
+    bruttokaltmiete_m_bg: float,
+    heizkosten_m_bg: float,
 ) -> float:
     """Calculate costs of living eligible to claim.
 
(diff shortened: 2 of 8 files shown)


src/_gettsim/demographic_vars.py
```diff
@@ -330,3 +332,30 @@ def birthdate_decimal(
     out = geburtsjahr + (geburtsmonat - 1) / 12
 
     return out
+
+
+@policy_info(skip_vectorization=True)
+def anz_paare_hh(
```
MImmesberger, line 338: This is a quick and dirty solution for the problem. We may want to generalize this, but I couldn't come up with other targets where this logic could be relevant.

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
```diff
@@ -21,71 +20,131 @@ def kinderzuschl_kost_unterk_m_bg(
...
+        bedarf_bg = (
+            anz_erwachsene_bg
+            * kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["single"]
+            + anz_kinder_bis_17_bg
+            * kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["kinder"]
+        )
+    else:
+        bedarf_bg = (
+            kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["paare"]
+            + anz_kinder_bis_17_bg
```
MImmesberger, line 83: Just discovered one weakness and wanted to point that out: This produces wrong results if there is a (married) couple and one of the has `kind=True` (and is below 18). IIRC we want to endogenize the `kind` column so I guess this is fine?

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
```diff
@@ -21,71 +20,131 @@ def kinderzuschl_kost_unterk_m_bg(
...
+    )
 
+    if anz_erwachsene_bg == 1:
+        bedarf_bg = (
+            anz_erwachsene_bg
+            * kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["single"]
+            + anz_kinder_bis_17_bg
+            * kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["kinder"]
+        )
+    else:
```
ChristianZimpelmann, line 80: This doesn't cover the case in which bg consists of only a child.

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
```diff
@@ -21,71 +20,131 @@ def kinderzuschl_kost_unterk_m_bg(
...
-    out = _kinderzuschl_wohnbedarf_eltern_anteil_bg * anteil_warmmiete_m_bg
+    return (
+        bruttokaltmiete_m_bg + heizkosten_m_bg
+    ) * _kinderzuschl_wohnbedarf_eltern_anteil_bg
 
-    return out
 
-
-def bruttokaltmiete_m_bg(
+def bruttokaltmiete_m_bg(  # noqa: PLR0913
```
ChristianZimpelmann, line 29: Wouldn't a simpler solution be:
- calculate `bedarf_kosten_der_unterkunft_m` on the individual level (if part of a couple in the same hh, divide `kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["paare"]` by two
- use aggregates of this variable on the hh and bg level to determine `bruttokaltmiete_m_bg` etc.

Then we wouldn't need the variable `anz_paare_hh`

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
```diff
@@ -21,71 +20,131 @@ def kinderzuschl_kost_unterk_m_bg(
...
+            kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["paare"]
+            + anz_kinder_bis_17_bg
+            * kinderzuschl_params["existenzminimum"]["kosten_der_unterkunft"]["kinder"]
+        )
 
-def heizkosten_m_bg(
+    return bruttokaltmiete_m_hh * (bedarf_bg / bedarf_hh)
+
+
+def heizkosten_m_bg(  # noqa: PLR0913
```
ChristianZimpelmann, line 90: For `kinderzuschlag`, we use the same factor for bruttokaltmiete and heizkosten which is based on the sum of `ex_min["kosten_der_unterkunft"]` and `ex_min["heizkosten"]`. It shouldn't matter much, but we should have an issue on that to check what is the correct way.

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
```diff
@@ -21,71 +20,131 @@ def kinderzuschl_kost_unterk_m_bg(
...
     bruttokaltmiete_m_hh: float,
-    _anteil_personen_in_haushalt_bg: float,
+    anz_kinder_bis_17_bg: int,
+    anz_kinder_bis_17_hh: int,
+    anz_erwachsene_bg: int,
+    anz_erwachsene_hh: int,
+    kinderzuschl_params: dict,
 ) -> float:
-    """Share of household's monthly rent attributed to the Bedarfsgemeinschaft.
+    """Monthly rent attributed to the Bedarfsgemeinschaft.
```
ChristianZimpelmann, line 38: Do we have any legal reference that it is done that way? Should reference it here.

ChristianZimpelmann (commented):
A few comments.

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
MImmesberger: Yes, maybe that's more useful in case we want to compute `bruttokaltmiete_m_eg` at some point.

We still need to determine whether a couple lives together in the same household, but we don't need the aggregation step of `anz_paare_hh`.

ChristianZimpelmann: yes, we probably need a variable `einstandspartner_in_household` or so which is non-standard. Not sure how difficult that is to generate.

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
ChristianZimpelmann: I think that should never be the case as `kind` refers to a dependent child

src/_gettsim/transfers/kinderzuschl/kost_unterk.py
MImmesberger: Shouldn't be a problem because this is part of `anz_paare_hh`

MImmesberger:
I checked the legal background again and it seems like Mietkosten are split up simply by the number of people in BG in comparison to the number of people in the household. (I probably have interpreted too much into @mjbloemer comment which was only referring to KiZ)

There is no great source for this, but one [ruling by the BSG](https://datenbank.nwb.de/Dokument/626096/) and page 7 of [this document](https://harald-thome.de/files/pdf/redakteur/KdU_Ordner/KdU%20Brandenburg%20Stadt%20-%2001.01.21.pdf) suggest this.

Closing this PR and continuing with Christian's implementation in #725.

PR #747, opened 2024-05-02
Move ALG 2 Mietkosten from hh to bg level

MImmesberger:

ALG2 Mietkosten are currently considered on the `hh` level. This PR changes this to the `bg` level. Additionally, I specify `bewohnt_eigentum` as a individual level variable that gets aggregated to the `hh` or `bg`level.


src/_gettsim_tests/test_data/grunds_im_alter/2022-07-01/hh_id_25.yaml
```diff
@@ -122,6 +122,12 @@ inputs:
       - 2006
   assumed: {}
 outputs:
+  bruttokaltmiete_m_bg:
```
MImmesberger, line 125: This test currently fails because `bruttokaltmiete_m_bg` and `heizkosten_m_bg` is wrong. This happens because the adult in this household is married to a child ( `kind=True`) below 18. (I pointed this bug out in #746)

I'm not sure what this test is trying to achieve. I'd vote for deleting it or skipping it until `kind` is endogenous.

ChristianZimpelmann: I would also delete the test. This person shouldn't count as `"dependent child living with parents"`

docs/gettsim_objects/input_variables.md
```diff
@@ -231,9 +231,9 @@ Size of household dwelling in square meters
 
 Type: float
 
-(bewohnt_eigentum_hh)=
+(bewohnt_eigentum)=
 
-## `bewohnt_eigentum_hh`
+## `bewohnt_eigentum`
```
ChristianZimpelmann, line 236: I would have thought that this is always a hh-wide variable (i.e. if somebody in the hh owns the apartment, it would count as owner occupied for everybody - even if e.g. the children do not own it themselves)

- But you are thinking about a case in which one adult in the hh owns the apartment and another adult (most likely in another bg) pays rent to him/her?
- What do we expect the user to enter here for children?

ChristianZimpelmann (commented):
nice!

src/_gettsim_tests/test_data/grunds_im_alter/2022-07-01/hh_id_25.yaml
hmgaudecker: Originally, the test was an adult + his/her kid:

```
git log src/_gettsim_tests/test_data/grunds_im_alter/2022-07-01/hh_id_25.yaml
git checkout c56937dc95808927db129ea4a24691e737019e12 -- src/_gettsim_tests/test_data/grunds_im_alter/2022-07-01/hh_id_25.yaml
```

And then the conversion to the pointer structure happened and we ended up with a couple. 

Just revert to the original test.

docs/gettsim_objects/input_variables.md
hmgaudecker: Tricky indeed. 

1. The typical case will be that there are 3 generations in one household, forming two `bg`en (1. grandparents, 2. parents + their kids).

   Say the grandparents own the house and there are no formal rent payments by the middle generation, let alone the kids. 

   Can we find out how that would work? E.g. 
   1. rent payments are imputed for the second `bg`
   2. interest payments + maintenance costs are distributed across all household members
   3. something else is being done

2. The other case would be when there are formal rent payments involved, but I think we can ignore that.

If 1.ii happens to be the case (what I would sort of hope), we can stick with the `hh` level, I think.

hmgaudecker (commented):
Very cool, thanks! 

Depending on what seems best to move this along, we can ignore the case with owner-occupied housing for now (but then leave it at the hh-level)

docs/gettsim_objects/input_variables.md
MImmesberger: Okay, then let's stick to `bewohnt_eigentum_hh` for now because the short-term goal is to implement the Kindergeldübertrag. I'll create an issue that addresses this.

MImmesberger:
Now that #746 will be continued in #725, we can move this implementation here to #725 as well (Christian already started the changes that I did here).

Closing this in favour of #725.

Issue #696, opened 2024-03-03
BUG: Several bugs in Grundsicherung im Alter


MImmesberger:
### Bug description

1. In the code, there is the following ToDo: 

> Grundsicherung im Alter is only paid if all adults in the household are retired. Other households get ALG 2 (This is a simplification by GETTSIM)

2. The wealth exemption differs for children depending on whether the individuals in the Einstandsgemeinschaft have a Kindergeld claim. [Source](https://www.lebenshilfe.de/informieren/arbeiten/grundsicherung-im-alter-und-bei-erwerbsminderung#:~:text=Seit%2001.01.2023%20gilt%3A,Höhe%20von%2010.000%20Euro%20berufen.)
3. The wealth exemption is 10,000 € per individuals since 2023.

MImmesberger:
Also, src/_gettsim_tests/test_data/grunds_im_alter/2022-07-01/skip_hh_id_25.yaml has a child and retiree in the same household. Because the condition to receive the transfer is `erwachsene_alle_rentner_hh=True`, the child has a claim on Grundsicherung im Alter.

Issue #622, opened 2023-07-25
ENH: eigenbedarf_gedeckt


hmgaudecker:
The tests for generation of `bg_id` in #601 set the variable `eigenbedarf_gedeckt`, which determines whether children continue to be part of a Bedarfsgemeinschaft or not. The variable should be calculated from comparing income/other transfers on the one hand  and levels of needs / costs of dwelling on the other hand, though.

![dag](https://github.com/iza-institute-of-labor-economics/gettsim/assets/158456/5b4d92e9-5521-43f1-b72f-9f37f6d7f148)

hmgaudecker:
Closing as this will not be implemented here, see #1010 / #1039.

PR #751, opened 2024-05-04
Kindergeldübertrag

### LINKED ISSUE #742: BUG: Kindergeldübertrag
### Bug description

The current version of GETTSIM doesn't support the Kindergeldübertrag.

The Kindergeldübertrag is relevant whenever children are not part of the Bedarfsgemeinschaft of their parents (e.g. due to enough income through Unterhalt). Then, the Kindergeld that is not necessary to cover the needs of the child are considered as parental income for Bürgergeld.

MImmesberger:

Closes #742 

The implementation will still rely on manually removing children from their Bedarfsgemeinschaft (via eigenbedarf_gedeckt) until the corresponding issue #622 has been addressed.

(This PR is based on #744)


src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
+"""Module for the calculation of the Kindergeldübertrag."""
```
MImmesberger, line 1: For some reason, using the correct Umlaut in the filename results in gibberish and I get a ruff error in the pre-commits (N999)

hmgaudecker: I think we are commonly doing it that way.

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
```diff
@@ -2,7 +2,42 @@
 from _gettsim.shared import policy_info
 
 
-def arbeitsl_geld_2_eink_m(  # noqa: PLR0913
+def arbeitsl_geld_2_eink_m(
+    _arbeitsl_geld_2_nettoeink_ohne_transfers_m: float,
+    kind_unterh_erhalt_m: float,
+    unterhaltsvors_m: float,
+    kindergeld_zur_bedarfdeckung_m: float,
```
hmgaudecker, line 9:
```suggestion
    kindergeld_zur_bedarfsdeckung_m: float,
```
Or did they get rid of the Genitiv in official documents? 😉

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2.py
```diff
@@ -319,15 +317,10 @@ def arbeitsl_geld_2_vor_vorrang_m_bg(  # noqa: PLR0913
...
-    kindergeld_m_bg
-        See :func:`kindergeld_m_bg`.
-    kind_unterh_erhalt_m_bg
-        See basic input variable
-        :ref:`kind_unterh_erhalt_m_bg <kind_unterh_erhalt_m_bg>`.
-    unterhaltsvors_m_bg
-        See :func:`unterhaltsvors_m_bg`.
     arbeitsl_geld_2_eink_m_bg
         See :func:`arbeitsl_geld_2_eink_m_bg`.
+    kindergeld_übertrag_m_bg
```
hmgaudecker, line 322:
```suggestion
    kindergeldübertrag_m_bg
```
One word, no abbreviation = no underscore.

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
```diff
@@ -2,7 +2,42 @@
...
+) -> float:
+    """SGB II income.
+
+    Relevant income according to SGB II. Does not include child benefit transfer
+    (Kindergeldübertrag).
```
hmgaudecker, line 14: Can you explain why? Intuitively, Kindergeldübertrag seems like a component of income to me, so I don't quite understand the structure of not including it there but deducting it from the payments in the end.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
...
+
+import numpy as np
+
+from _gettsim.shared import join_numpy, policy_info
+
+aggregate_by_p_id_kindergeldübertrag = {
+    "kindergeld_übertrag_m": {
+        "p_id_to_aggregate_by": "p_id_kindergeld_empf",
+        "source_col": "_kindergeld_kindbedarf_differenz_m",
```
hmgaudecker, line 10:
```suggestion
        "source_col": "_diff_kindergeld_kindbedarf_m",
```
Easier to read, I'd say?

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
...
+        "aggr": "sum",
+    },
+}
+
+
+def _kindergeld_per_child_m(
```
hmgaudecker, line 16:
```suggestion
def _mean_kindergeld_per_child_m(
```
I think we had that convention? 

And may I suggest separate functions for -2022, 2023+, mirroring the structure we have for Kindergeld?

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
...
+    kindergeld_m
+        See :func:`kindergeld_m`.
+    kindergeld_anz_ansprüche
+        See :func:`kindergeld_anz_ansprüche`.
+
+    Returns
+    -------
+
+    """
+    if kindergeld_anz_ansprüche == 0:
```
hmgaudecker, line 36: Can we ever end up here? Why?

Put differently, this seems like doubling up what is already happening via `join_numpy` below?

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
...
+    which the child is entitled to is the sum of the Kindergeld for all children divided
+    by the amount of children. Hence, the age of the child (in comparison to siblings)
+    does not matter.
+
+    Parameters
+    ----------
+    kindergeld_m
+        See :func:`kindergeld_m`.
+    kindergeld_anz_ansprüche
+        See :func:`kindergeld_anz_ansprüche`.
```
hmgaudecker, line 63: Remove

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
...
+    kindergeld_zur_bedarfdeckung_m: float,
+    eigenbedarf_gedeckt: bool,
+) -> float:
+    """Kindergeld that is used to cover the needs (SGB II) of the parent.
+
+    If a child does not need all of the Kindergeld to cover its own needs (SGB II), the
+    remaining Kindergeld is used to cover the needs of the parent (§ 11 Abs. 1 Satz 5
+    SGB II).
+
+
```
hmgaudecker, line 93: Add something like: `kindergeldübetrag` obtains by aggregating this function at the parent level.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,113 @@
+"""Module for the calculation of the Kindergeldübertrag."""
+
+import numpy as np
+
+from _gettsim.shared import join_numpy, policy_info
+
+aggregate_by_p_id_kindergeldübertrag = {
+    "kindergeld_übertrag_m": {
+        "p_id_to_aggregate_by": "p_id_kindergeld_empf",
```
hmgaudecker, line 9: Just thinking of a weird case here: what if Kindergeldempfänger is outside the BG? E.g. ex-partner. Is that even allowed? Do we have/need a check?

src/_gettsim_tests/test_data/kindergeld/2021/3_children.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - true
+      - true
+      - true
+    bruttolohn_m:
+      - 20000.0
+      - 20000.0
+      - 0.0
+      - 0.0
+      - 0.0
+    _zu_verst_eink_ohne_kinderfreib_y_sn:
```
hmgaudecker, line 55: I thought `sn` were only the parents, so should be zero for the kids?

src/_gettsim_tests/test_data/kindergeld/2021/3_children.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - 200000.0
+  assumed: {}
+outputs:
+  kindergeld_m:
+    - 663.0
+    - 0.0
+    - 0.0
+    - 0.0
+    - 0.0
+  _kindergeld_erstes_kind_m:
```
hmgaudecker, line 69: Why would this be the same for all members of the household?

hmgaudecker (commented):
Great work!

hmgaudecker:
Is that skipped test related to #696? I thought the ALG2 comment there referred to other adults (who could work in principle), but not underage children?

MImmesberger:
Yes, that's the bug we discovered some time ago. The skipped test shows that the `erwachsene_alle_rentner_hh` logic also fails for children because now they have a claim on the transfer themselves (I left a comment in the issue). 

We could fix this with a `and not kind` condition in `grunds_im_alter_m_eg`, or address the `erwachsene_alle_rentner_hh` when we fix #696.

src/_gettsim_tests/test_data/kindergeld/2021/3_children.yaml
MImmesberger: Because this is a parameter that is the same for any individual given the policy date (as, for example, `minijob_grenze`)

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
MImmesberger: I was leaning towards not counting Kindergeldübertrag as income of the parent (but of the child) but now that I'm thinking about it, I cannot defend that point.

Will switch to something like

```python
def arbeitsl_geld_2_eink_m(...):
   return arbeitsl_geld_2_eink_ohne_kindergeldübertrag_m + kindergeldübertrag_m
```

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: Yes because first `_mean_kindergeld_per_child_m` is being calculated (for any individual, so also for those without any children) and then the whole column is used by the `join_numpy`.
That's also what the tests say when removing that condition.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: The concept of Kindergeldübertrag is that the Kindergeldempfänger is in a different BG than the child (and hence there is Kindergeldübertrag). Are you thinking about different `fg`s?

I.e.
| p_id | hh_id | fg_id | bg_id | p_id_kindergeld_empf | p_id_elternteil_1 | p_id_elternteil_2 | kind  |
|------|-------|-------|-------|----------------------|-------------------|-------------------|-------|
| 0    | 0     | 0     | 0     | -1                   | -1                | -1                | false |
| 1    | 0     | 1     | 1     | -1                   | -1                | -1                | false |
| 2    | 0     | 0     | 2     | 1                    | 0                 | 1                 | true  |

I'd argue that this is not possible because then, the child should count towards the `fg` of `p_id==1`. But we don't have a test for that, we don't account for separated couples in the same household when creating `fg_id`.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
hmgaudecker: Cool, then maybe just switch the condition to go for `> 0` first, makes it easier to read (first what you care about, else other stuff)

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
hmgaudecker: Yes, I meant fg's.

Precondition for fg is living in the same household. I just wonder whether it is possible that the other parent gets the Kindergeld. And if not (what I'd kinda expect), whether we check that as a condition on the data.

hmgaudecker:
> Yes, that's the bug we discovered some time ago. The skipped test shows that the `erwachsene_alle_rentner_hh` logic also fails for children because now they have a claim on the transfer themselves (I left a comment in the issue). 
> 
> We could fix this with a `and not kind` condition in `grunds_im_alter_m_eg`, or address the `erwachsene_alle_rentner_hh` when we fix #696.

Cool, pick whatever is easier.

src/_gettsim_tests/test_data/kindergeld/2021/3_children.yaml
hmgaudecker: Is it worth testing that as an output then?

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: I'll create an issue on the creation of `fg_id`. Intuitively, the child should form an `fg` with the parent that gets the Kindergeld (but could be something else of course). It's not clear to me what GETTSIM is currently doing in that regard.

hmgaudecker: I am pretty sure the fg conditions on being part of the same household. My main question is whether Kindergeld does the same when parents are separated.

MImmesberger: Kindergeld does not affect `fg_id` currently

hmgaudecker: Let me rephrase, that was not clear: 

My main question is whether `kindergeldempfänger` conditions that parent being part of the same household when parents are separated — and we are talking about means-tested transfers (ofc parents receive kindergeld for kids at college even though they may not be in the same household).

*(I am just worried that the machinery here depends implicitly on this assumption being true, would like to make it explicit)*

src/_gettsim_tests/test_data/kindergeld/2021/3_children.yaml
MImmesberger: Wanted to test it somewhere, but making it part of an "economic" test case is probably not great. Should I move that to `test_kindergeld.py`?

hmgaudecker: I would be fine with testing it implicitly, but whatever you prefer!

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: Okay so my point was that the creation of `fg_id` in case parents have a child, are separated **but live in the same household** should react to who claims Kindergeld.

Your point is that if parents **live in different households** `p_id_kindergeldempfänger` should be compatible with the place of living of the child (if such a requirement exists, presumably it does) if one of the involved individuals gets transfers?

> (I am just worried that the machinery here depends implicitly on this assumption being true, would like to make it explicit)

We create neither of the variables endogenously. I'm not sure to which extent creating a consistency check here would create a circle in the DAG.

hmgaudecker: > Your point is that if parents **live in different households** `p_id_kindergeldempfänger` should be compatible with the place of living of the child (if such a requirement exists, presumably it does) if one of the involved individuals gets transfers?

Exactly.

> > (I am just worried that the machinery here depends implicitly on this assumption being true, would like to make it explicit)
> 
> We create neither of the variables endogenously. I'm not sure to which extent creating a consistency check here would create a circle in the DAG.

Would be a check pre-DAG,  just whether the data passed conforms to expectations.

MImmesberger: But how do we check pre-DAG whether the involved individuals get transfers?

If I understood it correctly, if all involved individuals don't have their SGB II Bedarf gedeckt, we want to allow for

| p_id | hh_id | fg_id | bg_id | p_id_kindergeld_empfänger | kind  |
|------|-------|-------|-------|---------------------------|-------|
| 0    | 0     | 0     | 0     | -1                        | false |
| 1    | 0     | 0     | 0     | 0                         | true  |
| 2    | 1     | 1     | 1     | -1                        | false |

but we don't want to allow for / it would be inconsistent if

| p_id | hh_id | fg_id | bg_id | p_id_kindergeld_empfänger | kind  |
|------|-------|-------|-------|---------------------------|-------|
| 0    | 0     | 0     | 0     | -1                        | false |
| 1    | 0     | 0     | 0     | 2                         | true  |
| 2    | 1     | 1     | 1     | -1                        | false |

But if all adults have their SGB II Bedarf gedeckt, the latter situation would be perfectly fine.

hmgaudecker: Fine, let's not get carried away here. It may be more about `alleinerziehend` status (or, more precisely, having `unterhaltsanspruch`) and `kindergeld`, basically it is baked in there already (deduction of 1/2 `kindergeld` from the aluminum payment). We should still think of a check for that, but no need to do it here. 

As always, closing an issue by thinking through it opens at least three new ones... 😂

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    return kindergeld_params["kindergeld"] if kindergeld_anz_ansprüche > 0 else 0.0
+
+
+@policy_info(skip_vectorization=True)
+def kindergeld_zur_bedarfsdeckung_m(
+    _mean_kindergeld_per_child_m: float,
+    p_id_kindergeld_empf: np.ndarray[int],
+    p_id: np.ndarray[int],
+) -> float:
```
hmgaudecker, line 73: Returns an array of floats, no?

hmgaudecker (approved):
Excellent! No show-stoppers left AFAICT!

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: I get the following error in the `_convert_data_to_correct_types` when using `numpy.ndarray[float]`:

```text
ValueError: The data types of the following columns are invalid:
E
E            - kindergeld_zur_bedarfsdeckung_m: The internal type numpy.ndarray[float] is not yet supported.
E
```

Other functions in the code use `numpy.ndarray[bool]` as expected return and this works. Any idea @lars-reimann?

lars-reimann: Looks like GETTSIM raises this error itself. Generally, one should never use floating point numbers to represent exact monetary value, for the reasons cited in the error message. But if a close approximation is fine, we could adjust the error.

hmgaudecker: The problem is that we check for the type of an input Series in `convert_series_to_internal_type`, which needs to be `float`, `bool` etc.:

```
 277             internal_type = None                                                                                                                                            
 278             if column_name in TYPES_INPUT_VARIABLES:                                                                                                                        
 279                 internal_type = TYPES_INPUT_VARIABLES[column_name]                                                                                                          
 280             elif (                                                                                                                                                          
 281                 column_name in functions_overridden                                                                                                                         
 282                 and "return" in functions_overridden[column_name].__annotations__                                                                                           
 283             ):                                                                                                                                                              
 284                 internal_type = functions_overridden[column_name].__annotations__["return"]                                                                                 
 285                                                                                                                                                                             
 286             # Make conversion if necessary                                                                                                                                  
 287             if internal_type and not check_series_has_expected_type(series, internal_type):                                                                                 
 288                 try:                                                                                                                                                        
 289                     data[column_name] = convert_series_to_internal_type(                                                                                                    
 290                         series, internal_type                                                                                                                               
 291                     )                                                                                                                                                       
 292                     collected_conversions.append(                                                                                                                           
 293                         f" - {column_name} from {series.dtype} "                                                                                                            
 294                         f"to {internal_type.__name__}"                                                                                                                      
 295                     )                                                                                                                                                       
 296                                                                                                                                                                             
 297                 except ValueError as e:                                                                                                                                     
 298                     breakpoint()                                                                                                                                            
 299  ->                 collected_errors.append(f" - {column_name}: {e}")                                                                                                       
(Pdb++) column_name
'kindergeld_zur_bedarfsdeckung_m'
(Pdb++) internal_type
numpy.ndarray[float]
(Pdb++) check_series_has_expected_type(series, internal_type)
False
(Pdb++) check_series_has_expected_type(series, float)
True
(Pdb++) functions_overridden[column_name].__policy_info__
*** AttributeError: 'function' object has no attribute '__policy_info__'
```

When we use the correct type annotation for a function that is already vectorised, this fails. @lars-reimann, can you think of a sensible solution here? Probably means making a case distinction regarding the definition of `internal_type` on line 284 above, depending on whether the function has the `skip_vectorization` decorator, though I am afraid the last line of my debugging shows we haven't applied it at that point.

> Other functions in the code use `numpy.ndarray[bool]` as expected return and this works.

Most likely we never use them as input columns in tests directly, so `convert_series_to_internal_type` never gets called. Although I have not checked.

@MImmesberger: I think the short-term hack here is to leave the annotation at `float`.

lars-reimann: The attribute added by the decorator is called `__info__`, not `__policy_info__`. Something like this should allow detecting whether the function is vectorized:

https://github.com/iza-institute-of-labor-economics/gettsim/blob/add6232460ac7addbd7f82ff21f0c3823e6e89a0/src/_gettsim_tests/test_docs.py#L139

Issue #713, opened 2024-03-18
BUG: Wohngeld anrechenbares Einkommen


MImmesberger:
### Bug description

To compute housing benefit substractions at the individual level, we currently use Einkommensteuer on the Steuernummer level as an approximation for "Steuern vom Einkommen". (see [§16 WoGG](https://www.gesetze-im-internet.de/wogg/__16.html)). 

#### Problem description

It's not clear what "Steuern vom Einkommen" actually is. Does this refer to Lohnsteuer only or are there other forms of income that are relevant here as well?

hmgaudecker:
For sure other forms of earned income will be relevant, too. I.e., income from self-employement etc.

hmgaudecker:
I'd think we'll be fine if we used:

```python
zu_versteuerndes_eink >= eink_st_params["eink_st_tarif"][0]
```

*(after parsing the piecewise linear structure, that may look different, but you'll get the idea)*

This is at the individual level and it encompasses all income that is relevant for paying taxes. I think this is not even affected by the Steuerklasse (or not much), so we can't do much wrong here.

Issue #710, opened 2024-03-17
BUG: Wohngeld Vorrangprüfung


MImmesberger:
### Bug description

If the individual is eligible for both Wohngeld and ALG2/Bürgergeld, Wohngeld has priority if the transfer is at least as high as the ALG2 transfer.

As an approximation for ALG2 eligibility, we currently compare Wohngeld on the household level with ALG2 on the Bedarfsgemeinschaft level in `wohngeld_vorrang_hh`.

It is, however, possible for some individuals in the household to receive Wohngeld and for some individuals to receive ALG2. The correct priority check (Vorrangprüfung) needs to be done on the individual level. For this, we need to know how the input variables (e.g. Kaltmiete) are broken down from the household to the individual level.

We started to discuss this [here](https://github.com/iza-institute-of-labor-economics/gettsim/issues/690#issuecomment-1975171783).

MImmesberger:
I have found some terminology that might be useful:

- When household = Bedarfsgemeinschaft (e.g. a married couple living together) we are speaking of "reine Wohngeldhaushalte"
- Whenever household > Bedarfsgemeinschaft where some household members receive Wohngeld and some receive ALG2 we have a "Mischhaushalt". The "Mischhaushalt" consists of the Bedarfsgemeinschaft and **exactly one** "wohngeldrechtlicher Teilhaushalt". [Terminology source](https://www.it.nrw/statistik/gesellschaft-und-staat/soziales/wohngeld)

Additionally, [this source](https://www.schleswig-holstein.de/DE/fachinhalte/W/wohnen/Downloads/faq_wohngeld.pdf?__blob=publicationFile&v=1) suggests that Haushaltsmitglieder in the Wohngeld context need to be either related (Parents, children, grandparents, aunt, uncle,...) or need to be an Einstandspartner of one of the family members. If that's not the case, we have two different wohngeldrechtliche Haushalte (even though they may be living in the same apartment).

This suggests for GETTSIM:
- We need to think about what the `hh_id` exactly indicates. If people living in the same apartment should have different `hh_id` (because we think of it in the wohngeldrechtlicher context), we should point this out somewhere).
- Grouping Wohngeld on the `hh` level is wrong. Rather, the grouping should be on a `wohngeldrechtlicher Teilhaushalt` level.

[...]

PR #753, opened 2024-05-07
Wohngeld on individual / BG level

### LINKED ISSUE #752: ENH: Improve current implementation of Vorrangprüfung for Wohngeld
This issue is based on #710. 

Wohngeld has priority over ALG2 if the recipients can cover their needs according to SGB II when receiving Wohngeld. 

The priority check that we have currently implemented follows this logic: 

1. Calculate Wohngeld on the Bedarfsgemeinschaft level. 
3. Check whether the Bedarfsgemeinschaft can cover its own needs (Regelbedarf) with Wohngeld. If not, the Bedarfsgemeinschaft is eligible for ALG2. 
3. Compute Wohngeld again for all individuals in the household that can cover their own needs with Wohngeld. This is the final Wohngeld amount that is paid out to the wohngeldrechtlicher Teilhaushalt.

Note: Because Wohngeld is nonlinear in the number of people in the wohngeldrechtlicher Teilhaushalt, there may be some individuals that pass the priority check, but cannot cover their needs with the Wohngeld calculated in point 3. In this sense, this implementation is an approximation of the actual Wohngeld.

MImmesberger:
Closes #752 

Also, this PR fixes a bug through which Kindergeldübertrag was too high (because Kindergeld was already counted as income when calculating the difference between Regelbedarf and income).


hmgaudecker:
I adjusted the title because the individual-level calculation of things is the core from my perspective, even though the main outputs may be at the bg level then.

MImmesberger:
Not to argue against the renaming but to make sure we're on the same page: I'd say that at the core, it's a BG-level calculation because the Vorrangprüfung happens at the BG level (to do it on the individual level, we would have to split up Regelbedarf or ALG2 on individual level).

hmgaudecker:
All good — nur to my mind that's just one of many calculations, other important ones happen at the level of the individual!

MImmesberger:
The Kinderwohngeld implementation now relies on the `eigenbedarf_gedeckt` bool (it basically turns on and off the consideration of Kinderwohngeld for the calculation of the Kindergeldübertrag.

If we ever make `eigenbedarf_gedeckt` endogenous, we need to calculate Wohngeld/ALG2 in two steps:

1. Calculate Wohngeld and ALG2 for the child, make Vorrangprüfung and calculate, based on this Kindergeldübertrag (based on the result for Wohngeld) and `bg_ id`.
2. Calculate ALG2 and Wohngeld for the parents given Wohngeld/Kindergeldübertrag of the child, and given `bg_id`.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -104,6 +104,7 @@ def kindergeld_zur_bedarfsdeckung_m(
 def _diff_kindergeld_kindbedarf_m(
     _arbeitsl_geld_2_eink_ohne_kindergeldübertrag_m_bg: float,
     arbeitsl_geld_2_regelbedarf_m_bg: float,
+    kinderwohngeld_m: float,
```
ChristianZimpelmann, line 107: why cannot we use `wohngeld_vor_vorrang_check_m` directly here?

As Kinderwohngeld isn't a term in the law (if I understood you correctly) I would try to avoid it.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -657,3 +657,40 @@ def wohngeld_vor_vermög_check_m_hh(
         )
 
     return out
+
+
+def kinderwohngeld_m(
```
ChristianZimpelmann, line 662: Should again probably contain `vor_vorrang_check`

ChristianZimpelmann (commented):
Thanks! Two quick comments.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: The `if kind and eigenbedarf_gedeckt else 0.0` is central here. `eigenbedarf_gedeckt` is our way of saying that the child actually receives Wohngeld (so it anticipates the Vorrangprüfung to not create a cycle in the DAG). If we use `wohngeld_vor_vorrang_check_m` directly when calculating the Kindergeldübertrag, we make errors if the child is not actually eligible for Wohngeld (because Eigenbedarf is not satisfied).

I agree that it would be ideal to avoid modelling KiWoG explitcitly (because it's just Wohngeld) but this does work only if we compute Wohngeld/ALG2 for parents and children separately, as mentioned [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/753#issuecomment-2098457626).

hmgaudecker:
> The Kinderwohngeld implementation now relies on the `eigenbedarf_gedeckt` bool (it basically turns on and off the consideration of Kinderwohngeld for the calculation of the Kindergeldübertrag).
> 
> If we ever make `eigenbedarf_gedeckt` endogenous, we need to calculate Wohngeld/ALG2 in two steps:
> 
>     1. Calculate Wohngeld and ALG2 for the child, make Vorrangprüfung and calculate, based on this Kindergeldübertrag (based on the result for Wohngeld) and `bg_ id`.
> 
>     2. Calculate ALG2 and Wohngeld for the parents given Wohngeld/Kindergeldübertrag of the child, and given `bg_id`.

Sure, I am sorry that this did not become clear all along. Do you see any issue with it?

src/_gettsim/transfers/wohngeld.py
MImmesberger: If anything, it's `nach_vorrang_check`

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
hmgaudecker: > if we compute Wohngeld/ALG2 for parents and children separately

Indeed that's how I had been thinking about it all the time. No other way of getting to `eigenbedarf_gedeckt` and we cannot expect users to supply that (outside of example households, that is).

MImmesberger:
> Sure, I am sorry that this did not become clear all along.

You probably made it clear, but I didn't thought about it deep enough :)

> Do you see any issue with it?

I think this goes along the lines of the discussion we had yesterday on Zoom (trying out different constellations of BGs to arrive at the actual BG/wohngeldrechtl. Teilhaushalt combination). We have to call the current version of GETTSIM several times to achieve exact results.

src/_gettsim/transfers/wohngeld.py
ChristianZimpelmann: Why? it uses the input `wohngeld_vor_vorrang_check_m`? And it represents potential Kinderwohngeld not the Kindergeld that is indeed paid out, right?

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
ChristianZimpelmann: I thought you could just use `wohngeld_vor_vorrang_check_m` as input variable here and then subtract only if it is a kind.  `eigenbedarf_gedeckt` shouldn't matter as out is set to 0 whenever  `eigenbedarf_gedeckt == False` anyway.

But this isn't important in the very short term.

MImmesberger: Ah you're right, we could do that!

src/_gettsim/transfers/wohngeld.py
MImmesberger: Because of `eigenbedarf_gedeckt` the outcome of the Vorrangprüfung is already specified. But given your comment above, we might get rid of this function anyway.

hmgaudecker:
> We have to call the current version of GETTSIM several times to achieve correct results.

That is true. 

> I think this goes along the lines of the discussion we had yesterday on Zoom (trying out different constellations of BGs to arrive at the actual BG/wohngeldrechtl. Teilhaushalt combination). 

To the extent that Wohngeld is linear in the number of people in the entire household (by Wohngeld def), we do not need to try out too many different combinations, right? 

1. Set `bg_id = fg_id` for all households without children or where children have no income.
2. For the remaining households,
    1. Calculate maximum WoG (always individual level, since linear) and BüG while ignoring childrens' income, setting `bg_id = fg_id`. *(This will give the set of potential BüG beneficiaries, we can ignore everyone else)*
    2. Set `bg_id = fg_id` if no eligibility for `BüG` even when children's income is ignored.
    3. For everyone else, create `eigenbedarf_gedeckt` when each child forms his/her individual `bg` *(requires maximum WoG from step 2.1 and BüG at the individual level)*. Create `bg_id` for the remaining households.
3. Stack the data from 1, 2.2, 2.3 and run GETTSIM for all households, using endogenous `bg`.

Probably not complete yet and my model of how this could work was definitely not complete! I think the upshot is there really is no way of representing it in one DAG when the functions have the same names.

If that makes sense, then to achieve the above 6 steps, we can either create a little helper script somewhere or document this. The planned interface change should make this sort of thing very natural.

hmgaudecker:
We'll know soonish, I hope. 

But in any case, I am not super-worried anymore, in the above thing it would mean doing lots of permutations on children in step 2.iii, which presumably will be a small set of people in most applications (colleagues at IAB might form an exception).

And you won't have to run steps 1 and 2 many times in any given application. Will be very natural with a revamped interface.

src/_gettsim/config.py
```diff
@@ -103,6 +103,11 @@ def set_array_backend(backend: str):
         "description": "A couple whose members are deemed to be responsible for each"
         " other.",
     },
+    "wth": {
```
hmgaudecker, line 106:
```suggestion
    "wthh": {
```
We'll hardly need it, better make it resemble the hh a little more.

docs/geps/gep-01.md
```diff
@@ -125,16 +125,22 @@ of aggregation.
...
 - `hh_id`: Haushalt, the relevant unit for Wohngeld. Encompasses more people than the
   Bedarfsgemeinschaft (e.g., possibly more than 2 generations).
+- `wthh_id`: Wohngeldrechtlicher Teilhaushalt, i.e. members of a household for whom the
+  priority check for Wohngeld/ALG2 yields the same result. This unit is based on the
+  priority check via `wohngeld_vorrang_bg` and `wohngeld_kinderzuschl_vorrang_bg`.
 - `fg_id`: Familiengemeinschaft. Maximum of two generations, the relevant unit for
   Bürgergeld / Arbeitslosengeld 2. Another way to think about this is the potential
   Bedarfsgemeinschaft before making checks for whether children have enough income fend
   for themselves. Subset of `hh`.
 - `bg_id`: Bedarfsgemeinschaft, i.e., Familiengemeinschaft plus for whether children
```
hmgaudecker, line 137:
```suggestion
- `bg_id`: Bedarfsgemeinschaft, i.e., Familiengemeinschaft excluding children who
```

docs/geps/gep-01.md
```diff
@@ -125,16 +125,22 @@ of aggregation.
...
   have enough income to fend for themselves. Subset of `fg_id`.
```
hmgaudecker, line 138:
```suggestion
  have enough income to fend for themselves (they will form separate `bg`s). Subset of `fg_id`.
```

src/_gettsim/config.py
```diff
@@ -75,6 +75,12 @@ def set_array_backend(backend: str):
         " unit for Wohngeld.",
         "potentially_endogenous": False,
     },
+    "wthh": {
+        "name": "wohngeldrechtlicher Teilhaushalt",
+        "description": "Members of a household for whom the Wohngeld priority check "
+        "yields the same result.",
```
hmgaudecker, line 81:
```suggestion
        "compared to Bürgergeld yields the same result ∈ {True, False}.",
```

src/_gettsim/groupings.py
```diff
@@ -11,6 +11,7 @@ def create_groupings() -> dict[str, Callable]:
         "eg_id": eg_id_numpy,
         "ehe_id": ehe_id_numpy,
         "sn_id": sn_id_numpy,
+        "wthh_id": wthh_id_numpy,
```
hmgaudecker, line 14: Please use same order as elsewhere.

src/_gettsim/groupings.py
```diff
@@ -211,3 +212,21 @@ def sn_id_numpy(
         next_sn_id += 1
 
     return numpy.asarray(result)
+
+
+def wthh_id_numpy(
```
hmgaudecker, line 217: The coverage error is a fluke, right?

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
```diff
@@ -3,40 +3,22 @@
...
-    """SGB II income without Kindergeldübertrag.
+    """SGB II income.
 
-    Relevant income according to SGB II. Does not include child benefit transfer
+    Relevant income according to SGB II. Includes child benefit transfer
     (Kindergeldübertrag).
 
+    Note: When aggregating this target to the household level, deduct
+    `_diff_kindergeld_kindbedarf_m_hh`. This is necessary because the Kindergeld
+    received by the child may enter `arbeitsl_geld_2_eink_m_hh` two times: once as
```
hmgaudecker, line 19:
```suggestion
    received by the child may enter `arbeitsl_geld_2_eink_m_hh` twice: once as
```

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -137,11 +143,11 @@ def _diff_kindergeld_kindbedarf_m(
     # the full amount for the child and one time the Kindergeldübertrag for the parent -
     # because the child doesn't drop out of Bedarfsgemeinschaft endogenously).
```
hmgaudecker, line 144: But this is happening now, right? Or will that be part of a different PR?

src/_gettsim/transfers/arbeitsl_geld_2/kost_unterk.py
```diff
@@ -29,8 +29,8 @@ def arbeitsl_geld_2_kost_unterk_m_bg_bis_2022(
 
 @policy_info(start_date="2023-01-01", name_in_dag="arbeitsl_geld_2_kost_unterk_m_bg")
 def arbeitsl_geld_2_kost_unterk_m_bg_ab_2023(
-    bruttokaltmiete_m_bg: float,
-    heizkosten_m_bg: float,
+    anteil_bruttokaltmiete_m_bg: float,
```
hmgaudecker, line 32:
```suggestion
    bruttokaltmiete_m_bg: float,
```
Or is there any issue with that? `anteil` sounds too much like a fraction compared to the amount. In the end, this is just a way to split up the `_hh` version; the other way round, aggregation should work as expected even. So just our usual aggregation logic, why give it a different name?

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -1,67 +1,97 @@
...
     -------
 
     """
-    # TODO(@MImmesberger): Allow for some individuals to receive Wohngeld and some
-    # individuals to receive ALG2 in the same household.
-    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/710
-    return wohngeld_nach_vermög_check_m_hh >= arbeitsl_geld_2_vor_vorrang_m_bg
+    return (
+        arbeitsl_geld_2_regelbedarf_m_bg
+        <= wohngeld_nach_vermög_check_m_bg + arbeitsl_geld_2_eink_m_bg
```
hmgaudecker, line 38:
```suggestion
        <= arbeitsl_geld_2_eink_m_bg + wohngeld_nach_vermög_check_m_bg
```

Easier to read.

I'd actually prefer to change the order of the arguments, too:

```py
    arbeitsl_geld_2_regelbedarf_m_bg: float,
    arbeitsl_geld_2_eink_m_bg: float,
    wohngeld_nach_vermög_check_m_bg: float,
```

(first the need, then what the need is compared against)

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -1,67 +1,97 @@
...
+    )
 
 
 def kinderzuschl_vorrang_bg(
```
hmgaudecker, line 42: Same as above

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -1,67 +1,97 @@
...
-    return _kinderzuschl_nach_vermög_check_m_bg >= arbeitsl_geld_2_vor_vorrang_m_bg
+    return (
+        arbeitsl_geld_2_regelbedarf_m_bg
+        <= _kinderzuschl_nach_vermög_check_m_bg + arbeitsl_geld_2_eink_m_bg
+    )
 
 
-def wohngeld_kinderzuschl_vorrang_hh(
-    wohngeld_nach_vermög_check_m_hh: float,
+def wohngeld_kinderzuschl_vorrang_bg(
```
hmgaudecker, line 68: Same as above

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -1,67 +1,97 @@
...
+    arbeitsl_geld_2_regelbedarf_m_bg
+        See :func:`arbeitsl_geld_2_regelbedarf_m_bg`.
+    arbeitsl_geld_2_eink_m_bg
+        See :func:`arbeitsl_geld_2_eink_m_bg`.
 
     Returns
     -------
 
     """
     sum_wohngeld_kinderzuschl = (
```
hmgaudecker, line 91: Just include in the one formula below.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -3,6 +3,14 @@
 from _gettsim.config import numpy_or_jax as np
 from _gettsim.piecewise_functions import piecewise_polynomial
 from _gettsim.shared import policy_info
+from _gettsim.transfers.wohngeld_formulas import (
+    _wohngeld_basisformel,
+    _wohngeld_einkommen_formula,
```
hmgaudecker, line 8:
```suggestion
    _wohngeld_einkommen_formel,
```
if anything.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -3,6 +3,14 @@
 from _gettsim.config import numpy_or_jax as np
 from _gettsim.piecewise_functions import piecewise_polynomial
 from _gettsim.shared import policy_info
+from _gettsim.transfers.wohngeld_formulas import (
```
hmgaudecker, line 6: I am not a fan of this structure. I know this module will grow large, but there is a lot of value in seeing the formula right next to the code. Having these internal functions instead of copying and pasting the code is good, though.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -13,37 +21,55 @@
...
+def wohngeld_m_wthh(
+    _wohngeld_nach_vermög_check_m_wthh: float,
     erwachsene_alle_rentner_hh: bool,
+    wohngeld_kinderzuschl_vorrang_wthh: bool,
+    wohngeld_vorrang_wthh: bool,
 ) -> float:
-    """Calculate final housing benefit on household level.
+    """Calculate housing benefit.
+
+    Wohngeld has priority over ALG2 if the recipients can cover their needs according to
```
hmgaudecker, line 32: This documentation belongs to some meta-document, has nothing to do with this particular function (other than this being step 3, right?)

src/_gettsim/transfers/wohngeld.py
```diff
@@ -13,37 +21,55 @@
...
     Returns
     -------
 
     """
-    if (
-        (not wohngeld_vorrang_hh)
-        and (not wohngeld_kinderzuschl_vorrang_hh)
-        or erwachsene_alle_rentner_hh
+    # TODO (@MImmesberger): Document approximation.
+    # TODO (@MImmesberger): Remove erwachsene_alle_rentner_hh condition once households
```
hmgaudecker, line 63: This confuses me. Isn't the problem rather that we don't have an SGB XII / WoG priority check just yet?

src/_gettsim/transfers/wohngeld_formulas.py
```diff
@@ -0,0 +1,331 @@
+from _gettsim.config import numpy_or_jax as np
+
+
+def _wohngeld_basisformel(
+    anz_personen: int,
+    einkommen_m: float,
+    miete_m: float,
+    params: dict,
+):
```
hmgaudecker, line 9:
```suggestion
) -> float:
```

hmgaudecker (commented):
Just needed for the rain to stop -- looks great! Mostly stylistic comments.

src/_gettsim/transfers/arbeitsl_geld_2/kost_unterk.py
MImmesberger: We want to keep `bruttokaltmiete_m_hh' as the input variable. Defining a function `bruttokaltmiete_m' that splits up Bruttokaltmiete by the number of people in the household works perfectly fine. However, If the user forgets to add any of the two variables as an input, there will be an error message that reports a cycle in the DAG rather than the missing input. 

But I agree, a different word than `anteil’ would be better.

hmgaudecker: I'd say we leave it at `bruttokaltmiete_m` then alongside opening an issue to catch such cycles.

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: That will be part of a different PR. The `eigenbedarf_gedeckt` condition makes sure that parents and children are not in the same BG, but we can throw `eigenbedarf_gedeckt` out and adjust the function such that we don't need it. Will make an issue.

src/_gettsim/transfers/wohngeld.py
MImmesberger: Yes I added it here because this functions yields the "main" target of the module and that way someone who uses the Wohngeld module extensively is likely to stumble over it. But we could also do this in the module docstring if you think that's a better fit?

src/_gettsim/transfers/wohngeld.py
MImmesberger: I'm not sure tbh, just tried to avoid touching that part because I don't know what the actual problem 2 years ago was when that check was introduced. It could be that the problems are resolved now that Wohngeld is calculated on a more granular level than `hh` and we can, thus, compare it to the SGB XII transfer.

src/_gettsim/groupings.py
MImmesberger: Yes, definitely

MImmesberger:
FYI: I'm still writing some tests that cover Mischhaushalte. Currently facing some bugs where GETTSIM and the calculator by BMWSB disagree but will fix that asap.

src/_gettsim/transfers/wohngeld.py
hmgaudecker: Yeah, sounds like a good place to put it.

src/_gettsim/transfers/wohngeld.py
hmgaudecker: Yeah, sounds like a good place to put it.

MImmesberger:
> Currently facing some bugs where GETTSIM and the calculator by BMWSB disagree but will fix that asap.

I noticed that the calculation of Mietkosten was wrong. This should be done on `hh` level and then disaggregated "pro-Kopf". (See for example page 30 [here](https://www.mhkbd.nrw/system/files/media/document/file/faq_wohngeld_2023.pdf))

I added Mischhaushalte to test the module, we get the same results as the BMWSB with its Wohngeld calculator.

I didn't implement #713 but I think we can do this in a different PR because this isn't central to what you're trying to do with the Wohngeld module, correct @ChristianZimpelmann ?

src/_gettsim/transfers/wohngeld.py
ChristianZimpelmann: I think it is still relevant since we haven't implemented any interaction between Grundsicherung im Alter and ALG2/Wohngeld.

In particular, I have no idea what happens if one partner is retired and eligible for Grundsicherung im Alter and the other one is not retired and eligible for ALG2 or eligible for Wohngeld.

hmgaudecker: Yes, this is what I had been trying to say -- the comment just is very opaque in this respect.

ChristianZimpelmann: Yes, I agree.

ChristianZimpelmann:
Looks great! I will try it out today.

> I didn't implement https://github.com/iza-institute-of-labor-economics/gettsim/issues/713 but I think we can do this in a different PR because this isn't central to what you're trying to do with the Wohngeld module, correct @ChristianZimpelmann ?

Yes, this shouldn't be relevant for now.

src/_gettsim/transfers/wohngeld.py
MImmesberger: Adjusted the todo note accordingly.

MImmesberger:
As discussed I added Wohngeld to the relevant income of the child to calculate Kindergeldübertrag. (We don't have official data or calculators for Kindergeldübertrag, that's why didn't catch this one before.)

hmgaudecker (approved):
Looks great to me -- will be away from the machine for a bit, so this is to signal I am very happy with this. Great work!

hmgaudecker:
^ Or did I miss anything in #710 that you don't tackle here?

MImmesberger:
No, we can close that one. But I'd suggest to leave #752 open (because we are not 100% sure that our implementation is correct)

hmgaudecker:
> I'd suggest to leave #752 open (because we are not 100% sure that our implementation is correct)

No. If new things come up, it is important to go through the process of explaining what is wrong. Better open a new issue for that then. There isn't a whole lot of information in that issue, either.

PR #755, opened 2024-05-13
Allow array returns when `skip_vectorization` is being used, enforce import convention.

hmgaudecker:

- Extract the `float` bit from typing info like `numpy.ndarray[float]`, so we can use proper return types.
- Add array types in some cases
- Enforce convention for importing numpy:
  - `np` means our own `numpy_or_jax`
  - `numpy` means numpy proper

### DIFF
--- a/src/_gettsim/interface.py
+++ b/src/_gettsim/interface.py
@@ -281,7 +281,12 @@ def _convert_data_to_correct_types(data, functions_overridden):
             column_name in functions_overridden
             and "return" in functions_overridden[column_name].__annotations__
         ):
-            internal_type = functions_overridden[column_name].__annotations__["return"]
+            func = functions_overridden[column_name]
+            if hasattr(func, "__info__") and func.__info__["skip_vectorization"]:
+                breakpoint()
+                internal_type = func.__annotations__["return"]
+            else:
+                internal_type = func.__annotations__["return"]
 
         # Make conversion if necessary
         if internal_type and not check_series_has_expected_type(series, internal_type):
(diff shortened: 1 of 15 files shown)


src/_gettsim/interface.py
```diff
@@ -384,23 +391,31 @@ class FunctionsAndColumnsOverlapWarning(UserWarning):
...
+            second_part = format_errors_and_warnings(
+                """
+                These are already present among the hard-coded functions of the taxes
+                and transfers system. If you want a data column to be used instead of
+                calculating it within GETTSIM you do not need to do anything. If you
+                want data columns to be calculated by hard-coded functions, remove them
+                from the *data* you pass to GETTSIM. You need to pick one option for
+                each column that appears in the list above.
+                """
+            )
```
lars-reimann, line 417: Is this duplication really worth it for correct pluralization? Always using the plural would not hurt readability of the error IMO.

lars-reimann (approved):
Looks great overall, just one question.

src/_gettsim/interface.py
hmgaudecker: I think so. When newcomers are lost (and this is a very common point to be confused), I think any little bit of showing that devs took care helps. We had it all along, too, just a different formatting now.

MImmesberger (commented):
Shouldn't we add support for the `numpy.ndarray[float]` return type here as well? We encountered that in #751. 

In my commit, I changed the return type from `float` to `numpy.ndarray[float]` for that function. If you don't want to tackle this right now, feel free to revert.

hmgaudecker:
> Shouldn't we add support for the `numpy.ndarray[float]` return type here as well? We encountered that in #751.

Sure, can't believe I missed that one! 

This shows that we'll need to update `convert_series_to_internal_type`, too, however...

hmgaudecker:
Great catch!

hmgaudecker:
Consider it a testimony to the quality of your work that I blindly thought you'd have to be right :smile:

Issue #697, opened 2024-03-06
BUG: Elterngeld eligibility


MImmesberger:
### Bug description

A quick look at `elternzeit_anspruch` makes me question whether eligibility for Elterngeld is correct. In particular, 

```python
m_elterngeld_mut_hh + m_elterngeld_vat_hh
            < elterngeld_params["max_monate_paar"]
```

sounds like both parents are able to take up Elterngeld in scenarios where only one parent can actually do so (e.g. suppose that `elterngeld_params["max_monate_paar"] = 14` and `m_elterngeld_mut_hh + m_elterngeld_vat_hh=13`).

MImmesberger:
Also, `alter_monate_jüngstes_mitglied_hh <= elterngeld_params["max_monate_paar"]` is being used to check whether children aren't too old for the parents to receive Elterngeld. Would be better to target the children directly (via `p_id_elternteil`).

hmgaudecker:
I don't think Elterngeld has ever been touched since the earliest days of GETTSIM. Or definitely does not live up to our quality standards.

Probably worth starting over at some point rather than patching here and there.

MImmesberger:
Closed via #756

PR #756, opened 2024-05-13
Update Elterngeld

### LINKED ISSUE #692: BUG: Add last year's income as an input variable for Elterngeld
### Bug description

As discussed in #690, GETTSIM uses this year's income as an approximation for last year's income when calculating Elterngeld. 

Instead, we should use last year's income as a basic input variable to correctly compute Elterngeld. At the same time, we should have the Elterngeld-relevant income as an output variable to facilitate computations in panel data.

ThOrtmann:
Closes #692, #697, #599

I have introduced some basic changes to better reflect the BEEG. These include a rework of some functions, as well as 1.5 new function

Updating the Functions of the Elterngeld to better reflect the BEEG


MImmesberger:
First of all, thanks @ThOrtmann for providing a good start for refactoring this module! We got rid of some major bugs and this module is much clearer now.

All of the existing tests for the Elterngeld are regression tests. Given the bugs we found, I just replaced their outcomes with the new ones. I also added tests for 2024 but didn't find a good source for test cases. The two Elterngeld calculators that I found do not agree on the numbers and @JakobWegmann told me that the official calculator is not reliable so I did regression tests here as well. However, all GETTSIM results are close to the results of the calculators (about +-20€).

@JakobWegmann, in case you have the time, it would be extremely helpful if you could have a look whether my implementation is in line with your understanding of the Elterngeld. (Doesn't have to be immediately, this is not time-critical)

src/_gettsim/transfers/elterngeld.py
```diff
@@ -1,347 +1,256 @@
...
     -------
 
     """
-    current_year = elterngeld_params["datum"].astype("datetime64[Y]").astype(int) + 1970
-    out = (
-        current_year - geburtsjahr
-        < list(elterngeld_params["geschw_bonus_altersgrenzen_kinder"].keys())[1]
-    )
+    if alleinerz:
+        out = monate_elterngeldbezug_fg <= elterngeld_params["max_monate_ind"]
```
JakobWegmann, line 146: I don't find it in the law, but the [official homepage](https://www.bmfsfj.de/bmfsfj/themen/familie/familienleistungen/elterngeld/elterngeld-73752) claims that alleinerziehend should be elterngeld_params["max_monate_paar"].

src/_gettsim/parameters/elterngeld.yaml
```diff
@@ -90,7 +105,7 @@ eink_schritt_korrektur:
   2007-01-01:
     scalar: 2
     reference: Art. 1 G. v. 05.12.2006 BGBl. I S. 2748
```
JakobWegmann, line 107: Shouldn't it be Art. 2 Absatz 2?

src/_gettsim/transfers/elterngeld.py
```diff
@@ -365,14 +274,14 @@ def elterngeld_anteil_eink_erlass(
...
     ):
         # Replacement rate is only lowered up to a specific value
         out = max(
             elterngeld_params["faktor"]
             - (
-                elterngeld_eink_relev_m
+                elterngeld_nettoeinkommen_vorjahr_m
                 - elterngeld_params["nettoeinkommen_stufen"]["upper_threshold"]
             )
             / elterngeld_params["eink_schritt_korrektur"],
```
JakobWegmann, line 287: Was it a conscious decision to smooth out the function here? Because the decrease should only happen every second Euro, resulting in [this ](https://de.wikipedia.org/wiki/Elterngeld_(Deutschland)#/media/Datei:Elterngeld.png) wonderful functional relationship. (Same applies to the functional form below 1000€)

src/_gettsim/transfers/elterngeld.py
```diff
@@ -384,106 +293,181 @@ def elterngeld_anteil_eink_erlass(
...
+    lohnst_m
+        See :func:`lohnst_m`.
+    soli_st_lohnst_m
+        See :func:`soli_st_lohnst_m`.
+
+    Returns
+    -------
+
+    """
+    return bruttolohn_m - lohnst_m - soli_st_lohnst_m
```
JakobWegmann, line 413: `lohnst_m` should be calculated without taking into account adaptions to the care insurance rate (having no child and being older than 23, living in Saxony). But I guess that is just impossible to incorprate?

"wobei die Höhe der Teilbeträge ohne Berücksichtigung der besonderen Regelungen zur Berechnung der Beiträge nach [§ 55 Absatz 3](https://www.buzer.de/55_SGB_11.htm) und [§ 58 Absatz 3 des Elften Buches Sozialgesetzbuch](https://www.buzer.de/58_SGB_11.htm) bestimmt wird."

docs/gettsim_objects/input_variables.md
```diff
@@ -247,11 +247,35 @@ Weekly working hours of individual
...
+Type: bool
+
+(elterngeld_nettoeinkommen_vorjahr_m)=
+
+## `elterngeld_nettoeinkommen_vorjahr_m`
+
+Approximation of the net wage in the 12 months before birth of youngest child (according
+to simplified calculation rules). You may let GETTSIM compute this variable via the
+`elterngeld_nettolohn_approximation_m` target in a separate run.
+
```
JakobWegmann, line 265: What is the reason that this is not calculated by default by GETTSIM?

src/_gettsim/transfers/elterngeld.py
JakobWegmann, line 146: I don't find it in the law, but the [official homepage](https://www.bmfsfj.de/bmfsfj/themen/familie/familienleistungen/elterngeld/elterngeld-73752) claims that alleinerziehend should be elterngeld_params["max_monate_paar"].

JakobWegmann (commented):
Impressive work! I only have some minor detail questions.

docs/gettsim_objects/input_variables.md
MImmesberger: In principle we have two options when we want to compute targets based on inputs from t-1: i) We compute everything in one GETTSIM call, i.e. the relevant target based on multiple inputs from the past period (wages, self-employed income, transfers, ... from past year / past 12 months), or ii) the user does this herself and calls GETTSIM two times. 

Hans-Martin and I agreed that the latter is much clearer than the former. It increases readability (1 input instead of 10ish, no Lohnsteuer function/module that computes Lohnsteuer just for the "Elterngeld wage" of the past 12 months, ...) and it is closer to reality (e.g. you need the income tax parameters of last year to compute the Elterngeld net wage).

src/_gettsim/parameters/elterngeld.yaml
MImmesberger: Artikel 1 is the "Gesetz zur Änderung des BEEG", §2 Abs. 2 is part of that.

src/_gettsim/transfers/elterngeld.py
MImmesberger: I'm not entirely sure if I'm missing your point, but `elterngeld_params["eink_schritt_korrektur"]` is 2 so dividing by it takes care of that I think?

src/_gettsim/transfers/elterngeld.py
MImmesberger: Oh yes, good point!

src/_gettsim/transfers/elterngeld.py
MImmesberger: Definitely not impossible. We would need to copy the functions and change the inputs of roundabout 10 functions of the Lohnsteuer module.

@hmgaudecker Given the other imprecisions in this module, do you think this is worth it? Maybe we should the new interface can provide a fancy way of handling this? Sounds very similar to the Bedarfsgemeinschaften issue to me.

hmgaudecker: Yes, rather keep it simple here. Just make a note on this!

docs/gettsim_objects/input_variables.md
```diff
@@ -247,11 +247,35 @@ Weekly working hours of individual
...
+Type: bool
+
+(elterngeld_nettoeinkommen_vorjahr_m)=
+
+## `elterngeld_nettoeinkommen_vorjahr_m`
+
+Approximation of the net wage in the 12 months before birth of youngest child (according
+to simplified calculation rules). You may let GETTSIM compute this variable via the
+`elterngeld_nettolohn_approximation_m` target in a separate run.
```
hmgaudecker, line 264:
```suggestion
`elterngeld_nettolohn_approximation_m` target in a separate run, which would typically be for the previous calendar year.
```

docs/gettsim_objects/input_variables.md
hmgaudecker: Yes, but useful to make that explicit. Also add a note somewhere (docstring of `elterngeld_nettolohn_approximation_m`, that for most datasets (i.e., all except for IAB/DRV data), the best approximation will be to just use the previous calendar year earnings, even if the earnings of the 12 months up to birth count for most people (i.e., those who only have income from dependent employment).

src/_gettsim/parameters/elterngeld.yaml
```diff
@@ -152,23 +167,73 @@ sozialv_pausch:
...
   description:
-    de: § 4 (3) BEEG
+    de: Basismonate ohne Partnermonate.
     en: null
+  reference: § 4 (3) BEEG
   2007-01-01:
     scalar: 12
     reference: Art. 1 G. v. 05.12.2006 BGBl. I S. 2748
+max_eink_vorj:
+  'description:':
```
hmgaudecker, line 195:
```suggestion
  description:
```

docs/gettsim_objects/input_variables.md
```diff
@@ -247,11 +247,35 @@ Weekly working hours of individual
...
+
+Type: float
+
+(elterngeld_zu_verst_eink_vor_geburt_sn)=
+
+## `elterngeld_zu_verst_eink_vor_geburt_sn`
```
hmgaudecker, line 270: What is the reason not to use `elterngeld_zu_verst_eink_vorjahr_sn` here? Has exactly the same problems as `elterngeld_nettoeinkommen_vorjahr_m`.

Name is also missing a `_y`, I guess?

src/_gettsim/parameters/elterngeld.yaml
```diff
@@ -152,23 +167,73 @@ sozialv_pausch:
...
+    reference: >-
+      Single: Art. 7 G. v. 27.03.2024 BGBl. 2024 I Nr. 107, Paar: Art. 1 G. v.
+      22.12.2023 BGBl. 2023 I Nr. 412
+max_arbeitsstunden_w:
+  name:
+    de: >-
+      Maximale Wochenstunden die bei Elterngeldbezug gearbeitet werden dürfen
+    en: >-
+      Maximum amount of weekly hours one is allowed to work, when receiving Elterngeld
+  descripition:
```
hmgaudecker, line 231:
```suggestion
  description:
```

src/_gettsim/synthetic.py
```diff
@@ -297,6 +297,7 @@ def create_constant_across_households_variables(df, n_adults, n_children, policy
         "m_schul_ausbild": 10.0,
         "m_kind_berücks_zeit": 24.0,
         "m_pfleg_berücks_zeit": 1.0,
+        "elterngeld_zu_verst_eink_vor_geburt_sn": 1000.0,
```
hmgaudecker, line 300: Why do we need this but not the net income?

src/_gettsim/transfers/elterngeld.py
hmgaudecker: Is this captured by the rounding rules? When `rounding == False`, we want to have a smooth relationship, else the sawtooth pattern (:scream:) from the pic Jakob linked.

hmgaudecker (commented):
Great work!

src/_gettsim/synthetic.py
MImmesberger: Elterngeld was not among the default targets. It is now!

src/_gettsim/transfers/elterngeld.py
MImmesberger: Jakob was of course right, we had only implemented the smoothed version. We now also have the correct calculation in there.

JakobWegmann: Very likely that I'm overlpooking something, but I think there are two things not correct here:
1. `elterngeld_params["prozent_korrektur"]` is missing for incomes larger than `upper_threshold`. I think it should be 

```
        out = max(
            elterngeld_params["faktor"]
            - (
                elterngeld_nettoeinkommen_vorjahr_m
                - elterngeld_params["nettoeinkommen_stufen"]["upper_threshold"]
            )
            / elterngeld_params["eink_schritt_korrektur"] * elterngeld_params["prozent_korrektur"],
            elterngeld_params["prozent_minimum"],
        )
```


2. But even then it would not be exact, as the rounding is only to the next Euro if I understand `_nettoeinkommen_minus_obere_lohnersatzanteil_grenze` correctly. For example with an `elterngeld_nettoeinkommen_vorjahr_m` of 1201€, GETTSIM would caclulate a replacement rate of 0.67-(1201-1200)/2 * 0.001 = 0.6695. But the correct replacement rate would be 0.67 as the replacement rate discontinuously jumps to 0.669 at an income of 1202€ but is 0.667 at an income of 1201€.
So for the exact calculation of the replacement rate `elterngeld_nettoeinkommen_vorjahr_m` would be rounded downwards to the next multiple of 2. (I would love to know why it was designed like that)

JakobWegmann:
One more general question: As parental leave benefit is calculated based on the state of the law at January 1st for the year of the respective claim, changes to the law that take place during the year (in the last years this happens regularly for the lohnsteuer), should not influence the calculation of parental leave benefit. Is this implementable/implemented?

docs/gettsim_objects/input_variables.md
hmgaudecker: It is not necessary. But it will typically be a reasonable approximation, unless you have spell data from the IAB.

Typical surveys will give you annual income in the past calendar year. For a kid born in 2019, you could use 2018 incomes or 2019 incomes. I would think that in the vast majority of cases, the error using 2018 incomes (and parameters) will be smallest because for a kid born in January 2019, you likely won't see any 2019 labor income etc.. So you'd go ahead and calculate `elterngeld_nettolohn_approximation_m` with 2018 data and policy env, then `elterngeld_m` in 2019.

But if you want to be super-precise, you could of course run `elterngeld_nettolohn_approximation_m` using the 2019-01-01 policy environment and 2018 data; then run it another time to calculate `elterngeld_m`.

hmgaudecker:
> One more general question: As parental leave benefit is calculated based on the state of the law at January 1st for the year of the respective claim, changes to the law that take place during the year (in the last years this happens regularly for the lohnsteuer), should not influence the calculation of parental leave benefit. Is this implementable/implemented?

As long as only net income is concerned, see the discussion [above](https://github.com/iza-institute-of-labor-economics/gettsim/pull/756#discussion_r1749264784). Else, we should make an issue for that, link from the docstring and if somebody needs that amount of detail, implement it!

docs/gettsim_objects/input_variables.md
JakobWegmann: Sorry, I still do not understand the core of the argument.

Why is requiring `elterngeld_nettolohn_approximation_m` better than requiring `elterngeld_bruttolohn_reference_period_m` as an input?

src/_gettsim/transfers/elterngeld.py
MImmesberger: Regarding the second point: Rounding should be implemented via the usual rounding mechanic:

```python
rounding:
  _untere_lohnersatzanteil_grenze_minus_nettoeinkommen:
    1980-01-01:
      base: 2
      direction: down
      reference: §2 (2) BEEG
  _nettoeinkommen_minus_obere_lohnersatzanteil_grenze:
    1980-01-01:
      base: 2
      direction: down
      reference: §2 (2) BEEG
```

It is not hard-coded in the function to get a smoothed out version for applications in computational projects.

Regarding the first point: Yes, I think you're right. I didn't spend too much time on the formula, hoping it was implemented correctly in the past...

JakobWegmann: Ah, that's so fancy!

docs/gettsim_objects/input_variables.md
MImmesberger: From my understanding:

1. You wouldn't need only the Bruttolohn but also an array of other variables (i.e. lagged values of all variables necessary to compute Lohnsteuer).
2. Depending on which parameters are necessary, you'd also need historical parameters in the GETTSIM call.

JakobWegmann: > You wouldn't need only the Bruttolohn but also an array of other variables (i.e. lagged values of all variables necessary to compute Lohnsteuer).

Which input variables are you thinking about? For all input variables I can come up with, only the state of the household at the time of birth matters. Withholding tax class is complicated as it is unclear which one is applied to the indvidiual if the household picked a different schedule between interviews.

> Depending on which parameters are necessary, you'd also need historical parameters in the GETTSIM call.

Which ones?

hmgaudecker: I think it would make for an interesting case study (maybe an MSc thesis in conjunction with your co-author, Jakob?) to find out which approximation would work best with a reasonable amount of work. After all, not everybody is paying Lohnsteuer only (at least those PV-things don't cause income tax to be used anymore...) and we would need to think of them, too. So we will never be able to be extremely precise in all use cases and I would guess most of the stuff we are discussing here is about eliminating second-order errors when we might still make (data-driven) first-order errors. But hard to find out by just discussing and without anyone actually crunching the data -- I'd certainly be happy to be proven wrong on this one!

MImmesberger: > For all input variables I can come up with, only the state of the household at the time of birth matters.

I didn't know that! Is that also true for place of residence (east/west) and Kinderfreibeträge (for Soli)?

JakobWegmann: The application form states: "Die entsprechenden Abzugsmerkmale werden den Lohn- oder Gehaltsabrechnungen entnommen. Grundsätzlich sind die Abzugsmerkmale maßgeblich, die in der überwiegenden Zahl der Monate des Bemessungszeitraums vor Geburt des Kindes gegolten haben."

MImmesberger: So we would need an array of historical parameters, right?

JakobWegmann: Only if a change in the parameters happened less than 6 months ago. But I'm convinced that it is good as it is.

hmgaudecker (approved):
Maybe add the discussion on timings / running GETTSIM more than once somewhere so that it does not get lost. Maybe a page in the docs and link from docstrings of `elterngeld_m` and `elterngeld_nettolohn_approximation_m`, description of input variable `elterngeld_nettoeinkommen_vorjahr_m` ?

docs/how_to_guides/calculating_elterngeld.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "sources exist, the basis is the average monthly gross income in the calendar year before\n",
+    "the birth of the child, 2023 in our example. We will stick to the first case here\n",
+    "because:\n",
+    "\n",
+    "> **Note**: The Elterngeld implementation in GETTSIM does currently not consider income\n",
+    "> from self-employment or other sources. See\n",
+    "> [this](https://github.com/iza-institute-of-labor-economics/gettsim/issues/613) issue.\n",
+    "\n",
+    "So let's stick to the case where income from dependent employment is the only source of\n",
+    "income. You can compute this directly if you have monthly information on gross income\n",
```
hmgaudecker, line 48: What is meant by "this"? And to the extent it's gross income, I would not need to calculate anything?

docs/how_to_guides/calculating_elterngeld.ipynb
```diff
@@ -0,0 +1,211 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Step 3: Compute Elterngeld\n",
+    "\n",
+    "Finally, we can compute Elterngeld using the target `elterngeld_m` and the net wage\n",
+    "information from step 2.\n",
+    "\n",
+    "The reason for not doing this in one GETTSIM call (together with step 2) is that \n",
+    "1. the correct policy date of the net wage calculation (step 2) is always January 1st of\n",
```
hmgaudecker, line 122: I thought @JakobWegmann linked a law/directive saying that the parameters throughout the larger part of the year before birth mattered? So for a kid born in March, it would be 2023. For a kid born in November, 2024? Again, I'd argue for keeping our the standard approximation using the previous year, but we should be precise here, also for our own sanity when reading it down the road.

hmgaudecker (commented):
The notebook is a very nice addition! I think, however,  we should be very precise there. Not fully sure whether that is the case just yet!

docs/how_to_guides/calculating_elterngeld.ipynb
MImmesberger: I guess this comes from me being used to DRV data (where you need to compute this based on Entgeltpunkte) 😄 
I'll adjust that.

docs/how_to_guides/calculating_elterngeld.ipynb
MImmesberger: One law, three different interpretations 😅 

They write:

>  Die Abzüge für Steuern werden einheitlich für Einkommen aus nichtselbstständiger und selbstständiger Erwerbstätigkeit auf Grundlage einer Berechnung anhand des am 1. Januar des Kalenderjahres vor der Geburt des Kindes für dieses Jahr geltenden Programmablaufplans [...] ermittelt.

My interpretation is that the relevant parameters are the ones of the year before (i.e. child born in December 2024, parameters of January 1st 2023 are considered). I'll adjust the notebook accordingly, but intervene if you understand the law differently.

hmgaudecker: That would be my interpretation of the text, too! @JakobWegmann, could you weigh in one last time?

pia-molitor: Yes, 1st of January of the calendar year before birth is correct (no matter the birth month).

docs/gettsim_objects/input_variables.md
hmgaudecker: Cool. Let's trust the law and 3 concurrent interpretations of that (including one real expert in @pia-molitor!) and consider this discussion closed for good. Will resolve so that it does not confuse people.

docs/how_to_guides/calculating_elterngeld.ipynb
JakobWegmann: > I thought @JakobWegmann linked a law/directive saying that the parameters throughout the larger part of the year before birth mattered?

For the sake of completeness: That was about the **input values** of the function, so Steuerklasse, place of residence, church tax,...

hmgaudecker: Ahhh. WTF. No wonder they are slow in the Elterngeldstelle.

hmgaudecker: Excellent, if all of you (@MImmesberger @JakobWegmann  and maybe also @pia-molitor ?) could just read through the notebook text and fix any errors in my summary (+maybe add a link to the application form, Jakob?), we'll be good to go!

MImmesberger: No complaints from my side.

JakobWegmann: Looks very good!

A direct link to a PDF: https://www.zbfs.bayern.de/imperia/md/content/zbfs_intranet/produktgruppe_i/formulare/beeg-antrag/2021_09_anlage_n_mit_info_.pdf

I find this part a bit confusing:
> The Lohnsteuerklasse is always set on a yearly basis; if it changes with the start of the calendar year a child is born, the date of birth matters. Up to June, the Lohnsteuerklasse from the previous calendar year will be relevant; for July or later, the Lohnsteuerklasse from the current calendar year will be relevant.

I would simply delete it. Eventually, also the Steuerklasse is a choice of the data preparer.

hmgaudecker: Depends on the dataset, I guess 😄

I don't care either way, your call @mimmesberger, and then go ahead and merge!

MImmesberger: I left it in there because it helped me to understand the purpose of the paragraph. I just added a "for example" to make clear that this reasoning also applies to the other variables.

Issue #707, opened 2024-03-09
ENH: Check whether `eg` and `ehe` groups consist of max 2 individuals


MImmesberger:
#### Is your feature request related to a problem?

As far as I can see, there are no checks whether groups created by `eg_id_numpy` and `ehe_id_numpy` are at most of size 2.

#### Describe the solution you would like to see

GETTSIM should throw an error if the user-provided inputs imply larger groups.

MImmesberger:
Related to #698

Issue #698, opened 2024-03-06
ENH: Consistency check of Ehepartner and Einstandspartner


MImmesberger:
We should check that Ehe- and Einstandspartner specified via `p_id_ehepartner` and `p_id_einstandspartner` are consistent. Two people can be Einstandspartner, but not Ehepartner. The other way around is not possible, i.e. the following is okay:

| `p_id` | `p_id_ehepartner` | `p_id_einstandspartner` |
|:------:|-------------------|:-----------------------:|
|    0   | -1                | 1                       |
|    1   | -1                | 0                       |
|    2   | -1                | -1                      |

But this is not:

| `p_id` | `p_id_ehepartner` | `p_id_einstandspartner` |
|:------:|-------------------|:-----------------------:|
|    0   | 1                | -1                       |
|    1   | 0                | -1                      |
|    2   | -1                | -1                      |

and also not:

| `p_id` | `p_id_ehepartner` | `p_id_einstandspartner` |
|:------:|-------------------|:-----------------------:|
|    0   | 1                | -1                       |
|    1   | 0                | 0                      |
|    2   | -1                | 1                      |

PR #765, opened 2024-06-13
KiZ only for BGs with parents.

### LINKED ISSUE #764: BUG: User input of `bg_id`
It seems like it is not possible for the user to set the `bg_id` themselves. This results in the following error:

```text
E           ValueError: The data types of the following columns are invalid:
E
E            - bg_id: The internal type numpy.ndarray[int] is not yet supported.
E
```

The error occurs as soon as `bg_id` is part of the inputs in the test files. 

The bug would be fixed if we include `bg_id` in `TYPES_INPUT_VARIABLES`, but I'm not sure whether this is intended.

MImmesberger:
A BG is eligible for KiZ only if parents are part of the child's BG and the BG has covered its SGB II needs.

In the current implementation, a child gets KiZ even if it is outside the parental BG (i.e. parents don't have their SGB II needs covered). I changed this such that a BG is only eligible for KiZ if there are more than two persons part of the BG (= parents are in BG).


hmgaudecker (approved):
Looks great, thanks! 

The one thing I wonder about is how many safety checks are reasonable. I.e., if in #766, someone puts two kids in the same BG, this will fail. Should we just make clear notes somewhere? Or add another check that there is at least one person in the BG who does not qualify as a child in the sense of the BKGG?

MImmesberger:
Consistency checks of user-provided grouping IDs would be a good idea. I would say we tackle this in a new PR together with #698 and #707, they are both concerned with checking consistency of user-provided grouping IDs.

hmgaudecker:
> Consistency checks of user-provided grouping IDs would be a good idea. I would say we tackle this in a new PR together with #698 and #707, they are both concerned with checking consistency of user-provided grouping IDs.
> 
> I can adjust the issues accordingly if you agree.

I would just make a new one. When do you ever get a chance for a simple PR closing three issues!

PR #766, opened 2024-06-13
Allow for user provided grouping IDs.

### LINKED ISSUE #764: BUG: User input of `bg_id`

MImmesberger:
Closes #764


src/_gettsim_tests/test_interface.py
```diff
@@ -685,6 +686,24 @@ def test_fail_if_cannot_be_converted_to_internal_type(
...
+            {"bg_id": bg_id_numpy},
+        ),
+        (
+            pd.DataFrame({"wthh_id": [1, 2, 3]}),
+            {"wthh_id": wthh_id_numpy},
+        ),
+    ],
+)
+def test_user_provided_groupings(data, functions_overridden):
+    """Test whether user-provided grouping IDs are admissible."""
```
hmgaudecker, line 703:
```suggestion
    """Test whether GETTSIM handles user-provided grouping IDs."""
```

src/_gettsim_tests/test_interface.py
```diff
@@ -685,6 +686,24 @@ def test_fail_if_cannot_be_converted_to_internal_type(
...
+            pd.DataFrame({"bg_id": [1, 2, 3]}),
+            {"bg_id": bg_id_numpy},
+        ),
+        (
+            pd.DataFrame({"wthh_id": [1, 2, 3]}),
+            {"wthh_id": wthh_id_numpy},
+        ),
+    ],
+)
+def test_user_provided_groupings(data, functions_overridden):
```
hmgaudecker, line 702:
```suggestion
def test_provide_groupings(data, functions_overridden):
```
Else one may think that this does some consistency checks, but it is really only about whether we handle that case at all, right?

hmgaudecker (approved):
Excellent, thanks! 

Just don't forget the CHANGES.md in both PRs.

Issue #763, opened 2024-06-04
BUG: Endogenous creation of `bg_id`


MImmesberger:
The current creation of `bg_id` is wrong because children with `eigenbedarf_gedeckt=True` will always have a different `bg_id` than their parents, even if the parents have `eigenbedarf_gedeckt=True` as well.

### Proposed solution

Parents should have the same `bg_id` as their children if all of them have `eigenbedarf_gedeckt=True`.

hmgaudecker:
Sounds okay as a short-term hack. But we'll have to be careful that
1. both parents always have the same value of `eigenbedarf_gedeckt` 
2. we treat children with `eigenbedarf_gedeckt == False` correctly (thinking of patchwork here).

Ultimately, however, `eigenbedarf_gedeckt` really needs to go -- it is just a synonym for "separate bg's" and with the suggested change, its name is not even related closely  to the content anymore.

I am rather convinced now that we won't be able to solve all of this with one pass of GETTSIM -- unless we find a way of working with temporary/candidate variables in a subgraph.

What I mean is that we'll need to check multiple things to determine the `bg_id`, which depend on the configurations of `bg`'s within a household. So we"ll have to check various candidate configurations (children out? Bürgergeld even relevant? Possible `wthh_id`s?), which determine the payments. Example from going through budget sets of exemplary households:
 
> Mir ist folgendes Problem aufgefallen:
> 
> - an dem Punkt, an dem Haushalt von "Elternteil ALG 2 + Kind WG" wechselt zu "gesamter Haushalt WG",  fällt das Einkommen des Haushaltes ein gutes Stück (obwohl das Bruttoeinkommen steigt)
> - Der Grund ist, dass beim Elternteil ALG 2 für eine Person verglichen wird mit WG für eine Person. Hier ist dann irgendwann WG günstiger (auch schon recht früh durch den Kindergeldübertrag). Ausgezahlt wird das Wohngeld dann aber nicht für zwei Einpersonenhaushalte (wie in der Günstigerprüfung angenommen), sondern das Wohngeld wird für beide Personen berechnet, was weniger ist und das verfügbare Einkommen reduziert.

In the end, this is sort of the same as calculating candidate benefits before checking eligibility / priority. We solve that by giving things names like `xxx_vor_vorrang_check`. But that is ugly and not feasible for this much more complex case. There might be a better way of solving both of these. 

[...]

PR #768, opened 2024-06-14
Create bg_id and wthh_id endogenously.

MImmesberger:

Create the bg_id and wthh_id for households with exactly 1 `fg` endogenously.

Based on the discussion in #763


src/_gettsim/groupings.py
```diff
@@ -230,3 +234,32 @@ def wthh_id_numpy(
...
+        Array of family group IDs.
+    """
+    unique_hh_ids = numpy.unique(hh_id)
+    error_msg = (
+        "There is at least one household with more than one `fg_id`. GETTSIM does "
+        "not support the endogenous creation of Bedarfsgemeinschaften in this case "
+        "yet. Please provide `bg_id` and `wthh_id` yourself."
+    )
+
+    for idx in unique_hh_ids:
```
hmgaudecker, line 260:
```suggestion
    for this_hh_id in unique_hh_ids:
```
Makes it clearer what the loop is over.

src/_gettsim/groupings.py
```diff
@@ -230,3 +234,32 @@ def wthh_id_numpy(
...
+        # Find all family group IDs for the current household ID
+        fg_ids_in_hh = fg_id[hh_id == idx]
+        # Check if all fg_ids are the same in this household
+        if len(numpy.unique(fg_ids_in_hh)) > 1:
+            raise ValueError(error_msg)
```
hmgaudecker, line 265: I think it is okay to fail on the first case here, but it would still be good to provide the faulty `hh_id` to the user in the error message.

src/_gettsim_tests/test_data/groupings/2023/skip_eltern_viele_kinder.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 info:
-  note: ''
+  note: Skipped because multiple fgs in hh.
```
hmgaudecker, line 3: I think this statement is wrong here and for the next few tests, so not repeating. Sometimes there are multiple households indeed, but only one fg per household.

src/_gettsim_tests/test_data/wohngeld/2006/hh_id_15.yaml
```diff
@@ -312,7 +312,16 @@ inputs:
       - true
       - true
       - true
-  assumed: {}
+  assumed:
```
hmgaudecker, line 315: Why not make it the same as in the next few tests, i.e., adding einstandspartner etc? No big deal, but might make people wonder who look at these one after the other.

src/_gettsim_tests/test_data/wohngeld/2023/skip_mischhaushalt_mit_einer_sgb_ii_bg.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 info:
-  note: https://www.bmwsb.bund.de/Webs/BMWSB/DE/themen/wohnen/wohnraumfoerderung/wohngeld/wohngeldrechner-2023-artikel.html
-  source: ''
+  note: Skipped because multiple fgs in hh.
```
hmgaudecker, line 3: Can't we supply the `bg_id` or would that run counter the purpose of the test in the first place? Same for the next test.

hmgaudecker (commented):
Just skimmed through it, great start!

src/_gettsim_tests/test_data/endogenous_bg_id/2023/single_parent_unterhalt_1000_1_child.yaml
hmgaudecker, line 1: Please make 9 tests out of this one and similar ones, else the yaml is illegible. Apologies, I did not mean my suggestion elsewhere to be taken that literally.

MImmesberger: Are you sure? I worked with both versions and as this is mainly about when household members switch BGs / wthh, I think it helps to have households that are similar (=everything the same, only wage is increasing across households) together and ordered.

hmgaudecker: Yes! I think this is a great example of where the current thing makes sense while developing (you are sure about the structure and unsure what is correct), but does not when some test starts to fail in 3 years from now (you'd look at the huge test file like I did a couple of hours ago and it takes you ages to realise the precise structure because that's not what our yaml structure is made for).

So feel free to change only when this is ready for review.

Structure should be clear when you nest one level deeper:

```
endogenous_bg_id/2023/single_parent_unterhalt_1000_1_child/
  - parental_income_0.yaml
  - parental_income_520.yaml
```

etc.

MImmesberger: Ah yes, makes sense!

hmgaudecker: Just realised it is confusing to have the two digits following each other in the directory name. Maybe change to:

```
endogenous_bg_id/2023/single_parent_1_child_unterhalt_1000/
  - parental_income_0.yaml
  - parental_income_520.yaml
```

? Also more sensible to have number of children first and then the alimony paid for them.

MImmesberger:
Improved version in #778

PR #769, opened 2024-06-17
Add Betreuungskosten pointer to synthetic data.

MImmesberger:
When creating synthetic data, the input variable `p_id_betreuungsk_träger` is currently set to 0 for everyone. I adjusted this to be the same as `p_id_kindergeld_empf`.


hmgaudecker (approved):
Makes perfect sense, thanks!

Issue #767, opened 2024-06-14
ENH: Consistency check of user-provided `bg_id`


MImmesberger:
Check whether the `bg_id` is sensible if provided by the user.

For example, we can check whether

- there are at most 2 adults in one BG (not counting their children)
- children are only in the same BG if their parents are part of it as well
- individuals with the same `bg_id` don't have different `wthh_id`s
- individuals with the same `sn_id`, `ehe_id` or `eg_id` have the same `bg_id`

MImmesberger:
Quick question regarding this: I planned to check the IDs before the DAG is created. But to test the specification of `bg_id` I either need to have the outcome of some tax functions available (e.g. kindergeld_anspruch). Or I test the specification via the inputs of `p_id_elternteil`, `p_id_ehepartner`, etc. which makes them mandatory inputs. Do you have any preference? @hmgaudecker

hmgaudecker:
Yes, those `p_id`s should be mandatory.

PR #771, opened 2024-06-25
Regelbedarf on individual level

### LINKED ISSUE #668: ENH: Remove hard-coding of SGB II age groups

`arbeitsl_geld_2_kindersatz_m_hh_ab_2011` currently does:

```py
        arbeitsl_geld_2_params["regelsatz"][6] * anz_kinder_bis_5_hh
        + arbeitsl_geld_2_params["regelsatz"][5] * anz_kinder_ab_6_bis_13_hh
        + arbeitsl_geld_2_params["regelsatz"][4] * anz_kinder_ab_14_bis_17_hh
        + arbeitsl_geld_2_params["regelsatz"][3] * anz_kinder_ab_18_bis_24_hh
```

Similarly, #582 introduced hard-coded `25`s in `groupings.py`

Include mappings of `regelbedarfsstufen` to min/max ages. In the above code snippet, work at the individual level.

MImmesberger:
Closes #668


MImmesberger:
I realised that we cannot compute Regelbedarf on individual level because of `_arbeitsl_geld_2_berechtigte_wohnfläche_bg`, which takes the number of individuals in the Bedarfsgemeinschaft as an input. 

However, it could be that our current understanding of the berechtigte Wohnfläche is wrong, i.e. that it should be calculated based on the amount of people in the Familiengemeinschaft. This would make sense because then, a child inside and outside the parental BG has the same Regelbedarf. (In the current implementation, the child's contribution to the Regelbedarf of the parental BG may be lower than its Regelbedarf in its own BG if there are more than 4 people in the parental BG). Will do some research regarding that.

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
```diff
@@ -0,0 +1,275 @@
...
+    anteile = arbeitsl_geld_2_params["anteil_regelsatz_kinder"]
+    regelsatz = arbeitsl_geld_2_params["regelsatz"]
+
+    if alter <= anteile[6]["max_alter"] and gleiche_fg_kindergeldempfänger_kind:
+        out = regelsatz * anteile[6]["anteil"]
+    elif alter <= anteile[5]["max_alter"] and gleiche_fg_kindergeldempfänger_kind:
+        out = regelsatz * anteile[5]["anteil"]
+    elif alter <= anteile[4]["max_alter"] and gleiche_fg_kindergeldempfänger_kind:
+        out = regelsatz * anteile[4]["anteil"]
+    else:
```
hmgaudecker, line 123: Just to double-check: Is it correct that level 3 is missing here but present in 2011+?

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
```diff
@@ -0,0 +1,275 @@
...
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+
+    Returns
+    -------
+    SGB II needs of child
+
+    """
+
+    kindersofortzuschl = arbeitsl_geld_2_params.get("kindersofortzuschl", 0.0)
```
hmgaudecker, line 154:
```suggestion
    out = arbeitsl_geld_2_params.get("kindersofortzuschl", 0.0)
```

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
```diff
@@ -0,0 +1,275 @@
...
+
+    if (
+        alter <= arbeitsl_geld_2_params["regelsatz"][6]["max_alter"]
+        and gleiche_fg_kindergeldempfänger_kind
+    ):
+        out = arbeitsl_geld_2_params["regelsatz"][6]["bedarf"] + kindersofortzuschl
```
hmgaudecker, line 160:
```suggestion
        out += arbeitsl_geld_2_params["regelsatz"][6]["bedarf"]
```
And so on.

src/_gettsim/transfers/kindergeld.py
```diff
@@ -174,3 +176,39 @@ def kind_bis_10_mit_kindergeld(
     """
     out = kindergeld_anspruch and (alter <= 10)
     return out
+
+
+@policy_info(skip_vectorization=True)
+def gleiche_fg_kindergeldempfänger_kind(
```
hmgaudecker, line 182:
```suggestion
def same_fg_as_kindergeldempfänger(
```
I think this should rather be English/Denglisch? Else `gleiche_fg_wie_kindergeldempfänger`, but in any case, the name should make it obvious that we are looking from the child level rather than symmetrically.

src/_gettsim_tests/test_data/arbeitsl_geld_2/2005/regelsatz_no_adult_in_bg.yaml
hmgaudecker, line 1: Can you explain in the notes why we now need two persons here? I'd rather stick to the old version. In case it is what I fear (`gleiche_fg_kindergeldempfänger_kind` relies on having a link to the `kindergeldempfänger`), that should not be the case. Should work with `p_id_kindergeldempfänger = -1`, too.

src/_gettsim_tests/test_data/arbeitsl_geld_2/2024/regelsatz_no_adult_in_bg.yaml
hmgaudecker, line 1: Same comment as for 2005.

src/_gettsim_tests/test_data/arbeitsl_geld_2/2005/regelsatz_no_adult_in_bg.yaml
hmgaudecker: Oh, I think I understand now where this is coming from. You interpreted this to be the test for "have a kid outside the bg, but inside the fg", right? 

Two things:
- Assuming that is correct, I just made that explicit in the name. However, note that `eigenbedarf_gedeckt` cannot be `false` for the kid then. I also made a copy with two adults to check my conjecture above (which seems to be correct at a glance).
- I am not sure this was the original intention of the test. @ChristianZimpelmann, can you chime in? I think we simply could not handle cases where we did not have adults in a standalone `bg` (e.g., 17-year-olds living on their own), but I might misremember.

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
```diff
@@ -0,0 +1,275 @@
...
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+
+    Returns
+    -------
+
+    """
+    if p_id_einstandspartner > 0:  # BG with 2 adults
+        out = arbeitsl_geld_2_params["regelsatz"] * (
+            arbeitsl_geld_2_params["anteil_regelsatz_erwachsene"]["zwei_erwachsene"]
```
hmgaudecker, line 205: Do I gather correctly that this gives the sum where it should give the average?

We should make a note that we want to return the average in order to have both adults symmetric rather than keeping the difference between RBS 1 and 2 around.

hmgaudecker (changes requested):
Great! I added a few tests, which are expected to fail partly because I did not make an effort to calculate the correct values. I think they show some things that we'll need to work on, however. Happy to chat!

src/_gettsim/transfers/kindergeld.py
```diff
@@ -174,3 +176,39 @@ def kind_bis_10_mit_kindergeld(
...
+    """
+    fg_id_kindergeldempfänger = join_numpy(
+        p_id_kindergeld_empf,
+        p_id,
+        fg_id,
+        value_if_foreign_key_is_missing=-1,
+    )
+
+    out = [
+        fg_id_empf == fg_id_child if fg_id_empf >= 0 else False
```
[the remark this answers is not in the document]
MImmesberger: `if fg_id_empf >= 0` this part is important because there are individual for whom `p_id_kindergeld_empf` is `-1`. Then, `fg_id_kindergeldempfänger` is `-1`.

We could use `np.where` if you like that more? But I figured it would be less readable.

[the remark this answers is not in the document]
hmgaudecker: I don't get it. `fg_id_empf` == `fg_id_kindergeldempfänger` in that snippet? How can it be -1 and >= 0 at the same time?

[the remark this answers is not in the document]
MImmesberger: Ah true, you were already one step ahead 😄 Then `out = fg_id_kindergeldempfänger == fg_id` is sufficient.

I think `fg_id_numpy` can never return `-1`. But we may want to check the input data. We can do this together with #767 and the other issues regarding consistency checks.

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
```diff
@@ -0,0 +1,275 @@
...
+    p_id_einstandspartner
+        See basic input variable :ref:`p_id_einstandspartner`.
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+
+    Returns
+    -------
+
+    """
+    if p_id_einstandspartner > 0:  # BG with 2 adults
```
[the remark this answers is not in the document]
MImmesberger: I'm afraid I can't quite follow here. It's true that the comments are confusing (because they refer to the BG composition while the function is on individual level.)

The first if condition checks whether the obs has an Einstandspartner. Given the correct `bg_id` this implies that both of them are in the same BG and, hence, each of them should get half of the couple Regelbedarf.

The second if condition checks whether the obs has a claim on the Kinderregelsatz. If not, the obs must be the only adult in the BG (and potentially a single parent).

I don't see how this function may create wrong output, but the object in your sentence "the regelsatz of the is returned" is missing, so maybe I didn't get your point. 😄 

The output of `regelsatz_fg_with_2_adults_1_kid_in_1_bg` looks fine to me (note that the test's targets are on BG level, while the function is on individual level)

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
MImmesberger: This whole level thing didn't exist before 2011 and I guess the values before 2011 were chosen to reflect the levels after the 2011 (in both cases, level 6, for example, refers to children younger than 6). The post-2011 level 3 didn't exist back then. Probably this is bad style and we should stick to something that doesn't try a fuzzy category matching before and after the reform.

src/_gettsim_tests/test_data/arbeitsl_geld_2/2005/regelsatz_no_adult_in_bg.yaml
MImmesberger: > I am not sure this was the original intention of the test.

We added this test because there was a bug that returned the Regelsatz of a couple if there were only children in the BG (or rather, what we currently call children via `kind`). But I will stick to your interpretation of the test and test the Regelsatz for children somewhere else.

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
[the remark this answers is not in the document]
hmgaudecker: Great, thanks. Indeed I interpreted the comment ("BG with two adults") more than the code, I think. That mutilated sentence should have included a "adults".

Do we have a check for the `einstandspartner` / `bg_id` - relationship in that data check list? Just in case a user seems to ignore that.

Thanks for pointing out that the tests are at the `bg` level. I did not realise that. They should move to the individual level then!

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
hmgaudecker: Yes, please check some old law and adjust accordingly. This made sense back in the pre-dates_active days, but not anymore.

src/_gettsim/transfers/arbeitsl_geld_2/sgb_ii_bedarf.py
MImmesberger: As discussed on the phone, the Regelsatz is on individual level already in the parameter files (as it is in the law). So the result here is correct.

MImmesberger:
I think this one is finished!

src/_gettsim/parameters/arbeitsl_geld_2.yaml
```diff
@@ -176,114 +176,240 @@ regelsatz:
     1: 364
     2: 328
     3: 291
-    4: 287
-    5: 251
-    6: 215
+    4:
+      min_alter: 14
+      max_alter: 17
+      bedarf: 287
```
hmgaudecker, line 182:
```suggestion
      betrag: 287
```
Or "satz", but "bedarf" seems very misleading. Sorry for noticing so late in the game!

src/_gettsim/transfers/arbeitsl_geld_2/bedarf.py
```diff
@@ -0,0 +1,285 @@
...
+    p_id_einstandspartner
+        See basic input variable :ref:`p_id_einstandspartner`.
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+
+    Returns
+    -------
+
+    """
+    if p_id_einstandspartner >= 0:  # BG with 2 adults
```
hmgaudecker, line 217:
```suggestion
    if p_id_einstandspartner >= 0:  # BG has 2 adults
```

src/_gettsim/transfers/arbeitsl_geld_2/bedarf.py
```diff
@@ -0,0 +1,285 @@
...
+        out = arbeitsl_geld_2_params["regelsatz"] * (
+            arbeitsl_geld_2_params["anteil_regelsatz_erwachsene"]["zwei_erwachsene"]
+        )
+    elif arbeitsl_geld_2_kindersatz_m == 0.0:  # BG with 1 adult
```
hmgaudecker, line 221:
```suggestion
    elif arbeitsl_geld_2_kindersatz_m == 0.0:  # This observation is not a child, so BG has 1 adult
```

src/_gettsim/transfers/arbeitsl_geld_2/bedarf.py
```diff
@@ -0,0 +1,285 @@
...
+        See basic input variable :ref:`p_id_einstandspartner`.
+    arbeitsl_geld_2_params
+        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
+
+    Returns
+    -------
+    float with the minimum needs of an household in Euro.
+
+    """
+    if p_id_einstandspartner >= 0:  # BG with 2 adults
```
hmgaudecker, line 256:
```suggestion
    if p_id_einstandspartner >= 0:  # BG has 2 adults
```

src/_gettsim/transfers/arbeitsl_geld_2/bedarf.py
```diff
@@ -0,0 +1,285 @@
...
+        out = arbeitsl_geld_2_params["regelsatz"][2]
+    elif arbeitsl_geld_2_kindersatz_m == 0.0:  # BG with 1 adult
```
hmgaudecker, line 258:
```suggestion
    elif arbeitsl_geld_2_kindersatz_m == 0.0:  # This observation is not a child, so BG has 1 adult
```

src/_gettsim/transfers/kindergeld.py
```diff
@@ -174,3 +176,34 @@ def kind_bis_10_mit_kindergeld(
...
+
+    """
+    fg_id_kindergeldempfänger = join_numpy(
+        p_id_kindergeld_empf,
+        p_id,
+        fg_id,
+        value_if_foreign_key_is_missing=-1,
+    )
+
+    return fg_id_kindergeldempfänger == fg_id
```
hmgaudecker, line 209: Simplified this "a bit" :wink:

src/_gettsim_tests/test_data/arbeitsl_geld_2/2005/regelsatz_fg_with_1_adult_1_kid_in_1_bg.yaml
hmgaudecker, line 1: For this and subsequent tests: Either 1. include more outputs (like regelbedarf) and rename, or 2. simplify the input data as much as possible (miete etc can go).

I would favor 1. 

Simplification should still happen, e.g., `p_id_ehepartner` and `gemeinsam_veranlagt` can go for sure, I think.

hmgaudecker (approved):
Looks great! I made a couple of changes and there are some little things left, but no show-stoppers. Thanks!

src/_gettsim/transfers/kindergeld.py
MImmesberger: A thanks, totally forgot that...

PR #772, opened 2024-06-26
Implement Mindesteinkommen for WoG.

### LINKED ISSUE #592: Minimum income check Wohngeld eligibility
There is an income check for Wohngeld. I think this is not implemented yet.

MImmesberger:
Closes #592

### DIFF
--- a/src/_gettsim/transfers/wohngeld.py
+++ b/src/_gettsim/transfers/wohngeld.py
@@ -763,6 +763,7 @@ def wohngeld_nach_vermög_check_m_bg(
 
 @policy_info(params_key_for_rounding="wohngeld")
 def wohngeld_vor_vermög_check_m_wthh(
+    wohngeld_mindesteinkommen_erreicht_wthh: bool,
     anz_personen_wthh: int,
     wohngeld_eink_m_wthh: float,
     wohngeld_miete_m_wthh: float,
@@ -776,6 +777,8 @@ def wohngeld_vor_vermög_check_m_wthh(
 
     Parameters
     ----------
+    wohngeld_mindesteinkommen_erreicht_wthh
+        See :func:`wohngeld_mindesteinkommen_erreicht_wthh`.
     anz_personen_wthh
         See :func:`anz_personen_wthh`.
     wohngeld_eink_m_wthh
@@ -789,16 +792,22 @@ def wohngeld_vor_vermög_check_m_wthh(
     -------
 
     """
-    return _wohngeld_basisformel(
-        anz_personen=anz_personen_wthh,
-        einkommen_m=wohngeld_eink_m_wthh,
-        miete_m=wohngeld_miete_m_wthh,
-        params=wohngeld_params,
-    )
+    if wohngeld_mindesteinkommen_erreicht_wthh:
+        out = _wohngeld_basisformel(
+            anz_personen=anz_personen_wthh,
+            einkommen_m=wohngeld_eink_m_wthh,
+            miete_m=wohngeld_miete_m_wthh,
+            params=wohngeld_params,
+        )
+    else:
+        out = 0.0
+
+    return out
 
 
 @policy_info(params_key_for_rounding="wohngeld")
 def wohngeld_vor_vermög_check_m_bg(
+    wohngeld_mindesteinkommen_erreicht_bg: bool,
     anz_personen_bg: int,
     wohngeld_eink_m_bg: float,
     wohngeld_miete_m_bg: float,
@@ -812,6 +821,8 @@ def wohngeld_vor_vermög_check_m_bg(
 
     Parameters
     ----------
+    wohngeld_mindesteinkommen_erreicht_bg
+        See :func`wohngeld_mindesteinkommen_erreicht_bg`.
     anz_personen_bg
         See :func:`anz_personen_bg`.
     wohngeld_eink_m_bg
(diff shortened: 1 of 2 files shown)


MImmesberger:
Blocked until #771 is merged.

hmgaudecker (commented):
Thanks! Just a wild thought, feel free to argue against it: From my perspective, this is fairly hard to understand now by just looking at the names. That is, you need to know that the "Mindesteinkommenscheck" also includes the "Vermögenscheck".

Can't we include both in a `wohngeldanspruch_erfüllt_wthh` Boolean, which we calculate upfront? And then calculate `wohngeldanspruch_wthh`, which will be zero for everyone where  `wohngeldanspruch_erfüllt_wthh == False`. This would be the one before checking priorities with Bürgergeld.

Would be much clearer from my perspective and should not be too hard to implement, I hope?

MImmesberger:
On 3 occasions, regression tests fail after switching to Lohnsteuer. Still investigating.

MImmesberger:
Not sure how official they are. Jakob didn't create test cases that reach back to 2009, right?

MImmesberger:
FYI: I removed two input variables because we can compute them now via our usual machinery.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -66,29 +66,134 @@ def wohngeld_m_wthh(
...
+    return out
+
+
+def wohngeld_anspruchshöhe_bg(
+    wohngeld_basisbetrag_m_bg: float,
+    wohngeld_anspruchsbedingungen_erfüllt_bg: bool,
+) -> float:
+    """Housing benefit after wealth and income check.
+
+    This target is used for the priority check calculation against ALG2 and
```
hmgaudecker, line 110:
```suggestion
    This target is used for the priority check calculation against Arbeitslosengeld 2.
```

src/_gettsim/transfers/wohngeld.py
```diff
@@ -66,29 +66,134 @@ def wohngeld_m_wthh(
...
+    Kinderzuschlag.
```
hmgaudecker, line 111:
```suggestion
```

src/_gettsim/transfers/wohngeld.py
```diff
@@ -66,29 +66,134 @@ def wohngeld_m_wthh(
...
+    if wohngeld_anspruchsbedingungen_erfüllt_wthh:
+        out = wohngeld_basisbetrag_m_wthh
+    else:
+        out = 0.0
+
+    return out
+
+
+def wohngeld_anspruchshöhe_bg(
+    wohngeld_basisbetrag_m_bg: float,
```
hmgaudecker, line 105: Remove the `wohngeld_basisbetrag` functions entirely and integrate them here. The point of the new structure is to only do the calculations if there actually is a claim on Wohngeld conditional on "absolute" checks like Vermögen, Mindesteinkommen.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -66,29 +66,134 @@ def wohngeld_m_wthh(
...
     """Calculate housing benefit subtractions on the individual level.
 
-    Note that eink_st_y_sn is used as an approximation for taxes on income (as mentioned
-    in § 16 WoGG Satz 1 Nr. 1).
-
     Parameters
     ----------
-    eink_st_y_sn
-        See :func:`eink_st_y_sn`.
+    lohnst_m
```
hmgaudecker, line 195: Oh, I am sorry if I suggested this! This will be wrong for the self-employed, I am afraid. Not sure how it is actually determined for those, but I'd rather leave it as it was.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -691,21 +796,18 @@ def wohngeld_miete_ab_2009_m_hh(  # noqa: PLR0912 (see #516)
...
+def wohngeld_vermögensgrenze_unterschritten_wthh(
     vermögen_bedürft_wthh: float,
     anz_personen_wthh: int,
     wohngeld_params: dict,
-) -> float:
-    """Preliminary housing benefit after wealth check.
+) -> bool:
+    """Wealth is below the threshold for housing benefit.
 
     This target is used for the actual Wohngeld calculation of the Bedarfsgemeinschaften
```
hmgaudecker, line 806: Delete this line, seems like a leftover.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -718,29 +820,25 @@ def _wohngeld_nach_vermög_check_m_wthh(
...
+def wohngeld_vermögensgrenze_unterschritten_bg(
     vermögen_bedürft_bg: float,
     anz_personen_bg: int,
     wohngeld_params: dict,
-) -> float:
-    """Preliminary housing benefit after wealth check.
+) -> bool:
+    """Wealth is below the threshold for housing benefit.
 
     This target is used for the priority check calculation against ALG2 and
```
hmgaudecker, line 837: Again, delete.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -833,6 +930,104 @@ def wohngeld_vor_vermög_check_m_bg(
...
+        See :func:`arbeitsl_geld_2_regelbedarf_m_wthh`.
+    wohngeld_einkommen_für_mindesteinkommen_check_m_wthh
+        See :func:`wohngeld_einkommen_für_mindesteinkommen_check_m_wthh`.
+
+    Returns
+    -------
+
+    """
+    return (
+        arbeitsl_geld_2_regelbedarf_m_wthh
```
hmgaudecker, line 960: Turn around, >= is much easier to parse given the function name.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -833,6 +930,104 @@ def wohngeld_vor_vermög_check_m_bg(
...
+        See :func:`arbeitsl_geld_2_regelbedarf_m_bg`.
+    wohngeld_einkommen_für_mindesteinkommen_check_m_bg
+        See :func:`wohngeld_einkommen_für_mindesteinkommen_check_m_bg`.
+
+    Returns
+    -------
+
+    """
+    out = (
+        arbeitsl_geld_2_regelbedarf_m_bg
```
hmgaudecker, line 989: Same.

src/_gettsim/transfers/wohngeld.py
```diff
@@ -833,6 +930,104 @@ def wohngeld_vor_vermög_check_m_bg(
...
+    )
+
+
+def wohngeld_mindesteinkommen_erreicht_bg(
+    arbeitsl_geld_2_regelbedarf_m_bg: float,
+    wohngeld_einkommen_für_mindesteinkommen_check_m_bg: float,
+) -> bool:
+    """Minimum income requirement for Wohngeld has been met.
+
+    Minimum income is defined via VwV 15.01 ff § 15 WoGG.
```
hmgaudecker, line 971: Align the docstrings of the two `wohngeld_mindesteinkommen_erreicht` functions.

hmgaudecker (commented):
Looks great overall! I am afraid you were a bit too greedy in a couple of places, though :wink:

docs/gettsim_objects/input_variables.md
```diff
@@ -150,14 +150,6 @@ Self-employed (main profession)
 
 Type: bool
 
-(hat_kinder)=
```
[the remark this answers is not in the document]
MImmesberger: Ah, I see, makes sense.

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
```diff
@@ -177,8 +193,8 @@ def arbeitsl_geld_2_nettoquote(  # noqa: PLR0913
     alg2_2005_bne = max(
         (
             bruttolohn_m
-            - (eink_st_y_sn / anz_personen_sn / 12)
-            - (soli_st_y_sn / anz_personen_sn / 12)
+            - lohnst_m
+            - soli_st_lohnst_m
             - sozialv_beitr_arbeitnehmer_m
             - arbeitsl_geld_2_params["abzugsfähige_pausch"]["werbung"]
```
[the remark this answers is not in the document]
MImmesberger: In our Lohnsteuer module both are separate (in the Einkommensteuer module as well) so we wouldn't make a mistake here. I'll ask Jakob whether we need to change this in the Lohnsteuer module.

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -13,7 +13,7 @@
 def wohngeld_vorrang_bg(
```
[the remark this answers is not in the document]
MImmesberger: I've been wondering about this for weeks, I think we can just use `wohngeld_kinderzuschl_vorrang_bg`. But this will be gone very soon anyhow.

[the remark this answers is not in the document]
hmgaudecker: Whatever is easier for you here.

src/_gettsim/transfers/arbeitsl_geld_2/arbeitsl_geld_2_eink.py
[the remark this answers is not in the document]
hmgaudecker: Ah, okay, my bad. Makes perfect sense as is -- these reduce the tax payments, but taxes will be covered by Arbeitslosengeld 2 anyhow. So in order to make recipients benefit, you need to deduct them from the relevant income...

src/_gettsim/transfers/wohngeld.py
MImmesberger: Sure, this will create a problem with the tests though because we didn't care about the Mindesteinkommen in the past. Many tests target what I had called "Basisbetrag", not the actual Anspruchshöhe.

I would add a `wohngeld_mindesteinkommen_erreicht_wthh=True` entry in all those tests and make a note at the top. If you prefer to rewrite all these tests such that they pass the Mindesteinkommen check (and are economically sensible), we can make an issue. I presume those are all regression tests.

src/_gettsim/transfers/wohngeld.py
MImmesberger: Hm then I don't understand your comment regarding switching to Lohnsteuer in `arbeitsl_geld_2_nettoeink_ohne_freibetrag_m`. There, the income for self-employed would be treated wrong as well (because we now subtract Lohnsteuer instead of `eink_st_y_sn`). Can you explain what you meant back then?

MImmesberger:
I'll start with a new PR given the requested changes back to `hat_kinder` and of Lohnsteuer back to Einkommensteuer. Will cherry-pick the commits here.

PR #778, opened 2024-07-06
Endogenous Bedarfsgemeinschaften

### LINKED ISSUE #763: BUG: Endogenous creation of `bg_id`

MImmesberger:
Closes #763

### DIFF
--- a/src/_gettsim/transfers/benefit_checks/benefit_checks.py
+++ b/src/_gettsim/transfers/benefit_checks/benefit_checks.py
@@ -7,15 +7,42 @@ aggregate_by_group_benefit_checks = {
         "source_col": "wohngeld_kinderzuschl_vorrang_bg",
         "aggr": "any",
     },
+    "alle_wohngeld_kinderzuschl_statt_arbeitsl_geld_2_fg": {
+        "source_col": "wohngeld_kinderzuschl_statt_arbeitsl_geld_2_endogen",
+        "aggr": "all",
+    },
 }
 
 
-def wohngeld_vorrang_bg(
-    arbeitsl_geld_2_regelbedarf_m_bg: float,
-    arbeitsl_geld_2_eink_m_bg: float,
-    wohngeld_anspruchshöhe_m_bg: float,
+def wohngeld_kinderzuschl_statt_arbeitsl_geld_2_endogen(
+    wohngeld_kinderzuschl_vorrangig_bg: bool,
+    wohngeld_kinderzuschl_günstiger: bool,
 ) -> bool:
-    """Check if housing benefit has priority.
+    """Individual receives Wohngeld and Kinderzuschlag instead of Arbeitslosengeld II.
+
+    Parameters
+    ----------
+    wohngeld_vorrangig_bg
+        See :func:`wohngeld_vorrangig_bg`.
+    wohngeld_kinderzuschl_günstiger
+        See :func:`wohngeld_kinderzuschl_günstiger`.
+
+    Returns
+    -------
+
+    """
+    return wohngeld_kinderzuschl_vorrangig_bg or wohngeld_kinderzuschl_günstiger
+
+
+def wohngeld_kinderzuschl_vorrangig_bg(  # noqa: PLR0913
+    arbeitsl_geld_2_regelbedarf_m_bg: float,
+    arbeitsl_geld_2_nettoeink_vor_abzug_freibetrag_m: float,
+    kindergeld_zur_bedarfsdeckung_m_bg: float,
+    kind_unterh_erhalt_m_bg: float,
+    unterhaltsvors_m_bg: float,
+    kindergeldübertrag_m_bg: float,
+) -> bool:
+    """Wohngeld and Kinderzuschlag have priority over Arbeitslosengeld II / Bürgergeld.
 
     Housing benefit has priority if the sum of housing benefit and income covers the
     needs according to SGB II of the Bedarfsgemeinschaft.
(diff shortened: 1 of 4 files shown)


MImmesberger:
I'm not too happy with the current implementation because I needed to copy+paste a lot of code in regarding SGB II Vermögensgrenzen and KiZ calculation. From a (very) quick first look it doesn't seem like we can move to individual-level calculation there, we should discuss this.

But it may be a good idea to postpone this to another PR (if its possible at all), this one is already quite complicated.

hmgaudecker (commented):
Just skimmed through a couple of things to keep the conversation up!

src/_gettsim/transfers/arbeitsl_geld_2/bedarf.py
```diff
@@ -3,6 +3,73 @@
...
+        arbeitsl_geld_2_nettoeink_vor_abzug_freibetrag_m
+        + kindergeld_zur_bedarfsdeckung_m
+        + kind_unterh_erhalt_m
+        + unterhaltsvors_m
+        + kindergeldübertrag_m
+        >= arbeitsl_geld_2_regelbedarf_m
+    )
+
+
+def arbeitsl_geld_2_eigenbedarf_gedeckt_und_kind_in_fg(
```
hmgaudecker, line 53: Why can't we just include this in the above function? I.e., do we ever need the `arbeitsl_geld_2_eigenbedarf_gedeckt` without the `ist_kind_in_fg` condition?

AFAICT:
- We are using the joint condition for `kinder_mit_gedecktem_bedarf_in_fg` and in `wohngeld_kinderzuschl_statt_arbeitsl_geld_2`
- We are replicating the joint condition *exactly* in `wthh_id_numpy` and `wohngeld_kinderzuschl_günstiger`
- We are implicitly replicating the joint condition in `bg_id_numpy` (there even is an issue to make it explicit)

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
```diff
@@ -125,39 +126,60 @@ def _diff_kindergeld_kindbedarf_m(  # noqa: PLR0913
...
     return out
+
+
+@policy_info(skip_vectorization=True)
+def _in_anderer_bedarfsgemeinschaft_als_kindergeldempfänger(
+    p_id: numpy.ndarray[int],
+    p_id_kindergeld_empf: numpy.ndarray[int],
+    bg_id: numpy.ndarray[int],
+) -> numpy.ndarray[bool]:
+    """True if the person is in a different Bedarfsgemeinschaft than the
```
hmgaudecker, line 172: I guess you are primarily thinking about cases where `fg_id` is the same here? Maybe make that explicit in the docstring.

src/_gettsim/config.py
```diff
@@ -157,7 +157,8 @@ def set_array_backend(backend: str):
     "p_id_ehepartner": int,
     "p_id_einstandspartner": int,
     "vermögen_bedürft": float,
-    "eigenbedarf_gedeckt": bool,
+    "arbeitsl_geld_2_eigenbedarf_gedeckt": bool,
```
hmgaudecker, line 160: I don't think this one is necessary anymore if we know `wohngeld_kinderzuschl_statt_arbeitsl_geld_2`?

src/_gettsim/transfers/benefit_checks/benefit_checks.py
```diff
@@ -1,97 +1,245 @@
...
+        See :func:`wohngeld_kinderzuschl_günstiger`.
+
+    Returns
+    -------
+
+    """
+    return wohngeld_kinderzuschl_vorrangig_bg or wohngeld_kinderzuschl_günstiger
+
+
+def wohngeld_kinderzuschl_statt_arbeitsl_geld_2(
```
hmgaudecker, line 33: We need to be really careful with the names here. It seems like this one applies to children only, but there is nothing in the name suggesting that.

Also, I am not sure we need / want the `wohngeld_kinderzuschl_statt_arbeitsl_geld_2_endogen` / `wohngeld_kinderzuschl_statt_arbeitsl_geld_2` distinction. Maybe that even applies to `bg_id` and `wthh_id` and their endogenous counterparts.

I would think we'll need two rounds of calling `compute_taxes_and_transfers`:
1. Calculate `bg_id`, `wthh_id`, `wohngeld_kinderzuschl_statt_arbeitsl_geld_2`
2. Given these three, we can just calculate `arbeitsl_geld_2`, `wohngeld`, `kinderzuschlag` without any checks on which one applies.

Trying to do both in one go would raise some circular graph error. I might be missing something (again), though.

src/_gettsim/transfers/kinderzuschl/kinderzuschl.py
```diff
@@ -6,40 +6,38 @@
...
     return out
 
 
-def _kinderzuschl_nach_vermög_check_m(
-    _kinderzuschl_nach_vermög_check_m_bg: float,
+def kinderzuschl_anspruchshöhe_m(
+    kinderzuschl_anspruchshöhe_m_bg: float,
     anz_personen_bg: int,
 ) -> float:
     """Kinderzuschlag on individual level.
```
hmgaudecker, line 43:
```suggestion
    """Kinderzuschlag distributed across all members of  a Bedarfsgemeinschaft, required for ...
```

src/_gettsim/transfers/arbeitsl_geld_2/kindergelduebertrag.py
MImmesberger: Yes, I think so. I'm actually not sure what happens with Kindergeldübertrag if both live in different households (and thus are in different fg's).

src/_gettsim/transfers/benefit_checks/benefit_checks.py
MImmesberger: Just a comment on this: The variable is not relevant for children only, it also determines the type of transfer for the adults.

If it is set endogenously to create candidates, it always results in `False` for the adults in the FG (hence your comment probably)

hmgaudecker: Exactly. The code needs that as a  description.

docs/gettsim_objects/input_variables.md
```diff
@@ -110,6 +110,15 @@ Identifier of person that claims Kindergeld for the particular child
 
 Type: int
 
+(wohngeld_kinderzuschl_statt_arbeitsl_geld_2)=
+
+## \`wohngeld_kinderzuschl_statt_arbeitsl_geld_2\`\`
+
+Individual may have claim on Wohngeld and Kinderzuschlag instead of Arbeitslosengeld II
```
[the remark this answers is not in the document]
MImmesberger: Maybe

```text
(beantragt_wohngeld_kinderzuschl_statt_arbeitsl_geld_2)=

## \`beantragt_wohngeld_kinderzuschl_statt_arbeitsl_geld_2\`\`

Individual applies for Wohngeld or Kinderzuschlag instead of Arbeitslosengeld II.

Type: bool
```
?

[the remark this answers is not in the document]
hmgaudecker: But not super-important given overall strategy, maybe add a `temp_` in the name.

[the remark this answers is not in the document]
MImmesberger: Ah, now I get it. 

In that case, the Anspruchshöhe for all transfers will be 0. However, if your income is small and that of your children is high, you might end up in a situation where your children form a wthh, and you have your own BG. Result is: Children get Wohngeld, you get nothing (as opposed to everyone in the same BG and the whole FG gets nothing).

Then, `beantragt_wohngeld_kinderzuschl_statt_arbeitsl_geld_2` for you is False, for your children it is True.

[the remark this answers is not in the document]
hmgaudecker: That was just an example for being ineligible, period. The problem I have with the name that it only makes sense for people eligible for at least one of the programmes.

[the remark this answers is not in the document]
hmgaudecker: But again, just don't worry much.

MImmesberger:
Sorry that this comes super late: While working on the test files, I noticed a major bug that is not easy to fix and would lead to jumps in disposable income. The reason is that Kindergeldübertrag depends on the amount of Wohngeld, but the Kindergeldübertrag currently uses a "candidate Wohngeld" function as an input, not the actual amount.

This is not easy to fix inside the DAG. I think it doesn't make sense to introduce even more candidate functions to make this work given that we discussed to move to several GETTSIM calls in the future. Hence, I would vote for not merging this one. We can use the test files and the logic developed in this PR in the PR that tackles this.

hmgaudecker:
The use case is so tiny that it won't be worth it to implement this. 

Rather, we should provide a recipe in a separate repo that 
1. generates all $$k$$ possible and feasible permutations of `bg_id` and `wthh_id`
2. runs GETTSIM $$k$$ times
3. Picks the correct configuration of `bg_id` and `wthh_id`

hmgaudecker:
[Link to follow-up in closed issue](https://github.com/iza-institute-of-labor-economics/gettsim/issues/779#issuecomment-3052722883)

Issue #533, opened 2023-02-16
ENH: Unify functions.py and functions_loader.py


lars-reimann:
#### Is your feature request related to a problem?

* [functions.py](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/src/_gettsim/functions.py) is used by [functions.md](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/docs/gettsim_objects/functions.md) to generate the documentation page [Policy functions in GETTSIM](https://gettsim.readthedocs.io/en/stable/gettsim_objects/functions.html). In works by importing all declarations in the appropriate Python modules.
* [functions_loader.py](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/src/_gettsim/functions_loader.py) contains a function `_convert_paths_to_import_strings(paths)` to discover built-in policy functions that can become part of the DAG. It works by recursively visiting Python files in a variety of directories and discovering functions using `importlib` and `inspect`.

Both seem to solve the same issue, namely providing a list of all built-in policy functions. First, this seems to be unnecessary duplication. Second, some functions that are available in the DAG might be missing from the documentation.

#### Describe the solution you would like to see

* **Option 1:** Use the approach of the function loader to populate the documentation. Advantage: No need to manually create and maintain a file that imports the functions. Disadvantage: Cannot simply use `.. automodsumm:: _gettsim.functions` in [functions.md](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/docs/gettsim_objects/functions.md) to generate the documentation. I don't know whether Sphinx offers a means to programmatically create this documentation page.

* **Option 2:** Use [functions.py](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/src/_gettsim/functions.py) in [functions_loader.py](https://github.com/iza-institute-of-labor-economics/gettsim/blob/main/src/_gettsim/functions_loader.py). This would simplify the logic because we no longer need to walk the Python files manually. Also, a missing import in `functions.py` would be easier to discover, since the function is missing from the DAG as well.

#### Interface breaking implications

No changes to the public interface.

hmgaudecker:
Very good analysis!

I would put this on hold until we have tackled GEP 6. I hope to be able to have a full draft by the end of next week.

It will probably make more sense to have the documentation by policy year then, else show a column with the result of `@dates_active` in the table.

hmgaudecker:
Outdated.

PR #780, opened 2024-07-17
Namespaces: Infrastructure

lars-reimann:

This PR introduces namespaces to GETTSIM's infrastructure.

- [x] Write `policy_function` decorator (rename `policy_info` and change behavior so that a `PolicyFunction` instance is returned). ~Apply to all TT functions.~ (that should be part of renamings)
- [x] Check that functions in module with same simple_name have the correct start_date, end_date specs (this was removed from the policy_info decorator).
- [x] Remove doubled levels in the functions tree automatically (to avoid writing functions in `__init__.py`).
- [x] Go over type hints for aggregation functions.
- [x] Refactor interface module.
- [x] Implement some safety checks 
  - [x] No function should have the same name as a module in the same directory
  - [x] No trailing underscores in module names (for [DAGS PR](https://github.com/OpenSourceEconomics/dags/pull/17))


lars-reimann:
@MImmesberger The function dictionary is currently structured like this:

image

Which levels of nesting should be removed? Based on our previous discussion, it's probably the first two, i.e. `_gettsim` and `social_insurance_contributions`/`transfers`/`taxes`/`demographic_vars`, right?

MImmesberger:
I think that, for example, `transfers.arbeitsl_geld.betrag` is perfect, so I would vote for keeping the `social_insurance_contributions/transfers/taxes/demographic_vars` level. The `_gettsim` should not be part of the target call.

hmgaudecker:
Looking at that, I'd almost think that `transfers` and `taxes` are superfluous, but not `demographics` *(I'd suggest renaming this)* and `social_insurance_contributions`, although we might be able to come up with a shorter name.

MImmesberger:
I agree, there should be no overlap between the elements one level below the `tax/transfer` level, so no need to explicitly distinguish. In that case, however, we should make sure that the module docstring what the module is about (e.g. it's not obvious whether the "Kinderbonus" is a tax deduction or a transfer).

Regarding the naming of `social_insurance_contributions`: In the google doc, I called this "sozialversicherungsbeitraege". But I just realized that I put this in the taxes namespace which is wrong. Should be its own (as it is the case currently in Lars' proposal).

hmgaudecker:
Agreed, let's use `sozialversicherungsbeitraege`. At some point we might rename `demographics` yet again, but that really should be a simple Search & Replace.

I'm all for good docstrings, but any tax deduction should go into the tax component it is deducted from, right?

MImmesberger:
> I'm all for good docstrings, but any tax deduction should go into the tax component it is deducted from, right?

Yes definitely. Was just thinking about making it as obvious as possible. In our case "Kinderbonus" is a transfer.

ChristianZimpelmann:
> sozialversicherungsbeitraege

More in line with the naming of modules might be `sozialv_beitraege`

hmgaudecker:
Good catch! Those are scheduled to be changed to 

- arbeitslosenversicherung
- einkommensgrenzen
- krankenversicherung
- pflegeversicherung
- rentenversicherung

But that was not impossible to know, ofc!

lars-reimann:
Should the renamings/removals be done programmatically or by changing the directory structure under `_gettsim`.

hmgaudecker:
Just came across #533 -- might that be fixed in passing here?

MImmesberger:
Have we already reached a consensus regarding the way users interact with GETTSIM's functions and targets after this PR? I think the case is pretty clear for `data` and `targets`, but I'm unsure about the policy environment.

**Data**
We surely want the user to use pandas Dataframes. This implies that the user will use variable names with double underscores as variable names (e.g. `transfers__arbeitslosengeld__betrag`). Internally, we can translate this to a nested dict structure that the result of `dags.concatenate_functions_tree` requires as an input. I have already done this in the last commits.

**Targets**
I figured that the same is true for targets. We don't want the user to type out a complicated nested dict, but rather use the strings.

**Policy Environment**
For the functions of the policy environment, we could either
1. Collect functions from all modules in a nested dict and carry this through the policy environment postprocessor. This has the downside(?) of users having to interact with nested dicts when they want to replace a function.
2. As above, but parse them back into a list (e.g. by renaming using the `__` notation). Then, recreate the nested dict just before the `dags.concatenate_functions_tree` call.

@hmgaudecker I guess you're leaning towards option 1, just wanted to be sure; Option 1 would require larger changes to the policy environment postprocessor.

src/_gettsim/functions/loader.py
```diff
@@ -187,9 +220,9 @@ def _create_policy_function_from_decorated_callable(
 
     # Only needed until the directory structure is cleaned up
```
hmgaudecker, line 221:
```suggestion
    # Only needed until the directory structure is cleaned up
    # TODO(@MImmesberger): Remove the removeprefix calls once the directory
    # structure is cleaned up```

File: src/_gettsim/functions/loader.py [hunk 2]
```diff
@@ -270,3 +303,11 @@ def _load_dicts_in_module(
         for name, member in inspect.getmembers(module)
         if isinstance(member, dict) and name.startswith(prefix_filter)
     ]
+
+
+def _fail_if_more_than_one_dict_loaded(dicts: list[dict]) -> None:
+    if len(dicts) > 1:
+        raise ValueError(
+            "More than one dictionary found in the module. "
```
hmgaudecker, line 311:
```suggestion
            "More than one dictionary found in the module:\n\n"
            f"{module}\n\n"
```

src/_gettsim/functions/loader.py
```diff
@@ -216,31 +249,31 @@ def load_internal_aggregation_dict(variant: _AggregationVariant):
...
+        module_name = _convert_path_to_module_name(path, package_root)
+        # TODO(@MImmesberger): Remove the removeprefix calls once the directory
+        # structure is cleaned up
+        clean_module_name = (
+            module_name.removeprefix("_gettsim__")
+            .removeprefix("taxes__")
+            .removeprefix("transfers__")
         )
+        tree_keys = get_path_from_qualified_name(clean_module_name)
+        dicts_in_module = _load_dicts_in_module(path, package_root, f"{variant}_")
```
hmgaudecker, line 272:
```suggestion
        derived_function_names = load_functions_to_derive(path, package_root, f"{variant}_")
```

Then make the check that only one dict is allowed in that function.

src/_gettsim/gettsim_typing.py
hmgaudecker, line 1: Very nice!

hmgaudecker (commented):
Looks very nice -- but far too long for the amount of time I have today. Will try to continue asap.

Thanks a lot!

MImmesberger:
So that we don't forget it: We should think about changing the `name_in_dag` keyword to something else. Maybe `simple_name` or `simple_name_in_dag` makes clearer that this isn't the qualified name.

hmgaudecker:
> So that we don't forget it: We should think about changing the `name_in_dag` keyword to something else. Maybe `simple_name` or `simple_name_in_dag` makes clearer that this isn't the qualified name.

Good one. But now that we switch, should not the full path be part of `policy_info`?

hmgaudecker:
I looked a little deeper now. Let's set aside a day in January where we'll look at this more closely. I do believe there is huge room for simplification in the whole function loader with the tree structure. That is, `policy_info` can just use the path to the module if no path is supplied. Maybe the `policy_info` decorator can even become a `policy_function` decorator returning the whole function so that `_create_policy_function_from_decorated_callable` can disappear completely.

src/_gettsim/functions/loader.py
```diff
@@ -186,10 +218,12 @@ def _create_policy_function_from_decorated_callable(
     """
 
     # Only needed until the directory structure is cleaned up
+    # TODO(@MImmesberger): Remove the removeprefix calls once the directory
```
hmgaudecker, line 221: I actually think we should do that right now. We will not merge this one into main, so we want to have the full machinery in place before tackling the renaming. Or am I missing something? Same for the other occurrence.

hmgaudecker: And that was the right thought at the time we wanted to merge into main!

MImmesberger: We haven't settled on a specific directory structure yet. I propose:

```text
|-- src
|     |-- gettsim
|     |       |-- ...
|     |-- _gettsim
|     |       |-- infrastructure (alternatively: ttsim?)
|     |       |        |-- aggregation_jax.py
|     |       |        |-- aggregation_numpy.py
|     |       |        |-- aggregation.py
|     |       |        |-- gettsim_typing.py
|     |       |        |-- piecewise_functiony.py
|     |       |        |-- policy_environment.py
|     |       |        |-- policy_environment_postprocessor.py
|     |       |        |-- interface.py
|     |       |        |-- shared.py
|     |       |        |-- time_conversion.py
|     |       |        |-- vectorization.py
|     |       |-- parameters
|     |       |-- taxes_and_transfers
|     |       |        |-- demographics
|     |       |        |-- einkommensteuer
|     |       |        |-- arbeitslosengeld_2
|     |       |        |-- wohngeld
|     |       |        |-- ...
|     |       |        |-- groupings.py
|     |       |-- config.py
|     |       |-- synthetic.py
|     |       |-- visualization.py
|     |-- _gettsim_tests
|     |       |-- ...
```

hmgaudecker: Great. Again, I think I would be even more radical, now that we are at it:

```text
|-- src
|     |-- gettsim
|     |       |-- ...
|     |-- _gettsim
|     |       |-- ttsim
|     |       |        |-- config.py
|     |       |        |-- aggregation_jax.py
|     |       |        |-- aggregation_numpy.py
|     |       |        |-- aggregation.py
|     |       |        |-- typing.py
|     |       |        |-- piecewise_functiony.py
|     |       |        |-- policy_environment.py
|     |       |        |-- policy_environment_postprocessor.py
|     |       |        |-- interface.py
|     |       |        |-- shared.py
|     |       |        |-- time_conversion.py
|     |       |        |-- vectorization.py
|     |       |        |-- visualize_dag.py
|     |       |-- de
|     |       |        |-- config.py
|     |       |        |-- parameters
|     |       |        |-- demographics
|     |       |        |-- einkommensteuer
|     |       |        |-- arbeitslosengeld_2
|     |       |        |-- wohngeld
|     |       |        |-- ...
|     |       |        |-- groupings.py
|     |       |        |-- synthetic.py
|     |-- _gettsim_tests
|     |       |-- ttsim
|     |       |        |-- ...
|     |       |-- de
|     |       |        |-- ...
```

That is,
- ttsim suggestion is cool, change taxes_and_transfers accordingly
- config and synthetic need to be split up among these two
- just typing, not gettsim_typing
- visualize_dag (we should probably rename the function, too?)
- make parameters part of de

Probably easiest to inject this as [step 3 in the grand PR](https://github.com/iza-institute-of-labor-economics/gettsim/pull/804#issue-2735500579) given the work you put in so far?

src/_gettsim/visualization.py
```diff
@@ -67,41 +73,66 @@ def plot_dag(
...
-    necessary_functions = {
-        f_name: f for f_name, f in functions_not_overridden.items() if (f_name in nodes)
-    }
+    )
+
+    _, necessary_functions = _filter_tree_by_name_list(
+        functions_not_overridden, dag.nodes
+    )
+    processed_functions = _round_and_partial_parameters_to_functions(
+        necessary_functions,
```
hmgaudecker, line 119:
```suggestion
        _filter_tree_by_name_list(...)[1],
```
and remove the above line.

hmgaudecker (commented):
Just a couple of small things while familiarising myself with it. Looks great in general!

src/_gettsim/shared.py
```diff
@@ -181,24 +187,116 @@ def format_list_linewise(list_):
...
+        if key in result and isinstance(result[key], dict) and isinstance(value, dict):
+            result[key] = merge_nested_dicts(result[key], value)
+        else:
+            result[key] = value
 
-    return sorted(set(user_input))
+    return result
+
+
+def _filter_tree_by_name_list(
```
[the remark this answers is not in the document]
MImmesberger: We use qualified names almost exclusively for comparing user-provided inputs (targets and data) with GETTSIMs DAG, mainly because I felt like string comparisons are more readable. We can get rid of almost all of them, though. 

One alternative would be something like

```python
from optree import tree_accessors

def func_1():
    """Some function."""
    pass

def func_2():
    """Some function."""
    pass

def func_3():
    """Some function."""
    pass

functions = {"module_1": {"submodule": {"func_1": func_1}}, "module_2": {"func_2": func_2}}
targets = {"module_1": {"submodule": {"func_1": None}}, "module_2": {"func_2": None}}

functions_in_targets = [
    accessor(functions)
    for accessor in tree_accessors(targets, none_is_leaf=True)
]
```

MImmesberger:
ToDos as discussed in person today:

- Write PolicyFunction decorator. Apply to all TT functions.
- Remove doubled levels in the functions tree (to avoid writing functions in `__init__.py`).
- Implement some safety checks (no function should have the same name as a module in the same directory, no trailing underscores in module names).
- Possibly: Don't use qualified names whenever possible (see above).

MImmesberger:
> This one is for the https://github.com/OpenSourceEconomics/dags/pull/17, I think!

Yes, but only the second check (no trailing underscores in module names).

hmgaudecker:
> Yes, but only the second check (no trailing underscores in module names).

Ah, I updated the list to make this apparent right away!

MImmesberger:
I'm thinking about whether we should introduce a data class to GETTSIM. 

I'm mainly thinking about the handling of derived functions here. In the case of time-conversions, we have two completely distinct pipelines for PolicyFunctions and data columns (strings). If we could use the same methods for both objects (`.leaf_name`, `.qualified_name`), we could greatly simplify the code for time conversions. Also, it would allow for "inheritance" of namespaces from source to derived functions. Currently, I do this manually by joining and splitting qualified name strings.

But, as this is purely internal, we could also postpone this.

hmgaudecker (commented):
A few comments while familiarising myself with this again.

src/_gettsim/interface.py
```diff
@@ -2,53 +2,72 @@
...
-    parse_to_list_of_strings,
+    get_path_from_qualified_name,
+    merge_nested_dicts,
+    partition_tree_by_reference_tree,
+    tree_to_dict_with_qualified_name,
+    tree_update,
 )
 
 
 def compute_taxes_and_transfers(  # noqa: PLR0913
```
hmgaudecker, line 53: I think we can schedule the most of this function's content for deletion
- Modifications can be moved out of the function, just leave input sanity checks (e.g., build_targets_tree, function partitions,  _convert_data_to_correct_types, check_minimal_specification)
- What is to be kept:
  - Call combine_policy_functions_and_derived_functions
  - Call _round_and_partial_parameters_to_functions (but we can apply that to all functions, right?)
  - Create the input structure tree (once!)
  - Concatenate the functions
  - Call the tax_transfer_function with the subset of the input data that is in input_structure
- Not all of this needs to happen in this PR, but let's redesign with that in mind in order to focus

MImmesberger: Thanks!

> Modifications can be moved out of the function, just leave input sanity checks

I'm a bit confused by this. Are the examples you list supposed to go or stay?

hmgaudecker: Most of the functionality in them can go, eventually. We will assume they are of the correct type, maybe check the bare necessities in order to avoid confusing errors down the line.

All these conversions can happen in earlier steps.

src/_gettsim/functions/loader.py
```diff
@@ -26,56 +33,188 @@ def load_functions_for_date(date: datetime.date) -> list[PolicyFunction]:
...
         The policy functions that are active at the given date.
     """
-    return [f for f in _load_internal_functions() if f.is_active_at_date(date)]
+    return _build_functions_tree_for_date(
+        date=date, package_root=RESOURCE_DIR, roots=PATHS_TO_INTERNAL_FUNCTIONS
+    )
 
 
-def _load_internal_functions() -> list[PolicyFunction]:
+def _build_functions_tree_for_date(
```
hmgaudecker, line 41: Remove lines 23--43. Given that this function is private, there is no need to differentiate (else one might have thought of loading from a different location). In any case, naming is confusing as it stands, so let's just simplify...

src/_gettsim/functions/policy_function.py
```diff
@@ -103,57 +74,106 @@ def original_function_name(self) -> str:
         """The name of the wrapped function."""
         return self.function.__name__
 
+    def set_qualified_name(self, qualified_name: str) -> None:
```
hmgaudecker, line 77: I am afraid we will need to get rid of this, sorry for suggesting it earlier.

The problem is that we cannot guarantee consistency between the tree structure and the qualified name. The `PolicyFunction` object does not know about the tree. If the structure of the tree changes (should not happen of course, but expect people to do crazy things ...), that will lead to errors that are very hard to debug. Plus, it creates state. First it is `None`, only later you call this function to set it to a string. 

The same holds for `leaf_name`, in a way, but that one we will need to keep. We could add a simple check that `tree_path[-1] == leaf.leaf_name` or so. 

The good thing I think we are only using it for a couple of errors in `interface.py`, which should be different for legibility reasons, anyhow. Maybe `".".join(tree_path)` ?

MImmesberger: Ah, had the same thought a few days ago, but forgot to ask you about it. 

For the `leaf_name`, it would be sufficient to add the check you suggested to the `upsert_function` method.

hmgaudecker: Excellent!

src/_gettsim/functions/loader.py
```diff
@@ -165,91 +221,104 @@ def _convert_path_to_module_name(path: Path, package_root: Path) -> str:
...
 
 
 _AggregationVariant: TypeAlias = Literal["aggregate_by_group", "aggregate_by_p_id"]
 
 
-def load_internal_aggregation_dict(variant: _AggregationVariant):
-    """
-    Load a dictionary with all aggregations by group or person that are defined for
-    internal functions.
+def load_internal_aggregation_tree(
```
hmgaudecker, line 268:
```suggestion
def load_one_aggregation_specs_tree(
```

And use that terminology throughout, as we already do in `PolicyEnvironment`.

Be consistent in the naming there, too: 

`_aggregate_by_group_specs` -> `_by_group_aggregation_specs_tree` etc.

src/_gettsim/policy_environment.py
```diff
@@ -257,6 +223,40 @@ def _parse_date(date):
     return date
 
 
+def _convert_all_functions_to_policy_functions(
```
MImmesberger, line 226: I think we don't even want to allow for user-provided functions trees that do not have PolicyFunctions (but callables) as leafs, right? Else, we would make the implicit assumption that the correct leaf name is the name of the last branch (which is the only option we allow, but still).

hmgaudecker: I think that's fine. If users want to do some simple modifications, they should be able to pass functions that are not decorated or explicitly creating a `PolicyFunction` instance.

But the module name should not exist here, it is not part of the `PolicyFunction` otherwise, right?

So the whole function would be: 

```python
function if isinstance(function, PolicyFunction) else PolicyFunction(function)
```

right?

MImmesberger: I think
```python
if isinstance(function, PolicyFunction):
    out = function
else:
    out = PolicyFunction(function, leaf_name=tree_path[-1])
```

src/_gettsim/functions/loader.py
MImmesberger: There is only one aggregation specs tree, so I wouldn't choose this function name. If you want to reflect the fact that this is not the final (augmented) tree, we could use `load_aggregations_tree_from_internal_modules`. But I think the status quo is fine as well.

src/_gettsim/policy_environment.py
```diff
@@ -2,193 +2,138 @@
...
+        )
 
+        # Read in parameters and aggregation specs
         self._params = params if params is not None else {}
-        self._aggregate_by_group_specs = (
-            aggregate_by_group_specs if aggregate_by_group_specs is not None else {}
-        )
-        self._aggregate_by_p_id_specs = (
-            aggregate_by_p_id_specs if aggregate_by_p_id_specs is not None else {}
+        self._aggregations_tree = (
```
hmgaudecker, line 76:
```suggestion
        self._aggregation_specs_tree = (
```
Just to be sure, same for the public property.

src/_gettsim/policy_environment.py
```diff
@@ -257,6 +221,34 @@ def _parse_date(date):
     return date
 
 
+def _convert_all_functions_to_policy_functions(
```
hmgaudecker, line 224:
```suggestion
def _convert_function_to_policy_function(
```

src/_gettsim/policy_environment.py
```diff
@@ -257,6 +221,34 @@ def _parse_date(date):
...
+    function: callable,
+) -> PolicyFunction:
+    """Add module name if missing.
```
hmgaudecker, line 227: Docstring seems outdated.

src/_gettsim/policy_environment.py
```diff
@@ -620,18 +619,24 @@ def transfer_dictionary(remaining_dict, new_dict, key_list):
...
 
-def get_by_path(data_dict, key_list):
-    """Access a nested object in root by item sequence."""
-    return reduce(operator.getitem, key_list, data_dict)
-
-
-def set_by_path(data_dict, key_list, value):
-    """Set a value in a nested object in root by item sequence."""
-    get_by_path(data_dict, key_list[:-1])[key_list[-1]] = value
+def _fail_if_name_of_last_branch_not_leaf_name_of_function(
```
hmgaudecker, line 626:
```suggestion
def _fail_if_name_of_last_branch_element_not_leaf_name_of_function(
```
I'd think of "branch" == "tree path". Please adjust below, too.

hmgaudecker (commented):
Very neat! I did not make it through interface and shared yet (but did have a close look recently), same holds for the tests.

hmgaudecker (commented):
Excellent! Nearing completion.

hmgaudecker:
Afterthought: Instead of [replacing flatten_dict by optree in dags](https://github.com/OpenSourceEconomics/dags/pull/17#issuecomment-2634745721), I think the better option is to add an explicit dependency here as well. The tools just serve different purposes.

In particular, `optree` does not support unflattening without a `treespec` that is easily done using `flatten_dict`. At a quick glance, that could happen both based on tuples (=`optree.path`s) or qualified names. Should remove quite a bit of boilerplate code.

While the project seems dormant ([PyPI](https://pypi.org/project/flatten-dict/), [github](https://github.com/ianlini/flatten-dict)), it seems easy enough to fork and adjust should need arise. Thoughts?

src/_gettsim/shared.py
```diff
@@ -19,189 +24,155 @@ def __repr__(self):
...
+    return nested_dict
 
 
-def _check_for_conflicts_in_time_dependent_functions(
-    dag_key: str, function_name: str, start: date, end: date
-):
-    """
-    Raises an error if a different time-dependent function has already been registered
-    for the given dag_key and their date ranges overlap.
+def merge_nested_dicts(base_dict: dict, update_dict: dict) -> dict:
```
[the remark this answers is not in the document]
hmgaudecker: Probably, yes.

Although it did occur to me that we want to change terminology a bit, I think. Let's re-use the "upsert" thing for "update or insert".

And then we would have `upsert_tree` and `upsert_path_and_value`.

src/_gettsim/functions/loader.py
```diff
@@ -105,168 +194,188 @@ def _find_python_files_recursively(roots: list[Path]) -> list[Path]:
...
+def _fail_if_leaf_name_is_module_name(
+    policy_functions: list[PolicyFunction],
+    module_name: str,
+) -> None:
+    """No PolicyFunction should have the same leaf name as its module name."""
+    leaf_names = [func.leaf_name for func in policy_functions]
+    if module_name in leaf_names:
+        msg = format_errors_and_warnings(
+            f"PolicyFunctions in module {module_name} have the same leaf name as their "
+            "module name."
```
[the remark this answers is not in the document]
hmgaudecker: I'd love to think that is the reason I found it hard to come up with something. I am afraid that was not the case, though!

src/_gettsim/functions/policy_function.py
```diff
@@ -39,57 +38,27 @@ def __init__(  # noqa: PLR0913
         self,
         function: Callable,
         *,
-        module_name: str = "",
-        function_name: str | None = None,
-        start_date: date | None = None,
-        end_date: date | None = None,
+        leaf_name: str,
+        start_date: date = "0001-01-01",
```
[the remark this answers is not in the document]
MImmesberger: Maybe we should define the defaults outside of the function then? Else, we would need to
apply the decorator, where we currently create an instance of `PolicyFunction`.

Then, the syntax gets more confusing than necessary, but maybe I'm just not used to it:

```python
# Convert Callable to PolicyFunction
converted_function = policy_function(leaf_name=some_callable.__name__)(some_callable)
```

src/_gettsim/combine_functions_in_tree.py
```diff
@@ -0,0 +1,768 @@
...
+        derived_from = (qualified_name_source_col, p_id_to_aggregate_by)
+
+    return DerivedFunction(
+        aggregate_by_p_id_func,
+        leaf_name=aggregation_target,
+        derived_from=derived_from,
+    )
+
+
+def _get_tree_path_from_source_col_name(
```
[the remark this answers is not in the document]
MImmesberger: This is here purely for getting annotations from `functions_tree` and `TYPES_INPUT_VARIABLES` (that's why we don't need this in `loader`). So this will go if [the dags issue](https://github.com/OpenSourceEconomics/dags/issues/21) is resolved.

However, in `_create_aggregation_functions` we can move to tree_paths of the source column rather than the qualified name. Then it gets much clearer, that this is not about the DerivedFunction arguments (but their annotations).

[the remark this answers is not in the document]
MImmesberger: In this case, I could imagine `dags` to infer missing annotations in the functions tree and add them (to the derived functions in GETTSIMs case). But this is not exactly what you had in mind in the type checking issue, right?

src/_gettsim/functions/policy_function.py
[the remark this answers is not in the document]
hmgaudecker: I don't quite get that. When do we ever create a PolicyFunction directly, without going through the decorator?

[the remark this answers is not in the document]
MImmesberger: Whenever the user creates an instance of `PolicyEnvironment` with their own functions
tree or upserts a function into the functions tree. We could of course be strict about
functions being `PolicyFunctions`.

src/_gettsim/combine_functions_in_tree.py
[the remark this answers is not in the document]
hmgaudecker: I think we have two separate things here:
1. What do we potentially do in GETTSIM and what in dags?
2. Do we need this specific function here at all and what about annotations

Re 
1. I do think that we should expose some more things in `dags.dag_tree`, just to make sure that we won't do anything subtly different in GETTSIM. From a glance, this includes:
   - `QUALIFIED_NAME_DELIMITER` (currently `_qualified_name_delimiter`)
   - `is_qualified_name` (currently `_is_qualified_name`)
   - `map_parameter` (currently `_map_parameter`; this seems to be doing partly what this function does, except it does not return a path in the optree sense but a qualified name?)
   - Already exposed, but partly duplicated here: Many of the types defined there (will have to wait for [#24](https://github.com/OpenSourceEconomics/dags/pull/24), though there probably won't be big changes)
2. There won't be missing annotations if a function is to be aggregated. Can't be that hard to write `-> int` or the like. Doing that on the dags side won't happen, I think.

Finally, yes, way too many conversions between qualified names and tree paths in that module!

src/_gettsim/interface.py
```diff
@@ -2,654 +2,451 @@
...
     # Add rounding to functions.
     if rounding:
-        functions = _add_rounding_to_functions(functions, params)
+        policy_functions_tree = optree.tree_map_with_path(
+            lambda path, x: _add_rounding_to_function(x, params, path),
+            policy_functions_tree,
+        )
 
     # Partial parameters to functions such that they disappear in the DAG.
     # Note: Needs to be done after rounding such that dags recognizes partialled
```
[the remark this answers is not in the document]
MImmesberger: I'm quite sure the reason why the order of operations matters here are the attributes of the `PolicyFunction` instance that get destroyed by partialling. The comment in the codebase reads similarly (even though there its about attributes of the partialled function.

But this is something I can figure out for sure. Will try tomorrow.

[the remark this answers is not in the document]
hmgaudecker: Yes, I remember that happening for `func.__info__` and friends. I hoped that this would not apply any more now that we have `policy_func.skip_vectorization` etc.

However, looking at it, we still need to store `self.__annotations__` etc., so the previous behaviour may still apply.

src/_gettsim/functions/policy_function.py
```diff
@@ -37,59 +39,29 @@ class PolicyFunction(Callable):
...
         self,
-        function: Callable,
         *,
-        module_name: str = "",
-        function_name: str | None = None,
-        start_date: date | None = None,
-        end_date: date | None = None,
+        function: Callable,
+        leaf_name: str,
+        start_date: date = DEFAULT_START_DATE,
```
hmgaudecker, line 45: Sorry, my bad. With "the above syntax" I meant calling the decorator like `dec(dec_args)(func)`, not defining defaults. We'd have to do the same for the other three... And you want to see the defaults right in the interface, e.g., when hovering over a use of it as decorator.

src/_gettsim/functions/loader.py
```diff
@@ -148,13 +196,13 @@ def _load_module(path: Path, package_root: Path) -> ModuleType:
...
 
-def _convert_path_to_module_name(path: Path, package_root: Path) -> str:
+def _convert_path_to_importable_module_name(path: Path, package_root: Path) -> str:
     """
     Convert an absolute path to a Python module name.
 
     Examples
     --------
-    >>> _convert_path_to_module_name(RESOURCE_DIR / "taxes" / "functions.py")
+    >>> _convert_path_to_importable_module_name(RESOURCE_DIR / "taxes" / "functions.py")
```
hmgaudecker, line 205: Example can't work. Also, I am not sure whether the function is strictly correct, in order to use it directly for imports it would have to return `_gettsim.taxes.functions`, right?

src/_gettsim/functions/policy_function.py
```diff
@@ -108,52 +77,97 @@ def is_active_at_date(self, date: date) -> bool:
...
+def policy_function(
+    *,
+    start_date: str = "0001-01-01",
+    end_date: str = "9999-12-31",
+    leaf_name: str | None = None,
+    params_key_for_rounding: str | None = None,
+    skip_vectorization: bool = False,
+) -> PolicyFunction:
+    """
+    Decorator that wraps a callable into a `PolicyFunction`.
```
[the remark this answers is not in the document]
hmgaudecker: I don't think "wraps" is the correct terminology here?

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
-        isinstance(i, pd.Series) for i in data.values()
-    ):
-        pass
-    else:
-        raise NotImplementedError(
-            "'data' is not a pd.DataFrame or a pd.Series or a dictionary of pd.Series."
+    if debug:
+        results = upsert_tree(
+            base_tree=results,
+            update_tree=data_tree_with_correct_types,
```
hmgaudecker, line 147:
```suggestion
            to_upsert=data_tree_with_correct_types,
```
A bit pedantic, but let's match the function definition here.

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
-    elif isinstance(data, dict) and all(
-        isinstance(i, pd.Series) for i in data.values()
-    ):
-        pass
-    else:
-        raise NotImplementedError(
-            "'data' is not a pd.DataFrame or a pd.Series or a dictionary of pd.Series."
+    if debug:
+        results = upsert_tree(
+            base_tree=results,
```
hmgaudecker, line 146:
```suggestion
            base=results,
```
Clear enough from the function definition / annotation. But mostly because of the next line.

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
+            series=data_leaf, internal_type=internal_type
+        ):
             try:
-                data[column_name] = convert_series_to_internal_type(
-                    series, internal_type
+                converted_leaf = convert_series_to_internal_type(
+                    series=data_leaf, internal_type=internal_type
+                )
+                data_tree_with_correct_types = upsert_path_and_value(
+                    tree=data_tree_with_correct_types,
```
hmgaudecker, line 217:
```suggestion
                    base=data_tree_with_correct_types,
```
match interface of `upsert_tree`.

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
+                    tree_path=accessor.path,
```
hmgaudecker, line 218:
```suggestion
                    path_to_upsert=accessor.path,
```

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
+                    value=converted_leaf,
```
hmgaudecker, line 219:
```suggestion
                    value_to_upsert=converted_leaf,
```

src/_gettsim/interface.py
```diff
@@ -2,661 +2,453 @@
...
+    )[0]
 
-    return {k: v for k, v in data.items() if k not in unnecessary_data}
+    # Convert to numpy.ndarray
+    return optree.tree_map(lambda x: x.to_numpy(), input_data_tree)
 
 
-def _round_and_partial_parameters_to_functions(functions, params, rounding):
-    """Create a dictionary of all functions that are available.
+def _round_and_partial_parameters_to_functions(
```
hmgaudecker, line 315: Should just be 
```suggestion
def _partial_parameters_to_functions(
```
the line doing the rounding can be done in the calling code.

Feel free to postpone to a different PR.

src/_gettsim/shared.py
```diff
@@ -19,189 +21,170 @@ def __repr__(self):
...
-        [
-            "{formatted_list}",
-        ]
-        """
-    ).format(formatted_list=formatted_list)
+    The path is a list of strings that represent the keys in the nested dictionary. If
+    the path does not exist, it will be created. If the path already exists, the value
+    will be updated.
+    """
+    update_dict = create_tree_from_path_and_value(tree_path, value)
```
hmgaudecker, line 139:
```suggestion
    to_upsert = create_tree_from_path_and_value(tree_path, value)
```

src/_gettsim/shared.py
```diff
@@ -19,189 +21,170 @@ def __repr__(self):
...
-    """Parse None, str, and list of strings to list of strings.
+def partition_tree_by_reference_tree(
+    tree_to_partition: NestedFunctionDict | NestedDataDict,
+    reference_tree: NestedFunctionDict | NestedDataDict,
+) -> tuple[
+    NestedFunctionDict | NestedDataDict,
+    NestedFunctionDict | NestedDataDict,
+]:
+    """
+    Partition a tree into two separate trees based on the presence of its leaves in a
```
hmgaudecker, line 151:
```suggestion
    Partition a tree into two based on the presence of its paths in a reference tree.
```

src/_gettsim/shared.py
```diff
@@ -19,189 +21,170 @@ def __repr__(self):
...
+    reference tree.
```
hmgaudecker, line 152:
```suggestion
```

hmgaudecker (approved):
Excellent! Some very small comments left, no show-stoppers though.

MImmesberger:
> In particular, optree does not support unflattening without a treespec that is easily done using flatten_dict.

I don't have a good intuition about the extent to which we should rely on `flatten_dict`. In principle, the new functions in `shared.py` emulate functionality of `flatten_dict`. 

i) One obvious candidate for the unflattening functionality of `flatten_dict` is `tree_structure_from_paths` because we only ever need it to create a dict from qualified names of DAG nodes.

ii) In the other cases, we never really unflatten in the `optree` version, but create a new tree from an empty dict using the tree upsert functions in `shared.py`. This wouldn't be necessary with `flatten_dict`. (Take for example `partition_tree_by_reference_tree`)

However, if we don't want to touch (ii), I would vote against using `flatten_dict` for (i) because of the complexity it entails to "learn" a new library. Just not worth it for the few lines it would save us.

src/_gettsim/combine_functions_in_tree.py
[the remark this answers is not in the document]
MImmesberger: Re 1: Makes sense. Do we do this now or later? (After the type hint PR at `dags`?)

src/_gettsim/functions/loader.py
```diff
@@ -89,14 +169,15 @@ def _find_python_files_recursively(roots: list[Path]) -> list[Path]:
...
-        Absolute paths to all discovered Python files.
+    Absolute paths to all discovered Python files.
     """
     result = []
 
     for root in roots:
         if root.is_dir():
-            modules = list(root.rglob("*.py"))
+            modules = [
+                file for file in root.rglob("*.py") if file.name != "__init__.py"
```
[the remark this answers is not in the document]
MImmesberger: The rule is exactly the other way around. [Link](https://docs.astral.sh/ruff/rules/negate-equal-op/)

MImmesberger:
We use `flatten_dict` now in some places where I thought the code gets simpler by doing that. LMK if you disagree for some function.

hmgaudecker:
Excellent, thanks! Just had a brief look on the phone — why do you go back to qualified names? flatten would just return the tree path we need, no?

MImmesberger:
Yes, in some places we need the qualified name, in other places we could use the qualified name or the path. I thought that consistency would be nice, i.e. knowing that an object flattened by `flatten_dict` will always have qualified names as keys. That way one doesn't have to look that up. 

But if you disagree, I can easily change that.

MImmesberger:
As far as I can tell, the only unresolved issue are the function attributes of `PolicyFunction`/`DerivedFunction` and the partialled functions. I didn't manage to do this today, but as we put the rounding step outside of the critical function, the priority is somewhat lower I'd say.

I will create an issue for that such that we can tackle it later.

hmgaudecker:
> Yes, in some places we need the qualified name, in other places we could use the qualified name or the path (just doesn't matter there). I thought that consistency would be nice, i.e. knowing that an object flattened by `flatten_dict` will always have qualified names as keys. That way one doesn't have to look that up.
> 
> But if you disagree, I can easily change that.

Yes, I do :wink:. I just changed that. It is clear enough from the arguments to `tree_flatten` whether we get path or a qualified name back.

Issue #627, opened 2023-07-26
ENH: Make DEFAULT_TARGETS dependent on the policy environment.


hmgaudecker:
#### Is your feature request related to a problem?

DEFAULT_TARGETS currently includes items that are not present in all years, which should throw an error. See [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/618#issuecomment-1651473444)

#### Describe the solution you would like to see

Make DEFAULT_TARGETS dependent on the policy date.

amageh:
Hi! I think this is an excellent idea. I want to add another related issue I came across when using `set_up_policy_environment` recently (I think it makes sense to add this here instead of opening a new issue, sorry if it was already raised somewhere else):

Maybe it would be good idea to add a warning to the function for dates that GETTSIM doesn't support or does not support very well. This applies to dates very far in the past but especially more recent policy environments. 

For instance, the most recent version of GETTSIM is from summer 2023 and it thus probably doesn't include parameters for 2024 yet. In my opinion it would make sense to issue a warning to users in these cases since I don't expect everyone will check for this themselves.

Issue #566, opened 2023-04-04
ENH: Expand and improve parameters for Unterhaltsvorschuss


ChristianZimpelmann:
#### Is your feature request related to a problem?

GETTSIM produces an error whenever `unterhaltsvors` is calculated before 2017 since parameters in `unterhaltsvors.yaml` are not specified for the time before.

Before #544, the parameters were defined from 2016 on, but we somehow lost 1 year. See #479.

#### Describe the solution you would like to see

- Look for law references for the time before
- Expand time frame
- When working on it: `altersgrenzen` should be a list instead of a dict as the keys are meaningless as far as I understand.

PR #785, opened 2024-08-03
Test whether `DEFAULT_TARGETS` can be computed with `create_synthetic_data` output

### LINKED ISSUE #733: ENH: Implement a test running GETTSIM over synthetic data

Implement a test running GETTSIM over synthetic data and check that no error is raised.

Currently, we only test for the correct creation of the synthetic data.

Following up on the discussion in #727

MImmesberger:
Closes #733

I let the test run through the years 2017-2024. Extending further back does not work because the Unterhaltsvorschuss isn't implemented prior to 2017 (as discussed in #566). 

I also added a `NotImplementedError` for the Unterhaltsvorschuss until 2016.

Related to #627

### DIFF
--- a/src/_gettsim_tests/test_synthetic.py
+++ b/src/_gettsim_tests/test_synthetic.py
@@ -293,7 +293,7 @@ def test_p_id_groups(fixture, expected, request):
 
 @pytest.mark.parametrize(
     "fixture, policy_date",
-    [("synthetic_data_couple_with_children", i) for i in range(2017, 2024)],
+    [("synthetic_data_couple_with_children", y) for y in range(2017, 2024)],
 )
 def test_default_targets(fixture, policy_date, request):
     policy_params, policy_functions = set_up_policy_environment(policy_date)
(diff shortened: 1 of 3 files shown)


src/_gettsim_tests/test_synthetic.py
```diff
@@ -286,3 +289,17 @@ def test_p_id_groups(fixture, expected, request):
     df = request.getfixturevalue(fixture)
     for col, values in expected.items():
         pd.testing.assert_series_equal(df[col], pd.Series(values, name=col))
+
+
+@pytest.mark.parametrize(
+    "fixture, policy_date",
+    [("synthetic_data_couple_with_children", i) for i in range(2017, 2024)],
```
hmgaudecker, line 296:
```suggestion
    [("synthetic_data_couple_with_children", y) for y in range(2017, 2024)],
```

More readable.

hmgaudecker (approved):
Great, thanks!

PR #786, opened 2024-08-07
Remove `eigenbedarf_gedeckt` from Kindergeldübertrag

### LINKED ISSUE #758: ENH: Remove `eigenbedarf_gedeckt` in Kindergeldübertrag
Kindergeldübertrag uses `eigenbedarf_gedeckt` to make sure that the child is not in the same Bedarfsgemeinschaft as parents (because then, Kindergeld may be counted as income twice). Conditioning on this exogenous variable is not necessary, we can either

- Make sure that the child is not in the same BG as parents via an endogenous variable
- Calculate income in a way that the double Kindergeld bug doesn't occur (preferred)

MImmesberger:
Closes #758

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
+- {gh}`786` Check directly whether child and parent are in same Bedarfsgemeinschaft for
+  Kindergeldübertrag ({ghuser}`MImmesberger`).
 - {gh}`772` Add Mindesteinkommen check to Wohngeld, calculate anz_eig_kind_bis_24
   instead of requiring it as an input variable ({ghuser}`MImmesberger`).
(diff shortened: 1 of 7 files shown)


hmgaudecker (commented):
Looks great, thanks! 

However, I think the scope should be larger. We should not require `eigenbedarf_gedeckt` as an input variable. Maybe I am missing again the current state of things, but with #771 merged, it should be just a matter of comparing income and needs at the individual level, no?

MImmesberger:
The only occurrence of `eigenbedarf_gedeckt` now is in the `bg_id` creation. I left it there intentionally because then we have a way to manually set the BGs until we get #763 right. If we replace it with the correct earnings check, we cannot compute scenarios where children + parents are in the same BG even when children have covered their needs for the time being (relevant for KiZ I think).

hmgaudecker:
Fine. Then maybe add a to-do next to the input variable and in the `bg_id` creation, both linking #763. Just so that this is very visible!

hmgaudecker (approved):

Issue #667, opened 2023-11-28
ENH: Class for policy environment


lars-reimann:
#### Is your feature request related to a problem?

A policy environment consists of functions and their parameters. It is usually created by `set_up_policy_environment`. However, `set_up_policy_environment` returns the parameters and functions separately in a tuple. Likewise, `compute_taxes_and_transfers` expects parameters and functions to be provided separately. This usually leads to code that is similar to this:

```py
policy_params, policy_functions = set_up_policy_environment(
    date=test_data.date
)

result = compute_taxes_and_transfers(
    data=df, params=policy_params, functions=policy_functions, targets=column
)
```

#### Describe the solution you would like to see

It would be cleaner to bundle functions and their parameters into an object, so they always get passed around together. Thus, I propose to add a `PolicyEnvironment` class.

The class can also provide methods for common use-cases like overwriting some specific function, while ensuring that functions and parameters don't get out of sync.

#### Interface breaking implications

* `set_up_policy_environment` would return a `PolicyEnvironment`.
* `compute_taxes_and_transfers` would take a `PolicyEnvironment` instead of the parameters and functions.

#### Describe alternatives you've considered

-

#### Additional context

Related to #538.

hmgaudecker:
Fixed by #787.

Issue #538, opened 2023-02-21
ENH: Class for policy functions


lars-reimann:
#### Is your feature request related to a problem?

Users can attach additional information to policy functions using decorators, which provides a nice interface for them to work with. Often, we cannot immediately transform the decorated function, however, so we need to store the arguments of the decorator somewhere. Since decorators can be applied in any order and can potentially wrap a function, adding a property to the decorated function and ensuring it gets carried along is the best solution. Currently, this data is stored in a dictionary called  [`__info__`](https://github.com/iza-institute-of-labor-economics/gettsim/pull/534).

Once, all decorators are processed, however, the dictionary is cumbersome to work with:
1. We need to ensure the dictionary exists, which is not the case for functions without decorators.
2. We need to ensure specific keys exist before accessing them or provide default values.
3. Any logic needs to be defined externally (e.g. we cannot just ask a policy function whether it is active at a given date).

#### Describe the solution you would like to see

Once all decorators are processed, create an instance of the following class:

```py
class PolicyFunction:
    def __init__(
        self,
        f: Callable,
        dag_key: str,
        start_date: datetime.date,
        end_date: datetime.date
    ):
        ...

    def is_active_at_date(date: datetime.date) -> bool:
        ... 
```

Any code downstream should work with policy functions instead of normal Python callables.

If https://github.com/OpenSourceEconomics/dags/issues/13 gets implemented, `PolicyFunction` could also extend `DagFunction` and pass the dag_key along to the super constructor.

#### Interface breaking implications

Only changes how we internally work with policy functions.

#### Additional context

Discussion started [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/534#discussion_r1109684925).

hmgaudecker:
Fixed by #787.

Issue #781, opened 2024-07-19
ENH: Interface, 2024 edition


hmgaudecker:
#### Is your feature request related to a problem?

GETTSIM's interface is hard to use and inflexible. We discussed the direction we should be heading at the 2024 workshop; this issue summarizes the discussion and opens it for community members who were not present. 

Several related issues:

- Aspects of #471 ([rendered version](https://gettsim--471.org.readthedocs.build/en/471/geps/gep-06.html))
- #378
- #449 
- #531 
- #532 
- #538 and #667 
- #640 

### Problem

GETTSIM's approach of being precise and requiring all relevant inputs means that simple questions can only be answered either with a huge amount of input data or with detailed knowledge of the DAG (e.g., how to cut off the pensions calculations part of the DAG when we have a sample of 30-year olds and ignore DI).

### Proposed solution

1. Provide a repo in the ecosystem with default values for broadly defined "personas"  (low-income households, rich people, families with children, pensioners, ...)
1. Provide an interactive interfaces for selecting subgraphs (more below)

### Problem

It is very hard to see how GETTSIM's requirements match what is in the data. For any reasonable application, the graph is far too fine-grained. It is very hard to see what kinds of intermediate nodes there might be in the data already

### Proposed solution

1. Simple stuff: A higher-level view of the graph (just the upper-level keys in the hierarchical structure, like `arbeitslosengeld_2`, `rentenversicherung`, ... width of edges mirrors the amount of links between individual functions, click on an area opens the internal structure), coloring of nodes following the upper-level keys.
1. Iterative protocol for setting up the graph
   1. User specifies the targets
   2. load policy functions (potentially including a reform)
   3. Select input variables (either just take all root nodes or interactive process outlined below)
   4. Use that directly if names in the data match, or provide a dictionary/yaml file (depending on calling language) with all required inputs, user can fill in values with the corresponding names in her dataset or a default value that will apply to all observations

### Problem

Some stuff needs to be done outside the graph (e.g., Günstigerprüfung with endogenous `bg_id`s). Sometimes, we want to use information from the data in order to calculate things (e.g., whether someone who has a Minijob contributes voluntarily to the pension insurance or not).

### Proposed solution

Set up sparse dataset with duplicated observations (0/1 take up in the above example; no concern if Minijob is irrelevant), potentially call GETTSIM on subsets of the data in sequence. *(needs more thought)*

## Interactive interface

- Notebook or other dashboard-style UI
[...]

PR #787, opened 2024-08-12
Model classes for policies

lars-reimann:

#538
#667

Define custom model classes for policy functions and policy environments. These are now used instead of ad-hoc data structures like dictionaries or lists. This also affects the main public interface:

* `compute_taxes_and_transfers` expects a `PolicyEnvironment` instead of separate arguments for `params` and `functions`
* `set_up_policy_environment` returns a `PolicyEnvironment` instead of a tuple of `params` and `functions`. There's now also a factory method `PolicyEnvironment.for_date` that can be used instead of `set_up_policy_environment`. We might eventually remove one of these options.

This change prepares later changes like the introduction of namespaces or handling parameters using the DAG.

### Todo


src/_gettsim/policy_environment.py
```diff
@@ -1,59 +1,166 @@
...
-def set_up_policy_environment(date):
-    """Set up the policy environment for a particular date.
 
-    Parameters
-    ----------
-    date : int, str, datetime.date
-        The date for which the policy system is set up.
+class PolicyEnvironment:
+    """
+    A container for policy functions and parameters. You should usually use the factory
```
hmgaudecker, line 29:
```suggestion
    A container for policy functions and parameters. 
    
    Almost always, instances are created with :PolicyEnvironment.for_date()`.
```

src/_gettsim/policy_environment.py
```diff
@@ -1,59 +1,166 @@
...
+    method :meth:`for_date` to create a policy environment instead of calling the
```
hmgaudecker, line 30:
```suggestion
```

src/_gettsim/policy_environment.py
```diff
@@ -1,59 +1,166 @@
...
+    constructor directly.
```
hmgaudecker, line 31:
```suggestion
```

src/_gettsim/policy_environment.py
hmgaudecker: I do appreciate the deeper explanation, but I also think it will be confusing to most potential readers.

src/_gettsim/policy_environment.py
```diff
@@ -1,59 +1,166 @@
...
 
-    """
-    # Check policy date for correct format and transfer to datetime.date
-    date = _parse_date(date)
+        return result
 
-    params = {}
-    for group in INTERNAL_PARAMS_GROUPS:
-        params_one_group = _load_parameter_group_from_yaml(date, group)
+    def replace_all_parameters(self, params: dict[str, Any]):
```
hmgaudecker, line 126: Just a mental note and link to #781 that we would probably want something like `upsert_params` to just change tiny portions. But precise interface will depend on how parsing of params and their inclusion in DAG will look like.

src/_gettsim/policy_function.py
```diff
@@ -0,0 +1,136 @@
...
+from typing import Any, TypeVar
+
+import numpy
+
+T = TypeVar("T")
+
+
+class PolicyFunction(Callable):
+    """
+    A function that computes an output vector based on some input vectors.
```
hmgaudecker, line 16:
```suggestion
    A function that computes an output vector based on some input vectors and/or parameters.
```

hmgaudecker (commented):
This is much, much nicer than what we have. Thanks a lot!!! 

I would appreciate some big picture documentation on how things fit together. Could also be more of a long-term vision of how PolicyEnvironment and PolicyFunctions should play together. Happy to chat and help develop that.

MImmesberger (commented):
I'm a big fan of these changes! Thank you! I only looked at the documentation and tried it out myself (haven't looked at the implementation in the source code).

I think this breaks most applications of GETTSIM, though. Given that, we need to think about whether we want to keep `set_up_policy_environment` at all.

Additionally, just some comments regarding the documentation. Happy to help if you don't find the time.

src/_gettsim/functions_loader.py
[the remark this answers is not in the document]
lars-reimann: The original function loader served two separate purposes:

1. It discovered internal/user functions. This is now included in the new function loader (and only this). This was used during `set_up_policy_environment`.
2. It preprocessed functions (vectorization etc.) and added derived functions. It still does this now. This was used during `compute_taxes_and_transfers`.

Fitting names would be

* `functions_loader` -> `functions_preprocessor`,
* `functions_loader_new` -> `functions_loader`.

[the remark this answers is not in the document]
hmgaudecker: Great, please go ahead and rename!

lars-reimann:
> Additionally, just some comments regarding the documentation. 

We'll probably need to do another pass over the tutorial once all the interface changes are implemented. Thanks for the suggestions, though, I'll apply these.

docs/tutorials/advanced_usage.ipynb
```diff
@@ -56,7 +56,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "policy_params, policy_functions = set_up_policy_environment(\"2020\")"
+    "environment = PolicyEnvironment.for_date(\"2020\")"
```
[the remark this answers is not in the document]
hmgaudecker: We should actually try to be precise everywhere, hence I'd suggest:

```suggestion
    "environment = PolicyEnvironment.for_date(\"2020-01-01\")"
```

But feel free to leave as you like.

docs/tutorials/policy_functions.ipynb
```diff
@@ -363,8 +304,7 @@
    "source": [
     "df = compute_taxes_and_transfers(\n",
     "    data=data,\n",
-    "    params=policy_params,\n",
-    "    functions=[policy_functions, arbeitsl_geld_2_m_bg, kindergeld_m],\n",
+    "    environment=environment.upsert_functions(arbeitsl_geld_2_m_bg, kindergeld_m),\n",
```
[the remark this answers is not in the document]
lars-reimann: I'd defer this part until the interface is finalized. Once parameters are also included in the DAG, this section might require updates.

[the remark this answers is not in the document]
lars-reimann: > I think the current default is that new functions are not vectorized. Imo it would make sense to change the default towards vectorization.

If users pass a `PolicyFunction`, they can set the `skip_vectorization` parameter appropriately. By default, it's `False`, so the function gets vectorized. If users pass another callable, a `PolicyFunction` is created, which always gets vectorized.

docs/how_to_guides/different_ways_to_load_policy_functions.ipynb
```diff
@@ -33,7 +33,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "policy_params, policy_functions = set_up_policy_environment(2020)"
+    "environment = set_up_policy_environment(2020)"
```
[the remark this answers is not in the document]
lars-reimann: I'd be fine with removing it. In my opinion, `PolicyEnvironment.for_date` reads better.

docs/tutorials/policy_functions.ipynb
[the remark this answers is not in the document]
hmgaudecker: No, because it would also replace (update) existing ones.

Although, thinking of it, Python's  `dict.update` has the same behavior, so maybe `update_policy_functions` is clear enough. I would indeed expand `functions`.

No reason to spend lots of time on it as things may change yet again once we include parameters in the DAG.

src/_gettsim/functions/policy_function.py
```diff
@@ -0,0 +1,137 @@
...
+        self.skip_vectorization: bool = _first_not_none_or_none(
+            skip_vectorization, info.get("skip_vectorization"), False
+        )
+
+        self.function = (
+            function if self.skip_vectorization else _vectorize_func(function)
+        )
+        self.module_name = module_name
+
+        self.name_in_dag: str = _first_not_none_or_none(
```
hmgaudecker, line 60: I'd prefer two versions: `_first_not_none` for the first three use cases (just replace the `return None` by raising an error) and `_first_not_none_or_none` for the rounding case. Took me a while to figure out what exactly is happening here.

src/_gettsim/functions/policy_function.py
```diff
@@ -0,0 +1,137 @@
...
+from collections.abc import Callable
+from datetime import date
+from typing import Any, TypeVar
+
+import numpy
+
+T = TypeVar("T")
+
+
+class PolicyFunction(Callable):
```
hmgaudecker, line 14: Even though the coverage report is probably nonsense, wouldn't a set of tests for setting up the instances of the class be useful?

src/_gettsim/functions/preprocessor.py
```diff
@@ -38,12 +38,18 @@
 )
 from _gettsim.time_conversion import create_time_conversion_functions
 
+from .policy_function import PolicyFunction
+
 if TYPE_CHECKING:
     from collections.abc import Callable
```
hmgaudecker, line 44: Unrelated, I know -- but why do we need the conditioning?

src/_gettsim/functions/preprocessor.py
```diff
@@ -127,7 +126,7 @@ def load_and_check_functions(
 
 
 def _create_derived_functions(
```
hmgaudecker, line 128: I think I'd prefer being stricter about `PolicyFunction`s here -- just replace `Callable` by that in all code underneath this call. Makes it clearer what is the purpose of things instead of relying on the sanitizing in `PolicyFunction.__init__`.

Also, instead of returning a tuple with three `dict[str, PolicyFunction]`s, let's just return a flat dictionary with all three tuple items. We never need a distinction of those again.

MImmesberger (approved):
Looks good to me!

In my commit, I just switched from `policy_functions` to `environment` in one cell in the basic tutorial. Still some adjustments left to do in the other tutorials (different ways to load policy functions notebook) but makes sense to postpone this.

src/_gettsim_tests/test_functions_loader.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 
 from _gettsim.config import RESOURCE_DIR
-from _gettsim.functions_loader import (
```
hmgaudecker, line 10: This will need updating for the new structure.

E.g., `_load_functions` is left in `functions.preprocessor` but should actually be deleted because it has moved to `functions.loader`. 

In the course of that, we should talk about the structure of those two modules. E.g., I would not expect a function `load_and_check_functions` in `functions.preprocessor`, I would have expected that as part of the setup of the `PolicyEnvironment` already? But I guess that one does not know about the targets yet?

We may also leave as is for the moment and do a quick cleanup PR ex post. But I think clearing up a little more would help for the changes ahead. Let's discuss.

hmgaudecker (commented):
Excellent, thanks a lot! 

A few smaller comments; then I realised the tests still depend on leftover code in `functions.preprocessor`. Let's discuss on Zoom what would be a good scope for changes in this PR.

src/_gettsim/functions/loader.py
```diff
@@ -0,0 +1,183 @@
...
+
+    return PolicyFunction(
+        function=function,
+        module_name=clean_module_name,
+    )
+
+
+def _is_function_defined_in_module(function: Callable, module_name: str) -> bool:
+    """Check if a function is defined in a specific module or only imported."""
+    return inspect.getmodule(function).__name__ == module_name
```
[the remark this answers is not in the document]
lars-reimann: Module names fully qualified, e.g. `_gettsim.social_insurance_contributions.eink_grenzen`.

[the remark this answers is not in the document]
hmgaudecker: Excellent, makes perfect sense! Apologies for consulting you instead of the Python docs.

src/_gettsim/functions/preprocessor.py
hmgaudecker: Update after talking on the phone, just for the public record: This will become something like `PolicyFunction | DerivedFunction`

hmgaudecker:
Is everything done so I should have a final look?

hmgaudecker:
Ah perfect -- just wasn't sure about the status. Thanks!

lars-reimann:
@hmgaudecker @MImmesberger This is ready for review again.

Since last time, I've mainly included the aggregation specs in the policy environment and added some tests for the methods of the policy environment.

hmgaudecker (commented):
Essentially done, just a few tiny comments and one possibly larger one re the `DerivedFunctions`. 

I think we can consider the Windows CI failure a fluke. Will be fixed by #794, anyhow.

MImmesberger (commented):
I think time-wise it doesn't make much sense for me to go through it line-by-line. Just briefly looked at the changes on the train and no complaints from my side.

hmgaudecker:
> I think we can consider the Windows CI failure a fluke. Will be fixed by #794, anyhow.

Saw a one-line fix in [another repo](https://github.com/OpenSourceEconomics/econ-project-templates/pull/162/commits/f582dff8ba824f9c3172a91523da612418296c77) and tried it out here.

src/_gettsim/functions/preprocessor.py
lars-reimann: Union types can be hard to use, since they only allow access to operations that are defined on all its elements. Otherwise, `isinstance` checks must be added, which can affect many locations in the code.

Generally, I don't see a major benefit of adding a completely separate class for `DerivedFunction` instead of using `PolicyFunction` for them, too: The fields `function`, `module_name`, and `function_name` are also required for derived functions. `start_date` and `end_date` are useful information. Rounding and vectorization are questionable, but could still enable some use-cases.

Tracking information could become part of `PolicyFunction` and be `None` for non-derived functions. If `DerivedFunction` was a subclass of `PolicyFunction`, we would first have to add an `isinstance` check, before we can access which function we derived from, which is redundant.

hmgaudecker: Thanks! From my perspective, it is mostly about communicating intent and source (written code vs. automatically generated). Granted, some of the `PolicyFunction`s are not about (public) policy at all, either, but for sure none of the `DerivedFunction`s is. So the name would be off.

I would be fine with just making `DerivedFunction` have the same signature as `PolicyFunction` and fill the fields with default values. 

This said, I think that `start_date` and `end_date` should be irrelevant at the point where `DerivedFunction`s come into play *(or am I missing something? Only happens once we condition on a particular point in time, no?)* so not accessing them might even be a helpful sanity check.

lars-reimann: > This said, I think that start_date and end_date should be irrelevant at the point where DerivedFunctions come into play

Yes, they would just be used for introspection.

lars-reimann: @hmgaudecker The class for derived functions is now added, too.

src/_gettsim/functions/derived_function.py
```diff
@@ -0,0 +1,44 @@
...
+        The function from which the new function is derived. If the function is derived
+        from a data column, this may be None.
+    """
+
+    def __init__(
+        self,
+        function: Callable,
+        function_name: str,
+        *,
+        derived_from: PolicyFunction | None = None,
```
hmgaudecker, line 32:
```suggestion
        derived_from: PolicyFunction | str | Tuple[str, str],
```
I think that should cover all cases when we are being explicit here.

src/_gettsim/policy_environment_postprocessor.py
```diff
@@ -630,30 +423,27 @@ def aggregate_by_group_func(source_col, group_id):
         else:
             raise ValueError(f"Aggr {agg_specs['aggr']} is not implemented.")
 
-    return aggregate_by_group_func
+    return DerivedFunction(
+        aggregate_by_group_func,
+        function_name=agg_col,
+    )
```
hmgaudecker, line 429:
```suggestion
        derived_from=derived_from,
    )
```

In if/else block above, incorporate:

```py
    if agg_specs["aggr"] == "count":
        derived_from = group_id
    else:
        derived_from = (source_col, group_id)
```

src/_gettsim/policy_environment_postprocessor.py
```diff
@@ -771,14 +562,20 @@ def aggregate_by_p_id_func(column, p_id_to_aggregate_by, p_id_to_store_by):
         else:
             raise ValueError(f"Aggr {agg_specs['aggr']} is not implemented.")
 
-    return aggregate_by_p_id_func
+    return DerivedFunction(
+        aggregate_by_p_id_func,
+        function_name=agg_col,
+    )
```
hmgaudecker, line 568:
```suggestion
        derived_from=derived_from,
    )
```

In if/else block above, incorporate:

```py
    if agg_specs["aggr"] == "count":
        derived_from = p_id_to_aggregate_by
    else:
        derived_from = (column, p_id_to_aggregate_by)
```

src/_gettsim_tests/test_policy_environment.py
hmgaudecker, line 1: Would it be worth it having an extra set of tests for the post-processing module?

hmgaudecker (commented):
Extremely nice, many thanks!!! 

Just a few things where I think we should be a bit more explicit about the `derived_from` attribute and other minor stuff. Almost there, thanks a lot!

src/_gettsim_tests/test_policy_environment.py
lars-reimann: The functions in there were not added by this PR, so writing tests is out of scope, in my opinion. Nevertheless, additional tests would be useful, of course.

hmgaudecker: Fair enough, I didn't quite realise that.

src/_gettsim/functions/derived_function.py
```diff
@@ -0,0 +1,44 @@
...
+            function,
+            module_name=derived_from.module_name if derived_from is not None else "",
+            function_name=function_name,
+            start_date=derived_from.start_date if derived_from is not None else None,
+            end_date=derived_from.end_date if derived_from is not None else None,
+            params_key_for_rounding=None,
+            skip_vectorization=True,
+        )
+
+        self.derived_from: PolicyFunction | str = derived_from
```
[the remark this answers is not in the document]
lars-reimann: That was just a leftover from an earlier design.

hmgaudecker (approved):
Great work, many thanks!

PR #788, opened 2024-08-15
Add earnings test for pension benefits

MImmesberger:

Until 2022, there was an earnings test for early retirees.

- If earnings are above some threshold, there is a 40% benefit deduction rate
- If earnings plus pension exceeds the highest labor market income of the last 15 years, there is a 100% deduction rate (Hinzuverdienstdeckel). For this, I added a new input variable `höchster_bruttolohn_letzte_15_jahre_y`.

This only applies for Vollrenten, we didn't implement Teilrenten yet.


src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    alter
+        See :func:`alter`.
+    ges_rente_regelaltersgrenze
+        See :func:`ges_rente_regelaltersgrenze`.
+
+    Returns
+    -------
+
+    """
+    if alter >= ges_rente_regelaltersgrenze:
```
MImmesberger, line 89: Here, I need the age with monthly precision. This hinges on the new interface (you discussed adding an `evaluation_date` argument from which we could calculate the age in #211 and #378).

It would be fine for me to keep it imprecise for now and tackle this later. We could also add another input variable `evaluation_month` or something like this in the meantime. What do you think @hmgaudecker ?

hmgaudecker: Yeah, just leave it for now. Would be an unfortunate point in time to tackle this. Linking #781 so we'll keep it in mind.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    out = ges_rente_vor_grundr_m + grundr_zuschlag_m if rentner else 0.0
+    return out
+
+
+@policy_info(
+    end_date="2022-12-31",
+    name_in_dag="ges_rente_vor_grundr_m",
+    params_key_for_rounding="ges_rente",
+)
+def ges_rente_vor_grundr_mit_hinzuverdienstgrenzen_m(
```
hmgaudecker, line 63: Let's remove this level. `vor_grundrente` is confusing the hell out of me, because it could be meant in a temporal sense or a DAG-sequential sense.

Just move the age check into `bruttorente_mit_hinzuverdienstgrenzen`.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
     return out
 
 
-@policy_info(params_key_for_rounding="ges_rente")
-def ges_rente_vor_grundr_m(
+@policy_info(end_date="2020-12-31")
+def ges_rente_m(ges_rente_vor_grundr_m: float) -> float:
```
hmgaudecker, line 23:
```suggestion
def ges_rente_m(bruttorente_m: float) -> float:
```

See below.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    return ges_rente_vor_grundr_m
```
hmgaudecker, line 24:
```suggestion
    return bruttorente_m
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+
+
+@policy_info(
+    start_date="2021-01-01",
+    name_in_dag="ges_rente_m",
```
hmgaudecker, line 29:
```suggestion
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    params_key_for_rounding="ges_rente",
+)
+def ges_rente_m_nach_grundr(
```
hmgaudecker, line 32:
```suggestion
def ges_rente_m(
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,104 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    ges_rente_vor_grundr_m: float,
```
hmgaudecker, line 33:
```suggestion
    bruttorente_m: float,
```

src/_gettsim/transfers/rente.py
```diff
@@ -68,40 +164,121 @@ def ges_rente_vor_grundr_m(
     return out
 
 
-@policy_info(end_date="2020-12-31")
-def ges_rente_m(ges_rente_vor_grundr_m: float) -> float:
-    return ges_rente_vor_grundr_m
+@policy_info(end_date="2022-12-31")
```
hmgaudecker, line 167:
```suggestion
@policy_info(start_date="2023-01-01")
def bruttorente_m(
    bruttorente_ohne_einkommensanrechnung_m: float,
):
    return bruttorente_ohne_einkommensanrechnung_m

@policy_info(end_date="2022-12-31", change_name="bruttorente_m")
```

src/_gettsim/transfers/rente.py
```diff
@@ -68,40 +164,121 @@ def ges_rente_vor_grundr_m(
...
-    return out
+
+    rentenabzug = max(
+        bruttorente_nach_teilweiser_einkommensanrechnung_m
+        + bruttolohn_m
+        - höchster_bruttolohn_letzte_15_jahre_m,
+        0.0,
+    )
+
+    return max(bruttorente_ohne_einkommensanrechnung_m - rentenabzug, 0.0)
```
hmgaudecker, line 250: max() is redundant, AFAICT

src/_gettsim/transfers/rente.py
```diff
@@ -68,40 +164,121 @@ def ges_rente_vor_grundr_m(
...
+
+    """
+    if bruttolohn_m <= ges_rente_params["hinzuverdienstgrenze"] / 12:
+        out = bruttorente_ohne_einkommensanrechnung_m
+    # Teilrente unter Hinzuverdienstdeckel; 40% deduction rate
+    elif (
+        bruttolohn_m + bruttorente_ohne_einkommensanrechnung_m
+        <= höchster_bruttolohn_letzte_15_jahre_m
+    ):
+        out = bruttorente_nach_teilweiser_einkommensanrechnung_m
```
hmgaudecker, line 208: I think the calculations are simple enough to include them here, more time-consuming to look all functions up than helpful to think at different levels. But I do not have a terribly strong opinion.

hmgaudecker (commented):
Great, thanks! I think the structure of the code can become somewhat clearer.

Have not looked into the laws etc. (i.e., whether the functions make sense) -- would be great if you could double-check that, @TeBackh !

src/_gettsim/transfers/rente.py
MImmesberger: This doesn't work because we already have another function `ges_rente_m`. I chose `ges_rente_mit_grundrente_m` for now.

MImmesberger:
While working on the comments I realized that the rule I implemented was partly introduced in 2017. Before that, there was a hard Hinzuverdienstgrenze above which pension benefits were set to 0. I'll implement the pre-2017 rule as well.

src/_gettsim/transfers/rente.py
hmgaudecker: I cannot look into it right now, but I am fairly sure the way I put it in the morning should with. No overlapping `ges_rente_m` at any point in time, that is.

MImmesberger:
Done!

docs/gettsim_objects/input_variables.md
```diff
@@ -555,6 +555,13 @@ implemented in 2006 (§ 237 SGB VI Abs. 5).
 
 Type: bool
 
+(höchster_bruttolohn_letzte_15_jahre_y)=
+
+## `höchster_bruttolohn_letzte_15_jahre_y`
```
hmgaudecker, line 560: Just curious: Say I'm early retired at 61 and my highest wage was at 46. What happens the next year? Does wage at 46 still count or does it drop? Put differently, is it "the 15 years before the first pension claim" or "the past 15 years from today"?

Please clarify in name and docstring.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    start_date="2021-01-01",
+    params_key_for_rounding="ges_rente",
+    name_in_dag="ges_rente_m",
+)
+def ges_rente_mit_grundrente_m(
+    bruttorente_m: float,
+    grundr_zuschlag_m: float,
+    rentner: bool,
+) -> float:
+    """Calculate total individual public pension including Grundrentenzuschlag. Is only
```
hmgaudecker, line 32:
```suggestion
    """Calculate total individual public pension including Grundrentenzuschlag.
```
No need to re-state what is obvious from the decorator.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    active after 2021 when Grundrente is in place.
```
hmgaudecker, line 33:
```suggestion
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    grundr_zuschlag_m
+        See :func:`grundr_zuschlag_m`.
+    rentner
+        See basic input variable :ref:`rentner <rentner>`.
+
+    Returns
+    -------
+
+    """
+    # Return 0 if person not yet retired
```
hmgaudecker, line 48:
```suggestion
```
Don't state the obvious.

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    out = bruttorente_m + grundr_zuschlag_m if rentner else 0.0
+    return out
+
+
+@policy_info(end_date="2020-12-31")
+def ges_rente_m(bruttorente_m: float) -> float:
+    return bruttorente_m
+
+
+@policy_info(
```
hmgaudecker, line 58: Move down (at least I thought that typically we order things as time progresses, not latest-on-top).

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+        bruttolohn_m <= ges_rente_params["hinzuverdienstgrenze"] / 12
+    ):
+        out = bruttorente_basisbetrag_m
+    else:
+        out = 0.0
+
+    return out
+
+
+def bruttorente_basisbetrag_m(
```
hmgaudecker, line 196: Pls set some starting date here, e.g. this version cannot exist before 3 Oct 1990

src/_gettsim/transfers/rente.py
```diff
@@ -19,8 +19,181 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
     ges_rente_zugangsfaktor: float,
     entgeltp_ost_update: float,
```
hmgaudecker, line 198: Apologies I this comes up only now: Why is this the "_update" version? AFAIU, the update is what happens because of *concurrent* wage earnings. How can they affect pension payments in real time?

Would affect all tests added here and probably more.

hmgaudecker (commented):
Cool, thanks! Just a few minor things left!

docs/gettsim_objects/input_variables.md
MImmesberger: Good catch, it's 15 years before pension claiming.

src/_gettsim/transfers/rente.py
MImmesberger: I'll set it to January 1992 (SGB VI came into effect then).

src/_gettsim/transfers/rente.py
MImmesberger: Oh yes this bug was in there for ages. Fixed!

hmgaudecker: Cool, thanks! While you are at it, could you rename in `entgeltp_..._updated` ? Otherwise, not so obvious whether we are talking about the sum or the diff.

MImmesberger: Yes, makes sense

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
 
 
-@policy_info(params_key_for_rounding="ges_rente")
-def ges_rente_vor_grundr_m(
-    ges_rente_zugangsfaktor: float,
-    entgeltp_ost_update: float,
-    entgeltp_west_update: float,
+@policy_info(end_date="2020-12-31")
+def ges_rente_m(bruttorente_m: float) -> float:
+    return bruttorente_m
```
hmgaudecker, line 24: Why do we nee the `is_rentner` check in the function directly below, but not here? Should be consistent.

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+
+@policy_info(
+    end_date="2016-12-31",
+    name_in_dag="bruttorente_m",
+    params_key_for_rounding="ges_rente",
+)
+def bruttorente_mit_harter_hinzuverdienstgrenze_m(
+    alter: int,
+    ges_rente_regelaltersgrenze: float,
+    bruttolohn_m: float,
```
hmgaudecker, line 64:
```suggestion
    bruttolohn_y: float,
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
-        See :func:`entgeltp_ost_update`.
-    entgeltp_west_update
-        See :func:`entgeltp_west_update`.
-    rentner
-        See basic input variable :ref:`rentner <rentner>`.
+    alter
+        See basic input variable :ref:`alter <alter>`.
+    ges_rente_regelaltersgrenze
+        See :func:`ges_rente_regelaltersgrenze`.
+    bruttolohn_m
```
hmgaudecker, line 78:
```suggestion
    bruttolohn_y
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+        See basic input variable :ref:`bruttolohn_m <bruttolohn_m>`.
```
hmgaudecker, line 79:
```suggestion
        See basic input variable :ref:`bruttolohn_y <bruttolohn_y>`.
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
-
-    if rentner:
-        out = (
-            entgeltp_west_update * ges_rente_params["rentenwert"]["west"]
-            + entgeltp_ost_update * ges_rente_params["rentenwert"]["ost"]
-        ) * ges_rente_zugangsfaktor
+    # TODO (@MImmesberger): Use age with monthly precision.
+    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/781
+    if (alter >= ges_rente_regelaltersgrenze) or (
+        bruttolohn_m <= ges_rente_params["hinzuverdienstgrenze"] / 12
```
hmgaudecker, line 92:
```suggestion
        bruttolohn_y <= ges_rente_params["hinzuverdienstgrenze"]
```

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+    start_date="2017-01-01",
+    end_date="2022-12-31",
+    name_in_dag="bruttorente_m",
+    params_key_for_rounding="ges_rente",
+)
+def bruttorente_mit_hinzuverdienstdeckel_m(  # noqa: PLR0913
+    alter: int,
+    ges_rente_regelaltersgrenze: float,
+    bruttorente_basisbetrag_m: float,
+    bruttolohn_m: float,
```
hmgaudecker, line 111:
```suggestion
    bruttolohn_y: float,
```

Should be the same in the function, maybe a bit more difficult here, maybe even postpone via a to-do (we will be able to do that for params eventually, too). But we should not have any time unit conversions within functions.

src/_gettsim/transfers/rente.py
```diff
@@ -19,88 +19,224 @@ def sum_ges_rente_priv_rente_m(priv_rente_m: float, ges_rente_m: float) -> float
...
+def bruttorente_basisbetrag_m(
+    ges_rente_zugangsfaktor: float,
+    entgeltp_ost: float,
+    entgeltp_west: float,
     rentner: bool,
+    ges_rente_params: dict,
 ) -> float:
-    """Calculate total individual public pension including Grundrentenzuschlag. Is only
-    active after 2021 when Grundrente is in place.
+    """Old-Age Pensions claim without Grundrentenzuschlag. The function follows the
```
hmgaudecker, line 202: Should not have a reference to Grundrente (let's try to avoid anachronisms)

hmgaudecker (approved):
Just some tiny stuff left, but no show-stoppers. 

I merged the pre-commit autoupdates in here -- apologies that it ended up being larger than expected, but I was too far along to stop and do elsewhere.

src/_gettsim/transfers/rente.py
MImmesberger: I think we can go with `bruttolohn_y` here and avoid parameter conversions. This just requires to split the function up into several functions (because we need automatic time conversions across them); will do this soon.

PR #799, opened 2024-11-11
Improve pixi setup

timmens:
- Add testing dependencies to default environment
- Make sure the correct kaleido dependency is installed on Windows/Unix
- Add task `tests`, so that `pixi run tests` gives one the option to run the tests on Python 3.11 and 3.12
- Set 3.12 as the upper bound for the default environment Python version (as long as we don't test 3.13, we should probably not use it in the development environment?).
- Use pdbp in pytest
- Remove artifacts from previous packaging workflow


hmgaudecker:
Btw, do we need MANIFEST.in anymore?

timmens:
No, we do not. And, of course, we need to actually use `pdbp`; thank you!

hmgaudecker (approved):
Very nice, thanks!

PR #803, opened 2024-12-05
Update eink_st parameters and rounding rules

mjbloemer:
This PR updates income tax parameters and law references for 2024.

Gesetz zur steuerlichen Freistellung des Existenzminimums 2024 has been published: [G. v. 05.12.2024 BGBl. 2024 I Nr. 386](https://www.recht.bund.de/bgbl/1/2024/386/VO.html). New parameter for Grundfreibetrag. New law reference for the full income tax tarif.

Additionally, this PR applies the correct rounding rules for Ehegattensplitting. Before, we only rounded `eink_st_y_sn`. It seems like in practice, there should be some intermediate rounding steps. In particular, in the splitting case, `eink_st_y_sn` should be just the doubled individual income tax which is rounded down.

Note: The parameter file seems to be inconsistent regarding whether the end of the interval or the beginning of the following interval must be used:
- 2020-01-01: End of the interval.
- 2021-01-01: End of the interval.
- 2022-01-01: End of the interval.
- 2023-01-01: Start of following interval.
- 2024-01-01: Start of following interval.

Here, I just keep the logic of the existing parameters for 2024. To be precise, the function should be checked, and it should be documented in the `eink_st_tarif` `note` key which value is the parameter in the yaml file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
+- {gh}`803` Apply correct rounding rules for Ehegattensplitting
+  ({ghuser}`MImmesberger`).
 - {gh}`797` Update infrastructure to use pixi and modern pre-commit hooks
   ({ghuser}`timmens`).
 - {gh}`787` Model classes for policy functions and policy environments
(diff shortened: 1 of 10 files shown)


MImmesberger:
Thanks for updating the reference!

Regarding the inconsistencies: I'll update the parameters in the next few days via this PR.

MImmesberger:
I updated the parameters.

Along the way I noticed some discrepancies to the BMF calculator. These were due to incorrect rounding rules we applied. So I added a fix here.

MImmesberger (approved):

PR #804, opened 2024-12-12
Namespaces

hmgaudecker:
This PR collects the different components of changing GETTSIM's internal DAG from a flat namespace to a nested one. This is a very large change that will be made in multiple PRs. Each PR individually should not change the main branch, however.

The different PRs are (to be updated)
- [x] Infrastructure #780 
- [x] Use qualified names for rounding keys #824 
- [x] Renaming #805
- [x] Fixing the check regarding overlapping periods of policy functions #830 
- [x] Getting rid of TYPES_INPUT_VARIABLES and other cruft #833 
- [x] Use relative paths wherever possible #957 
- [x] First pass restructuring directories to separate ttsim and gettsim (#850)
- [x] #908 
- [x] #853
- [x] #918, #883
- [x] #920
- [x] Required interface changes (parts of #781)
- [x] Convert test files from qualified name dicts to nested structure dicts #884 
- [x] Make ttsim a separate package.

PR #809, opened 2025-01-02
2025 social security and pension parameters

Eric-Sommer:
addresses #806 by adding changes to social security contributions and pensions .

To be discussed:
There is no differentiation anymore between East and West Germany with regard to social security and pensions. The easy way (as done here) is to set the `east` and `west` parameter identical henceforth. Alternatively, we could create new functions which ignore the `wohnort_ost` variable.

### Todo

### DIFF
--- a/src/_gettsim/parameters/sozialv_beitr.yaml
+++ b/src/_gettsim/parameters/sozialv_beitr.yaml
@@ -272,6 +272,11 @@ beitr_satz:
     ges_krankenv:
       mean_zusatzbeitrag: 0.017
     reference: BAnz AT 31.10.2023 B3
+  2025-01-01:
+    deviation_from: previous
+    ges_krankenv:
+      mean_zusatzbeitrag: 0.025
+    reference: BAnz AT 07.11.2024 B4
 beitr_bemess_grenze_m:
   name:
     de: Beitragsbemessungsgrenzen für die Sozialversicherungsbeiträge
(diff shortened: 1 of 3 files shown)


src/_gettsim/parameters/sozialv_beitr.yaml
```diff
@@ -565,6 +570,14 @@ beitr_bemess_grenze_m:
       west: 7550
       ost: 7450
     reference: V. v. 24.11.2023 BGBl. 2023 I Nr. 322
+  2025-01-01:
+    ges_krankenv:
+      west: 6150
```
MImmesberger, line 575: I think this should be 5,512.5€ (66,150€ / 12)? See [here](https://www.bmas.de/DE/Service/Gesetze-und-Gesetzesvorhaben/sozialversicherungs-rechengroessenverordnung-2025.html) for example.

MImmesberger (commented):
Thank you, looks great!

I think the increase of the Pflegeversicherungsbeitrag (0.02 pp) is still missing. Or is this not official yet?

Regarding the differentiation of east/west. We should definitely create new functions that do not require the `wohnort_ost` Variable. However, we should postpone this for now as #804 makes huge changes to nearly every part of the codebase (i.e. changes the module these function's live in). I created #810 such that we don't forget.

src/_gettsim/parameters/sozialv_beitr.yaml
Eric-Sommer: you are right, good catch, thanks!

Eric-Sommer:
pflegeversicherung increase is implemented as well

MImmesberger (approved):
Thanks a lot!

Issue #823, opened 2025-03-11
ENH: Move rounding specs to decorator


MImmesberger:
#### Is your feature request related to a problem?

Rounding specs are currently stored as parameters in the corresponding yaml files.

#### Describe the solution you would like to see

Specs can now become part of the decorator itself, possibly via a rounding spec dataclass.

#### Interface breaking implications

The advantage of the current implementation is that we can apply different rounding specs for different policy evaluation dates without changing the corresponding function. However, I'm not aware of any change in rounding rules (and doubt that they will occur without major changes to the underlying policy). Hence, we can get rid of this feature / just write a new function if it ever becomes relevant.

Edit: Kinderzuschlag Nettoeinkommen Eltern changed its rounding rules in 2019.

MImmesberger:
Done in #859

PR #824, opened 2025-03-11
Use qualified name for rounding keys in params files

MImmesberger:

Uses the qualified name instead of the leaf name to look for rounding specs in the params file. This is a temporary solution until we have tackled #823.

### DIFF
--- a/src/_gettsim/interface.py
+++ b/src/_gettsim/interface.py
@@ -373,7 +373,7 @@ def _add_rounding_to_function(
     """
     func = copy.deepcopy(input_function)
     nice_name = ".".join(path)
-    leaf_name = path[-1]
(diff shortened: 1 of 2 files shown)


src/_gettsim/interface.py
```diff
@@ -373,36 +373,36 @@ def _add_rounding_to_function(
     """
     func = copy.deepcopy(input_function)
     nice_name = ".".join(path)
-    leaf_name = path[-1]
+    qualified_name = "__".join(path)
```
hmgaudecker, line 376: Don't make a habit out of this (=use stuff from dags/for now flatten dict), but for this quick'n'dirty one out is perfectly fine!

hmgaudecker (approved):
Thanks!

PR #805, opened 2024-12-19
Namespaces: Renamings

MImmesberger:
### What problem do you want to solve?

This PR provides the necessary renamings of taxes and transfers functions for #804. 

ToDo:
- [x] Create new directory structure
- [x] Rename all function arguments
- [x] Set namespace of basic input variables
- [x] Update `pyproject.toml` to reflect new file structure
- [x] Make sure tests run (#841)
- [x] `kinderfreibetragempfänger` $\rightarrow$ `kinderfreibetragsempfänger`
- [x] Link issue #842 in relevant docstrings

PR #828, opened 2025-03-11
Move grouping functions to their namespace

MImmesberger:
This PR moves the grouping functions from `groupings.py` to their correct namespaces. It introduces a new class `GroupByFunctions` and makes sure that their new path in the functions/data tree can be inferred.


src/_gettsim/demographics.py
```diff
@@ -391,7 +390,7 @@ def ist_kind_mit_erwerbseinkommen(
     return out
 
 
-@grouping_function()
+@groupby_function()
 def ehe_id(
     demographics__p_id: numpy.ndarray[int],
     demograpics__p_id_ehepartner: numpy.ndarray[int],
```
hmgaudecker, line 396:
```suggestion
    demographics__p_id_ehepartner: numpy.ndarray[int],
```

src/_gettsim/transfers/arbeitslosengeld_2/id.py
hmgaudecker, line 1: I know I suggested this, but looking at it, I think `group_by_ids` would be even better.

src/_gettsim/combine_functions_in_tree.py
```diff
@@ -489,6 +492,106 @@ def aggregate_by_group_func(source_col, group_id):
     )
 
 
+def get_groupby_id(
```
hmgaudecker, line 495: I do not quite understand why we need a different mechanism here than for policy functions, related to your commit message *("group-id-to-source-col match. Need to think about how to handle group IDs from input data")*. 

Could you elaborate? I think writing it down might help.

hmgaudecker: Ah, I think I am slowly getting it. Yes, just hard-code.

MImmesberger:
There are tests in `test_vectorization.py` that still fail. No idea why as of now but they already failed in #805 so we should address it there.

src/_gettsim_tests/test_vectorization.py
```diff
@@ -382,6 +385,7 @@ def test_unallowed_operation_wrapper(func):
             for pf in tree_flatten(
                 load_functions_tree_for_date(datetime.date(year=year, month=1, day=1))
             )[0]
+            if not isinstance(pf, GroupByFunction)
```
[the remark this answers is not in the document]
hmgaudecker: We never vectorized the aggregation functions -- basically vectorization does to a whole vector at once what you you write just for a scalar. So it cannot possibly work out of the box for the aggregation stuff, which necessarily works on a vector to start with.

Or am I missing something?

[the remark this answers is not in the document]
MImmesberger: Yes, that's how I understood it. I would have expected the test to filter out functions that have the `skip_vectorization = True` attribute. (Aggregation functions are not yet part of the functions tree here)

However, it seems like that the `make_vectorizable` function tested here is closely related to our efforts towards making the package JAX-compatible. It complains about a typical JAX incompatibility an it is currently never used anywhere (just for the test).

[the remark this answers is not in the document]
hmgaudecker: Sounds reasonable. Fine to ignore for the moment!

src/_gettsim/config.py
```diff
@@ -71,43 +71,50 @@ def set_array_backend(backend: str):
 SUPPORTED_GROUPINGS = {
```
hmgaudecker, line 71: So we leave this one for lack of time right now, right? Maybe incorporate the scheduling for deletion in #829?

src/_gettsim/loader.py
```diff
@@ -58,12 +56,12 @@ def load_functions_tree_for_date(date: datetime.date) -> NestedFunctionDict:
...
 
-def get_active_policy_functions_tree_from_module(
+def get_active_functions_tree_from_module(
     path: Path,
     package_root: Path,
     date: datetime.date,
-) -> dict[str, PolicyFunction]:
-    """Extract all active PolicyFunctions from a module.
+) -> dict[str, PolicyFunction | GroupByFunction]:
+    """Extract all GroupByFunctions and active PolicyFunctions from a module.
```
hmgaudecker, line 64:
```suggestion
    """Extract all active PolicyFunctions and GroupByFunctions from a module.
```

src/_gettsim/loader.py
```diff
@@ -76,7 +74,8 @@ def get_active_policy_functions_tree_from_module(
 
     Returns
     -------
-    A nested dictionary of active PolicyFunctions with their leaf names as keys.
+    A nested dictionary of GroupByFunctions and active PolicyFunctions with their
```
hmgaudecker, line 77:
```suggestion
    A nested dictionary of active PolicyFunctions and GroupByFunctions with their
```

src/_gettsim/shared.py
```diff
@@ -385,3 +385,100 @@ def format_key_path(key_tuple: tuple[str, ...]) -> str:
...
     _assert_valid_gettsim_pytree(tree, current_key=())
+
+
+def get_group_by_id_path(
+    target_path: tuple[str],
+    group_by_functions_tree: NestedFunctionDict,
+) -> tuple[str] | None:
+    """Get the group_by_id for an aggregation target.
+
+    The group_by_id is the id of the group over which the aggregation is performed. If
```
hmgaudecker, line 396:
```suggestion
    The group by-identifier is the id of the group over which the aggregation is performed. If
```
Same in the rest here and in other docstrings. We don't want to document a variable that is only used in the function body.

hmgaudecker (approved):
Thanks a lot, excellent work! 

I did not mananage to look at shared / test_shared in detail now. But can do that tomorrow. If you want to merge it already, please go ahead!

src/_gettsim/config.py
MImmesberger: Yes, it prevents users from doing shenanigans until we're done with #829.

src/_gettsim_tests/test_vectorization.py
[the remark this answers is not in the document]
hmgaudecker: Okay, figured this one mostly out, I think.

The vectorisation is now immediately tied to the policy function decorator. The tests, however, rely on triggering it manually. I just copied the functions over into the test module, without the decorator.

hmgaudecker:
> There are tests in `test_vectorization.py` that still fail. No idea why as of now but they already failed in #805 so we should address it there.

Okay, figured this one out, I think.

The vectorisation is now immediately tied to the policy function decorator. The tests, however, rely on triggering it manually. I just copied the functions over into the test module, without the decorator.

src/_gettsim_tests/test_vectorization.py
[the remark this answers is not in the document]
hmgaudecker: Related to [how handling of the vectorisation of policy functions has changed](https://github.com/iza-institute-of-labor-economics/gettsim/pull/828#issuecomment-2720045045). This is entirely expected.

src/_gettsim/gettsim_typing.py
```diff
@@ -12,11 +12,10 @@
 
 from _gettsim.aggregation import AggregateByGroupSpec, AggregateByPIDSpec
 from _gettsim.config import numpy_or_jax as np
-from _gettsim.functions.derived_function import DerivedFunction
-from _gettsim.functions.policy_function import PolicyFunction
+from _gettsim.function_types import DerivedFunction, GroupByFunction, PolicyFunction
 
 NestedFunctionDict = dict[
```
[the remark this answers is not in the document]
timmens: Yes, I had to use `typing.Mapping`; see [here](https://github.com/OpenSourceEconomics/dags/blob/d4e4b0b268be13472444cd9e291202513d0b1bcb/src/dags/dags_typing.py#L17) for example.

But, does mypy not raise an error here?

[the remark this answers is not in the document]
hmgaudecker: I am sure it will once we remove the super-permissive flags we are setting. Right now, it is close to useless.

src/_gettsim/shared.py
```diff
@@ -385,3 +385,109 @@ def format_key_path(key_tuple: tuple[str, ...]) -> str:
...
+    for g in SUPPORTED_GROUPINGS:
+        if target_path[-1].endswith(f"_{g}") and g == "hh":
+            # Hardcode because hh_id is not part of the functions tree
+            return ("demographics", "hh_id")
+        elif target_path[-1].endswith(f"_{g}"):
+            return _select_group_by_id_from_candidates(
+                candidate_paths=[p for p in group_by_paths if p[-1] == f"{g}_id"],
+                target_path=target_path,
+            )
+    raise ValueError(f"No group-by-identifier found for target {target_path}.")
```
hmgaudecker, line 427: Just flagging that I changed the behaviour here. Should add a test for that.

src/_gettsim/interface.py
```diff
@@ -526,24 +533,33 @@ def _fail_if_group_variables_not_constant_within_groups(
...
+    flat_data_tree = flatten_dict.flatten(data_tree)
 
     def faulty_leaf(path, leaf):
-        leaf_name = path[-1]
-        for grouping in SUPPORTED_GROUPINGS:
-            id_name = f"{grouping}_id"
-            if leaf_name.endswith(grouping) and id_name in grouping_ids_in_data_tree:
-                # Retrieve the corresponding group ID series from the data tree.
-                group_id_series = grouping_ids_in_data_tree.get(id_name)
+        group_by_id = get_group_by_id_path(
```
hmgaudecker, line 549: Me changing the behavior of `shared.get_group_by_id_path` led to lots of errors here. Let's discuss that, I am not perfectly sure yet what is the best way out.

hmgaudecker (commented):
Excellent! I updated the behaviour because our prime example of non-unique group-by identifiers was not quite covered yet. Please double-check that.

Just to be sure: Things are not working yet.

MImmesberger:
Looks good, thanks!

Issue #604, opened 2023-07-18
ENH: Type annotations in aggregations


hmgaudecker:
#### Is your feature request related to a problem?

-  User-provided inputs:  We currently only have a placeholder for that case. See TODO in `functions_loader.py` / `_create_one_aggregation_func`.
- `test_full_taxes_and_transfers` / `test_data_types` skips aggregated types.

PR #841, opened 2025-03-17
Namespaces: Turn tests back on

MImmesberger:
This PR is based on #805 and turns the tests back on which were previously skipped.

Todos:
- [x] Turn all tests on
- [x] Rename test dirs


src/_gettsim/interface.py
```diff
@@ -513,7 +513,7 @@ def _fail_if_data_tree_not_valid(data_tree: NestedDataDict) -> None:
     """
     assert_valid_gettsim_pytree(
         tree=data_tree,
-        leaf_checker=lambda leaf: isinstance(leaf, pd.Series | np.ndarray),
+        leaf_checker=lambda leaf: isinstance(leaf, pd.Series | np.ndarray | list),
```
hmgaudecker, line 516: We should rather adjust the loader as opposed to being less strict just for the tests, I think. Should return a dict of arrays, see below.

src/_gettsim_tests/_policy_test_utils.py
```diff
@@ -1,158 +1,111 @@
...
     out = []
 
-    for test_file in root.glob("**/*.yaml"):
-        if _is_skipped(test_file):
-            continue
-
-        with test_file.open("r", encoding="utf-8") as file:
-            test_data: dict[str, dict] = yaml.safe_load(file)
+    inputs: NestedDataDict = raw_test_data.get("inputs", {})
+    input_tree: NestedDataDict = merge_trees(
```
hmgaudecker, line 86:
```suggestion
    input_tree: NestedDataDict = optree.treee_map(
        np.array,
        merge_trees(
            inputs.get("provided", {}), inputs.get("assumed", {}
        ),
    )
```

Like this? (untested)

hmgaudecker (commented):
Just a couple of small suggestions! Might be a good idea to explain (very briefly!!!) in a comment here why we are moving from two classes (`PolicyTestSet` and `PolicyTestData`) to a single one.

src/_gettsim_tests/_policy_test_utils.py
MImmesberger: I did it like this:

```python
    input_tree: NestedDataDict = flatten_dict.unflatten(
        {
            k: np.array(v)
            for k, v in flatten_dict.flatten(
                merge_trees(inputs.get("provided", {}), inputs.get("assumed", {}))
            ).items()
        }
    )
```

I didn't manage to register lists as nodes for optree, I think one has to create a custom data object to do that. (But didn't spend too much time on this.)

MImmesberger:
Just to be sure: We want to turn `name_clashes` for `dags.dag_tree.create_dag_tree` and `dags.concatenate_functions_tree` off, right? With 'raise' it throws an error because some of our namespaces start with the name of another namespace and have the same leaf names, e.g. `sozialversicherung__rente__altersrente__langjährig` and `sozialversicherung__rente__altersrente` both have `altersgrenze_vorzeitig` as leafs.

Tbh, I'm not 100% sure what this check is supposed to do.

hmgaudecker:
> Just to be sure: We want to turn `name_clashes` for `dags.dag_tree.create_dag_tree` and `dags.concatenate_functions_tree` off, right? With 'raise' it throws an error because some of our namespaces start with the name of another namespace and have the same leaf names, e.g. `sozialversicherung__rente__altersrente__langjährig` and `sozialversicherung__rente__altersrente` both have `altersgrenze_vorzeitig` as leafs.
> 
> Tbh, I'm not 100% sure what this check is supposed to do.

Can't check in detail right now, but IIRC this should mirror the `columns_overriding_functions` from earlier GETTSIM versions. It definitely should not raise when the same leaf name is used in different namespaces. Could you push a version that raises such an error and point me to it?

MImmesberger:
The current version does raise it. Run 
`pytest src/_gettsim_tests/test_aggregate_by_p_id.py`. 

(Maybe the test path is incorrect typed it out of memory, current not on my machine)

hmgaudecker:
> `pytest src/_gettsim_tests/test_aggregate_by_p_id.py`



> > Just a couple of small suggestions! Might be a good idea to explain (very briefly!!!) in a comment here why we are moving from two classes (PolicyTestSet and PolicyTestData) to a single one.
> 
> So before we did the following:
> 
>     1. Create an `PolicyTestData` object that contains the content of an entire test file
> 
>     2. Store all `PolicyTestData` objects in a list
> 
>     3. Create an `PolicyTestSet` object that takes the list of `PolicyTestData`. Here, every test file is broken down into multiple  (input_data x one expected output column) tuples via the `parametrize_args` property of the class.
> 
> 
> I thought this is a bit complicated. We now have only one class (PolicyTest) which contains exactly what we need for one test (input columns, one expected output column). Everything else is done outside of classes (especially this parametrize functionality).

Thanks! Let's see when things are back how speed looks like. I guess if there are many tests with 5 output columns, each  previously ran once, now they will run 5 times? Might not be a big deal though.

MImmesberger:
No, a test did run once per output column before as well. So hopefully no decrease in speed!

hmgaudecker:
> Tbh, I'm not 100% sure what this check is supposed to do.

Okay, simply too much time has passed since and it seems like I misremembered the resolution we had found on the following. Say we have:

```
einkommensteuer
  einkünfte
    aus_selbstständiger_arbeit
    aus_vermietung_und_verpachtung
```

We wanted to be able to write

```
einkünfte__aus_selbstständiger_arbeit
```

inside `einkommensteuer`, i.e., use relative paths anywhere inside the tree. This means, however, that we needed to make sure that this is unique anywhere.

We wrote #805 like this option did not exist; I do think now that this is a good thing even if we need to disable `E501` globally.

I set `name_clashes="ignore"` and this will definitely be the behaviour we want. Will probably remove that option in dags.

MImmesberger:
Another small bug came up: Because we now apply parameters to every function in the functions tree (rather than to those that are part of the DAG), we now encounter situations where we cannot build the entire functions tree because rounding parameters are specified only after a certain date. 

I'd vote for doing something like this in the parameter file.

```yaml
rounding:
  einkommensteuer__betrag_mit_kinderfreibetrag_y_sn:
    1900-01-01:
      base: 1
      direction: down
      reference: Reference for rounding specs before 2002 currently unknown.
      issue_link: xxxxx
    2002-01-01:
      base: 1
      direction: down
      reference: § 32a Abs. 1 S. 6 EStG
```

This makes sense because I think it's quite unlikely that rounding rules did change, we just don't have the reference. Also, we plan to change this machinery anyhow (see #823).

hmgaudecker:
> This makes sense because I think it's quite unlikely that rounding rules did change, we just don't have the reference. Also, we plan to change this machinery anyhow (see #823).

Yes, do whatever is quickest now without making #823 more difficult to implement. Any hack will do, we won't release before implementing #823, anyhow.

src/_gettsim/transfers/wohngeld/einkommen.py
```diff
@@ -90,22 +90,22 @@ def einkommen_m_bg(
 
 @policy_function()
 def abzüge_vom_einkommen_für_steuern_sozialversicherung_m(
-    einkommensteuer__betrag_y_sn: float,
+    einkommensteuer__betrag_m_sn: float,
```
hmgaudecker, line 93: Why convert? I would leave it at its natural unit.

src/_gettsim_tests/_policy_test_utils.py
```diff
@@ -1,158 +1,163 @@
...
+        raw_test_data: The raw test data.
 
-        inputs: dict[str, dict] = test_data["inputs"]
-        inputs_provided: _ValueDict = inputs.get("provided", {})
-        inputs_assumed: _ValueDict = inputs.get("assumed", {})
-        outputs: _ValueDict = test_data["outputs"]
+    Returns:
+        A list of PolicyTest objects.
+    """
+    out = []
```
hmgaudecker, line 112: Move to line 135 so that it is clear that everything befor that is just a prelude.

And please make an issue to move to a optree-based solution for comparing trees down there.

src/_gettsim_tests/test_aggregate_by_p_id.py
```diff
@@ -1,36 +1,33 @@
...
-        result[column],
-        test_data.output_df[column],
-        check_dtype=False,
-        atol=1e-1,
-        rtol=0,
-    )
+    flat_result = dt.flatten_to_qual_names(result)
+    flat_expected_output_tree = dt.flatten_to_qual_names(test.expected_output_tree)
+
+    for result, expected in zip(
```
hmgaudecker, line 30: Add to the prior issue that we move to comparing the trees directly rather than array-by-array. Should be pretty straightforward using `tree_map_with_path`, I'd hope...

src/_gettsim/interface.py
```diff
@@ -40,8 +39,6 @@
     get_path_for_group_by_id,
     merge_trees,
     partition_tree_by_reference_tree,
-    qualified_name_reducer,
-    qualified_name_splitter,
 )
```
hmgaudecker, line 44: Just a note that we should eventually move to `targets` meaning `targets_tree_paths`. Pretty unnatural to be using a dict at the lowest level here! *(but of course the best we could think of before diving deep into `dags.tree`)*. Same for the input structure elsewhere.

hmgaudecker (commented):
Excellent, thanks a lot! I did go through in detail once more since any larger renamings involving paths will become more painful after the final step. So there are a few simple renamings and a couple of larger things (`abgeltungssteuer`, restructure `vorsorgeaufwendungen`).

I am getting clearer about how to handle relative paths in dags. I do think we should move some stuff from demographics to the top-level namespace then. So probably wait until then with the last step?

MImmesberger:
> I do think we should move some stuff from demographics to the top-level namespace then. So probably wait until then with the last step?

Yes, we should rewrite the internal checks whether an argument is qualified or not first. Currently, it does not use the functions or data tree as reference.

src/_gettsim/taxes/demographics/demographics.py
```diff
@@ -236,6 +236,7 @@ def alter_monate(geburtsdatum: numpy.datetime64, elterngeld_params: dict) -> flo
     return out.astype(float)
 
 
+@policy_function()
 def ist_kind_mit_erwerbseinkommen(
```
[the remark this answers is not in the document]
MImmesberger: Wohngeld namespace would be even better (because its used only there).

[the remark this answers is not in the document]
hmgaudecker: Sure, I did not check that!

src/_gettsim_tests/_policy_test_utils.py
MImmesberger: Which tree comparison do you mean in this file?

hmgaudecker: No comparison here, but we should not need that complex code lines 135+, which just seems to be converting data structures.

MImmesberger: Ah that comes for free when we do tests based on entire test files rather than column-by-column.

src/_gettsim/taxes/einkommensteuer/abzüge/vorsorgeaufwendungen.py
```diff
@@ -7,22 +7,22 @@
     params_key_for_rounding="eink_st_abzuege",
 )
 def vorsorgeaufwendungen_y_sn_bis_2004(
-    vorwegabzug_lohnsteuer_2004er_regime_y_sn: float,
-    sozialversicherung__kranken__beitrag__betrag_versicherter_y_sn: float,
-    sozialversicherung__rente__beitrag__betrag_versicherter_y_sn: float,
+    vorwegabzug_y_sn_lohnsteuer_2004er_regime: float,
```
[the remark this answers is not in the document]
MImmesberger: I have rewritten the module to get rid of all formulas. I think this makes it easier and
avoids the naming patterns you suggested. But feel free to disagree if you think the new
structure is confusing.

In general, I prefer something like `vorsorgeaufwendungen_vor_günstigerprüfung_y_sn`
over `vorsorgeaufwendungen_regime_nach_2004_y_sn`. Would be good to do this as well for
`vorsorgeaufwendungen_regime_bis_2004_y_sn`, but I'd want to avoid to read up on the
details right now...

src/_gettsim/taxes/einkommensteuer/abzüge/vorsorgeaufwendungen.py
```diff
@@ -33,9 +33,9 @@ def vorsorgeaufwendungen_y_sn_bis_2004(
 
     """
     return formel_2004er_regime(
-        vorwegabzug_lohnsteuer=vorwegabzug_lohnsteuer_2004er_regime_y_sn,
-        krankenversicherung_beitrag_arbeitnehmer=sozialversicherung__kranken__beitrag__betrag_versicherter_y_sn,
-        rentenversicherung_beitrag_arbeitnehmer=sozialversicherung__rente__beitrag__betrag_versicherter_y_sn,
+        vorwegabzug_lohnsteuer=vorwegabzug_y_sn_lohnsteuer_2004er_regime,
+        krankenversicherung_beitrag_arbeitnehmer=sozialversicherung__kranken__beitrag__betrag_versicherter_m_sn,
```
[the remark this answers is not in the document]
MImmesberger: Oh yes, when I saw we were a bit too greedy with the changes, I fully reverted to the
old version and missed that part.

src/_gettsim/taxes/einkommensteuer/abzüge/vorsorgeaufwendungen.py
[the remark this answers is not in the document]
hmgaudecker: Using `vor_günstigerprüfung` for periods where there is no such thing just confuses the hell out of me. I think most of the problem actually came from that and from trying to press the favorability check to have the same inputs 2005-2019. Let's just split that up. So here we go (apologies for having to scroll to the right...): 

| Old name                                                          | New name                                                      | Comment                                                              |
| ----------------------------------------------------------------- | ------------------------------------------------------------- | -------------------------------------------------------------------- |
| `vorsorgeaufwendungen_y_sn_bis_2004`                              |                                                               |                                                                      |
| `vorsorgeaufwendungen_y_sn_ab_2005_bis_2019`                      | `vorsorgeaufwendungen_y_sn_ab_2005_bis_2009`                  | Split up, continue to do only Günstigerprüfung here                  |
| `vorsorgeaufwendungen_y_sn_ab_2005_bis_2019`                      | `vorsorgeaufwendungen_y_sn_ab_2010_bis_2019`                  | Split up, continue to do only Günstigerprüfung here                  |
| `vorsorgeaufwendungen_y_sn_ab_2020`                               |                                                               |                                                                      |
| `vorsorgeaufwendungen_regime_bis_2004_y_sn`                       |                                                               | Just move up in the module                                           |
| `vorsorgeaufwendungen_vor_günstigerprüfung_y_sn_ab_2005_bis_2009` | `vorsorgeaufwendungen_globale_kappung_y_sn`                   | No renaming                                                          |
| `vorsorgeaufwendungen_vor_günstigerprüfung_y_sn_ab_2010`          | `vorsorgeaufwendungen_keine_kappung_krankenversicherung_y_sn` | No renaming                                                          |
| `altersvorsorge_y_sn_phase_in`                                    |                                                               |                                                                      |
| `altersvorsorge_y_sn_volle_anrechnung`                            |                                                               |                                                                      |
| `vorwegabzug_lohnsteuer_2004er_regime_y_sn`                       | `vorwegabzug_lohnsteuer_y_sn`                                 | Though the old name does make me wonder whether we need a start date |

hmgaudecker (commented):
Excellent, thanks a lot! 

One last thing, couldn't comment there:

`minijob_grenze_einheitlich` -> `minijob_grenze_fixer_betrag` (the result is also uniform after linking it to the min wage.)

docs/gettsim_objects/input_variables.md
```diff
@@ -325,9 +325,9 @@ Month of retirement
 
 Type: int
 
-(elterngeld__bisheriger_bezug_m)=
+(elterngeld__anzahl_bisheriger_bezugsmonate)=
 
-## `elterngeld__bisheriger_bezug_m`
+## `elterngeld__anzahl_bisheriger_bezugsmonate`
```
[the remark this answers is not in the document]
MImmesberger: This is only about the anzahl + monate pattern, right?

If not: How do you think about `anzahl_kinderfreibeträge` and `anzahl_personen_hh`? I think for both, keeping `anzahl_` aids clarity.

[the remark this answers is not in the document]
hmgaudecker: > This is only about the anzahl + monate pattern, right?

Yes, sorry for not being clear!

src/_gettsim/taxes/einkommensteuer/abzüge/vorsorgeaufwendungen.py
```diff
@@ -7,51 +7,74 @@
...
+    leaf_name="vorsorgeaufwendungen_y_sn",
+    params_key_for_rounding="eink_st_abzuege",
+)
+def vorsorgeaufwendungen_y_sn_ab_2005_bis_2009(
+    vorsorgeaufwendungen_regime_bis_2004_y_sn: float,
+    vorsorgeaufwendungen_globale_kappung_y_sn: float,
+) -> float:
+    """Vorsorgeaufwendungen from 2005 to 2009.
+
+    With the 2005 reform, no taxpayer was supposed to be affected negatively.
```
hmgaudecker, line 38: Please remove cruft from the docstrings. Outdated docs are worse than no docs!

hmgaudecker (commented):
Wonderful, thanks! Updated dags to drop soon-ish...

src/_gettsim/interface.py
```diff
@@ -68,23 +65,24 @@ def compute_taxes_and_transfers(
     debug : bool
         The debug mode does the following:
             1. All necessary inputs and all computed variables are returned.
-            2. If an exception occurs while computing one variable, the exception is
-                skipped.
+            2. Exceptions that occur while computing one variable are ignored. This
```
MImmesberger, line 68: Just a note that this is currently not true. Which exceptions should be skipped? Only those that are raised when calling `tax_transfer_function(input_data_tree)` or earlier ones as well?

hmgaudecker: Oh, in that case just delete. I simply updated the text so that it would be comprehensible, didn't check whether it is still true...

MImmesberger: Oops missed that.

MImmesberger:
We're now back to using qualified names everywhere. All tests pass -- and are much faster now 🙂

src/_gettsim/interface.py
```diff
@@ -185,35 +198,40 @@ def _convert_data_to_correct_types(
...
         # Look for column in TYPES_INPUT_VARIABLES
-        if qualified_name in flat_internal_types:
-            internal_type = flat_internal_types[qualified_name]
+        if name in TYPES_INPUT_VARIABLES:
+            internal_type = TYPES_INPUT_VARIABLES[name]
         # Look for column in functions_tree_overridden
-        elif qualified_name in flat_functions:
-            func = flat_functions[qualified_name]
+        elif name in functions_overridden:
+            func = functions_overridden[name]
```
MImmesberger, line 211: These lines here are a bit ugly, mainly because `func.skip_vectorization = True` does
not mean that annotations are in array form (in particular true for all derived
function). I think we talked about touching annotations soon anyhow, right (-> jax and #604)?
Then we could leave it as it is for now.

src/_gettsim/shared.py
```diff
@@ -453,31 +465,43 @@ def _select_group_by_id_from_candidates(
...
+                candidate_names_in_matching_namespace=candidate_names_in_matching_namespace,
+                all_candidate_names=candidate_names,
+                target_name=target_name,
             )
     else:
-        return candidate_paths[0]
+        return candidate_names[0]
 
 
 def _fail_with_ambiguous_group_by_identifier(
```
hmgaudecker, line 487:
```suggestion
def _fail_because_of_ambiguous_group_by_identifier(
```

src/_gettsim/time_conversion.py
```diff
@@ -230,8 +231,8 @@ def d_to_w(value: float) -> float:
 
 
 def create_time_conversion_functions(
-    functions_tree: NestedFunctionDict,
-    data_tree: NestedDataDict,
+    functions: NestedFunctionDict,
```
hmgaudecker, line 234:
```suggestion
    functions: QualNameFunctionDict,
```

etc.

Time to turn mypy on, I think :smile:.

hmgaudecker (commented):
Beautiful! Just some tiny changes, mostly a suggestion for the global namespace.

MImmesberger:
I have adjusted the namespaces. Needed to change the infrastructure a tiny bit to allow for aggregation targets to be in different namespaces than aggregation sources, but I think this is something we should do anyhow.

src/_gettsim/taxes/household_characteristics.py
hmgaudecker, line 1: So these can't just live in the global namespace?

hmgaudecker (approved):
Excellent! Thanks!

src/_gettsim/taxes/household_characteristics.py
MImmesberger: Yes, they're currently in the global namespace!

PR #850, opened 2025-04-02
Introduce TTSIM/DE distinction

MImmesberger:

This PR implementes the distinction between TTSIM (basically the infrastructure) and DE (the German taxes and transfers) components of GETTSIM. This was discussed [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/780#discussion_r1894663083).

In particular, I

- Move modules from `_gettsim` to `_gettsim/ttsim/` or `_gettsim/de/`
- Remove the `taxes` and `transfers` subdirs
- Split up `config.py` into a TTSIM and a DE part
- Adjust the loader accordingly
- Also split up tests in TTSIM and DE parts.

For tests, the distinction is not always super sharp. There are some tests that test a specific feature of the infrastructure (e.g. rounding), but do this by loading the functions tree from the DE part. Still, I chose to label those tests as `ttsim`. 

Similarly, we don't test `aggregate_by_p_id` directly in the `ttsim` part, but do it by testing specific components of the TT system. I put them in the `de` dir.


hmgaudecker (approved):
Excellent, thanks! Maybe take another look before merging to check that I did not mess anything up!

PR #854, opened 2025-04-07
Add PolicyInputs, change Type Hierarchy

### LINKED ISSUE #833: ENH: Remove TYPES_INPUT_VARIABLES / add PolicyInput class
### Problem

We currently require information that is manually maintained:
1. `TYPES_INPUT_VARIABLES` in `config.py` for checking the types of data provided to GETTSIM / creating aggregation functions
2. `input_variables.md`, which is basically the same list for the docs
3. In the docstrings, references to functions and input variables, even though these 
   a. are created automatically via dags
   b. don't work with aggregations / time period changes.

### Solution requirements

1. Inputs should appear in the namespace where they end up, rather than in a central place
2. Inputs should be able to change over the years
3. No duplicate information required

### Suggested solution

Define a class `PolicyInput` with an associated decorator `policy_input`. Both mirror their policy-function counterparts.

Usage example in the `elterngeld` namespace:

```python
@policy_input(start_date="2007-01-01")
def nettoeinkommen_vorjahr_m() -> float:
    """Net income before the birth of youngest child relevant for calculating Elterngeld.

     Note: You may let GETTSIM compute this variable via the 
    `elterngeld__nettoeinkommen_approximation_m` target in a separate run, which would
     typically be for the previous calendar year.
    """
    return -np.inf
```

This allows us to:
1. Use inputs and their types for aggregations before knowing about the user data
2. Document input variables at the place where they belong using the same mechanism as for policy functions
3. Have everything documented *in principle*, so we should just get rid of the `Parameters` and `Returns` parts in the docstrings for now (remember they don't work anyhow ATM); we'll need to find a way for generating them programmatically down the road)
4. Fail gracefully early on if these are not provided by users.

timmens:
This is a huge PR, which started innocently as a fix to #833.

In the end, it turned out to be very difficult to change things locally, so during the intense sprint during the week 7-12 April 2025, this ended up including the following:
- Updated type hierarchy (`TTSIMObject` as basic building block, `PolicyInput` and `TTSIMFunction` inheriting from that, `TTSIMFunction` has further subclasses for policy, aggregation, ...).
- Further separation of tests in ttsim / _gettsim. Including Middle Earth Taxes an Transfers SIMulator METTSIM as tiny example for the ttsim-side of tests (#856) and sensible structure for `_gettsim_tests` (#858)
- Sensible treatment of Einnahmen / Einkünfte (#862)
- Specify rounding in a dataclass to be provided in the decorators rather than referencing the yaml files from there (#859)
- Improve structure for AggregationSpecs, including an Enum for the type of Aggregations (#860)


src/ttsim/aggregation.py
```diff
@@ -31,14 +33,69 @@
...
+            func = grouped_mean
+        elif self.aggr == AggregationType.MAX:
+            func = grouped_max
+        elif self.aggr == AggregationType.MIN:
+            func = grouped_min
+        elif self.aggr == AggregationType.ANY:
+            func = grouped_any
+        elif self.aggr == AggregationType.ALL:
+            func = grouped_all
+        else:
```
hmgaudecker, line 74: This can go

src/ttsim/combine_functions.py
```diff
@@ -369,81 +364,31 @@ def _create_one_aggregate_by_group_func(
...
             source=source,
             wrapped_func=wrapped_func,
         )
-        if source
-        else None
-    )
+    else:
+        qual_name_source = None
 
     return DerivedAggregationFunction(
```
hmgaudecker, line 386: Can we do the same trick as for the `AggregateByGroupSpec` / `CountByGroupSpec` here? I.e., make a case distinction above and return either `DerivedAggregationFunction` or `DerivedCountFunction`

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+
+def policy_input(
+    *,
+    start_date: str | datetime.date = "1900-01-01",
+    end_date: str | datetime.date = "2100-12-31",
+    params_key_for_rounding: str | None = None,
+    skip_vectorization: bool = False,
+) -> PolicyInput:
+    """
+    Decorator that makes a `PolicyInput` from a function.
```
hmgaudecker, line 250:
```suggestion
    Decorator that makes a (dummy) function a `PolicyInput`.
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+
+    **Dates active (start_date, end_date, leaf_name):**
```
hmgaudecker, line 252:
```suggestion
    **Dates active (start_date, end_date):**
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+
+    Specifies that a PolicyInput is only active between two dates, `start` and `end`.
+
+    Note that even if you use this decorator with the `leaf_name` argument, you must
```
hmgaudecker, line 256:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+    ensure that the function name is unique in the file where it is defined. Otherwise,
```
hmgaudecker, line 257:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+    the function would be overwritten by the last function with the same name.
```
hmgaudecker, line 258:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+    function:
+        The function to wrap. Argument values of the `@policy_function` are reused
+        unless explicitly overwritten.
+    start_date:
+        The date from which the function is active (inclusive).
+    end_date:
+        The date until which the function is active (inclusive).
+    params_key_for_rounding:
+        The key in the params dictionary that should be used for rounding.
+    skip_vectorization:
```
hmgaudecker, line 202:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+        Whether the function should be vectorized.
```
hmgaudecker, line 203:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+    """
+
+    def __init__(
+        self,
+        *,
+        function: Callable,
+        start_date: datetime.date,
+        end_date: datetime.date,
+        params_key_for_rounding: str | None,
+        skip_vectorization: bool | None,
```
hmgaudecker, line 213:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -173,7 +184,123 @@ def wrapper_vectorize_func(*args, **kwargs):
...
+    ):
+        self.skip_vectorization: bool = skip_vectorization
```
hmgaudecker, line 215:
```suggestion
```

src/ttsim/function_types.py
```diff
@@ -241,10 +368,7 @@ def __init__(
         self,
         *,
         function: Callable,
-        source_function: PolicyFunction
-        | DerivedTimeConversionFunction
-        | DerivedAggregationFunction
-        | None = None,
+        source_function: TTSIMFunction | None = None,
```
hmgaudecker, line 371:
```suggestion
        source_function: TTSIMFunction,
```

src/ttsim/function_types.py
```diff
@@ -283,10 +407,7 @@ def __init__(
...
```
hmgaudecker, line 410:
```suggestion
        source_function: TTSIMFunction,
```

tests/ttsim/test_compute_taxes_and_transfers.py
```diff
@@ -458,7 +458,7 @@ def test_aggregate_by_group_specs_agg_not_impl():
         "module_name": {
             "betrag_m_hh": AggregateByGroupSpec(
                 source="betrag_m",
-                aggr="aggr_not_implemented",
+                aggr=AggregationType.NOT_IMPLEMENTED,
```
hmgaudecker, line 461: Just pass some invalid input or remove the test

src/ttsim/aggregation.py
```diff
@@ -31,14 +33,69 @@
...
+    """
+
+    COUNT = "count"
+    SUM = "sum"
+    MEAN = "mean"
+    MAX = "max"
+    MIN = "min"
+    ANY = "any"
+    ALL = "all"
+    NOT_IMPLEMENTED = "aggr_not_implemented"
```
hmgaudecker, line 48:
```suggestion
```

hmgaudecker (commented):
Great!

hmgaudecker (approved):
Let's merge this monster, will fix remaining tests separately (just an addition of new columns for incomes).

PR #859, opened 2025-04-08
Rounding spec dataclass, fixes #823

### LINKED ISSUE #823: ENH: Move rounding specs to decorator

hmgaudecker:
Improve the handling of rounding.


MImmesberger:
I rewrote GEP 5 to a large degree, you should have a look at that. Also, I deleted an entire section from GEP 3 which didn't make sense their (IMHO). Feel free to disagree there.

The implementation works great.

MImmesberger (approved):

PR #860, opened 2025-04-08
Fix #814

### LINKED ISSUE #814: Add CountByGroupSpec, CountByPIDSpec; use enums for AggregateByXSpec
In #780, we introduce

```python
@dataclass
class AggregateByGroupSpec:
    """
    A container for aggregate by group specifications.
    """

    aggr: str
    source_col: str | None = None
```

In order to get better typing, we should remove the `None` option for the source col. Way to go is a separate class for counting, which is the only case how it can become `None`.

Same for the PID version.

The type annotation for the `aggr` function currently is a string and should be an enum-style type.

timmens:

### Todo


hmgaudecker (approved):
Excellent, many thanks!

PR #863, opened 2025-04-09
Add join_jax

mj023:

`join_numpy` is not jit compatible, I replaced it with a generic join function that uses either Jax or Numpy depending on the Flag in config.py. The Input Checks are now resposibility of the user of the function. If there are duplicate Private Keys the first encountered value will be used. If a foreign key is missing the user-supplied default value will be used.

### Todo

- [x] Write `join` thet supports both Jax and Numpy
- [x] Add test


hmgaudecker (approved):
Excellent, thank you!!!

PR #864, opened 2025-04-09
Fix vectorization

### LINKED ISSUE #515: ENH: Enable Jax

We merged #425 before Jax is fully usable in GETTSIM. This issue is to remind ourselves of what needs to be done still. 

Core thing is to change `src._gettsim.functions_loader.vectorize_func` to something like:

```python
def _vectorize_func(func):
    backend = "jax" if USE_JAX else "numpy"
    return make_vectorizable(func, backend)
```

Doing so currently leads to failures. Fixing these failures requires knowing which function arguments are to be vectorized over and which are not (some ideas for a decorator in GEP 6).

timmens:
## Fix vectorization

In this PR, we create the necessary infrastructure such that

1. Individual policy functions are dynamically re-written to be vectorizable
2. The compute taxes and transfer function is potentially jax-jitted

Closes #515

For better error messages and full support of vectorization see #869 and #868 

### Changes

- Add mypy dependency to dev environment
- Add "test" feature to jax environment (can then select py312-jax after running `pixi run tests`)
- Adds `vectorization_strategy` attribute to `PolicyFunction`
  Depending on the vectorization strategy, functions are vectorized via "loop" (`np.vectorize`), "vectorize" (vectorized by dynamically re-writing the source code, or "not_required" (left as they are for functions that are already written in a vectorized fashion)
- Removes type conversions in the return statement of policy functions, as this cannot be handled by the vectorization-compiler (see #869)
- Re-writes some source code such that the vectorization-compiler works
- Implements lambda-functions as named functions and throws an error in the vectorization-compiler if lambda-functions are found
- Removes possibly to set the array backend
- Sets array backend to JAX if JAX is installed, otherwise uses numpy
- Re-writes `sum_by_p_id` aggregation function for jax to be jittable
- Adds the boolean argument `jit` to `compute_taxes_and_transfers` , which then jax-jit's the internal function
- Adds test module `test_jax_jit_kindergeld.py` which contains a unit-level test for a policy function, and an end-to-end test for compute taxes and transfers


MImmesberger (approved):
Thanks! Lets see whether we can get rid of the test failures on the other branch.

PR #865, opened 2025-04-10
Make top-level namespace include potential derived functions

MImmesberger:

The top-level namespace did not account for derived functions that are created later. We now add place-holders when creating the top-level namespace for those derived functions.

Because tests don't work currently, I'm not 100% sure that the changes don't crash time conversions (but I'm optimistic that they didn't). Have to keep that in mind if we run into bugs when wrapping up #804.

### DIFF
--- a/src/ttsim/compute_taxes_and_transfers.py
+++ b/src/ttsim/compute_taxes_and_transfers.py
@@ -37,7 +37,7 @@ from ttsim.shared import (
     merge_trees,
     partition_by_reference_dict,
 )
-from ttsim.time_conversion import _TIME_UNITS
+from ttsim.time_conversion import TIME_UNITS
 from ttsim.typing import (
     check_series_has_expected_type,
     convert_series_to_internal_type,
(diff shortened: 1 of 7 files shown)


src/ttsim/compute_taxes_and_transfers.py
```diff
@@ -88,9 +90,7 @@ def compute_taxes_and_transfers(
     _fail_if_environment_not_valid(environment)
 
     # Transform functions tree to qualified names dict with qualified arguments
-    top_level_namespace = set(environment.raw_objects_tree.keys()) | set(
-        environment.aggregation_specs_tree.keys()
-    )
+    top_level_namespace = _get_top_level_namespace(environment)
```
hmgaudecker, line 93:
```suggestion
    top_level_namespace = _get_top_level_namespace(
        environment=environment,
        supported_time_conversions=supported_time_conversions,
        supported_groupings=supported_groupings,
    )
```

src/ttsim/time_conversion.py
```diff
@@ -447,25 +447,28 @@ def _create_time_conversion_functions(
     name: str, func: PolicyFunction | None = None
```
hmgaudecker, line 447:
```suggestion
    name: str, ttsim_object: TTSIMObject
```

hmgaudecker (commented):
Excellent, thanks! Just fix the tests for time conversions and then it's ready to go!

PR #866, opened 2025-04-11
Add tests for aggregations and infrastructure for METTSIM testing

MImmesberger:

- Add tests for start and end dates of aggregation by group and p_id
- Fixes a start date/end date bug in aggregation by p_id
- Adds infrastructure for METTSIM tests (but doesn't add test cases yet because tests don't run)


hmgaudecker (approved):
:+1:

PR #867, opened 2025-04-11
Vectorize `piecewise_polynomial`

mj023:

Vectorize `piecewise_polynomial` and all other functions that are needed to calculate the parameters.

### Todo

- [x] Vectorize `piecewise_polynomial`
- [ ] ~~Change all functions that check the parameters~~


mj023:
Some warnings are showing up that are connected to `piecewise_polynomial`, even though the tests are not failing. The reason for these should be investigated before merging, maybe NAN caused by -inf lower threshold.

hmgaudecker:
Excellent analysis, thank you!

mj023:
Thanks for the tip! Tim is absolutely right. The behavior of the old function was looping around and returning the lower intercept of the highest interval. This also doesn't seem right, maybe we should return NAN in cases where the requested value is out of bounds so users know something is up?

hmgaudecker:
Sorry for being slow to respond. This check is already there in `get_piecewise_parameters`, which makes sure that the entire real line is covered. The problem is that we do not mark the cutoffs as a set of piecewise parameters in `arbeitsl_geld.yaml`, so that never gets called. Not sure we even want a `piecewise_constant` function though, that whole thing might be a simple `searchsorted`.

So please just ignore that particular issue, it will be improved in the near future by virtue of actually parsing the YAML contents to parameters / a better implementation of `arbeitslosengeld`.

hmgaudecker:
>     * [ ]  Change all functions that check the parameters

Related to my previous message, this is probably unnecessary. We will do the parameter checks when processing the yaml files. The `piecewise_polynomial` function will then require a custom object for the parameters, so the above situation won't occur anymore.

mj023:
Okay, then I think it's enough if we mention this behaviour in the docstring, just in case someone uses it without using `get_piecewise_parameters`. I had to convert some more inputs from lists to arrays, but now everything should be ready for merging.

hmgaudecker (approved):
Wonderful, thanks a lot!

PR #871, opened 2025-04-13
Require dashed iso date everywhere. Fixes #795

### LINKED ISSUE #795: Require precise date for setting up a policy environment
Frequently, the policy environment changes throughout the year. Currently, we can call it with an integer for a year, a month, or a precise date. The former two may create ambiguity (we probably default to 1 Jan if a year is provided, but I would actually have to look that up).

Just as for data, we should require precision from the user instead of making assumptions.

- [x] Change parser
- [x] Check docs
- [x] Fix tutorials

hmgaudecker:
# What problem do you want to solve?

Unify handling of dates to remove ambiguity and code duplication.


hmgaudecker:
*(has exactly the same pattern of test failures as `collect-components-of-namespaces`. I.e., `einkommensteuer` inputs are missing.*

src/ttsim/shared.py
```diff
@@ -13,23 +14,28 @@
...
+_DASHED_ISO_DATE_REGEX = re.compile(r"\d{4}-\d{2}-\d{2}")
 
 
-def validate_dashed_iso_date(date: str | datetime.date):
-    if not _DASHED_ISO_DATE.match(date):
-        raise ValueError(f"Date {date} does not match the format YYYY-MM-DD.")
+def to_datetime(date: datetime.date | DashedISOString):
+    if isinstance(date, datetime.date):
+        return date
+    if isinstance(date, str) and _DASHED_ISO_DATE_REGEX.match(date):
```
MImmesberger, line 33:
```suggestion
    if isinstance(date, str) and _DASHED_ISO_DATE_REGEX.fullmatch(date):
```

Else, we match only the beginning of `date`, not the entire thing.

MImmesberger (approved):
Looks good!

PR #872, opened 2025-04-14
Remove custom type conversions.

hmgaudecker:
We put some effort into trying to convert types. However, the code was a mess and it would be a pain to maintain it. What Python/Pandas/Numpy/Jax do is more than good enough for GETTSIM, too. Now that we have the explicitly annotated `policy_inputs`, it will be easy to check and throw errors if users want to be strict.

This PR removes the code which has been stale for the last week, anyhow.


MImmesberger (approved):
LGTM!

PR #875, opened 2025-04-15
Fix Vectorization 

hmgaudecker:
This is just #864 by @timmens, the merge did not work out and we need to re-do that.


hmgaudecker (commented):
Check looks good.

PR #876, opened 2025-04-16
Interface: Add DataFrame to NestedDataDict conversion

MImmesberger:

Users can easily create a NestedDataDict by providing a mapper from the paths used in the TTSIM instance to a column in the DataFrame or a pandas Series or a single value.

Todo:
- [ ] Add input template for users. Args: date, NestedTargetDict
- [x] Refactor `df_to_data_tree` such that it calls `compute_taxes_and_transfers` directly.
- [ ] Decide whether we want to allow `pd.Series` inputs and which types of data we want to allow for broadcasting. Probably we want to be way more restrictive than I currently am.


MImmesberger:
@hmgaudecker 
Regarding the input template: I think it makes sense to wait with this until we have broken up `compute_taxes_and_transfers`. We basically need to call `dags.tree.create_input_structure_tree` on the functions tree with partialled parameters.

Also, could you have a look at the type hints? I tried many versions but can't get it to work... When I run `pytest`, I get the error below. If it helps, you can look at the diff of [1accf2e](https://github.com/iza-institute-of-labor-economics/gettsim/pull/876/commits/1accf2efbf8c608d7ce2fc38ef498acb67c7942d) (version without the type hints, works) and [d8b859d](https://github.com/iza-institute-of-labor-economics/gettsim/pull/876/commits/d8b859d537ab9c24d767f33e6b23d129a15542af) (version with type hints).

```shell
ERROR: while parsing the following warning configuration:

  ignore::ttsim.compute_taxes_and_transfers.FunctionsAndColumnsOverlapWarning

This error occurred:

Traceback (most recent call last):
  File "/Users/marvin/GitHub/gettsim/.pixi/envs/default/lib/python3.12/site-packages/_pytest/config/__init__.py", line 1918, in parse_warning_filter
    category: type[Warning] = _resolve_warning_category(category_)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marvin/GitHub/gettsim/.pixi/envs/default/lib/python3.12/site-packages/_pytest/config/__init__.py", line 1956, in _resolve_warning_category
    m = __import__(module, None, None, [klass])
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marvin/GitHub/gettsim/src/ttsim/__init__.py", line 15, in <module>
    from ttsim.prepare_data import create_data_tree_from_df
  File "/Users/marvin/GitHub/gettsim/src/ttsim/prepare_data.py", line 5, in <module>
    from ttsim.typing import NestedDataDict, NestedInputToSeriesNameDict
ImportError: cannot import name 'NestedDataDict' from 'ttsim.typing' (/Users/marvin/GitHub/gettsim/src/ttsim/typing.py)
```

src/_gettsim/interface.py
```diff
@@ -0,0 +1,83 @@
...
+    create_data_tree_from_df,
+    set_up_policy_environment,
+)
+from ttsim.typing import NestedDataDict, NestedInputToSeriesNameDict
+
+
+def quickrun(
+    date: str,
+    df: pd.DataFrame,
+    input_tree_to_column_map: NestedInputToSeriesNameDict,
```
hmgaudecker, line 16:
```suggestion
    inputs_tree_to_df_columns: NestedInputsPathsToDfColumns,
```
- plural consistent for inputs/targets
- df should appear again
- map seems unnecessary
- Type hint: We'll need another round to make them consistent, anyhow. Not too important. But `SeriesName` is not a concept I'd like to see introduced :wink:.

hmgaudecker (commented):
> Regarding the input template: I think it makes sense to wait with this until we have broken up `compute_taxes_and_transfers`. We basically need to call `dags.tree.create_input_structure_tree` on the functions tree with partialled parameters.

Yes!

> Also, could you have a look at the type hints? I tried many versions but can't get it to work... 

The trick is to use `from __future__ import annotations`. Here is what [you.com came up with](https://you.com/search?q=how+to+specify+pattern+in+pyproject.toml+to+ignore+mypy+error%3A++error%3A+Missing+type+parameters+for+generic+type+%22dict%22++%5Btype-arg%5D&tbm=youchat&fromExtension=true&cid=c0_57814a97-feea-4a2b-b6e8-913fcab0852e):

> 3. Use from __future__ import annotations
>
> If you are using Python 3.7+ and want to keep the import inside the TYPE_CHECKING block, you can enable deferred evaluation of type annotations using from __future__ import annotations. This makes all type hints lazy (i.e., they are not evaluated at runtime), which can help avoid runtime import issues. For example:

src/ttsim/prepare_data.py
```diff
@@ -0,0 +1,119 @@
...
+                "n1": {
+                    "n2": pd.Series([1, 2, 3]),
+                    "n3": pd.Series([4, 5, 6]),
+                },
+                "n4": pd.Series([3, 3, 3]),
+            }
+
+
+    """
+    _fail_if_df_has_bool_or_numeric_column_names(df)
```
hmgaudecker, line 65: Why do we care? All we need is that we can apply our mapping, no?

src/ttsim/prepare_data.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return dt.unflatten_from_qual_names(name_to_input_series)
+
+
+def _fail_if_mapper_has_incorrect_format(
+    input_tree_to_column_map: NestedInputToSeriesNameDict,
+) -> None:
+    """Fail if the input tree to column name mapping has an incorrect format."""
+    if not isinstance(input_tree_to_column_map, dict):
+        msg = "The input tree to column mapping must be a dictionary."
```
hmgaudecker, line 90: Add a hint on how to get the format? I.e. via `create_input_structure`?

src/ttsim/prepare_data.py
```diff
@@ -0,0 +1,119 @@
...
+        raise TypeError(msg)
+
+    inputs = optree.tree_flatten(input_tree_to_column_map, none_is_leaf=True)[0]
+    if not all(isinstance(x, str | int | bool) for x in inputs):
```
hmgaudecker, line 94: So here we allow ints and bools... Should we add a check that all tree paths consist of strings only? In the previous line, could also use some flattening from `dags.tree`, not sure whether that would make for a more helpful error message directly.

src/ttsim/typing.py
```diff
@@ -25,6 +25,7 @@
     QualNamePolicyInputDict = Mapping[str, PolicyInput]
 
     # Specialise from dags' NestedInputDict to GETTSIM's types.
+    NestedInputToSeriesNameDict = Mapping[str, Any | "NestedInputToSeriesNameDict"]
     NestedDataDict = Mapping[str, pd.Series | "NestedDataDict"]
```
hmgaudecker, line 29: Just a heads-up that the current type `NestedDataDict` is nonsense (we never use series) and will change via #879.

hmgaudecker (commented):
Thanks a lot, looks great!!! And apologies for the piecemeal review, I pressed the wrong buttons in Cursor's interface.

Just a remark explaining my changes to imports:
- Within the ttsim package, always use imports from modules themselves. I.e., never use `from ttsim import ...`
- From tests/ttsim, use `from ttsim import ...` whenever possible, i.e., use module-specific imports only for objects not exported via the main `ttsim` namespace.
- From anything that is `gettsim`, only ever use `from ttsim import ...`. If tempted to use something else, we'll need to adjust the global namespace of `ttsim`.  Exception: `ttsim.typing`.

src/_gettsim/interface.py
```diff
@@ -0,0 +1,83 @@
...
+from _gettsim.config import RESOURCE_DIR, SUPPORTED_GROUPINGS
+from ttsim import (
+    compute_taxes_and_transfers,
+    create_data_tree_from_df,
+    set_up_policy_environment,
+)
+from ttsim.typing import NestedDataDict, NestedInputToSeriesNameDict
+
+
+def quickrun(
```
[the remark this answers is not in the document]
MImmesberger: I'm fine with oss; seems to be a common abbreviation: https://en.wikipedia.org/wiki/One-stop_shop

src/ttsim/prepare_data.py
MImmesberger: We allow for numeric or boolean values as values of the `inputs_tree_to_df_columns`
mapping. Those are broadcasted to match the length of the dataframe. If we allow for
numeric or bool column names, the mapping would be ambiguous; we'd have to make
assumptions about what the user wants.

src/ttsim/prepare_data.py
MImmesberger: > Should we add a check that all tree paths consist of strings only?

I don't see a reason for that. I could imagine users defining boolean namespaces and I
don't see a big issue if they choose to do so, or am I missing something? Internally, we
convert everything to strings anyhow.

hmgaudecker: I guess it would be a problem if we ever j need a round trip, i.e., from qualified names back to tree paths. I could imagine subtle errors creeping up and do not see much of a reason to allow anything but strings.

hmgaudecker: I guess it would be a problem if we ever j need a round trip, i.e., from qualified names back to tree paths. I could imagine subtle errors creeping up and do not see much of a reason to allow anything but strings.

hmgaudecker (approved):
Excellent, thank you!

hmgaudecker:
*(just to be sure, I think the two to-dos left in the original post should be tackled elsewhere)*

PR #858, opened 2025-04-08
Fix test namespace

felixschmitz:
### What problem do you want to solve?

Will close #852. Adapts tests to match GETTSIM src structure.

### Todo

- [ ] Pick an appropriate title.
- [ ] Put `Closes #XXXX` in the first PR comment to auto-close the relevant issue once
      the PR is accepted. This is not applicable if there is no corresponding issue.
- [ ] Document PR in CHANGES.md.

PR #877, opened 2025-04-18
Fix JAX-Jit tests

timmens:

Tests in `test_jax_jit_kindergeld.py` were failing because policy functions were not jittable.

#### Non-Hashable Function in `jit`

The policy functions were non-jittable because the dataclasses were non-frozen and had the equality argument set to True. This implies that the dataclass get an equality method which compares the fields. To not break the equality/hash contract (a == b implies hash(a) == hash(b)), a dataclass with equality method that is not frozen has a deactivated hash. This does not work with `jax.jit`, because for caching JAX requires a hash of the object. By freezing the dataclasses they get their hash back, and everything works nicely again with JAX.

### Todo

- [x] Freeze ttsim_objects dataclasses and update post init of `TTSIMFunction` to be compatible 
- [ ] Understand why list `single_test` in  `kindergeld_policy_test` fixture has only one entry, although the yaml file says there are two outputs
- [ ] Fix `test_compute_taxes_and_transfers_kindergeld`

### DIFF
--- a/src/ttsim/ttsim_objects.py
+++ b/src/ttsim/ttsim_objects.py
@@ -63,7 +63,7 @@ class FKType(StrEnum):
     MUST_NOT_POINT_TO_SELF = "must not point to self"
 
 
-@dataclass
+@dataclass(frozen=True)
 class TTSIMObject:
     """
     Abstract base class for all TTSIM Functions and Inputs.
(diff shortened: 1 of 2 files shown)


hmgaudecker:
Thanks! 

> Understand why list `single_test` in  `kindergeld_policy_test` fixture has only one entry, although the yaml file says there are two outputs

We switched the strategy of testing so that we convert the entire outputs to DataFrames and check at once instead of doing it column by column. First implementation by @felixschmitz in #858 and fully switched on at some later point, probably in #873.

hmgaudecker (approved):
Excellent, thanks a lot!

PR #878, opened 2025-04-22
Use `ast.unparse` instead of `astor.code_gen.to_source`

hmgaudecker:
In limited set of experiments, it produced exactly the same result. `ast.unparse` is available since Python 3.9, so it's fine to use.


timmens (approved):
Very very nice!

I am quite shocked that I did not find this despite extensive googling and LLM queries.

PR #880, opened 2025-04-22
Validate params files

hmgaudecker:
- [x] Add a json (yaml) schema based on GEP-03 
- [x] Make sure manual validation of parameters passes
- Once agreement is reached that this is the correct schema, 
  - [x] make a pre-commit hook out of this
  - [ ] make the document an appendix to GEP-03

We might want to do the last things only after renaming / replacing params in functions by custom types.


hmgaudecker:
Note that the most substantive change is to break up `sozialv_beitr.yaml` into multiple files. Most other changes were just omissions of descriptions / names or misspelled units.

src/_gettsim/sozialversicherung/midijob.py
```diff
@@ -96,9 +70,9 @@ def midijob_faktor_f_mit_minijob_steuerpauschale_bis_2004(
 
     # Sum over the shares which are specific for midijobs.
     pausch_mini = (
-        sozialv_beitr_params["ag_abgaben_geringf_jahresanfang"]["ges_krankenv"]
-        + sozialv_beitr_params["ag_abgaben_geringf_jahresanfang"]["ges_rentenv"]
-        + sozialv_beitr_params["ag_abgaben_geringf_jahresanfang"]["st"]
+        ges_krankenv_params["arbeitgeberpauschale_bei_geringfügiger_beschäftigung"]
```
MImmesberger, line 73: Wow GETTSIM still has some secrets. That's the first time I encountered the `access_different_date` mechanic. The key `ag_abgaben_geringf_jahresanfang` does not exist, it derives the value from `ag_abgaben_geringf`. When processing parameters, `_jahresanfang` overrides the policy_date to `01-01-XXXX`.

That's something we'll definitely get rid of (because eventually we want to pass only the 'active'  params), right?

If so, we'd have to add the `jahresanfang` params to the params yaml.

src/ttsim/params-schema.json
```diff
@@ -0,0 +1,91 @@
...
+            ]
+          },
+          "type": {
+            "type": "string"
+          },
+          "reference_period": {
+            "type": "string",
+            "enum": ["Year", "Quarter", "Month", "Week", "Day"]
+          },
+          "access_prior_parameters": {
```
MImmesberger, line 49: We never use this anywhere. Should we get rid of it?

src/ttsim/params-schema.json
```diff
@@ -0,0 +1,91 @@
...
+              "scalar": { "type": ["number", "null"] },
+              "reference": { "type": ["string", "null"] },
+              "note": { "type": ["string", "null"] },
+              "deviation_from": { "type": ["string", "null"] },
+              "unit": { "type": "string" },
+              "type": { "type": "string" },
+              "min_alter": { "type": "integer" },
+              "max_alter": { "type": "integer" },
+              "betrag": { "type": "number" },
+              "lower_threshold": { "type": ["number", "string"] },
```
MImmesberger, line 78: I think this will always be a number (or inf), no?

src/ttsim/params-schema.json
```diff
@@ -0,0 +1,91 @@
...
+          "^(19|20|21)[0-9]{2}-[0-9]{2}-[0-9]{2}$": {
+            "type": "object",
+            "properties": {
+              "scalar": { "type": ["number", "null"] },
+              "reference": { "type": ["string", "null"] },
+              "note": { "type": ["string", "null"] },
+              "deviation_from": { "type": ["string", "null"] },
+              "unit": { "type": "string" },
+              "type": { "type": "string" },
+              "min_alter": { "type": "integer" },
```
MImmesberger, line 75:
```suggestion
              "min_alter": { "type": "number" },
```
?
Thinking about weird things like the Rente f. Arbeitslose and its 58 1/2 age threshold here.

src/ttsim/params-schema.json
```diff
@@ -0,0 +1,91 @@
...
+                "type": "integer",
+                "minimum": 1
+              }
+            },
+            "required": ["reference_period", "number_of_lags"],
+            "additionalProperties": false
+          }
+        },
+        "patternProperties": {
+          "^(19|20|21)[0-9]{2}-[0-9]{2}-[0-9]{2}$": {
```
MImmesberger, line 66:
```suggestion
          "^(19|20|21)[0-9]{2}-(0[1-9]|1[0-2])-[0-9]{2}$": {
```

Month should not be larger than 12

MImmesberger (commented):
Looks good! The json scheme is just for the hook, we keep the `.yaml`s, right?

src/_gettsim/parameters/ges_rente.yaml
```diff
@@ -2396,6 +2413,11 @@ abzugsrate_hinzuverdienst:
   name:
     de: Abzugsrate für Hinzuverdienst vor der Regelaltersgrenze
     en: Deduction rate for additional earnings during early retirement
+  description:
```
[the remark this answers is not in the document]
hmgaudecker: Yes, we have always required it per [GEP 3](https://gettsim.readthedocs.io/en/stable/geps/gep-03.html), but omissions only became obvious now. I just felt it was out of the scope of this PR to dig out the precise details for all parameters.

src/_gettsim/parameters/wohngeld.yaml
```diff
@@ -1883,7 +1883,7 @@ vermögensgrundfreibetrag:
     en: Since 2009, wealth might be a reason not to grand housing benefit.
   unit: Euro
   1970-01-01:
-    scalar: inf
+    scalar: 999999999
```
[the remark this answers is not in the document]
hmgaudecker: No, because of the validator. But of course it is better to fix that one!

src/_gettsim/sozialversicherung/midijob.py
hmgaudecker: It's actually the taxes / transfers system, which has the secrets :laughing:.

For some things, the values from the beginning of the year matter so that things depending on it do not have to be updated during the year (searching the codebase it seems to be mostly for midijobs). So both parameters are active! We'll definitely want to keep this, though mechanics might see improvements, of course.

src/ttsim/params-schema.json
hmgaudecker: Good one. Yes, this is a leftover from the days where we had the idea of calculating relevant net wages for Arbeitslosengeld I (or the like) in prior periods with one GETTSIM call instead of calling it twice.

hmgaudecker: Was actually just outdated in GEP 3, which the schema is based on. We now use `access_different_date` for this, which can take on the values `vorjahr` and `jahresanfang`.

MImmesberger (approved):
No complaints

PR #885, opened 2025-04-26
Fix ruff complaints.

MImmesberger:

Fix ruff complaints by

- Ignoring `synthetic.py` and its test file because those will be rewritten soon. Same is true for `test_docs.py`.
- Removing old tests:
   - We don't need to test for type conversions anymore.
   - We don't need to explicitly test for the handling of (qualified) source column names for aggregation functions because those are handled by `dags` now.
- Some minor adjustments to the rest of the code


hmgaudecker (approved):
Excellent, thanks!

PR #886, opened 2025-04-26
Fix docs build

MImmesberger:

Rough fix of the readthedocs build.

- Remove list of "typical outputs" (because `_gettsim.functions` does not exist anymore)
- Remove list of all policy functions (because `_gettsim.functions` does not exist anymore)
- Move outdated tutorials to a different folder. (all of them rely on `create_synthetic_data` or the visualisation mechanic, none of them works anymore (and we have rewrite them soon anyhow))


hmgaudecker (approved):
Sweet, thanks!

PR #888, opened 2025-04-28
Infer groupings from objects tree

MImmesberger:

Infer groupings from the objects tree. This needs to be done by looking for names in the top-level namespace that end with "_id". Filtering for `group_creation_functions` does not work because this would miss `hh_id`.

**Changes**

- Removed explicit `groupings` argument from `compute_taxes_and_transfers`
- Added the `grouping_levels` property to the policy environment.
- Moved the `_fail_if_group_ids_are_outside_top_level_namespace` check to the policy environment.

### DIFF
--- a/tests/ttsim/mettsim/inputs.py
+++ b/tests/ttsim/mettsim/inputs.py
@@ -7,8 +7,8 @@ def p_id() -> int:
 
 
 @policy_input()
-def hh_id() -> int:
-    """Household ID."""
+def kin_id() -> int:
+    """Kinstead ID."""
 
 
 @policy_input(foreign_key_type=FKType.MUST_NOT_POINT_TO_SELF)
--- a/tests/ttsim/mettsim/inputs.py
+++ b/tests/ttsim/mettsim/inputs.py
@@ -8,7 +8,7 @@ def p_id() -> int:
 
 @policy_input()
 def hh_id() -> int:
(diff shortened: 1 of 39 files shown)


tests/ttsim/mettsim/inputs.py
```diff
@@ -8,7 +8,7 @@ def p_id() -> int:
 
 @policy_input()
 def hh_id() -> int:
-    """Household id (will delete once fam_id is enough)."""
+    """Household ID."""
```
hmgaudecker, line 11: So, can't we get rid of this now? Or rename it, in case we want to test something that is a policy input only.

hmgaudecker (commented):
Excellent! Just one question left -- I guess finding another name is the way to go.

tests/ttsim/mettsim/inputs.py
hmgaudecker: I'd love to keep GETTSIM and METTSIM fully separate, so that when looking at things we are very sure where we are. Especially given the ad-hoc history of adding `hh`.

My favourite LLM-answer to a household-like concept in middle earth would be 

> Kinstead: A combination of "kin" and "stead," this name highlights the familial ties and the dwelling place of a family. 

with associated `kin` / `kin_id`. But feel free to use something else. In any case, it should be a very quick search & replace, I'd hope.

PR #889, opened 2025-04-29
Update loader, fixes #830

### LINKED ISSUE #830: BUG: _fail_if_multiple_policy_functions_are_active_at_the_same_time needs to be checked at namespace level
### Bug description

- The check used to be at the global level
- Moved it to the module level in #780, that is wrong now we decided that namespaces are at the directory level
- So need to check there.

hmgaudecker:
Update the loader so we first get all objects including module names and original function names. 

Then do checks based on that and filter for active objects in policy environment.

### DIFF
--- a/src/ttsim/policy_environment.py
+++ b/src/ttsim/policy_environment.py
@@ -265,7 +265,10 @@ class ConflictingTimeDependentObjectsError(Exception):
 
         have overlapping start and end dates. The following functions are affected:
 
-          {"\n          ".join(overlapping_objects)}
+          {
+            '''
+          '''.join(overlapping_objects)
+        }
 
         Overlap from {self.overlap_start} to {self.overlap_end}."""
 
--- a/src/ttsim/policy_environment.py
+++ b/src/ttsim/policy_environment.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import copy
 import datetime
+import itertools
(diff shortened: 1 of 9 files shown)


hmgaudecker:
@MImmesberger, please ignore the failing mypy stuff; I based it on #887 and then decided to reduce the scope relative to what I had originally planned, so that other one will take a little longer.

src/ttsim/policy_environment.py
```diff
@@ -162,7 +164,10 @@ def set_up_policy_environment(
     # Check policy date for correct format and convert to datetime.date
     date = to_datetime(date)
 
-    functions_tree = load_objects_tree_for_date(resource_dir=resource_dir, date=date)
+    # Will move this line out eventually. Just include in tests, do not run every time.
```
MImmesberger, line 167: You mean we should move this check to a test? Should we do it right now?

MImmesberger (approved):
Looks good. Testing fails on Linux only can you have a look? Can't reproduce that.

src/ttsim/policy_environment.py
hmgaudecker: No, *right* now we should not change functionality. But with the upcoming interface, we'll either run it or not one level up. Or just add a test for that indeed, let's discuss then.

PR #891, opened 2025-05-01
Prepare for jitting / vectorization of GETTSIM

hmgaudecker:

A brief shot at trying to do the same for GETTSIM what #879 did for METTSIM.

- Good news: `test_full_taxes_and_transfers` runs nicely!
- Bad news: Many tests fail because the logic in dividing up the taxes and transfers function is not elaborate enough. In the first pass, we are trying to build all ids. However, in many cases we are missing the required input data (made-up example: `Einkommensteuer` tests may require calculation of `sn_id`, but won't have all inputs required for `bg_id`).

Solving this should be doable (first set up the entire graph, then check which ids are needed), but the required functions are buried inside of `compute_taxes_and_transfers` so we should not waste time on that before implementing the new interface.


src/_gettsim_tests/utils.py
```diff
@@ -51,7 +56,7 @@ def __init__(
         date: datetime.date,
     ) -> None:
         self.info = info
-        self.input_tree = input_tree
+        self.input_tree = optree.tree_map(np.array, input_tree)
```
MImmesberger, line 59: Wouldn't we want `np.array([1, 2, 3])` instead of `[np.array(1), np.array(2),
np.array(3)]` here?

MImmesberger (approved):
I think the logic of calculating IDs beforehand and calculate the rest of the DAG afterwards fails for stuff like `bg_id` and `wthh_id`. Their dependencies consist basically of almost the entire DAG. (That's probably what you meant, I just want to make sure we're one the same page). So users would not get speed benefit if their data is large, but only if they want to call GETTSIM repeatedly on the same (stable - with respect to BD and WTHH) household constellation.

hmgaudecker:
> I think the logic of calculating IDs beforehand and calculate the rest of the DAG afterwards fails for stuff like `bg_id` and `wthh_id`. Their dependencies consist basically of almost the entire DAG. (That's probably what you meant, I just want to make sure we're one the same page). So users would not get speed benefit if their data is large, but only if they want to call GETTSIM repeatedly on the same (stable - with respect to BD and WTHH) household constellation.

And yes, I'd think that will be a fairly common use case. And of course, one could do 3+ calls of GETTSIM; just that we can't / probably don't want to bake that into `gettsim.oss`.

PR #894, opened 2025-05-05
Replace Proxy Renteneinkommen Vorjahr with policy input

MImmesberger:

- Replaces the proxy for Renteneinkommen Vorjahr with policy input `renteneinkünfte_vorjahr_m`.
- Removes tests of the proxy


MImmesberger:
For Elterngeld, we put zvE in the Elterngeld namespace because zvE was defined relative to the date of birth of the child.

Here, pension income is reported by the Finanzamt to the DRV and stems from the income tax returns, so I'm in favour of keeping it in this namespace.

PR #896, opened 2025-05-06
Remove fake dependencies, fixes #666.

### LINKED ISSUE #666: BUG: Scalar instead of list returned if function inputs are parameters only

hmgaudecker:
Noted in the process of parameter conversion that all tests pass when fake dependencies are removed. As that PR is going to be large, I move this out to a separate one.

### DIFF
--- a/src/_gettsim/sozialversicherung/rente/altersrente/besonders_langjährig/besonders_langjährig.py
+++ b/src/_gettsim/sozialversicherung/rente/altersrente/besonders_langjährig/besonders_langjährig.py
@@ -8,10 +8,7 @@ from ttsim import policy_function
     end_date="2014-06-22",
     leaf_name="altersgrenze",
 )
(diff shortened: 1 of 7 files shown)


MImmesberger (approved):
Always great to fix issues accidentally 😃

PR #897, opened 2025-05-06
Unify parsing of params

hmgaudecker:
Unify the parsing of parameters, as outlined in GEP 6.

- [ ] Basic structure for METTSIM
  - Add `TTSIMParam` dataclass including basic subclasses (`ScalarTTSIMParam`, `DictTTSIMParam`, `ListTTSIMParam`). 
  - Clean up many inconsistencies in params definition
  - Put infrastructure into place for METTSIM to use the new machinery
- [ ] Add `piecewise_constant`, `piecewise_linear`, `piecewise_quadratic` tests to METTSIM
- [ ] Implement automatic parsing of piecewise param types
- [ ] Add `CustomTTSIMParam` test to METTSIM
- [ ] Convert GETTSIM parameter files
  1. Rename parameters within the current structure, add `type` property everywhere, make sure they conform to requirements (e.g., homogeneity of dicts)
  2. Break up files one-by-one and use new infrastructure.


hmgaudecker:
I think this is done -- the last open to-do is more about the interface than parsing of parameters. Should we merge, @MImmesberger ?

PR #898, opened 2025-05-06
Unify parsing of basic METTSIM params, update "piecewise" infrastructure

hmgaudecker:
Implement the first set of to-dos in #897.

One main open question for the review is whether we want to mark parameters in some way to distinguish them in function arguments from data columns / policy functions.

One possibility could be to prefix with `param_` or the like.

- Pro: More obvious
- Cons: More verbose, more difficult to see the similarity when searching (else can easily search for the entire string)

I tend towards the current implementation of not explicitly distinguishing. Clear names should be easy enough to do (@MImmesberger, please do not take METTSIM as the example when thinking this through, parameter names clearly have not gotten as much love as they will be getting in GETTSIM).


src/ttsim/ttsim_params.py
```diff
@@ -0,0 +1,143 @@
...
+from ttsim.shared import to_datetime, validate_date_range
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+    from ttsim.config import numpy_or_jax as np
+    from ttsim.typing import DashedISOString
+
+DEFAULT_START_DATE = datetime.date(1900, 1, 1)
+DEFAULT_END_DATE = datetime.date(2100, 12, 31)
```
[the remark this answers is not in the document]
hmgaudecker: Will delete, just a leftover from just copying the file.

src/ttsim/policy_environment.py
```diff
@@ -56,6 +69,7 @@ def __init__(
         self,
         raw_objects_tree: NestedTTSIMObjectDict,
         params: dict[str, Any] | None = None,
+        params_tree: FlatTTSIMParamDict | None = None,
```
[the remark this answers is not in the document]
hmgaudecker: Good one. I thought the same and called `dt.unflatten_from_tree_paths` in the return statement of `active_ttsim_params_tree`, but did not adjust the type hints. A bit surprising this did not get flagged...

src/ttsim/policy_environment.py
```diff
@@ -576,16 +773,23 @@ def set_date_to_beginning_of_year(date: datetime.date) -> datetime.date:
...
+                        out_params[param] = _parse_raw_parameter_group(
+                            raw_group_data=raw_group_data,
+                            date=new_date,
+                            group=group,
+                            parameters=[param],
                         )[param]
                     elif "." in policy_in_place["deviation_from"]:
+                        assert (  # noqa: PT018
+                            group == "arbeitsl_geld_2"
+                            and param == "eink_anr_frei_kinder"
```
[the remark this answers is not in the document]
hmgaudecker: Yes, this is just left in place to ensure a piecemeal transition will be possible for GETTSIM. Will be solved via explicit conversion, no reason to put the extra logic into the parameter files anymore.

src/ttsim/ttsim_params.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass(frozen=True)
+class ListTTSIMParam(TTSIMParam):
+    """
+    A TTSIM parameter directly read from a YAML file that is a list.
+    """
+
+    value: list[float] | list[int] | list[bool]
+
+
+def _convert_and_validate_dates(
```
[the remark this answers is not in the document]
hmgaudecker: :heavy_check_mark:

src/_gettsim_tests/utils.py
```diff
@@ -114,8 +114,14 @@ def execute_test(test: PolicyTest, jit: bool = False) -> None:
     flat_expected_output_tree = dt.flatten_to_qual_names(test.expected_output_tree)
 
     if flat_expected_output_tree:
-        expected_df = pd.DataFrame(flat_expected_output_tree)
-        result_df = pd.DataFrame(flat_result)
+        expected_df = pd.DataFrame.from_dict(
```
hmgaudecker, line 117: > Fix bug that caused Dataframe creation to fail if only one target and one individual in test data.

Turns out this brought back the behaviour in #666 / #896 though / made it problematic again. For our computations, it is perfectly fine to have scalar outputs of functions, these will just need to be broadcasted in the output step.

Can you please 
1. revert, 
2. add a test case for the behaviour that led you to make this change, 
3. and then we'll see about it?  Likely a special casing here, potentially relevant for other ways to output things.

Thanks!

MImmesberger: I did this because in the former version, parsing to pd.DataFrame did not
work with tests that have only one expected output column (e.g. `{"a": np.array(1)}` as expected output tree). I'll see whether we can get the best of both worlds.

hmgaudecker: Yes, I got that! Apparently we did not have any test like that so far, would be great if some failed.

But it will be just a matter of fixing the output formatting, no big deal.

src/ttsim/typing.py
```diff
@@ -36,3 +40,25 @@
...
+        str
+        | dict[
+            datetime.date,
+            dict[
+                str | int,
+                str
+                | float
+                | int
+                | bool
+                | list[float | int | bool]
```
MImmesberger, line 54:
```suggestion
```

MImmesberger (approved):
We definitely need a nice documentation about how to specify a piecewise polynomial
function in GETTSIM, it was hard to specify even the simple case for METTSIM. Its not obvious what is optional and what is not.

I tried to think about some checks of the user input / yamls that catch the most obvious errors. For example, its easy to accidentally introduce jumps in the piecewise function (but users might want to do exactly that -> can't check for that), forgetting to define it on the real line (do we check that?), or having duplicate int keys in the spec (I think its not possible to check for this, right?).
There might be more common errors, but a good documentation might help in most cases already.

src/_gettsim_tests/utils.py
hmgaudecker: Thanks! 

Looking at the failure, I believe we could just temporarily pass `p_id` as the index. It is guaranteed to exist and to have the correct length. And then call `.reset_index(drop=True)` immediately.

hmgaudecker:
> We definitely need a nice documentation about how to specify a piecewise polynomial function in GETTSIM, it was hard to specify even the simple case for METTSIM. Its not obvious what is optional and what is not.

Agreed. Most of that code has not been touched in 5 years.

A problem is that we decided to have the special keys `note` and `reference` in the `YYYY-MM-DD` keys, so we are stuck with dicts and can't have lists, which would have helped on the duplicate keys problem. Still surprised there are no complaints by checkers... But we are stuck with that format, it has worked well enough and will continue to do so.

It would be great if you could start with a rough draft, just bullets, of the documentation and just flag the points where you struggled. Best to do it soon, as the memory is still fresh.

> its easy to accidentally introduce jumps in the piecewise function (but users might want to do exactly that -> can't check for that),

Indeed, I do think this happened to lawmakers in the past!

src/_gettsim_tests/utils.py
hmgaudecker: I just added another to-do to the description of #897 that should take care of this in a unified manner. Comments welcome, feel free to edit there!

MImmesberger:
I think we can merge!

PR #900, opened 2025-05-07
Rename GETTSIM params and fix yaml validation issues

hmgaudecker:
Next set of to-dos from #897.

- Rename the parameters in GETTSIM's yaml files
- Restructure where useful (often, moving from scalars to dicts does wonders for readability)
- Add now-required unit, reference_period, type keywords

- [x] abgelt_st.yaml
- [x] arbeitsl_geld_2.yaml
- [x] arbeitsl_geld.yaml
- [x] arbeitslosenversicherung.yaml
- [x] eink_st_abzuege.yaml
- [x] eink_st.yaml
- [x] elterngeld.yaml
- [x] erwerbsm_rente.yaml
- [x] erziehungsgeld.yaml
- [x] geringfügige_einkommen.yaml
- [x] ges_krankenv.yaml
- [x] ges_pflegev.yaml
- [x] ges_rentenv.yaml
- [x] ges_rente.yaml
- [x] grunds_im_alter.yaml
- [x] __init__.py
- [x] kindergeld.yaml
- [x] kinderzuschl.yaml
- [x] lohnst.yaml
- [x] soli_st.yaml
- [x] unterhaltsvors.yaml
- [x] unterhalt.yaml
- [x] wohngeld.yaml


src/_gettsim/parameters/wohngeld.yaml
```diff
@@ -18,13 +18,14 @@ faktor_berechnungsformel:
...
+    en: Parameters a, b, c of the Wohngeld formula
   description:
     de: >-
       WoGG - Anlage 2 (bis 2019 Anlage 1) (zu § 19 (1)). Seit 2022 auch §24
       Wohngeldverordnung (WoGV).
-    en: null
-  note: The keys from 1 to 12 below refer to the household size.
+      Die Schlüssel von 1 to 12 beziehen sich auf die Haushaltsgröße.
+    en: The keys from 1 to 12 below refer to the household size.
+  type: require_converter
```
hmgaudecker, line 28: @MImmesberger: In case you find some time before tomorrow afternoon (I won't find much), the to-do *"Add CustomTTSIMParam test to METTSIM"* means anything that is not a scalar / flat homogenous dict / piecewise polynomial. For example the nested dict found here. Would be great to have something like that in METTSIM so we can test there.

MImmesberger: Will do it tomorrow before noon.

MImmesberger: Added an orc hunting bounty in [97ff74a](https://github.com/iza-institute-of-labor-economics/gettsim/pull/900/commits/97ff74ad2921f7bb7eae59aa71f9725b22f297bf). I probably did not specify the type hint for the parameter correctly. Also I'm not sure what the content of the `type` key in the yaml should be (didn't look at this PR just yet). So you may have to update that.

MImmesberger:
How far do we want to go when it comes to merging parameter specs? My suggestion: If it relates to the same concept and is defined at the same place (same § or even same Absatz), it should be in one spec.

Example 1:

In [d528ac7](https://github.com/iza-institute-of-labor-economics/gettsim/pull/900/commits/d528ac7492c8b29396ed869c70b79cbfcc4e64b5) I merged all parameters related to the ALG2 Regelsatz calculation into the `regelsatz` key. Downside: non-homogeneous dicts. 

The new structure looks like this (everything except for `basissatz` was under different parameter keys before). It looks complicated but the Bedarfsstufen starting in 2011 are not much nicer.

```yaml
  2005-01-01:
    basissatz: 338
    anteil_vom_regelsatz_bei_zwei_erwachsenen: 0.9
    anteil_vom_regelsatz_bei_weiteren_erwachsenen: 0.8
    anteil_vom_regelsatz_für_kinder:
      zwischen_14_und_24_jahre:
        min_alter: 14
        max_alter: 24
        anteil: 0.8
      zwischen_6_und_13_jahre:
        min_alter: 6
        max_alter: 13
        anteil: 0.6
      bis_5_jahre:
        min_alter: 0
        max_alter: 5
        anteil: 0.6
```

Example 2:

This one is a bit more difficult. I would like to merge `vermögensgrundfreibetrag_obergrenze`, `vermögensgrundfreibetrag`, `schonvermögen_bürgergeld`, `vermögensfreibetrag_austattung`, `vermögensfreibetrag_kind` under `vermögensfreibetrag`. All defined in §12 Abs. 2 SGB II and the underlying concept "Vermögensfreibetrag" is a convolute of all of these factors. However, the resulting dict could be overloaded.

src/_gettsim/parameters/arbeitsl_geld_2.yaml
```diff
@@ -87,30 +89,35 @@ eink_anr_frei:
...
+      Obere Einkommensgrenze bei anzurechnendem Einkommen, Intervall 3,
+      Bedarfsgemeinschaft mit Kindern. Gilt statt der Parameter in
+      parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg für Leistungsbezieher, die
+      mit minderjährigem Kind in BG leben oder die ein minderjähriges Kind haben
+      (zweiteres wird aktuell noch nicht im Code beachtet). § 30 SGB II. Seit 01.04.2011
+      § 11b SGB II (neugefasst durch B. v. 13.05.2011 BGBl. I S. 850. Artikel 2 G. v.
+      24.03.2011 BGBl. I S. 453). Ersetzt Parameter in
+      parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg, falls nicht vorhanden
     en: null
   type: piecewise_linear
```
hmgaudecker, line 111: This must become `require_converter` because we will get rid of the `deviation_from` method. Ignore for now.

hmgaudecker:
> How far do we want to go when it comes to merging parameter specs? My suggestion: If it relates to the same concept and is defined at the same place (same § or even same Absatz), it should be in one spec.
>
> Alternative: A spec-driven approach, i.e. try to find a balance between a grouping that makes sense and the need of parameter conversions.

I'd say more like the latter, i.e., simplicity of the code. As soon as we have heterogeneous  parameters, we cannot use the built-in machinery, so there simply is more code to maintain.  

> Example 1:
> 
> In [d528ac7](https://github.com/iza-institute-of-labor-economics/gettsim/pull/900/commits/d528ac7492c8b29396ed869c70b79cbfcc4e64b5) I merged all parameters related to the ALG2 Regelsatz calculation into the `regelsatz` key. Downside: non-homogeneous dicts.
> 
> The new structure looks like this (everything except for `basissatz` was under different parameter keys before). It looks complicated but the Bedarfsstufen starting in 2011 are not much nicer.

Agreed this is better

> Example 2:
> 
> This one is a bit more difficult. I would like to merge `vermögensgrundfreibetrag_obergrenze`, `vermögensgrundfreibetrag`, `schonvermögen_bürgergeld`, `vermögensfreibetrag_austattung`, `vermögensfreibetrag_kind` under `vermögensfreibetrag`. All defined in §12 Abs. 2 SGB II and the underlying concept "Vermögensfreibetrag" is a convolute of all of these factors. However, the resulting dict could be overloaded.

My initial reaction was: Why would a dict with 4-5 entries be overloaded?

But then I looked into the structure... Probably leave it more like the current thing, though we need to be much clearer what the years of birth mean. I guess "1948" is "1948 and later"? Then we should change the parameter once this becomes irrelevant (I guess in 2015 or so, i.e. `ceil(1948+65.x)` ?)

src/_gettsim/parameters/arbeitsl_geld_2.yaml
```diff
@@ -87,90 +89,145 @@ eink_anr_frei:
...
       upper_threshold: 1500
   2023-07-01:
     reference: G. v. 20.12.2022 BGBl. I S. 2328.
     note: Bürgergeld-Gesetz
-    deviation_from: arbeitsl_geld_2.eink_anr_frei
+    deviation_from: arbeitsl_geld_2.parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg
     4:
       upper_threshold: 1500
-regelsatz:
+regelsatz_anteilsbasiert:
```
hmgaudecker, line 123: @MImmesberger, just a quick heads-up that we should have different parameter names in case the internal structure changes completely over time. Somewhat similar to the `dates_active + leaf_name`, just that we don't have a leaf name here (could mimic through converter functions, though).

tests/ttsim/test_compute_taxes_and_transfers.py
```diff
@@ -882,3 +987,33 @@ def test_get_top_level_namespace(environment, time_units, expected):
...
+    processed_params_tree = _process_params_tree(
+        params_tree=params_tree,
+        params_functions=params_functions,
+    )
+    expected = {
+        "some_converting_params_func": ConvertedParam(
+            some_float_param=1,
+            some_bool_param=False,
+        ),
+        "some_scalar_params_func": 1,
```
MImmesberger, line 1014: This line reveals that the return type hint for `_process_params_tree` is wrong. Probably we should change the type hint rather than the output of this function, correct?

tests/ttsim/test_policy_environment.py
```diff
@@ -555,3 +570,15 @@ def test_active_ttsim_objects_tree(
...
     assert accessor(functions_last_day).__name__ == function_name_last_day
     assert accessor(functions_next_day).__name__ == function_name_next_day
+
+
+def test_get_params_contents_with_updated_previous(
+    some_params_spec_with_updates_previous,
+):
+    params_contents = _get_params_contents(some_params_spec_with_updates_previous)
+    expected = {
+        "updates_previous": True,
```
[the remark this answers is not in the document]
MImmesberger: (or rather: remove it from the result, such that we can remove it from expected 😄 )

We don't need it after this step, right? A bit confusing having `updates_previous` around under `param.value`, especially if users don't know where it came from or what it meant.

tests/ttsim/test_compute_taxes_and_transfers.py
hmgaudecker: Done in 9bbfb1517f7ca49e6e80ee8fdd033d2768b4b28f

tests/ttsim/test_policy_environment.py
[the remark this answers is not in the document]
hmgaudecker: Yes! Very good! Done in d3aa1962a.

hmgaudecker (commented):
Much improved so far! Just leaving a checkpoint.

MImmesberger:
I noticed that it might make sense to consolidate the `ges_rentenv`, and `ges_rente` yamls into `rente` and `grundrente` files (but maybe you're doing that right now). 

Also, I deleted lots of parameters that are relevant to the calculation of the aktueller Rentenwert, which is just a parameter in GETTSIM. We never made this endogenous and I think we shouldn't start to do so.

hmgaudecker:
> I noticed that it might make sense to consolidate the `ges_rentenv`, and `ges_rente` yamls into `rente` and `grundrente` files (but maybe you're doing that right now).

No, that'll be the next PR, when we will move the files from their own directory into the ones where they belong in the tree. This will involve lots of breaking up, which will destroy history. This is why I wanted to limit the scope to renaming / internal structure here.

> Also, I deleted lots of parameters that are relevant to the calculation of the aktueller Rentenwert, which is just a parameter in GETTSIM. We never made this endogenous and I think we shouldn't start to do so.

Could you revert that one, please? I agree with you for the moment, but can imagine lots of application where it will become relevant.

hmgaudecker:
Great teamwork, thanks @MImmesberger! 

Do you want to have a brief look at the changes in [449873f9](https://github.com/iza-institute-of-labor-economics/gettsim/pull/900/commits/449873f9c970f3e43a5b60eb5274b8afcec1384f] (made before merging #904) and [9d8ae7a05](https://github.com/iza-institute-of-labor-economics/gettsim/pull/900/commits/9d8ae7a058496be5ca50eeec86032b1745b66afb) (last thing I did)  before I merge?

MImmesberger:
Yes! Will do in an hour or so

MImmesberger (approved):
Looks good! Is it intentional that most of the suggestions from the last review in #904 are not implemented? The commit where I expected those to be implemented changes only two lines ([4fcc163](https://github.com/iza-institute-of-labor-economics/gettsim/pull/904/commits/4fcc1639525f50d13a37b95a123753ce8330f588))

hmgaudecker:
Shoot, that was just my inability to handle GitHub's interface. Will try to get them back after the seminar...

PR #902, opened 2025-05-09
Add params_function

hmgaudecker:
Next to-do from #897, puts the last bit of infrastructure in place.

- Introduces `ParamsFunction`, similar to `PolicyFunction` but only for parsing params
- These are read into the usual tree upfront, but are executed before the DAG of `PolicyFunction`s & friends
- The partialling is done with the outputs of this, and all parameters removed from the tree that are marked as `require_conversion`.

In addition, we add stuff like including the params tree in checks for tree sanity etc.

Testing could certainly be improved, feel free to add some when you look at it @MImmesberger :innocent:. But I can do that next week, too.


tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
```diff
@@ -1,23 +1,48 @@
...
+    peasant_hunter: float
+
+
+@dataclass(frozen=True)
+class BountyPerOrcs:
+    small_orc: int
+    large_orc: BountyPerLargeOrcs
+
+
+@params_function()
```
MImmesberger, line 24: I am a bit surprised that there is no difference between functions that process params
(like this one) and functions that return a scalar param only. The former will never be
a target, the latter may be. Should we enforce this?

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
```diff
@@ -1,23 +1,48 @@
...
+
+from ttsim import params_function, policy_function
+
+if TYPE_CHECKING:
+    from ttsim.typing import RawParamsRequiringConversion
+
+
+@dataclass(frozen=True)
+class BountyPerLargeOrcs:
+    noble_hunter: float
```
MImmesberger, line 14: Shouldn't this inherit from a TTSIM Object? Besides the scalars, we lose all information defined in the yaml.

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
```diff
@@ -1,23 +1,48 @@
...
 
 
 @policy_function(vectorization_strategy="vectorize")
 def amount(
     small_orcs_hunted: int,
     large_orcs_hunted: int,
     parent_is_noble: bool,
-    orc_hunting_bounty: DictTTSIMParam,
+    bounty_per_orc: BountyPerOrcs,
 ) -> float:
```
MImmesberger, line 41: I see why we want to do it this way in GETTSIM's source code. However, I'm wondering
whether we can still support just throwing a nested dict here (and in the params yaml).
Ofc typing will not work, but most users wouldn't want to do this anyhow. Is there
something I'm missing here?

src/ttsim/policy_environment.py
```diff
@@ -100,6 +107,16 @@ def params(self) -> dict[str, Any]:
         """The parameters of the policy environment."""
         return self._params
 
+    @property
+    def params_tree(self) -> dict[str, Any]:
+        """The parameters of the policy environment."""
+        return self._params_tree
+
+    @property
+    def combined_tree(self) -> dict[str, Any]:
```
MImmesberger, line 116: This should only ever be called on the flat versions, right?

MImmesberger (approved):
Cool, thanks!

MImmesberger (approved):
Cool, thanks!

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
hmgaudecker: Now I am confused what difference there could possibly be? All functions we have so far that return a scalar for sure will become a `params_function`. Sure, you may want to request them as outputs, but why would you want to have an entire column filled with a single value? The only reason might be when it changes over time and at some point starts depending on the data. But that will just continue to work.

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
hmgaudecker: That continues to be supported, but I want to show off good practice here!

src/ttsim/policy_environment.py
hmgaudecker: Good one. Just opposite, copy & paste error!

hmgaudecker:
> Edit: Still have to look into the tests.

Thanks! Just add to #900 then, will merge into that one.

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
MImmesberger: My point is that `bounty_per_orc` returns a `BountyPerOrc` object, while other `params_function` may only return a scalar. It makes sense to request the latter as a target, but I'm not sure about the former. If we allow for that, it definitely changes how we define GETTSIM's output (e.g. type hint is wrong).

src/ttsim/policy_environment.py
MImmesberger: Oh this was not about the type hint, but about the implementation of the function. For nested dicts, we can't have overlapping keys at the top level, which may be the case at some point, right? We should use `merge_trees` here.

tests/ttsim/mettsim/orc_hunting_bounty/orc_hunting_bounty.py
hmgaudecker: I was trying to say that scalars won't make much sense, either, except for special cases. 

This said, for tree output, anything goes. In case of df output, we should throw an error when requesting dict conversion. See [ii](https://github.com/iza-institute-of-labor-economics/gettsim/pull/897#issue-3041987819) under that to-do; that will still need to be settled.

src/ttsim/policy_environment.py
hmgaudecker: Excellent, yes! Done in f3276cd69. Would be good to add a test for this.

PR #905, opened 2025-05-12
Make ID creation functions jittable

mj023:

Make the ID creation functions in Gettsim and Mettsim jittable.

### Todo

- [x] Make Gettsim ID's jittable
- [x] Make METTSIM ID's jittable.
- [x] Change tests to the new ID values

### DIFF
--- a/src/_gettsim/ids.py
+++ b/src/_gettsim/ids.py
@@ -31,7 +31,7 @@ def ehe_id(
         + np.minimum(p_id, p_id_ehepartner_or_own_p_id) * n
     )
 
-    return __reorder_ids(result)
+    return _reorder_ids(result)
 
 
 @group_creation_function()
@@ -61,7 +61,7 @@ def fg_id(
             familie__p_id_elternteil_2 == p_id[i], i, p_id_elternteil_2_loc
         )
 
-    children = np.isin(p_id, familie__p_id_elternteil_1) + np.isin(
+    children = np.isin(p_id, familie__p_id_elternteil_1) | np.isin(
         p_id, familie__p_id_elternteil_2
     )
 
@@ -74,17 +74,17 @@ def fg_id(
         + np.minimum(p_id, arbeitslosengeld_2__p_id_einstandspartner) * n,
     )
 
-    fg_id = __assign_parents_fg_id(
+    fg_id = _assign_parents_fg_id(
         fg_id, p_id, p_id_elternteil_1_loc, hh_id, alter, children, n
     )
-    fg_id = __assign_parents_fg_id(
+    fg_id = _assign_parents_fg_id(
         fg_id, p_id, p_id_elternteil_2_loc, hh_id, alter, children, n
     )
 
-    return __reorder_ids(fg_id)
+    return _reorder_ids(fg_id)
 
 
-def __assign_parents_fg_id(
+def _assign_parents_fg_id(
     fg_id: np.ndarray,
     p_id: np.ndarray,
     p_id_elternteil_loc: np.ndarray,
@@ -143,7 +143,7 @@ def bg_id(
         fg_id,
     )
 
-    return __reorder_ids(bg_id)
+    return _reorder_ids(bg_id)
 
 
 @group_creation_function()
@@ -166,7 +166,7 @@ def eg_id(
         + np.minimum(p_id, p_id_einstandspartner__or_own_p_id) * n
     )
 
-    return __reorder_ids(result)
+    return _reorder_ids(result)
 
 
 @group_creation_function()
@@ -187,7 +187,7 @@ def wthh_id(
         hh_id + offset,
         hh_id,
     )
-    return __reorder_ids(wthh_id)
+    return _reorder_ids(wthh_id)
 
 
 @group_creation_function()
@@ -214,10 +214,10 @@ def sn_id(
         + np.minimum(p_id, p_id_ehepartner_or_own_p_id) * n
     )
 
-    return __reorder_ids(result)
+    return _reorder_ids(result)
 
 
-def __reorder_ids(ids: np.ndarray) -> np.ndarray:
+def _reorder_ids(ids: np.ndarray) -> np.ndarray:
     """Make ID's consecutively numbered."""
     sorting = np.argsort(ids)
     ids_sorted = ids[sorting]


mj023:
Some of the tests are failing, because I still have to adjust the expected values to the new IDs, but others, like `test_groupings[groupings/2023-01-01/mehrere_haushalte_durchmischt.yaml]`, are failing, because they do not use ordered, continuously numbered ```p_id``` values. I can probably come up with some kind of workaround, but it will be computationally expensive and requiring them to be sorted and continuously numbered does not seem unreasonable to me, what do you think @hmgaudecker ?

mj023:
It was only problematic for the `fg_id` function, so I decided to just lookup the correct indices for `familie__p_id_elternteil`. This is O(n²), if it turns out to be problematic (when the number of p_ids is huge), we can switch to a hash table, but it's probably overkill for now. The only restriction is now that `p_id` values should not be higher than 2³² or we get an overflow with int64.

In 4fbb2275ae454f49471530c2ec1c5105a8077bb4 I encountered some test where parents were married to their children, I assumed that was a mistake and fixed them.

hmgaudecker:
Thanks a lot!! 
 
> It was only problematic for the fg_id function, so I decided to just lookup the correct indices for familie__p_id_elternteil. This is O(n²), if it turns out to be problematic (when the number of p_ids is huge), we can switch to a hash table, but it's probably overkill for now. The only restriction is now that p_id values should not be higher than 2³² or we get an overflow with int64.

Great, thanks! And agreed.

However, we should check these things upfront. Should we add this to the `group_creation_function` decorator, which can then set up a mechanism to perform that check? If there is no performance penalty to large `num_segments`, we could even use these maximum values for that and get rid of the computation-before-jitting-stuff, which in its current naive form is causing problems when we only need a subset of group ids (lots of those GETTSIM test failures, but also the METTSIM tess in #897). What do you think?

> In [4fbb227](https://github.com/iza-institute-of-labor-economics/gettsim/commit/4fbb2275ae454f49471530c2ec1c5105a8077bb4) I encountered some test where parents were married to their children, I assumed that was a mistake and fixed them.

For sure. @MImmesberger, do we still have that open issue for sanity checks? If so, could you please add this to the list?

hmgaudecker:
> > It was only problematic for the fg_id function, so I decided to just lookup the correct indices for familie__p_id_elternteil. This is O(n²), if it turns out to be problematic (when the number of p_ids is huge), we can switch to a hash table, but it's probably overkill for now. The only restriction is now that p_id values should not be higher than 2³² or we get an overflow with int64.
> 
> Great, thanks! And agreed.
> 
> However, we should check these things upfront. Should we add this to the `group_creation_function` decorator, which can then set up a mechanism to perform that check? If there is no performance penalty to large `num_segments`, we could even use these maximum values for that and get rid of the computation-before-jitting-stuff, which in its current naive form is causing problems when we only need a subset of group ids (lots of those GETTSIM test failures, but also the METTSIM tess in #897). What do you think?

Apologies, this was a bit cryptic I guess. I changed the base to the current "HEAD" so the issue becomes apparent. In the tests, we are currently pre-computing the ids so we have the num_segments parameter. Search for `def execute_tests` and then see what happens under `if IS_JAX_INSTALLED:`. This is done for all groupings, which is problematic when we only need a subset of targets that don't require all groupings. In those cases, the input data will often not include the dependencies of the `group_creation_function`s. Happens all the time in GETTSIM, now also in a couple of METTSIM test cases.

It would be best if we could get rid of this altogether. But we need the `num_segments`. 

How about changing the group ids upon return, so that they are consecutive numbers? We could then simply pass the number of observations as `num_segments`. There might be a small performance penalty, but we could do so in a way that users may override it. 

Other ideas welcome!

src/_gettsim/ids.py
```diff
@@ -22,27 +20,20 @@ def ehe_id(
...
-    result: list[int] = []
-
-    for index, current_p_id in enumerate(map(int, p_id)):
-        current_p_id_ehepartner = int(familie__p_id_ehepartner[index])
-
-        if current_p_id_ehepartner >= 0 and current_p_id_ehepartner in p_id_to_ehe_id:
-            result.append(p_id_to_ehe_id[current_p_id_ehepartner])
-            continue
+    n = np.max(p_id) + 1
+    familie__p_id_ehepartner = np.where(
```
hmgaudecker, line 24:
```suggestion
    p_id_ehepartner_or_own_p_id = np.where(
```
This is what happens, right? Also, please don't re-use function input names for other things, makes it hard to read (for me).

src/_gettsim/ids.py
```diff
@@ -55,103 +46,53 @@ def fg_id(  # noqa: PLR0912
...
+    children = np.isin(p_id, familie__p_id_elternteil_1) + np.isin(
+        p_id, familie__p_id_elternteil_2
+    )
+    fg_id = np.where(
+        arbeitslosengeld_2__p_id_einstandspartner < 0,
+        p_id + p_id * n,
+        np.maximum(p_id, arbeitslosengeld_2__p_id_einstandspartner)
+        + np.minimum(p_id, arbeitslosengeld_2__p_id_einstandspartner) * n,
+    )
+    fg_id = np.where(
```
hmgaudecker, line 70:
```suggestion
    # TODO (@MImmesberger): Check correct conditions for grown up children
    # https://github.com/iza-institute-of-labor-economics/gettsim/pull/509
    # TODO(@MImmesberger): Remove hard-coded number
    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/668
    fg_id = np.where(
```
Still relevant!

tests/ttsim/mettsim/group_by_ids.py
```diff
@@ -10,23 +10,11 @@ def sp_id(
     """
     Compute the spouse (sp) group ID for each person.
     """
-    p_id_to_sp_id: dict[int, int] = {}
-    next_sp_id: int = 0
-    result: list[int] = []
+    n = 1000000
```
hmgaudecker, line 13: Why do we need this line? Would be good to showcase a way without any such hard-coded numbers in case people want to get inspiration from here. Same solution as in GETTSIM, I guess.

src/_gettsim_tests/test_groupings.py
```diff
@@ -30,27 +29,3 @@ def test_groupings(test: PolicyTest):
         flat_result.values(), flat_expected_output_tree.values()
     ):
         assert_array_almost_equal(result, expected, decimal=2)
-
-
-def test_fail_to_compute_sn_id_if_married_but_gemeinsam_veranlagt_differs():
```
hmgaudecker, line 35: So this should be added to the list of input data checks, right? @MImmesberger, is it already there? *(sorry, too time-pressed that I cannot check any more myself)*

tests/ttsim/mettsim/group_by_ids.py
```diff
@@ -40,53 +28,35 @@ def fam_id(
...
-
-            if child_age < 25 and len(child_p_id_children) == 0:
-                p_id_to_fam_id[current_p_id_child] = next_fam_id
-
-        next_fam_id += 1
-
-    # Compute result vector
-    result = [p_id_to_fam_id[current_p_id] for current_p_id in map(int, p_id)]
-    return np.array(result)
+    n = 1000000
```
hmgaudecker, line 31: Same.

hmgaudecker (commented):
Very nice! Just a few small suggestions.

src/_gettsim_tests/test_groupings.py
MImmesberger: Yes it's already in the meta issue

mj023:
Thanks for the clarification, I understand the problem now. 
With number of observations you mean the number of unique p_ids or number of fg_ids? Number of p_ids might work, if the num_segments argument is allowed to be bigger than the actual number of segments in the data, I will try that out first. Otherwise we might be able to avoid num_segments altogether by writing our own segment_max, this would take a lot more time though.

hmgaudecker:
The thing is timing. The best value would be the maximum of the group ids. But we only know that once we have computed them. 

To avoid that, I'd use the length of the data, which is an upper bound when redefining the contents as suggested above.

mj023:
Your idea is probably the best then, I did not realize that segment_sum was implemented so that it returns an array the size of the highest segment_id. I will change all ID's to be consecutive numbers starting at 0. Using the length of the inputs as num_segments will be fine, I don't think using a little bit too much memory there will be a problem.

hmgaudecker:
> I did not realize that segment_sum was implemented so that it returns an array the size of the highest segment_id. 

Me neither! Did you, @timmens ?

> I will change all ID's to be consecutive numbers starting at 0. Using the length of the inputs as num_segments will be fine, I don't think using a little bit too much memory there will be a problem.

Excellent!

timmens:
[This](https://github.com/iza-institute-of-labor-economics/gettsim/pull/879#issuecomment-2840899378) is my last update.

hmgaudecker:
> [This](https://github.com/iza-institute-of-labor-economics/gettsim/pull/879#issuecomment-2840899378) is my last update.

Sure, this was just meant FYI!

mj023:
I think the first solution is good. Might make it easier for people to write their own ID functions that depend on other ID's, it's much easier if you know they are consecutively numbered starting at 0. It's also probably just better if num_segments does not get too big.

I tried removing the precomputed ID's from the tests, but for some reason I still get many KeyErrors, but strangely only in GETTSIM not METTSIM.

mj023:
In this specific case you might be able to convert the dictionary to an array, use a base year e.g. 1900 and then use `geburtsjahr` like an index by subtracting the base year. Otherwise you probably have to loop through all dictionary entries, or if you can somehow get a sorted array you can use jax.numpy.searchsorted(). Quite annoying, that there are no lookup tables in JAX.

Incidentally a case like this might also be the reason why all the tests are failing. I now updated `execute_test` to actually jit when Jax is installed. The missing num_segments was only because jit was always set to false.

src/_gettsim/ids.py
```diff
@@ -55,103 +46,53 @@ def fg_id(  # noqa: PLR0912
...
-            continue
-
-        p_id_to_fg_id[current_p_id] = next_fg_id
-
-        current_hh_id = int(hh_id[index])
-        current_p_id_einstandspartner = int(
-            arbeitslosengeld_2__p_id_einstandspartner[index]
+    n = np.max(p_id) + 1
+
+    familie__p_id_elternteil_1_loc = familie__p_id_elternteil_1
```
hmgaudecker, line 51:
```suggestion
    p_id_elternteil_1_loc = familie__p_id_elternteil_1
```
No need to carry around the namespace for local variables. 

- A couple of comments in the body of this function would help future readability
- The conditions in `np.where` for the second and third statements assigning to `fg_id` seem complicated enough to factor out into a private function?

hmgaudecker:
Looks great and I think it can be merged once the comments in the code are addressed. We'll probably want to postpone this mechanism here:

> > I think there are two ways out:
> > 
> >     * creating internal `[group]_id` columns, which basically call `reorder_ids` on the existing ones and sticking with one global `num_segments=len(p_id)`
>
> I think the first solution is good. Might make it easier for people to write their own ID functions that depend on other ID's, it's much easier if you know they are consecutively numbered starting at 0. It's also probably just better if num_segments does not get too big.

until we are in the process of redesigning the interface (early June).

However, please revert the changes to the test cases that changed id's to 0-based consecutive numbers, so that we don't forget. Please also open an issue outlining the solution strategy.

src/_gettsim/ids.py
mj023: I added the new function and also added some comments to some of the other functions.

mj023:
I think I addressed all your comments and I also reverted the test cases, that test the ID's. Was this correct, or was I supposed to revert the reordering and the changes to the test execution too?

hmgaudecker:
Perfect! 

We just want to support these cases in Jax, so we should keep them around. Test execution is perfect as is.

PR #908, opened 2025-05-13
Move GETTSIM params files, use new machinery

hmgaudecker:
Another to-do from #897. 

- Move the parameters from `src/_gettsim/parameters` into the correct location of the tree.
- Break up if necessary so that each parameter ends up at the correct location
- Use new machinery. When touching the functions, remove the parts of the docstrings referencing other functions.
- Once done, remove the old machinery.

- [x] abgelt_st (this PR)
- [x] arbeitsl_geld (#912)
- [x] arbeitsl_geld_2 (#912)
- [x] arbeitslosenversicherung (#914)
- [x] eink_st (this PR)
- [x] eink_st_abzuege (this PR)
- [x] elterngeld (#933)
- [x] erwerbsm_rente (#914)
- [x] erziehungsgeld (#933)
- [x] geringfügige_einkommen (#914)
- [x] ges_krankenv (#914)
- [x] ges_pflegev (#914)
- [x] ges_rente (#914)
- [x] ges_rentenv (#914)
- [x] grunds_im_alter (#936)
- [x] kindergeld (this PR)
- [x] kinderzuschl (#927)
- [x] lohnst (#931)
- [x] soli_st (this PR)
- [x] unterhalt (#936)
- [x] unterhaltsvors (#936)
- [x] wohngeld (#932)

Also:
- ~Move parameter classes to `param_types.py`~ *Edit HMG: I don't think we want to do that. Would just bring back the whole namespace issue. As long as we don't have to import to other modules, we are fine.*


hmgaudecker:
@MImmesberger, I got started, a bit of initial hiccups, but it makes the code much, much nicer! 

Now we are getting rid of those special parsing functions, we will need the evaluation date in there. I'd propose adding `evaluationsjahr` (and potentially day, month) to the global params namespace. See `src/_gettsim/einkommensteuer/abzüge/vorsorgeaufwendungen.py:rate_abzugsfähige_altersvorsorgeaufwendungen` for how that may look like in action.

We could even use a `datetime.date`, but then we would need to make sure that people don't throw it at `policy_functions` if Jax is used.

What do you think?

*Aside re process: Once we are in parallel working mode, we should make little PRs against this one, reviewing was a mess in that last one.*

MImmesberger:
Yes that sounds good. I'm asking myself whether `evaluationsjahr` should be a regular input or passed as an argument of `compute_taxes_and_transfers` directly. From an interface perspective, I like the latter more, mostly because specifying `evaluationsjahr`, `evaluationsmonat` and `evaluationstag` independently as ints feels weird given our ruke on how to specify the date of the policy environment. 

Anyhow, we need month and day as well for #211.

hmgaudecker:
Yeah, let's maybe push the precise handling of dates off for a moment.

I'd think that the standard behaviour would be that the date passed to `create_policy_environment` would be used, but this could be overridden, of course. Precise way will depend a little on the new interface.

hmgaudecker:
@MImmesberger, asking for a review so you please have a check, not in order to merge yet. 

From now on, let us please make smallish PRs against this one when converting parameter files.

This said, I won't be able to do much the next couple of days due to the CRC conference!

MImmesberger:
Yes, I will try to go over the remaining ones as much as possible.

src/_gettsim/einkommensteuer/abzüge/alter.py
```diff
@@ -0,0 +1,91 @@
...
+    einkommensteuer__einkünfte__aus_nichtselbstständiger_arbeit__bruttolohn_y: float,
+    einkommensteuer__einkünfte__aus_kapitalvermögen__kapitalerträge_y: float,
+    einkommensteuer__einkünfte__aus_selbstständiger_arbeit__betrag_y: float,
+    einkommensteuer__einkünfte__aus_vermietung_und_verpachtung__betrag_y: float,
+    altersentlastungsbetrag_altersgrenze: int,
+    maximaler_altersentlastungsbetrag_gestaffelt: dict[int, float],
+    altersentlastungsquote_gestaffelt: dict[int, float],
+) -> float:
+    """Calculate tax deduction allowance for elderly since 2005."""
+    bins = sorted(maximaler_altersentlastungsbetrag_gestaffelt.keys())
```
MImmesberger, line 61: In principle we could apply `birth_year_based_phase_inout` here as well. Should we refactor?

hmgaudecker: Yeah, as long as we can merge tonight :smile: 

There will be a bunch of other things where we can apply the ConsecutiveInt types (Kindergeld in the old days, ..), just searching for `gestaffelt` turns up quite some stuff.

MImmesberger: I didn't know there was a hard deadline 😄 

But then let's merge, I'll write an issue. Maybe its best to tackle this when Max touches some of this anyhow.

MImmesberger (approved):

Issue #910, opened 2025-05-16
ENH: Use ALG 2 Kindersofortzuschlag for KiZ


MImmesberger:
#### Is your feature request related to a problem?

There's a period of 6 months where the Sofortzuschlag is relevant for KiZ explicitly. We have the same parameter in the ALG2 and the KiZ parameters, in the latter we just set it to 0 once its not relevant for KiZ anymore.

#### Describe the solution you would like to see

We should use the ALG2 parameter and handle this in `maximales_nettoeinkommen_m_bg` explicitly (split up into two and use `start_date`/`end_date`).

MImmesberger:
Closed via #912

Issue #782, opened 2024-07-23
ENH: Regional differences in KdU parameters


MImmesberger:
#### Is your feature request related to a problem?

This is just for future reference because this came up a couple of times (and recently @MartinWiegandecon asked me about it): Currently we use placeholder values for the maximum rent per m2 and maximum apartment size in m2. In fact, these parameters differ across municipalities (maximum rent per m2) and across Bundesländer (apartment size).

#### Describe the solution you would like to see

Choose (empirical) averages for the parameters.

#### Supporting variation across municipalities

@MartinWiegandecon sent me [this](https://harald-thome.de/informationen/bundesweite-dienstanweisungen-kdu.html) overview for 2024. However, finding out which parameters were used by the municipalities in the past is not trivial (if possible at all).

MImmesberger:
Also, Heizkosten might differ regionally. We haven't implemented any Angemessenheitsprüfung there.

PR #912, opened 2025-05-16
Move ALG 1 + 2 params to namespace

### LINKED ISSUE #826: ENH: Use ALG2 Erwachsenensatz target for KiZ
The KiZ currently computes the Regelsatz for adults independently. We can remove much of the code by using the ALG2 targets.

#### Note

We should check whether the KiZ law just refers to the ALG2 rules or codifies its own calculation rules for the Regelsatz (which happen to be the same as in ALG2). In case of the latter, the code should stay.

MImmesberger:

Moves params of ALG1 and ALG2 to their namespace.

Closes #826, #910


MImmesberger:
Its hard to test whether all changes are correct given the `assert "deviation_from" not in current_spec` check that fails for almost all tests, so we might discover some bugs later when all specs are migrated.

inputs.yaml
hmgaudecker, line 1: What is this file?

src/_gettsim/arbeitslosengeld_2/params.yaml
hmgaudecker, line 1: I tried to avoid the generic `params.yaml` files and tried to divide them up by name of the Python module where the contents are used, particularly for large sets of parameters See `einkommensteuer` / `abzüge`, for example. For smaller parameter sets, a file with the same name as the directory may do, too. 

I think that is nicer, but it is open for debate, of course. In any case, we should strive to be consistent here.

hmgaudecker (commented):
Thanks, just had a brief look with a couple of very high-level comments!

inputs.yaml
MImmesberger: Oops this shouldn't be under version control 😅

hmgaudecker:
> Its hard to test whether all changes are correct given the `assert "deviation_from" not in current_spec` check that fails for almost all tests, so we might discover some bugs later when all specs are migrated.

Sorry, my bad of not writing a proper migration guide. 

This should not happen once you have migrated a yaml-file to the new structure. Then also remove it from the list in `config.py` and the file from the directory.

In almost all cases, just replace `deviation_from: previous` by `updates_previous: true`, but you did that already.

Of course, you picked the single case where we are using a different value for `updates_previous` (see line 746 of `policy_environment.py`)... That will need a special conversion function, which takes the two parameters (`parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg` and `raw_parameter_anrechnungsfreies_einkommen_mit_kindern_in_bg`) as inputs and gives the combination back. Apologies for not mentioning that specifically or suggesting to start with anything but Arbeitslosengeld 2...

LMK if other issues arise!

MImmesberger:
Thanks, I'll look into that.

And I think my issue was that there is a mistake in the soli params on the base branch of this PR. Didn't realise you already touched that. Will be an easy fix probably, will do that in a few minutes.

hmgaudecker:
Ah, I might have rushed that one. Sorry!!!

MImmesberger:
> Of course, you picked the single case where we are using a different value for updates_previous (see line 746 of policy_environment.py)... That will need a special conversion function, which takes the two parameters (parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg and raw_parameter_anrechnungsfreies_einkommen_mit_kindern_in_bg) as inputs and gives the combination back. Apologies for not mentioning that specifically or suggesting to start with anything but Arbeitslosengeld 2...

I took the easy route here and just used the raw version of `parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg`. You can take a look at commit [dac07ca](https://github.com/iza-institute-of-labor-economics/gettsim/pull/912/commits/dac07caeb75cd55a5b209f0149a0764bf329f647). LMK if you're fine with being a bit lazy here.

src/_gettsim/arbeitslosengeld_2/einkommen.py
```diff
@@ -226,28 +234,12 @@ def nettoquote(
...
-        See :func:`nettoquote`.
-    arbeitsl_geld_2_params
-        See params documentation :ref:`arbeitsl_geld_2_params <arbeitsl_geld_2_params>`.
-
-    Returns
-    -------
-
-    """
+    """Share of income which remains to the individual."""
     out = piecewise_polynomial(
```
hmgaudecker, line 240:
```suggestion
    return piecewise_polynomial(
```

Let's get rid of these when we touch them.

src/_gettsim/arbeitslosengeld_2/einkommen.py
```diff
@@ -226,28 +234,12 @@ def nettoquote(
...
         x=einkommensteuer__einkünfte__aus_nichtselbstständiger_arbeit__bruttolohn_m,
-        parameters=arbeitsl_geld_2_params[
-            "parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg"
-        ],
+        parameters=parameter_anrechnungsfreies_einkommen_ohne_kinder_in_bg,
         rates_multiplier=nettoquote,
     )
     return out
```
hmgaudecker, line 245:
```suggestion
```

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+    reference: Artikel 1 G. v. 20.07.2006 BGBl. I S. 1706.
+  2008-01-01:
+    1947: 33800
+    1948: 9750
+    1958: 9900
+    1964: 10050
+    note: Artikel 1 G. v. 20.04.2007 BGBl. I S. 554.
+  2023-01-01:
+    reference: G. v. 20.12.2022 BGBl. I S. 2328.
+    note: Bürgergeld-Gesetz
```
hmgaudecker, line 34:
```suggestion
    note: Abolished in the Bürgergeld-Gesetz
```

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+    1947: 520
+    1948: 200
+    reference: Artikel 1 G. v. 30.12.2002 BGBl. I S. 4619.
+  2006-08-01:
+    1947: 520
+    1948: 150
+    reference: Artikel 1 G. v. 20.07.2006 BGBl. I S. 1706.
+  2023-01-01:
+    reference: G. v. 20.12.2022 BGBl. I S. 2328.
+    note: Bürgergeld-Gesetz
```
hmgaudecker, line 55:
```suggestion
    note: Abolished in the Bürgergeld-Gesetz
```

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+schonvermögen_bürgergeld:
+  name:
+    de: Schonvermögen beim Bürgergeld
+    en: Protected wealth for the citizen's income
+  description:
+    de: >-
+      Seit der Einführung des Bürgergelds zum 01.01.2023 gilt ein Vermögensfreibetrag
```
hmgaudecker, line 62: Not necessary to fix now, but in the future we should strive to write these texts in a way that they are future-proof...

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+      person.
+  unit: Euros
+  reference_period: null
+  type: dict
+  2023-01-01:
+    während_karenzzeit: 40000
+    normaler_satz: 15000
+    reference: G. v. 20.12.2022 BGBl. I S. 2328.
+    note: Bürgergeld-Gesetz
+vermögensfreibetrag_austattung:
```
hmgaudecker, line 80: Move up to the group before `schonvermögen_bürgergeld`.

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+  unit: Euros
+  reference_period: null
+  type: scalar
+  2005-01-01:
+    value: 750
+    reference: Artikel 1 G. v. 24.12.2003 BGBl. I S. 2954.
+  2023-01-01:
+    reference: G. v. 20.12.2022 BGBl. I S. 2328.
+    note: Ceased to exist with Bürgergeld-Gesetz
+vermögensgrundfreibetrag_je_kind:
```
hmgaudecker, line 96: Move up to the group before `schonvermögen_bürgergeld`.

src/_gettsim/arbeitslosengeld_2/freibeträge_vermögen.py
hmgaudecker, line 1: Please use the occasion to set the start date for all functions in the file. Maybe double-check the rest in the directory, too.

src/_gettsim/arbeitslosengeld_2/freibeträge_vermögen.py
```diff
@@ -109,33 +57,17 @@ def maximaler_grundfreibetrag_vermögen(
 
 @policy_function(start_date="2023-01-01")
 def freibetrag_vermögen_in_karenzzeit_bg(
```
hmgaudecker, line 59: I am not fan of calling the per-person parameter `schonvermögen` and the aggregated one `freibetrag_vermögen`, unless this is what happens in the law. 

Also, can we rename to `vermögensfreibetrag` everywhere or is a different term used in SGB II ?

src/_gettsim/arbeitslosengeld_2/freibeträge.yaml
```diff
@@ -0,0 +1,262 @@
...
+    1948: 200
+    reference: Artikel 1 G. v. 30.12.2002 BGBl. I S. 4619.
+  2006-08-01:
+    1947: 520
+    1948: 150
+    reference: Artikel 1 G. v. 20.07.2006 BGBl. I S. 1706.
+  2023-01-01:
+    reference: G. v. 20.12.2022 BGBl. I S. 2328.
+    note: Bürgergeld-Gesetz
+schonvermögen_bürgergeld:
```
hmgaudecker, line 56:
```suggestion
schonvermögen_bürgergeld:
```
Only realised ex post, see some comments below, but `bürgergeld` really should not appear in this name. Maybe `vermögensfreibeträge_nach_karenzzeit`?

src/_gettsim/arbeitslosengeld_2/kosten_der_unterkunft.yaml
```diff
@@ -0,0 +1,68 @@
+---
+mietobergrenze_pro_qm:
```
hmgaudecker, line 2: This parameter is pretty crazy... For sure let it start in 2005 only, but I am honestly not sure what to do with it. For sure it does not make sense to have a flat version for the entire country which does not change over the years. Could you set up an issue if there isn't one already, please?

src/_gettsim/arbeitslosengeld_2/kosten_der_unterkunft.yaml
```diff
@@ -0,0 +1,68 @@
...
+    de: >-
+      Eine Mietwohnung darf für einen Single 45 Quadratmeter (+15 für jede weitere
+      Person) groß sein.
+    en: >-
+      A rental apartment may be 45 square meters for a single person (+15 for each
+      additional person).
+  unit: Square Meters
+  reference_period: null
+  type: dict
+  1984-01-01:
```
hmgaudecker, line 36:
```suggestion
  2005-01-01:
```

src/_gettsim/arbeitslosengeld_2/kosten_der_unterkunft.yaml
```diff
@@ -0,0 +1,68 @@
...
+    single: 45
+    je_weitere_person: 15
+parameter_berechtigte_wohnfläche_eigentum:
+  name:
+    de: Berechtigte Eigentumsfläche für ALG2-Empfänger*innen
+    en: Property space eligible for ALG2-recipients
+  description:
+    de: >-
+      Ab 2023 gilt diese Obergrenze nur für Eigenheime. Eigentumswohnungen dürfen nur
```
hmgaudecker, line 45: Please add an issue that we need to add an input variable starting 2023...

src/_gettsim/arbeitslosengeld_2/regelbedarf.py
```diff
@@ -1,6 +1,171 @@
...
+    regelsatz: float
+    min_alter: int
+    max_alter: int
+
+
+@dataclass(frozen=True)
+class RegelsatzNachRegelbedarfsstufen:
+    """Regelsatz as a fraction of the Basissatz."""
+
+    regelbedarfsstufe_1: RegelbedarfsstufeErwachsener
```
hmgaudecker, line 99:
```suggestion
    rbs_1: RegelbedarfsstufeErwachsener
```
Should be fine here, after all we are only working around not being able to do `1: ...` directly.

src/_gettsim/arbeitslosengeld_2/regelbedarf.py
```diff
@@ -96,47 +225,29 @@ def mehrbedarf_alleinerziehend_m(
...
-    anteile = arbeitsl_geld_2_params["regelsatz_anteilsbasiert"][
-        "anteil_vom_basissatz_für_kinder"
-    ]
+    """Basic monthly subsistence / SGB II needs of children until 2010."""
+    basissatz = regelsatz_anteilsbasiert.basissatz
 
     if (
-        alter >= anteile["zwischen_14_und_24_jahre"]["min_alter"]
-        and alter <= anteile["zwischen_14_und_24_jahre"]["max_alter"]
+        alter >= regelsatz_anteilsbasiert.kind.zwischen_14_und_24_jahre.min_alter
```
hmgaudecker, line 234: Is there a specific term used in the law? It really is ugly to have the age appearing twice here.

src/_gettsim/arbeitslosengeld_2/regelbedarf.py
```diff
@@ -96,47 +225,29 @@ def mehrbedarf_alleinerziehend_m(
 def kindersatz_m_bis_2010(
```
hmgaudecker, line 225:
```suggestion
def kindersatz_m_anteilsbasiert(
```
Reason is that the `ab_2011` does not make sense when describing a 2022 change. Please change those two accordingly, adding `_nach_regelbedarfsstufen`

src/_gettsim/sozialversicherung/arbeitslosen/arbeitslosengeld.py
```diff
@@ -1,22 +1,30 @@
...
 
 
-@policy_function(vectorization_strategy="loop")
+@policy_function(end_date="1998-07-31", leaf_name="betrag_m")
+def betrag_m_not_implemented() -> float:
+    """Calculate individual unemployment benefit."""
+    raise NotImplementedError("Not implemented yet.")
+
+
+@policy_function(vectorization_strategy="loop", start_date="1998-08-01")
```
hmgaudecker, line 15:
```suggestion
@policy_function(start_date="1998-08-01", vectorization_strategy="loop")
```
important things first :wink:. Though I don't even see why this cannot be vectorized, can probably be deleted with the change suggested below.

src/_gettsim/sozialversicherung/arbeitslosen/arbeitslosengeld.py
```diff
@@ -1,22 +1,30 @@
...
     einkommen_vorjahr_proxy_m: float,
-    arbeitsl_geld_params: dict,
+    satz: dict[str, float],
 ) -> float:
     """Calculate individual unemployment benefit."""
 
     if einkommensteuer__anzahl_kinderfreibeträge == 0:
-        arbeitsl_geld_satz = arbeitsl_geld_params["satz"]["allgemein"]
+        arbeitsl_geld_satz = satz["allgemein"]
     elif einkommensteuer__anzahl_kinderfreibeträge > 0:
```
hmgaudecker, line 26:
```suggestion
    else:
```

src/_gettsim/arbeitslosengeld_2/kosten_der_unterkunft.yaml
MImmesberger: There's issue #782. The parameters should be on the municipal level but it seems like
they are impossible to get from public sources / internet, especially historical parameters. Maybe we can approach the
Statistische Landesämter at some point (or you know someone who might have the data)?

hmgaudecker: Great. And no, I do not have great ideas right now, either...

Issue #907, opened 2025-05-13
ENH: Some more uses of piecewise polynomial for converting parameters


hmgaudecker:
After #897, it will be much easier to convert parameters. This means we can use piecewise_polynomial in more cases instead of specifying lots of numbers (often months expressed as fractions of years...) in the yaml files *(and yes, I know that means reverting to the initial implementation in many cases, which I had asked to be changed... 😇  But we did not have the required machinery in place back then)*.

- [ ] Erwerbsminderungsrente:
  - `zurechnungszeitgrenze`
  - `altersgrenze_abschlagsfrei`
- [ ] Altersrente
  - `regelaltersgrenze`
  - `altersgrenze_für_frauen_abschlagsfrei`
  - `altersgrenze_für_frauen_vorzeitig`
  - `altersgrenze_rente_wegen_arbeitslosigkeit_abschlagsfrei`
  - `altersgrenze_rente_wegen_arbeitslosigkeit_vorzeitig`
  - `altersgrenze_langjährig_versicherte_abschlagsfrei`
  - `altersgrenze_besonders_langjährig_versicherte`

Possibly more.

MImmesberger:
I'm not sure that the infrastructure is quite there yet to transform the Altersgrenzen. They depend on birth year and month.

`piecewise_polynomial` takes floats as arguments. We either have two ints or `datetime.date` (taking only birth year and month as input -- else we'd have to define a piecewise constant with many many constant parts as the day of birth does not matter).

hmgaudecker:
> Edit: Ofc assuming that we don't want something like 1953 + 1/12 (=February 1953) as thresholds. But maybe the easiest solution is the smartest...

This is it, and it is actually how the original implementation had sometimes looked like *(I pushed against it back in the day because it would have blown up `compute_taxes_and_transfers` in the days prior to active periods & `params_function`s)*. The thing is we will get rid of many of these fractions--which are currently plentiful in the yaml files--because the thresholds are natural numbers and fractions are only implied. Maybe this won't work in all cases, though.

In any case, let's please not take the above list literally. We should try to stick close to the law. 
- If the law specifies a "formula", it will usually be best to use `piecewise_polynomial`.
- If the law specifies a mapping directly, let's copy that.

MImmesberger:
Superseded by #938.

PR #914, opened 2025-05-16
Move Sozialversicherung params to namespace

MImmesberger:

Move Sozialversicherung params to their namespace


MImmesberger:
Have we settled on how to deal with parameters that change their type (e.g. from float to dict). In principle, we can deal with that easily via our decorator machinery. However, the `type` in the `.yaml` file will be wrong. So we always need distinct entries in the yaml file for them? @hmgaudecker

hmgaudecker:
> Have we settled on how to deal with parameters that change their type (e.g. from float to dict). In principle, we can deal with that easily via our decorator machinery. However, the `type` in the `.yaml` file will be wrong. So we always need distinct entries in the yaml file for them? @hmgaudecker 

Yes, I think different parameters are the clearest. The corresponding function should change, too. We often supported that in the old days by setting a component of a parameter to zero or by having `dict.get()` in those functions. If you see one of those, please correct. 

I have been using the terminology of `_einheitlich` / `_gestaffelt` in many cases. *(especially the second one might not always make sense)*

MImmesberger:
Finished everything but the Altersgrenzen, see discussion in #907.

I planned to implement #907 here already because touching it twice is probably inefficient. But if we don't want to adjust `piecewise_polynomial` right now, I can also add an intermediate solution (or we keep the current infrastructure a little longer).

MImmesberger:
Also, just noticed that we can get rid of many intermediate `params_function`s that just duplicate later `policy_function`s (e.g. for einheitliche Beitragssätze). Will implement this later.

hmgaudecker:
> Finished everything but the Altersgrenzen, see discussion in #907.
> 
> I planned to implement #907 here already because touching it twice is probably inefficient. But if we don't want to adjust `piecewise_polynomial` right now, I can also add an intermediate solution (or we keep the current infrastructure a little longer).

Great, go ahead — no need to adjust p_p, I think?

src/_gettsim/sozialversicherung/rente/grundrente/grundrente.py
```diff
@@ -155,41 +112,23 @@ def basisbetrag_m(
...
+    35 years (420 months).
     """
 
     # Winsorize Bewertungszeiten and Zugangsfaktor at maximum values
     bewertungszeiten_monate_wins = min(
         bewertungszeiten_monate,
-        ges_rente_params["grundrente_berücksichtigte_wartezeit"]["max"],
+        berücksichtigte_wartezeit["max"],
     )
     ges_rente_zugangsfaktor_wins = min(
```
hmgaudecker, line 129:
```suggestion
    zugangsfaktor = min(
```

src/_gettsim/sozialversicherung/rente/grundrente/grundrente.py
```diff
@@ -204,22 +143,7 @@ def basisbetrag_m(
 def durchschnittliche_entgeltpunkte(
```
hmgaudecker, line 143: I have to stop now, but we really should not have one function (input) `mean_entgeltpunkte` and another one `durchschnittliche_entgeltpunkte`. I believe we settled on the `mean_` convention.

src/_gettsim/sozialversicherung/rente/grundrente/wartezeit.yaml
```diff
@@ -0,0 +1,21 @@
+---
+berücksichtigte_wartezeit:
```
hmgaudecker, line 2:
```suggestion
berücksichtigte_wartezeit_monate:
```
We are using that for the column names these are compared to. Also,  `wartezeit` is measured in years in the more common contexts, so I think it is helpful to be explicit here.

hmgaudecker (commented):
Great work!

I looked mostly at `rente` so far, mostly high-level comments, as usual some unrelated to the PR -- hopefully still useful!

MImmesberger:
> I had not seen https://github.com/iza-institute-of-labor-economics/gettsim/issues/907#issuecomment-2888464342 on the phone... But should be fine to go ahead with the simple solution!

Your last comment implies that we don't want to use piecewise linear here. The reason is that the statutory retirement ages are specified in large tables exactly like we do it here in GETTSIM.

hmgaudecker:
Great, maybe then the only thing would be to switch to dicts with year, month entries instead of fractions?

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
```diff
@@ -0,0 +1,395 @@
...
+      §77 Abs. 2 Nr. 2 SGB VI
+      Factor with which the Zugangsfaktor / "access factor" for receiving the statutory
+      pension is increased/decreased for each month of less/more work
+  unit: null
+  reference_period: Year
+  type: dict
+  2001-01-01:
+    vorzeitiger_renteneintritt: 0.036
+    späterer_renteneintritt: 0.06
+parameter_beitragspflichtiges_durchschnittsentgelt:
```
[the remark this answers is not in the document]
MImmesberger: Would it be difficult to allow for automatic time conversions of parameters? Shouldn't
be, right? Then we could call this `beitragspflichtiges_durchschnittsentgelt_y`
directly.

src/_gettsim/sozialversicherung/rente/beitrag/beitragsbemessungsgrenze.yaml
```diff
@@ -1,82 +1,47 @@
 ---
-parameter_beitragssatz:
+parameter_beitragsbemessungsgrenze_einheitlich:
```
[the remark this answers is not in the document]
MImmesberger: No because this produces a `ConflictingNamesError`. I'm realizing right now that I don't
get why this is even there. We shouldn't mind names that are used as leaf names and as parameter names as
long as they're not active at the same time, right?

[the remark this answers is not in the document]
hmgaudecker: Could you elaborate? I don't quite get why this is a problem (or not anymore). Or just implement and let me check which problem(s) arise(s).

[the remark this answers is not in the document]
hmgaudecker: Okay, so the problem is that in `fail_because_of_clashes`, we (manually) apply the active periods, but do not check whether elements of `orig_params_tree` are empty. This check should be done; the use case here shows that this is an important thing to have.

Let's think through how to best do that tomorrow.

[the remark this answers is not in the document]
MImmesberger: Basically its test case 1 and 2 in `test_fail_because_of_conflicting_names`.

What we currently have: `fail_because_of_clashes` fails if something defined in a `.yaml` file has the same path + leaf name as a policy function.

What we want: i) Fail if the path + name of a param in a `.yaml` file duplicates the path + original function name of some some policy function. ii) Fail if path + name of a param in a `.yaml` file duplicates the path + leaf name of some policy function if they are both active at the same time (i.e. param has more entries than `note` and `reference` in the relevant period).

src/_gettsim/sozialversicherung/geringfügige_einkommen.yaml
```diff
@@ -0,0 +1,146 @@
...
+    de: §8 (1a) SGB IV
+    en: §8 (1a) SGB IV
+  unit: null
+  reference_period: null
+  type: dict
+  2022-10-01:
+    zähler: 130
+    nenner: 3
+    reference: Art. 7 G. v. 28.06.2022 BGBl. I S. 969
+arbeitgeberpauschale_lohnsteuer:
```
hmgaudecker, line 124:
```suggestion
minijob_arbeitgeberpauschale_lohnsteuer:
```
The names of the yaml file does not make it into the params tree. Better be explicit.

src/_gettsim/sozialversicherung/kranken/beitrag/arbeitgeberpauschale_geringfügig.yaml
```diff
@@ -0,0 +1,24 @@
+---
+arbeitgeberpauschale_bei_geringfügiger_beschäftigung:
```
hmgaudecker, line 2:
```suggestion
minijob_arbeitgeberpauschale:
```
symmetric to previous comment, just different namespace.

src/_gettsim/sozialversicherung/rente/beitrag/beitragsbemessungsgrenze.yaml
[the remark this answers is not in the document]
hmgaudecker: Why would we care about the original function name? Not sure right now.

But good hint at that test, could you maybe add a few parameterisations that would help us do that?

Re implementation, I am mostly thinking about the extent to which it'll be possible to avoid code duplication between `fail_because_of_clashes` and the actual processing.

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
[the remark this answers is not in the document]
hmgaudecker: Not really, I think. Basically a matter of extending the machinery we have to function arguments, too?

src/_gettsim/sozialversicherung/rente/beitrag/beitragsbemessungsgrenze.yaml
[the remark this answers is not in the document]
hmgaudecker: > > Why would we care about the original function name? Not sure right now.
> 
> That's optional because we never merge the trees (why not?)

`compute_taxes_and_transfers._process_params_tree` is a bit simpler this way, but we could certainly do the filtering in another way. However, to the extent that we do not want to / cannot copy the entire `leaf_name` machinery to params, we probably want to keep it as is. The use case we are debating here kind of shows it.

> > Re implementation, I am mostly thinking about the extent to which it'll be possible to avoid code duplication between fail_because_of_clashes and the actual processing.
> 
> When it comes to checking params against functions, we could just do it after filtering for the active elements of each tree. But I'd like it more to have a general approach that tests everything against everything.

I think the difference is that `fail_because_of_clashes` checks for all periods, not only the currently active one. In a future interface, it will be optional to call it. Shouldn't be too hard to sort that out, but I won't be involved today! :shrug:

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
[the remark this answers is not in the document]
hmgaudecker: Update: I spent half an hour on this, but I am pretty sure it is not worth it. Makes sense for scalar parameters, but for anything else we may do **very** unexpected things unless we are extremely careful.

Since there are only very few of these guys (we might be inconsistent in naming, but still):

image

I don't think working through all corner cases will be worth it. 

If you have a different opinion, @MImmesberger, LMK!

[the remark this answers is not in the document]
MImmesberger: Yes, let's use conversions via `params_function` (or explicit conversions)  for now.

src/_gettsim/sozialversicherung/rente/grundrente/persönliche_entgeltpunkte.yaml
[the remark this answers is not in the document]
MImmesberger: We could do that, but persönliche Entgeltpunkte is the official term (persEP = EP x
Zugangsfaktor x Rentenwert).

src/_gettsim/sozialversicherung/rente/grundrente/grundrente.py
```diff
@@ -94,55 +70,36 @@ def anzurechnendes_einkommen_m(
...
 
     # Calculate relevant income following the crediting rules using the values for
     # singles and those for married subjects
     # Note: Thresholds are defined relativ to rentenwert which is implemented by
     # dividing the income by rentenwert and multiply rentenwert to the result.
     if familie__anzahl_personen_ehe == 2:
-        einkommensanr_params = ges_rente_params[
-            "grundrente_anzurechnendes_einkommen_mit_partner"
-        ]
+        params = anzurechnendes_einkommen_mit_partner
```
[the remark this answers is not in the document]
MImmesberger: Added a regression test.

src/_gettsim/sozialversicherung/rente/grundrente/persönliche_entgeltpunkte.yaml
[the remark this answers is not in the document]
hmgaudecker: Yes, I could imagine that as a function name, but not really as the file name?

MImmesberger:
Another thing came up: We can't require pure parameters as targets, they need to be wrapped in at least a `params_function`.

Before, we did things like this:

```python
@policy_function(
    end_date="1989-12-17",
    leaf_name="altersgrenze",
    vectorization_strategy="not_required",
)
def altersgrenze_ohne_staffelung(ges_rente_params: dict) -> float:
    return ges_rente_params["altersgrenze_rente_wegen_arbeitslosigkeit_abschlagsfrei"]
```

Now, I'd like to get rid of the entire function and just call the parameter `altersgrenze`.

I can imagine that this is not too hard to do. Make the `_fail_if_root_nodes_are_missing` a bit more lenient by looking in the params tree as well, filter `flat_targets` before throwing it into  `dags` for regular functions, look for remaining targets in the params file. Do you agree?

Ofc we could take a preliminary solution for now (i.e. the thing above) and do it later properly.

hmgaudecker:
> filter `flat_targets` before throwing it into `dags` for regular functions, look for remaining targets in the params file

Gave this a shot in #922. 

Note: Just making `_fail_if_root_nodes_are_missing` more lenient is not enough because dags will fail down the road.

src/_gettsim/sozialversicherung/rente/altersrente/für_frauen/altersgrenze.yaml
```diff
@@ -0,0 +1,344 @@
...
+    value: 60.0
+  1989-12-18:
+    reference: Rentenreformgesetz 1992. BGBl. I S. 2261 1989 § 41
+    note: Increase of early retirement age from 60 to 62 for birth cohort 1949-1952.
+    entry_age_old_regime: 60.0
+    entry_age_new_regime: 62.0
+    max_birthyear_old_regime: 1948
+    min_birthyear_new_regime: 1953
+    1949:
+      1: 60.083333
```
hmgaudecker, line 292: Unless we change this to piecewise_linear, I'd suggest:
```suggestion
      1: 
        years: 60
        months: 1
```

More verbose, but clearer it seems?

src/_gettsim/sozialversicherung/rente/altersrente/regelaltersrente/altersgrenze.yaml
```diff
@@ -0,0 +1,106 @@
...
+      Increase of the early retirement age from 65 to 67 for birth cohort 1947-1964.
+      Vertrauensschutz (Art. 56) applies for birth cohorts before 1955 who were in
+      Altersteilzeit before January 1st, 2007 or received "Anpassungsgeld für
+      entlassense Arbeitnehmer des Bergbaus".
+    entry_age_old_regime: 65.0
+    entry_age_new_regime: 67.0
+    max_birthyear_old_regime: 1946
+    min_birthyear_new_regime: 1964
+    1947:
+      base_age: 65
```
hmgaudecker, line 57:
```suggestion
      years: 65
```

src/_gettsim/sozialversicherung/rente/altersrente/regelaltersrente/altersgrenze.yaml
```diff
@@ -0,0 +1,106 @@
...
+      months_to_add: 1
```
hmgaudecker, line 58:
```suggestion
      months: 1
```

Or any reason to add the extra words? But see general comment below.

src/_gettsim/sozialversicherung/rente/altersrente/regelaltersrente/regelaltersrente.py
```diff
@@ -50,47 +46,79 @@ def altersgrenze_mit_staffelung(geburtsjahr: int, ges_rente_params: dict) -> flo
...
-    Returns
-    -------
-    Eligibility as bool.
+    return sozialversicherung__rente__mindestwartezeit_erfüllt
 
-    """
 
-    return sozialversicherung__rente__mindestwartezeit_erfüllt
+@params_function(start_date="2007-04-20")
+def piecewise_params_regelaltersgrenze(
```
hmgaudecker, line 66: As is, I don't see a benefit of specifying this as piecewise linear. There is no linear function at all, right? Specified as piecewise constant.

I thought the specification would be:

- $(-\inf, 1946]$: $65$
- $[1947, 1957]$: $65 + (x-1946) / 12$
- $[1958, 1963]$: $66 + (x-1958) / 6$
- $[1964, \inf]$: $67$

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
```diff
@@ -0,0 +1,395 @@
+---
+zugangsfaktor_veränderung_pro_jahr:
+  name:
+    de: Veränderung des Zugangsfaktors pro Jahr
```
hmgaudecker, line 4:
```suggestion
    de: Veränderung des Zugangsfaktors pro Jahr relativ zu ... 
```
Not sure of the precise term to put there :innocent:

src/_gettsim/sozialversicherung/rente/beitrag/beitragsbemessungsgrenze.yaml
```diff
@@ -179,7 +144,7 @@ parameter_beitragsbemessungsgrenze:
     ost: 6700
     reference: §3, §4 V. v. 30.11.2020, BGBl. I S. 2612.
   2022-01-01:
-    deviation_from: previous
+    updates_previous: true
```
hmgaudecker, line 147:
```suggestion
```
Both values are changed, confusing.

hmgaudecker (commented):
Very nice! As usual, some unrelated comments, but most should be relevant right here, I think!

src/_gettsim/sozialversicherung/rente/altersrente/regelaltersrente/regelaltersrente.py
MImmesberger: Ah yes I didn't consider `piecewise_constant`. Makes total sense to use this instead.

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
MImmesberger: It's an absolute change, not a relative one. In the end it's just a factor without a scale, so I'm not sure what to put here.

src/_gettsim/sozialversicherung/rente/altersrente/regelaltersrente/regelaltersrente.py
hmgaudecker: As discussed, in case we stick to the current way of defining things for all cohorts in the yaml-file, I don't think we should use it at all. I would just have `require_converter` fill up a dict for all birth years from 1900 until today.

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
hmgaudecker: But it is relative to an age (NRA?). "Jahr" is too ambiguous, could be a calendar year, ...

src/_gettsim/sozialversicherung/rente/altersrente/altersrente.py
```diff
@@ -51,320 +37,96 @@ def betrag_m_mit_grundrente(
...
 
     if sozialversicherung__rente__bezieht_rente:
         out = (
-            sozialversicherung__rente__entgeltpunkte_west
-            * ges_rente_params["parameter_rentenwert"]["west"]
-            + sozialversicherung__rente__entgeltpunkte_ost
-            * ges_rente_params["parameter_rentenwert"]["ost"]
-        ) * zugangsfaktor
+            (
+                sozialversicherung__rente__entgeltpunkte_west
```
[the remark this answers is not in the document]
MImmesberger: No there's no difference after 2023-07

src/_gettsim/sozialversicherung/rente/altersrente/für_frauen/altersgrenze.yaml
MImmesberger: Yes, all Altersgrenzen should be like this eventually. I haven't touched Frauen, langjährig and besonders langjährig yet (just created the files)

src/_gettsim/sozialversicherung/rente/altersrente/rentenformel.yaml
MImmesberger: Ah yes

```suggestion
    de: Veränderung des Zugangsfaktors je Jahr vorzeitigem/verzögerten Renteneintritt relativ zur Regelaltersgrenze
```

We should add in the description that there are many exceptions to this rule.

hmgaudecker: Thanks! Preferred it a bit shorter, but should be fine now.

src/_gettsim/sozialversicherung/kranken/beitrag/einkommen.py
```diff
@@ -48,50 +39,28 @@ def einkommen_regulär_beschäftigt_m(
     )
 
 
-@policy_function()
+@policy_function(start_date="1990-01-01")
 def bemessungsgrundlage_selbstständig_m(
```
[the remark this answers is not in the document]
hmgaudecker: Not a great idea of mine because the `bezugsgröße` is a parameter applying to everyone, while this one is person-specific.

hmgaudecker (approved):
Excellent, thanks!

src/_gettsim/sozialversicherung/minijob.yaml
```diff
@@ -0,0 +1,123 @@
...
+    de: §8 (1a) SGB IV
+    en: §8 (1a) SGB IV
+  unit: null
+  reference_period: null
+  type: dict
+  2022-10-01:
+    zähler: 130
+    nenner: 3
+    reference: Art. 7 G. v. 28.06.2022 BGBl. I S. 969
+minijob_arbeitgeberpauschale_lohnsteuer:
```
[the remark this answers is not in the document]
JakobWegmann: Tough question!

The calculation is completely independent of the laws in the Lohnsteuer module and therefore fits much better to Minijobs. On the other hand, as a user with less knowledge on the Germany tax and transfer system I would search for 'minijob_arbeitgeberpauschale_lohnsteuer' in the Lohnsteuer module. 

So I think I would move 'minijob_arbeitgeberpauschale_lohnsteuer' to the Lohnsteuer module.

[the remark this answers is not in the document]
hmgaudecker: Thanks, done in #934 !

PR #916, opened 2025-05-19
Active periods for params

hmgaudecker:
As realised in #914, we want to be able to interchangeably use parameters and functions with the same leaf name. Use case there:
```
('sozialversicherung', 'rente', 'beitrag', 'beitragsbemessungsgrenze_m')
```
sometimes depends on the data (different values east/west), sometimes it is a common scalar.

Before this PR, this will fail because we check existence of the above path in the `orig_params_tree` against the same path in the `orig_ttsim_objects_tree` without checking for active periods. In this PR, we will adjust the check.

### DIFF
--- a/src/ttsim/policy_environment.py
+++ b/src/ttsim/policy_environment.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import copy
 import datetime
 import itertools
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
 import dags.tree as dt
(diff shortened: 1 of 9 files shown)


src/ttsim/policy_environment.py
```diff
@@ -264,26 +265,21 @@ def set_up_policy_environment(
...
 ) -> None:
-    """Fail because of clashes of names.
+    """Fail because active periods of TTSIM objects / parameters overlap.
 
-    Two scenarios are checked:
-    - Within TTSIMObjects, active periods of a path + leaf name could overlap
-    - There may be name clashes involving parameters (parameters from multiple files
-      that end up having the same path or parameters named in the same way as a path +
-      leaf name).
+    Checks that objects or parameters with the same path are not active at the same
```
MImmesberger, line 274: We should make clear that this is not about the path in the `orig_...` dict. Maybe `Checks that objects or parameters with the same qualified name are not active at the same time`?

MImmesberger (approved):
I slightly updated the Minijob-Arbeitgeberbeitrag policy functions such that we don't have to skip the test.

Rest looks good imo

PR #923, opened 2025-05-21
Restructure types

hmgaudecker:
Quick one that came up while writing the docs. The type hierarchy was a mess; the TTSIMXyz-Terminology just cluttered things.

The important distinction is between
- things working with data columns
- ParamFunctions 
- ParamObjects

ParamFunctions and ParamObjects are very different whereas ParamFunctions are included in the same pytree as ColumnFunctions and PolicyInputs. Before this PR, all of this was happily conflated...


src/ttsim/typing.py
```diff
@@ -21,47 +21,60 @@
...
-
-    QualNameProcessedParamDict = Mapping[str, Any]
-    """A mapping of qualified names to processed TTSIM parameters."""
+    FlatOrigParamSpecs = dict[tuple[str, ...], OrigParamSpec]
+    """Flat tree of yaml contents; the last element of the key is the leaf name."""
+    OrigParam = dict[str | int, Any]
+    """The original contents of YYYY-MM-DD key, after minimal processing."""
+    NestedParams = Mapping[str, ParamObject | "NestedParams"]
+    """Tree mapping TTSIM paths to parameters."""
+    QualNameParams = Mapping[str, ParamObject]
```
MImmesberger, line 75: We never use the `QualNameParams` type hint and I don't understand the difference to
`QualNameProcessedParams`. Copy/paste error?

src/ttsim/typing.py
```diff
@@ -21,47 +21,60 @@
...
+    """Mapping of qualified name paths to parameters."""
+    QualNameProcessedParams = Mapping[str, Any]
```
MImmesberger, line 77:
```suggestion
    QualNameProcessedParams = Mapping[str, ParamObject]
```
?

src/ttsim/policy_environment.py
```diff
@@ -106,12 +107,12 @@ def params(self) -> dict[str, Any]:
...
 
     @property
-    def params_tree(self) -> NestedTTSIMObjectDict:
+    def params_tree(self) -> NestedColumnObjectsParamFunctions:
         """The parameters of the policy environment."""
         return self._params_tree
 
     @property
-    def combined_tree(self) -> NestedTTSIMObjectDict:
+    def combined_tree(self) -> NestedColumnObjectsParamFunctions:
```
MImmesberger, line 115: Type hint contains functions only, not params. Need a new type hint in `typing.py` (if I
didn't miss it).

src/ttsim/loader.py
```diff
@@ -7,36 +7,38 @@
...
+        FlatColumnObjectsParamFunctions,
+        FlatOrigParamSpecs,
         OrigParamSpec,
     )
 
 
-def orig_ttsim_objects_tree(root: Path) -> FlatTTSIMObjectDict:
+def orig_tree_with_column_objects_param_functions(
+    root: Path,
+) -> FlatColumnObjectsParamFunctions:
```
MImmesberger, line 25: Similar to before: This also returns `ParamObject` which is not clear by looking at the
type hint (also applies to the other `FlatColumnObjectsParamFunctions` type hints in
this file.)

MImmesberger (approved):
Cool, makes it much clearer! 

There's some inconsistency regarding the type hint of dicts that have either a ParamObject, a ColumnFunction, or a ParamsFunction as leaf. Maybe `NestedParamObjectParamFunctionColumnObject`? If this is too long, `NestedTTSIMObjects` might do here (because it contains basically all objects `ttsim` creates).

src/ttsim/typing.py
hmgaudecker: No, but we can delete this one.

src/ttsim/typing.py
hmgaudecker: Nope, the `ParamObjects` really only exist temporarily, mostly for eventual documentation purposes (they are specialised to one year in the `value` attribute, but retain all the info from the header and eventual notes). After processing, we are left with scalar values, homogenous dicts, PiecewisePolynomialParameters, and whatever may come out of the `param_function`s. So no type hints possible on that one.

src/ttsim/policy_environment.py
```diff
@@ -125,11 +126,11 @@ def grouping_levels(self) -> tuple[str, ...]:
         )
 
     def upsert_objects(
-        self, tree_to_upsert: NestedTTSIMObjectDict
+        self, tree_to_upsert: NestedColumnObjectsParamFunctions
```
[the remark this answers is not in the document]
hmgaudecker: Good point, but it could be policy inputs, too... Will change anyhow, let's leave it for now.

src/ttsim/loader.py
hmgaudecker: No, it only returns `ParamFunction`s. See line 87; `ParamObject`s always come from the yaml files.

PR #927, opened 2025-05-24
Move Kinderzuschlag parameters

hmgaudecker:
Move the Kinderzuschlag parameters to new machinery.

Also implements the formula for 2024+, adds start_date everywhere.


src/_gettsim/kinderzuschlag/kinderzuschlag.py
MImmesberger, line 1: I think we should keep the "main" function at the top of the file (`betrag_m_bg`). Do
you agree?

src/_gettsim/param_types.py
MImmesberger, line 1: Nice, I'll add putting all param dataclasses to the TODOs of #908.

MImmesberger (approved):

PR #933, opened 2025-05-26
Move Erziehungsgeld + Elterngeld

MImmesberger:
### Todo

- [x] Erziehungsgeld
- [x] Elterngeld
- [x] Find out whether we lost the pre-2011 values of max_zu_versteuerndes_einkommen_vorjahr_nach_alleinerziehendenstatus along the way
- [x] Find out what to do with `elterngeld_params["datum"]`

### DIFF
--- a/src/_gettsim/config.py
+++ b/src/_gettsim/config.py
@@ -8,7 +8,6 @@ GETTSIM_ROOT = Path(__file__).parent.resolve()
 INTERNAL_PARAMS_GROUPS = [
     "unterhalt",
     "unterhaltsvors",
-    "wohngeld",
     "grunds_im_alter",
 ]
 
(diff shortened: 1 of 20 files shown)


src/_gettsim/erziehungsgeld/erziehungsgeld.py
```diff
@@ -2,14 +2,82 @@
...
+class EinkommensgrenzeErziehungsgeldNachFamilienstand:
+    alleinerziehend: EinkommensgrenzeErziehungsgeldGegebenFamilienstandNachSatzArt
+    paar: EinkommensgrenzeErziehungsgeldGegebenFamilienstandNachSatzArt
+
+
+@dataclass(frozen=True)
+class EinkommensgrenzeErziehungsgeld:
+    regulär: EinkommensgrenzeErziehungsgeldNachFamilienstand
+    reduziert: EinkommensgrenzeErziehungsgeldNachFamilienstand
+    maximalalter_reguläres_limit_monate: int
```
hmgaudecker, line 33: Why not keep this as a scalar parameter? Is often used without the other components and it is not an income threshold.

Issue #938, opened 2025-05-27
ENH: Apply `birth_year_based_phase_inout`, `ConsecutiveInt`-types and `piecewise_constant` to more parameters


MImmesberger:
#### Is your feature request related to a problem?

While working on #908, we introduced new parameter types. We didn't apply them to all old (yaml) parameters.

#### Describe the solution you would like to see

Birth year based phase-in/out could be applied to the following parameters:

- `zurechnungszeitgrenze` (#940)
- `altersgrenze_allgemein` (EM Rente) (#940)

ConsecutiveInt-types could be useful for:
- `altersentlastungsquote_gestaffelt` (#939)
- `maximaler_altersentlastungsbetrag_gestaffelt` (#939)
- `raw_max_miete_m_nach_baujahr`
- `kindergeld__satz_gestaffelt` (#939)

Piecewise constant is useful for:
- `parameter_behindertenpauschbetrag` (done in #897)

MImmesberger:
Not quite. The current `birth_year_based_phase_inout` is basically a birth year based phase in/out **of age thresholds**. For `maximaler_altersentlastungsbetrag_gestaffelt` and `altersentlastungsquote_gestaffelt` we need a birth year based phase in of (generic) scalars.

But we could (and probably should) do this via a `ConsecutiveInt`-type as well. But then we should rename `birth_year_based_phase_inout` such that it becomes clear what the object being phased in is.

PR #939, opened 2025-05-28
Apply `ConsecutiveInt`-type to EstG Abzüge and Kindergeld

MImmesberger:

Partly closes the `ConsecutiveInt`-type section of #938 (everything except for `raw_max_miete_m_nach_baujahr`).

To be discussed: Do we want to find a general solution for my provisional implementation of `get_consecutive_int_1d_lookup_table_with_filled_up_tails`? I don't think we need it somewhere else (as of now) so I think dealing with it right there is fine?


hmgaudecker (approved):
Very nice! A couple of small simplifications should be possible.

PR #941, opened 2025-05-28
Single tree in policy environment

hmgaudecker:
The two trees inside the `PolicyEnvironment` caused constant headaches. This PR includes all elements in a single tree, we do not need a class `PolicyEnvironment` any more.

This allows for goodies like overwriting `param_function`s out of the box.

Also prepares for the new interface by naming the intermediate steps of `compute_taxes_and_transfers` accordingly.


src/ttsim/typing.py
MImmesberger, line 1: Just flagging that the ParamObject import is not correct.

src/ttsim/policy_environment.py
```diff
@@ -47,143 +47,75 @@
...
         NestedParamObjects,
+        NestedPolicyEnvironment,
         OrigParamSpec,
     )
 
 
-class PolicyEnvironment:
-    """
-    A container for policy functions and parameters.
+def grouping_levels(policy_environment: NestedPolicyEnvironment) -> tuple[str, ...]:
```
MImmesberger, line 59:
```suggestion
def grouping_levels(policy_environment: QualNamePolicyEnvironment) -> tuple[str, ...]:
```

src/ttsim/policy_environment.py
```diff
@@ -225,19 +157,63 @@ def set_up_policy_environment(
...
+    -------
+    converted_tree
+        The converted tree.
+
+    """
+    converted = optree.tree_map(
+        lambda leaf: _convert_to_policy_function_if_callable(leaf),
+        tree,
     )
+    fail_if_name_of_last_branch_element_not_leaf_name_of_function(converted)
```
MImmesberger, line 187: Maybe move this one level up, the nesting doesn't make sense to me.

src/ttsim/compute_taxes_and_transfers.py
```diff
@@ -89,91 +96,76 @@ def compute_taxes_and_transfers(
...
+    # Super-ugly, will be refactored
+    all_targets = dt.qual_names(targets_tree)
+    function_targets = [t for t in all_targets if t in _required_column_functions]
+    param_targets = [
+        t
+        for t in set(all_targets) - set(function_targets)
+        if t in _policy_environment_with_derived_functions
+    ]
+    # Will just return these.
+    own_targets = set(all_targets) - set(function_targets) - set(param_targets)  # noqa: F841
```
MImmesberger, line 143: Not sure what this is, data inputs that are returned without GETTSIM doing anything?
Cannot work because of `_fail_if_targets_not_in_policy_environment`.

MImmesberger (commented):
Cool, the new logic of `column_objects_param_function` seems to pay off now. 

Time conversion of scalar parameters is really useful. However, I'm wondering whether we should not allow it for `bool`? It would divide 0/1 by the time converter unit which almost always doesn't make sense.

Also, I'm not sure about getting rid of the `PolicyEnvironment` class entirely. I get why you got rid of it. However, users may find it useful to be able to have methods available to disassemble the tree (e.g. in params functions/objects and column functions; or in different namespaces, see Jürgens comment in the GEP7 PR). Also its convenient to have the upserting mechanic as a method, not as a function. 
This is ofc not of primary concern right now, just wanted to make sure we don't revert back in this regard at some point.

src/ttsim/compute_taxes_and_transfers.py
hmgaudecker: Yes, just a convenience if you want to have some input columns back. Was a constant request / annoyance.

hmgaudecker: Fixed the `fail` function, but did not adjust the behaviour yet.

PR #943, opened 2025-05-30
Remove all occurances of `vectorization_strategy='loop'`

mj023:

Some GETTSIM Functions still use numpy vectorization, indicated by `vectorization_strategy='loop'`. These should be modified to be either vectorized by default or use GETSSIM's vectorization functionality. 

### Todo

- [ ]  Remove `vectorization_strategy='loop'`

  - [ ] `miete.py`
  - [ ] `regelbedarf.py`
  - [ ] `alter.py`
  - [ ] `behinderung.py`
  - [ ] `erziehungsgeld.py`
  - [ ] `kindergeld.py`
  - [ ] `einkommen.py`
  - [ ] `lohnsteuer.py`
  - [ ] `midijob.py`
  - [ ] `solidaritätszuschlag.py`
  - [ ] `arbeitslosengeld.py`
  - [ ] `beitrag.py`
  - [ ] `altersgrenzen.py`
  - [ ] `grundrente.py`
  - [ ] `unterhaltsvorschuss.py`
  - [ ] `einkommen.py`
  - [ ] `wohngeld.py`

### DIFF
--- a/src/_gettsim/individual_characteristics.py
+++ b/src/_gettsim/individual_characteristics.py
@@ -2,19 +2,19 @@ from __future__ import annotations
 
 import datetime
 
-import numpy
+from ttsim.config import numpy_or_jax as np
 
 from ttsim import policy_function
 
 
-@policy_function(vectorization_strategy="loop")
+@policy_function()
 def geburtsdatum(
     geburtsjahr: int,
     geburtsmonat: int,
     geburtstag: int,
-) -> numpy.datetime64:
+) -> np.datetime64:
     """Create date of birth datetime variable."""
-    return numpy.datetime64(
+    return np.datetime64(
         datetime.datetime(
             geburtsjahr,
             geburtsmonat,
--- a/src/_gettsim/wohngeld/miete.py
+++ b/src/_gettsim/wohngeld/miete.py
@@ -172,28 +172,30 @@ def min_miete_m_hh(
     start_date="1984-01-01",
     end_date="2008-12-31",
     leaf_name="miete_m_hh",
-    vectorization_strategy="loop",
+    
 )
 def miete_m_hh_mit_baujahr(
-    mietstufe: int,
-    wohnen__baujahr_immobilie_hh: int,
-    anzahl_personen_hh: int,
-    wohnen__bruttokaltmiete_m_hh: float,
-    min_miete_m_hh: float,
+    mietstufe: np.ndarray,
+    wohnen__baujahr_immobilie_hh: np.ndarray,
+    anzahl_personen_hh: np.ndarray,
+    wohnen__bruttokaltmiete_m_hh: np.ndarray,
+    min_miete_m_hh: np.ndarray,
     max_miete_m_lookup: dict[int, ConsecutiveInt2dLookupTableParamValue],
-) -> float:
+) -> np.ndarray:
     """Rent considered in housing benefit calculation on household level until 2008."""
+
+    # TODO(@mj023): Use N-dimensional Lookup Table once available
+
+    keys = sorted(max_miete_m_lookup.keys())
     selected_bin_index = np.searchsorted(
-        np.asarray(sorted(max_miete_m_lookup.keys())),
+        np.asarray(keys),
         wohnen__baujahr_immobilie_hh,
         side="left",
     )
-    constr_year = list(max_miete_m_lookup.keys())[selected_bin_index]
-    lookup = max_miete_m_lookup[constr_year]
-    max_miete_m = lookup.values_to_look_up[
-        anzahl_personen_hh - lookup.base_to_subtract_rows,
-        mietstufe - lookup.base_to_subtract_cols,
-    ]
+    full_lookup_table = np.stack([max_miete_m_lookup[key].values_to_look_up for key in keys], axis=0)
+    full_lookup_col_substract = np.asarray([max_miete_m_lookup[key].base_to_subtract_cols for key in keys])
+    full_lookup_row_substract = np.asarray([max_miete_m_lookup[key].base_to_subtract_rows for key in keys])
+    max_miete_m = full_lookup_table[selected_bin_index, anzahl_personen_hh - full_lookup_row_substract[selected_bin_index], mietstufe - full_lookup_col_substract[selected_bin_index]]
     return max(min(wohnen__bruttokaltmiete_m_hh, max_miete_m), min_miete_m_hh)
 
 


src/_gettsim/wohngeld/miete.py
```diff
@@ -172,27 +172,38 @@ def min_miete_m_hh(
...
     leaf_name="miete_m_hh",
-    vectorization_strategy="loop",
 )
 def miete_m_hh_mit_baujahr(
-    mietstufe: int,
-    wohnen__baujahr_immobilie_hh: int,
-    anzahl_personen_hh: int,
-    wohnen__bruttokaltmiete_m_hh: float,
-    min_miete_m_hh: float,
+    mietstufe: np.ndarray,
```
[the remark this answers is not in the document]
mj023: I was not sure, after vectorization these are technically all arrays, so the type hints should probably reflect that?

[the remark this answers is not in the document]
hmgaudecker: Yes, but we are writing everything on the individual level only. That's how people looking at the code should think about it, the vectorisation is basically hidden.

src/_gettsim/individual_characteristics.py
```diff
@@ -2,19 +2,18 @@
...
+@policy_function()
 def geburtsdatum(
     geburtsjahr: int,
     geburtsmonat: int,
     geburtstag: int,
-) -> numpy.datetime64:
+) -> np.datetime64:
     """Create date of birth datetime variable."""
-    return numpy.datetime64(
+    return np.datetime64(
```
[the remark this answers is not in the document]
hmgaudecker: Just to be sure: This was an extremely useful mishap :laughing: I would not have checked again otherwise, had given up on this a while ago!

src/_gettsim/wohngeld/miete.py
```diff
@@ -172,27 +172,38 @@ def min_miete_m_hh(
...
+        np.asarray(keys),
         wohnen__baujahr_immobilie_hh,
         side="left",
     )
-    constr_year = list(max_miete_m_lookup.keys())[selected_bin_index]
-    lookup = max_miete_m_lookup[constr_year]
-    max_miete_m = lookup.values_to_look_up[
-        anzahl_personen_hh - lookup.base_to_subtract_rows,
-        mietstufe - lookup.base_to_subtract_cols,
+    full_lookup_table = np.stack(
```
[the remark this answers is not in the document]
mj023: I moved it, but had to create a new Dataclass to be able to pass all the values. We can probably remove this once the N-Dim-LookupTable is implemented.

src/_gettsim/individual_characteristics.py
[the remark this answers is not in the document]
hmgaudecker: So we'd need to tackle that one before the Jax tests will run, right? 

Please revert this change and let's do it once we have converted the interface, at least in a rudimentary fashion.

src/_gettsim/lohnsteuer/lohnsteuer.py
```diff
@@ -2,68 +2,240 @@
...
+
+@policy_function(start_date="2015-01-01")
+def tarif_klassen_5_und_6(
+    einkommen_y: float,
+    einkommensteuer__parameter_einkommensteuertarif: PiecewisePolynomialParamValue,
+    parameter_max_lohnsteuer_klasse_5_6: PiecewisePolynomialParamValue,
+) -> float:
+    """Lohnsteuer for Lohnsteuerklassen 5 and 6."""
+
+    lohnsteuer_klasse_5_6 = basis_für_klassen_5_6(
```
hmgaudecker, line 113:
```suggestion
    basis = basis_für_klassen_5_6(
```

src/_gettsim/lohnsteuer/lohnsteuer.py
```diff
@@ -2,68 +2,240 @@
...
+        einkommen_y, einkommensteuer__parameter_einkommensteuertarif
+    )
+    max_lohnsteuer = piecewise_polynomial(
+        einkommen_y, parameter_max_lohnsteuer_klasse_5_6
+    )
+    min_lohnsteuer = (
+        einkommensteuer__parameter_einkommensteuertarif.rates[0, 1] * einkommen_y
+    )
+    return np.minimum(np.maximum(min_lohnsteuer, lohnsteuer_klasse_5_6), max_lohnsteuer)
```
hmgaudecker, line 122:
```suggestion
    return np.minimum(np.maximum(min_lohnsteuer, basis), max_lohnsteuer)
```

src/_gettsim/lohnsteuer/lohnsteuer.py
```diff
@@ -2,68 +2,240 @@
...
-        einkommensgrenzwerte_steuerklassen_5_6=einkommensgrenzwerte_steuerklassen_5_6,
-        steuerklasse=steuerklasse,
+
+    if steuerklasse == 1 or steuerklasse == 2 or steuerklasse == 4:
+        out = basistarif
+    elif steuerklasse == 3:
+        out = splittingtarif
+    else:
+        out = tarif_klassen_5_und_6
+    out = out / 12
```
hmgaudecker, line 140:
```suggestion
```
and call the function `betrag_y`

src/_gettsim/lohnsteuer/lohnsteuer.py
```diff
@@ -2,68 +2,240 @@
...
+    einkommen_y: float,
+    einkommensteuer__parameter_einkommensteuertarif: PiecewisePolynomialParamValue,
+    kinderfreibetrag_soli_y: float,
+) -> float:
+    """Lohnsteuer in the Basistarif deducting the Kindefreibetrag."""
+    einkommen_abzüglich_kinderfreibetrag_soli = np.maximum(
+        einkommen_y - kinderfreibetrag_soli_y, 0
+    )
+    return piecewise_polynomial(
+        einkommen_abzüglich_kinderfreibetrag_soli,
```
hmgaudecker, line 155:
```suggestion
        x=einkommen_abzüglich_kinderfreibetrag_soli,
```
Please call by name wherever n arguments > 1. Apologies for being pedantic here.

src/_gettsim/wohngeld/miete.py
```diff
@@ -172,27 +197,27 @@ def min_miete_m_hh(
...
         side="left",
     )
-    constr_year = list(max_miete_m_lookup.keys())[selected_bin_index]
-    lookup = max_miete_m_lookup[constr_year]
-    max_miete_m = lookup.values_to_look_up[
-        anzahl_personen_hh - lookup.base_to_subtract_rows,
-        mietstufe - lookup.base_to_subtract_cols,
+    max_miete_m = max_miete_m_lookup.lookup_table[
+        selected_bin_index,
+        anzahl_personen_hh
```
hmgaudecker, line 218:
```suggestion
        anzahl_personen_hh - max_miete_m_lookup.lookup_base_to_subtract_rows[selected_bin_index],  # noqa: E501
```
Out of the top of my head, much easier to read despite the length.

src/_gettsim/wohngeld/miete.py
```diff
@@ -172,27 +197,27 @@ def min_miete_m_hh(
...
+        - max_miete_m_lookup.lookup_base_to_subtract_rows[selected_bin_index],
```
hmgaudecker, line 219:
```suggestion
```

hmgaudecker:
> > * in `lohnsteuer`, rewrite the internal `lohnsteuerformel` so that it we have more `policy_functions`. That is, have policy functions `basistarif`, `splittingtarif`, `tarif_klassen_5_und_6`, then just pick the right element from these three in `betrag_m` and `betrag_mit_kinderfreibetrag_m`.
> 
> I rewrote everything, but I found no way to make them `policy_function` without doubling the code, because the income calculation with the kinderzuschlag is different. I think the old way with a lohnsteuerformel() helper function might be better because now you have to remember to change 2 functions if you make a change to the Lohnsteuerformel.

That is precisely what we have the `leaf_name=...` argument to `policy_function`. Allows you to change the output of a function. So the leaf names in the DAG will just be the same regardless of the implementation and the aggregator does not have to change. @MImmesberger, can you have a look, please? Also, please do check my other comments tagging you so they don't get lost -- once done, please give green light for merging!

src/_gettsim/wohngeld/miete.py
mj023: I could not get this to work with `#noqa`, skipping the formatting worked though.

src/_gettsim/individual_characteristics.py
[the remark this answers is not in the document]
hmgaudecker: Oh yes, just realised that. Anyhow, let's go ahead and merge this once the Lohnsteuer module is cleaned up and then discuss how to best move forward. I'd hope it'll be not too many and fairly simple changes like in b59f61cd, but finding the offending code may require quite some detective work...

hmgaudecker:
Changed my mind: Will merge now, little things can be applied as patches ex post. Don't want to diverge further from #948.

MImmesberger:
> > > * in `lohnsteuer`, rewrite the internal `lohnsteuerformel` so that it we have more `policy_functions`. That is, have policy functions `basistarif`, `splittingtarif`, `tarif_klassen_5_und_6`, then just pick the right element from these three in `betrag_m` and `betrag_mit_kinderfreibetrag_m`.
> > 
> > 
> > I rewrote everything, but I found no way to make them `policy_function` without doubling the code, because the income calculation with the kinderzuschlag is different. I think the old way with a lohnsteuerformel() helper function might be better because now you have to remember to change 2 functions if you make a change to the Lohnsteuerformel.
> 
> That is precisely what we have the `leaf_name=...` argument to `policy_function`. Allows you to change the output of a function. So the leaf names in the DAG will just be the same regardless of the implementation and the aggregator does not have to change. @MImmesberger, can you have a look, please? Also, please do check my other comments tagging you so they don't get lost -- once done, please give green light for merging!

I'm sorry but I cannot find the issue in the code, maybe was looking in the wrong places. Is this still relevant? The functions in `lohnsteuer.py` look good, is the issue somewhere else?

hmgaudecker:
> > > > * in `lohnsteuer`, rewrite the internal `lohnsteuerformel` so that it we have more `policy_functions`. That is, have policy functions `basistarif`, `splittingtarif`, `tarif_klassen_5_und_6`, then just pick the right element from these three in `betrag_m` and `betrag_mit_kinderfreibetrag_m`.
> > > 
> > > 
> > > I rewrote everything, but I found no way to make them `policy_function` without doubling the code, because the income calculation with the kinderzuschlag is different. I think the old way with a lohnsteuerformel() helper function might be better because now you have to remember to change 2 functions if you make a change to the Lohnsteuerformel.
> > 
> > 
> > That is precisely what we have the `leaf_name=...` argument to `policy_function`. Allows you to change the output of a function. So the leaf names in the DAG will just be the same regardless of the implementation and the aggregator does not have to change. @MImmesberger, can you have a look, please? Also, please do check my other comments tagging you so they don't get lost -- once done, please give green light for merging!
> 
> I'm sorry but I cannot find the issue in the code, maybe was looking in the wrong places. Is this still relevant? The functions in `lohnsteuer.py` look good, is the issue somewhere else?

Ah, my bad! I thought that was one pre-2015 and one post-2015 regime, but we are starting only in 2015...

PR #945, opened 2025-06-01
Move param dataclasses to param_types.py

MImmesberger:

Move parameter dataclasses to `param_types.py` and create `SatzMitAltersgrenzen` class.

### DIFF
--- a/src/_gettsim/arbeitslosengeld_2/regelbedarf.py
+++ b/src/_gettsim/arbeitslosengeld_2/regelbedarf.py
@@ -2,13 +2,22 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+from _gettsim.param_types import (
+    Altersgrenzen,
+    BerechtigteWohnflächeEigentum,
+    SatzMitAltersgrenzen,
+    SGBIIRegelsatzAnteilErwachsen,
+    SGBIIRegelsatzAnteilKind,
+    SGBIIRegelsatzAnteilKindNachAlter,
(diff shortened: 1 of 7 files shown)


MImmesberger:
> Looking at the changes, I still think the original version is mostly better. What do you think?

Yes felt a bit weird having to add words to clarify the transfer, but certainly helps to minimize duplications (see ALG2 thing). But I like the original version more too.

> maybe we even want to expose them as a special type similar to ConsecutiveInt... and can use them outside of arbeitslosengeld_2, too?

Yes, will do.

hmgaudecker:
> helps to minimize duplications (see ALG2 thing).

Ah, because we need for SGB XII, too? Is that actually a reference to SGB II in the law or did we take a shortcut at some point?

MImmesberger:
Yes, but apparently the other way around. SGB II points to SGB XII.

MImmesberger:
> Altersgrenzen and SatzMitAltersare nice in the param_types location, maybe we even want to expose them as a special type similar to ConsecutiveInt... and can use them outside of arbeitslosengeld_2, too?

Eventually, I didn't do this. First, because TTSIM-users may want to use their own language/words for the keywords of the dataclasses. The denglish mix of `grundsicherung__regelbedarfsstufen.rbs_4.age_thresholds.min_age` felt wrong. Second, the structure and classes are so simple that we can just leave them at `param_types` -- equally simple I think.

MImmesberger:
No, back then it was part of SGB II

hmgaudecker (approved):
Wonderful, thanks a lot!

PR #947, opened 2025-06-01
METTSIM: Test overwriting scalar param with TTSIMArray

MImmesberger:

Test that we can overwrite a scalar parameter with a TTSIMArray.

I haven't added a test of overwriting a `param_function` because we could only ever overwrite those if they returned a scalar. And this should never happen, right?

### DIFF
--- a/tests/ttsim/test_compute_taxes_and_transfers.py
+++ b/tests/ttsim/test_compute_taxes_and_transfers.py
@@ -120,9 +120,9 @@ def some_converting_params_func(
 
 @param_function()
 def some_param_function_taking_scalar(
-    some_int_param: int, some_float_scalar: float, some_bool_scalar: bool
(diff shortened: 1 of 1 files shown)


hmgaudecker:
Thanks! 

Overwriting scalar `param_functions` should work in exactly the same way, I think. 

In principle, that is true for non-scalar parameters and non-scalar functions, too -- but you are right that this would lead to downstream errors unless one also overwrites the functions that consume the parameters / function outputs.

hmgaudecker:
> And this should never happen, right?

Maybe I didn't quite get that. Why would this never happen? For example, I just merged your PR doing exactly that for the Minijobgrenze :wink:

tests/ttsim/test_create_data_tree.py
MImmesberger, line 1: This is from a different PR; you can ignore that file.

Issue #951, opened 2025-06-04
ENH: Interface sketch


hmgaudecker:
Just want to jot down some ideas for discussion as I don't have enough time today for putting them into [GEP-7](https://github.com/iza-institute-of-labor-economics/gettsim/pull/855). #948 has a prototype for a new interface that is DAG-based, but still messy (some elements have 70+ characters, that was on purpose though). 

What is clear is that the interface-DAG is complex enough to be structured more clearly. Once more, as with the taxes & transfers-DAG, we should turn to a tree-dags for rescue. That hierarchy could looks something like this:

```yaml
- input_data:
  - tree  # ✅ 
  - df_and_mapper:  # ✅ 
    - df  # ✅ 
    - mapper  # ✅ 
  - df_with_nested_columns
  - flat
  - qname
- targets:
  - tree  # ✅ 
  - qname  # ✅ 
  - flat
  - tree_with_map_to_df  # partly implemented, cannot pass this yet instead of tree.
- policy_environment  # ✅ 
- processed_data  # ✅ 
- orig_policy_objects:  # ✅ 
  - column_objects_and_param_functions  # ✅ 
  - param_specs  # ✅ 
- combined_environment:  # ✅ 
  - with_derived_functions_and_input_nodes  # ✅ 
  - with_processed_params_and_scalars  # ✅ 
  - with_partialled_params_and_scalars  # ✅ 
  - tax_transfer_dag  # ✅ 
  - tax_transfer_function  # ✅ 
- names:
  - target_columns  # ✅ 
  - target_params  # ✅ 
  - targets_from_input_data  # ✅ 
  - top_level_namespace  # ✅ 
  - grouping_levels  # ✅ 
  - processed_data_columns  # ✅ 
  - root_nodes
- raw_results:
  - columns  # ✅ 
  - params  # ✅ 
  - input_columns  # ✅ 
  - combined  # ✅ 
- results:
  - tree  # ✅ 
  - df_with_mapper
  - df_with_nested_columns
  - flat
  - qname
- templates:
  - map_tree_paths_to_to_dtypes
- dates:
  - policy_str
  - policy
  - evaluation_str
  - evaluation
- fail_if:  # ✅ 
  - any_paths_are_invalid  # ✅ 
  - data_tree_is_invalid  # ✅ 
  - ...
- warn_if:  # ✅ 
  - ttsim_objects_and_data_columns_overlap  # ✅ 
```

[...]

PR #953, opened 2025-06-05
Hierarchical interface

hmgaudecker:
Add a prototype of the new interface, implementing the sketch at #951 and discussions mostly in #970, #976. Many more improvements. We now have two DAGs:
- The interface DAG (new)
- The taxes & transfers DAG (which already used to be there)


hmgaudecker:
> For some reason I cannot comment `param_objects.py` and `rounding.py`: I find it a bit confusing that those are in `tt_dag_elements` because they're technically not part of the DAG.

I'd love it if we found a better name for the namespace. Better = shorter + more precise. Maybe `tt_leaf_objects`? The important feature is that these are leaves of the various trees we have. Thereby, users may actually want to work with (some of) them directly, as opposed to all other objects, which are exclusively to be worked with via the interface-DAG.

hmgaudecker:
Thanks!!! 
> Looks great, the heavy use of Dags takes some getting used to, but it gets much clearer after looking at the visualization of the `taxes_transfer_dag`. Just two things, one regarding the structure of the modules and one regarding the jitting.

>     1. The  `main` function from `__init__.py` is supposed to be the function the user executes, right? I would not expect such an important function to be defined in an `__init__`-File, maybe it could be moved to it's own file. This might be on me, but I just now realized that this function exists.

Apologies for asking you to look at this prototype. Already on its way out from there.

>     2. If one uses the `main` function, then one would not profit from jitting, because I think we execute the DAG again and create a 'new' function every time we call `main`. This will most likely trigger recompilation, slowing down everything considerably. This would be solved by returning f instead of executing it on the inputs.

Exactly -- you could just ask for the target "tax_transfer_function" and execute that as often as you want with different, but compatible data. Either on your own, or you plug it into a second call where you provide this "tax_transfer_function" as a root node in order to profit from the post-processing functionality.

hmgaudecker:
I tried to get the plotting back to some extent, which actually uncovered a few missing things. Still has the Claude-generated code, though. @MImmesberger, could you maybe
1. Move the check in lines 86+ from `src.ttsim.plot_dag` to a proper test? In GETTSIM, we should be systematic about policy environment dates eventually.
2. Bring back the goodies from the old plotting days like selectors, (dummy) url, improved layout ... (lines 134+ `src.ttsim.plot_dag_old`) as you see fit?
3. If you feel like it, add some color coding / do the layout by group (einkommensteuer in one block of the graph etc.). But not a big deal, this can wait.

I won't be able to do much today, but I have a pretty good idea of fixing the the error handling thing above. If you have time, the other thing that would be great to start looking at would be the "n ways to create a particular node". The example would be to obtain `input_data__flat` (see #958) directly from whatever input data is provided.

PR #958, opened 2025-06-08
Reorder ID's in Input Data

mj023:
Users can provide arbitrary ID's as Input, we want these to be consecutively numbered starting at 0, so we have to convert them internally.

I also added the `reorder_ids` function to `ttsim.shared` because this is now the third place where we are using it.

### Todo

- [x] Reorder Input Data
- [x] Add old ID's back to the Output data


mj023:
The tests are currently failing, this is because of an error in the base branch. I fixed that error locally and then the tests all had positive results.

src/_gettsim/ids.py
```diff
@@ -31,7 +32,7 @@ def ehe_id(
         + np.minimum(p_id, p_id_ehepartner_or_own_p_id) * n
     )
 
-    return _reorder_ids(result)
+    return reorder_ids(result)
```
hmgaudecker, line 35: This will be easy to forget. Could we maybe run this automatically via the `@group_creation_function` decorator?

src/ttsim/interface_dag_elements/processed_data.py
```diff
@@ -26,6 +27,20 @@ def processed_data(
...
         A DataFrame.
     """
-    return {
-        k: np.asarray(v) for k, v in dt.flatten_to_qual_names(input_data__tree).items()
-    }
+    flattened_tree = dt.flatten_to_qual_names(input_data__tree)
+    processed_input_data = {}
+    old_p_ids = flattened_tree["p_id"]
+    new_p_ids = reorder_ids(np.asarray(flattened_tree["p_id"]))
+    for k, v in flattened_tree.items():
```
hmgaudecker, line 34:
```suggestion
    for path, data in input_data__flat.items():
```

src/ttsim/interface_dag_elements/processed_data.py
```diff
@@ -26,6 +27,20 @@ def processed_data(
     Returns:
         A DataFrame.
     """
-    return {
-        k: np.asarray(v) for k, v in dt.flatten_to_qual_names(input_data__tree).items()
-    }
+    flattened_tree = dt.flatten_to_qual_names(input_data__tree)
+    processed_input_data = {}
+    old_p_ids = flattened_tree["p_id"]
```
hmgaudecker, line 32:
```suggestion
    old_p_ids = flattened_tree[("p_id",)]
```
and so on.

src/ttsim/interface_dag_elements/processed_data.py
```diff
@@ -26,6 +27,20 @@ def processed_data(
...
+    new_p_ids = reorder_ids(np.asarray(flattened_tree["p_id"]))
+    for k, v in flattened_tree.items():
+        if k.endswith("_id"):
```
hmgaudecker, line 35:
```suggestion
        if path[-1].endswith("_id"):
```

src/ttsim/interface_dag_elements/results.py
```diff
@@ -17,15 +17,18 @@
 
 
 @interface_function()
-def tree(raw_results__combined: QNameData) -> NestedData:
+def tree(raw_results__combined: QNameData, input_data__tree: NestedData) -> NestedData:
```
hmgaudecker, line 20:
```suggestion
def tree(raw_results__combined: QNameData, input_data__flat: FlatData) -> NestedData:
```
Reverse of above. Then use `path = dt.tree_path_from_qual_name(k)` inside the loop in order to use the same conditions as there.

mj023:
I have added the reordering to the `@group_creation_function` decorator, similar to how the vectorization is handled, and I enabled it by default. I also added the new `FlatData` Variable you proposed. Tests are of course still failing, but it's for the same reason as before.

src/ttsim/tt_dag_elements/column_objects_param_function.py
```diff
@@ -18,7 +18,7 @@
 )
 
 from ttsim.config import IS_JAX_INSTALLED
-from ttsim.interface_dag_elements.shared import to_datetime
+from ttsim.interface_dag_elements.shared import reorder_ids, to_datetime
```
hmgaudecker, line 21: I'd say that this module the correct place to define it now. Clearly a feature of one node in the taxes-transfers-DAG, fine to import it into `process_data`.  What do you think?

src/ttsim/tt_dag_elements/column_objects_param_function.py
```diff
@@ -399,17 +399,33 @@ def group_creation_function(
...
+        The date until which the function is active (inclusive).
+    reorder:
+        Whether the created Group ID's should be reordered to be
+        consecutively numbered starting from 0.
     """
     start_date, end_date = _convert_and_validate_dates(start_date, end_date)
 
     def decorator(func: GenericCallable) -> GroupCreationFunction:
         _leaf_name = func.__name__ if leaf_name is None else leaf_name
+        func_with_reorder = lambda **kwargs: reorder_ids(func(**kwargs))
```
hmgaudecker, line 423: Very nice!

src/ttsim/interface_dag_elements/shared.py
```diff
@@ -30,6 +32,29 @@ def to_datetime(date: datetime.date | DashedISOString) -> datetime.date:
...
+
+    sorting = np.argsort(ids)
+    ids_sorted = ids[sorting]
+    index_after_sort = np.arange(ids.shape[0])[sorting]
+
+    # Look for difference from previous entry in sorted array
+    diff_to_prev = np.where(np.diff(ids_sorted) >= 1, 1, 0)
+
+    # Sum up all differences to get new id
+    cons_ids = np.concatenate((np.asarray([0]), np.cumsum(diff_to_prev)))
```
hmgaudecker, line 53:
```suggestion
    consecutive_ids = np.concatenate((np.asarray([0]), np.cumsum(diff_to_prev)))
```
`cons` is hard-wired to mean `consumption` in my old brain...

PR #960, opened 2025-06-12
Allow setting backend at runtime

hmgaudecker:
We currently use global settings for numpy/jax based on whether jax is installed. This is unsatisfactory from many reasons.

This PR allows setting `backend: Literal["numpy", "jax"]` as part of the inputs of the interface DAG. It defaults to `"numpy"`. All operations on columns that require a module-level call to `np` / `jnp` take a parameter `xnp`, which can be either one depending on the `backend` setting.

- [x] Remove all imports of `config.numpy_or_jax`
- [x] Defer vectorization until the creation of the specialised environment (backend is unknown when decorator is applied)
- [x] Add command-line option to tests for using jax
- [x] Make sure ttsim tests pass
- [x] Make sure GETTSIM test pass
- [x] Remove imports of `IS_JAX_INSTALLED` and adjust testing infrastructure to be based on command-line option everywhere
- [x] Use `jaxtyping` instead of `numpy.ndarray` for type hints.


hmgaudecker:
Btw, both of you: Forget about the mypy errors, I will tackle them later on but would love to have some feedback!

hmgaudecker:
> Should be fixed now, it was just a problem with the test execution. The cached policy environment did not receive a backend, so it defaulted to numpy. I hope it was okay for me to just quickly commit this change to your branch.

Excellent, thanks!!! I might have searched forever for that one... No worries at all re committing to the branch.

hmgaudecker:
@MImmesberger, please double-check my change in [d0a147](https://github.com/iza-institute-of-labor-economics/gettsim/pull/960/commits/d0a147b6f45bfe99b485e55cfd2a244b08e3e7d4). The previous version did not work in jax (iteration there is over are 1-element arrays rather than scalars) and I removed the checks re numeric contents in arrays (superfluous) and the possibility to return lists (cannot happen except for the case where a parameter happens to return a list of length n obs, right?).

MImmesberger:
> cannot happen except for the case where a parameter happens to return a list of length n obs, right?

Yes, that's correct, and the `TypeError` it raises now is more telling than the `ValueError`.

src/ttsim/interface_dag_elements/orig_policy_objects.py
```diff
@@ -48,13 +49,31 @@ def column_objects_and_param_functions(
     root:
         The resource directory to load the ColumnObjectParamFunctions tree from.
     """
-    return {
+
+    @policy_input()
```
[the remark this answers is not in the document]
hmgaudecker: Good point. On the one hand, I completely agree. On the other hand, I really don't want to bloat the types. Let's postpone until we have some experience, will open an issue momentarily.

src/ttsim/interface_dag_elements/policy_environment.py
```diff
@@ -74,6 +79,7 @@ def policy_environment(
         right=_active_param_objects(
             orig=orig_policy_objects__param_specs,
             date=date,
+            xnp=xnp,
         ),
     )
```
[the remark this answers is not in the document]
hmgaudecker: Yes, still missing the dates handling. 

```mermaid
graph TD;
    date_str --> date;
    date --> evaluation_date;
    evaluation_date_str --> evaluation_date;
    date --> policy_date;
    policy_date_str --> policy_date;
```

All these would live in the interface DAG, the evaluation date would also go into the taxes & transfers DAG. The str variables would be inputs, the others functions. Raise an error if both date and either specific date are passed.

src/ttsim/interface_dag_elements/orig_policy_objects.py
```diff
@@ -48,13 +49,31 @@ def column_objects_and_param_functions(
...
+    out = {
         k: v
         for path in _find_files_recursively(root=root, suffix=".py")
         for k, v in _tree_path_to_orig_column_objects_params_functions(
             path=path, root=root
         ).items()
     }
+    # Add backend so we can decide between numpy and jax for aggregation functions
+    assert "backend" not in out
+    out[("backend",)] = backend
```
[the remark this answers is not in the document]
hmgaudecker: These are only the placeholders, right? And they should be filled up here. But probably we could just add to `policy_environment`, too.

src/ttsim/plot_dag.py
```diff
@@ -27,10 +27,14 @@
...
 def plot_tt_dag(
-    with_params: bool, inputs_for_main: dict[str, Any], title: str, output_path: Path
+    with_params: bool,
+    inputs_for_main: dict[str, Any],
+    title: str,
+    output_path: Path,
 ) -> None:
     """Plot the taxes & transfers DAG, with or without parameters."""
-
+    if "backend" not in inputs_for_main:
```
mj023, line 36: Do we need this? I think the same check is already in `main`.

src/ttsim/tt_dag_elements/piecewise_polynomial.py
```diff
@@ -47,59 +51,58 @@ class RatesOptions:
...
-    thresholds : np.array
-                A one-dimensional array containing the thresholds for all intervals.
-    coefficients : np.ndarray
-            A two-dimensional array where columns are interval sections and rows
-            correspond to the coefficient of the nth polynomial.
-    intercepts : np.ndarray
-        The intercepts at the lower threshold of each interval.
-    rates_multiplier : np.ndarray
-                       Multiplier to create individual or scaled rates.
+    parameters:
```
mj023, line 67: Maybe it's somewhere else now, but I think it might be better to explain here what the parameters look like. I mean what rates (?), thresholds and intercepts are.

mj023, line 67: Maybe it's somewhere else now, but I think it might be better to describe here what the parameters look like. I mean what rates (?), thresholds, intercepts are.

src/ttsim/plot_dag.py
mj023, line 36: Do we need this? I think the same check is already in `main`.

mj023 (approved):

src/ttsim/plot_dag.py
hmgaudecker: Probably not, this part of the code was really just a hack to get some visualisation out. As @MImmesberger is working on a proper implementation in #964, we can ignore it here.

PR #963, opened 2025-06-12
Check array length when converting to df.

MImmesberger:

Adresses this comment by @hmgaudecker when checking whether the resuls tree can be converted to DF.

```quote
    # TODO: HM doesn't think this will work as is, we'll need to check the length of
    # the data. Someone might request a policy parameter that is a 3-element array.
```


hmgaudecker (approved):
Thanks! I made two smallish changes -- eventually we may end up with processed_data directly from any kind of input, so better depend on that instead of some input_data element. And outputs can never be Series, so I removed that.

PR #964, opened 2025-06-12
Add plotting options and check for completeness of DAG

MImmesberger:

This PR does two things:

- Adds some options to plotting we have supported in the past
- Test completeness of the DAG for a set of dates. Fails if any input is not specified as param or column object.

### DIFF
--- a/src/ttsim/interface_dag_elements/specialized_environment.py
+++ b/src/ttsim/interface_dag_elements/specialized_environment.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import datetime
 import functools
 from types import ModuleType
 from typing import TYPE_CHECKING, Literal
@@ -40,14 +39,6 @@ if TYPE_CHECKING:
         QNameSpecializedEnvironment2,
         UnorderedQNames,
     )
-
-
-_DUMMY_COLUMN_OBJECT = ColumnObject(
-    leaf_name="dummy",
-    start_date=datetime.date(1900, 1, 1),
-    end_date=datetime.date(2099, 12, 31),
-    description="Dummy column object",
-)
 
 
 @interface_input(in_top_level_namespace=True)
@@ -276,9 +267,11 @@ def with_partialled_params_and_scalars(
             for arg in [
                 a
                 for a in get_free_arguments(func)
-                if not isinstance(
-                    with_processed_params_and_scalars.get(a, _DUMMY_COLUMN_OBJECT),
-                    ColumnObject,
+                if (
+                    a in with_processed_params_and_scalars
+                    and not isinstance(
+                        with_processed_params_and_scalars[a], ColumnObject
+                    )
                 )
             ]:
                 partial_params[arg] = with_processed_params_and_scalars[arg]


src/ttsim/plot_dag.py
```diff
@@ -123,16 +191,93 @@ def plot_full_interface_dag(output_path: Path) -> None:
...
+    inputs_for_main: dict[str, Any],
+) -> list[str]:
+    """Get all targets from the original policy objects / params functions."""
+    return main(
+        inputs=inputs_for_main,
+        targets=["targets__qname"],
+    )["targets__qname"]
+
+
+def specialized_environment_for_targets(
```
hmgaudecker, line 207: Let's see whether we can recycle more code here. The stuff in `specialized_environment` is the most complex part of the code lots of internal (and-probably-still-changing...) logic. Let's chat about that!

src/_gettsim_tests/test_policy.py
```diff
@@ -22,10 +30,55 @@
...
 @pytest.mark.parametrize(
     "test",
     POLICY_TEST_IDS_AND_CASES.values(),
     ids=POLICY_TEST_IDS_AND_CASES.keys(),
 )
 def test_policy(test: PolicyTest, backend: Literal["numpy", "jax"]):
     execute_test(test=test, root=GETTSIM_ROOT, backend=backend)
+
+
+@pytest.mark.parametrize("date", [f"{year}-01-01" for year in range(2015, 2025)])
```
hmgaudecker, line 55: We'll need to sort by start (and potentially end) dates in the original objects here. Sometimes we have pretty large changes within a year.

src/ttsim/plot_dag.py
```diff
@@ -1,150 +1,316 @@
...
-            "The full interface DAG should include all root nodes but requires inputs:"
-            f"\n\n{format_list_linewise(args.keys())}",
+
+
+def _get_tt_dag_with_node_metadata(
+    environment: NestedPolicyEnvironment,
+    node_selector: _QNameNodeSelector | None = None,
+    include_params: bool = True,
+    include_other_objects: bool = True,
+) -> nx.DiGraph:
```
hmgaudecker, line 153: Please propagate to the main interface. This will exclude stuff like the backend, num_segments, etc.

src/ttsim/plot_dag.py
```diff
@@ -159,69 +145,49 @@ def plot_interface_dag(show_node_description: bool = False) -> go.Figure:
...
-    ]
+    qname_environment = dt.flatten_to_qnames(environment)
+    qnames_to_plot = list(qname_environment)
+    if not include_params:
+        qnames_to_plot = [
+            qn
+            for qn, v in qname_environment.items()
+            if not isinstance(v, (ParamObject, ParamFunction))
+        ]
+    if not include_other_objects:
```
MImmesberger, line 163: `other_objects` is the backend literal and module?

src/_gettsim_tests/test_policy.py
MImmesberger: You mean creating the dates from the start dates in the original objects?

src/ttsim/interface_dag_elements/targets.py
```diff
@@ -15,13 +15,19 @@
         OrderedQNames,
     )
 
+NamespaceSelector = Literal["all"] | str  # noqa: PYI051
```
hmgaudecker, line 18:
```suggestion
```

src/ttsim/interface_dag_elements/targets.py
```diff
@@ -15,13 +15,19 @@
...
+    column names.
+    """
 
 
 @interface_function()
 def qname(tree: NestedTargetDict) -> OrderedQNames:
-    """All targets in their qualified name-representation."""
-    return dt.qual_names(tree)
+    """Targets in their qualified name-representation."""
+    return dt.qnames(tree)
```
hmgaudecker, line 33:
```suggestion
    return dt.flatten_to_qnames(tree)
```
We'll should not get rid of the leaves here or make that clear in the name, I think.

src/ttsim/plot_dag.py
```diff
@@ -231,14 +197,15 @@ def _get_tt_dag_with_node_metadata(
...
 
-    if not include_param_functions:
-        selected_dag.remove_nodes_from(
-            [
-                qn
-                for qn, n in specialized_environment.items()
-                if not isinstance(n, ColumnObject)
-            ]
-        )
+    # HM: Shouldn't be necessary; these should just show up as input nodes, right?
```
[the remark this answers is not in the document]
hmgaudecker: So why are we doing the `qnames_to_plot` thing above? Isn't this just duplicating the logic from there? If we cannot reuse, just use the entire DAG up to this point and then kick out nodes ex post only.

src/ttsim/plot_dag.py
```diff
@@ -1,150 +1,316 @@
...
+        )(dummy)
+    if isinstance(obj, InterfaceInput):
+        return interface_function(
+            leaf_name=obj.leaf_name,
+            in_top_level_namespace=obj.in_top_level_namespace,
+        )(dummy)
+    return dummy
+
+
+def create_dag_with_selected_nodes(
```
hmgaudecker, line 271:
```suggestion
def _create_dag_with_selected_nodes(
```

src/ttsim/testing_utils.py
```diff
@@ -222,3 +230,52 @@ def _get_policy_test_from_raw_test_data(
...
+            "targets__qname": list(qname_environment),
+        },
+        targets=[tgt],
+    )[tgt]
+    all_nodes = {
+        qn: dummy_callable(n) if not callable(n) else n for qn, n in env.items()
+    }
+    f = dags.concatenate_functions(
+        functions=all_nodes,
+        targets=qnames_policy_inputs,
```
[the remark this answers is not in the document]
hmgaudecker: Nope, I messed up!

hmgaudecker (approved):
Thanks!!! 

Just a few tiny things left, I think.

src/ttsim/plot_dag.py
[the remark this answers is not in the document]
hmgaudecker: Yeah, we can make that assumption for the sake of our own sanity should we ever try to understand this code again.

src/_gettsim_tests/test_policy.py
```diff
@@ -46,11 +78,20 @@ def test_policy(test: PolicyTest, backend: Literal["numpy", "jax"]):
...
     """Test that GETTSIM's policy environment contains all root nodes of its DAG."""
+    if date.year < 2015:
+        pytest.skip(
+            "Policy environment for dates before 2015 are not complete. See issue #962."
+        )
+
     check_env_completeness(
         name="GETTSIM",
-        date_str=date,
+        date_str=date.isoformat(),
```
hmgaudecker, line 95: Looks like we want to require just `date` there?

tests/ttsim/test_mettsim.py
```diff
@@ -40,6 +50,28 @@ def orig_mettsim_objects():
     )
 
 
+def dates_in_orig_mettsim_objects() -> list[datetime.date]:
+    objects = get_orig_mettsim_objects()
```
hmgaudecker, line 54: `orig_objects = ` 

Else a bit too generic.

src/ttsim/interface_dag_elements/specialized_environment.py
```diff
@@ -46,6 +46,7 @@
     leaf_name="dummy",
```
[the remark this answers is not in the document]
MImmesberger: I actually don't know why it's there. Can you add it or explain it here?

PR #969, opened 2025-06-17
Include failures and warnings in `main()` calls by default

hmgaudecker:
- So far, we always required explicitly asking for `fail_if` or `warn_if` nodes or they were included when asking for all targets. These behaviours do not change.
- This PR adds a switch `fail_and_warn: bool = True` to `main()`, which all failures and warnings that can be computed from the relevant graph (i.e., no extra nodes are added)
- The mechanism is to add `FailOrWarnFunction(InterfaceFunction)` and a corresponding decorator. It has additional attributes `include_if_any_element_present`, `include_if_all_elements_present`. For example, we want to run `fail_if__data_paths_are_missing_in_paths_to_mapped_df_column_names` only if we ask for `results__df_with_mapper` (else we would fail even if we only ask for `results__tree`, which may include dict nodes etc because all inputs to `fail_if__data_paths_are_missing_in_paths_to_mapped_df_column_names` are there). We then specify: `@fail_or_warn_function(include_if_all_elements_present=["results__df_with_mapper"])`


src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -431,7 +435,12 @@ def group_variables_are_not_constant_within_groups(
         raise ValueError(msg)
 
 
-@interface_function()
+@fail_or_warn_function(
```
MImmesberger, line 438: The interface elements specified in the decorator need to be qualified. Not relevant as long as we don't have elements of `fail_if` in there, but we might want to at least enforce that all names are qualified?

src/ttsim/interface_dag_elements/interface_node_objects.py
```diff
@@ -183,3 +185,66 @@ def inner(func: GenericCallable) -> InterfaceFunction:  # type: ignore[type-arg]
...
+
+
+@dataclass(frozen=True)
+class FailOrWarnFunction(InterfaceFunction):  # type: ignore[type-arg]
+    """
+    Base class for all functions operating on columns of data.
+    """
+
+    include_if_any_element_present: Iterable[str]
+    include_if_all_elements_present: Iterable[str]
```
MImmesberger, line 197: We never want both `include_if_all_elements_present` and `include_if_any_element_present` to be specified, right? We should add a check.

src/ttsim/interface_dag_elements/fail_if.py
hmgaudecker: Yeah, I guess just give a meaningful error in case that thing can't be found. Whether because it is a relative path or anything else. Would you mind giving that a shot?

hmgaudecker:
> Cool! I'm not sure how useful `include_if_all_elements_present` is (the example here would work just as fine with `include_if_any_element_present`), have you had something specific in mind?

Just wanted to go for the general solution and not bother with it again. Simple enough. The two conditions are linked with `or`, so it is perfectly fine to specify both.

src/ttsim/interface_dag_elements/fail_if.py
MImmesberger: (Will do tomorrow, just created #974 in case you want to merge before I got to it)

hmgaudecker:
@MImmesberger, would be useful to take another look -- I got carried away a little, in hindsight should have made a new PR.

But we now have `('templates', 'input')` as a potential target and allow for not passing any data / targets. Pretty useless for getting the computed result out, but required for the inputs template.  The nice thing about the latter is that the template immediately specialises when you provide data / targets, we are just re-using the entire machinery. See `tests/ttsim/interface_dag_elements/test_templates.py`

src/ttsim/interface_dag_elements/templates.py
```diff
@@ -0,0 +1,30 @@
...
+    from ttsim.interface_dag_elements.typing import (
+        NestedInputStructureDict,
+        OrderedQNames,
+        QNameSpecializedEnvironment0,
+        UnorderedQNames,
+    )
+
+
+@interface_function()
+def inputs(
```
MImmesberger, line 19: The naming is very confusing, looks like it's a template for the `inputs` argument of
`main`. 

I'm fine with `templates` as the namespace, but maybe we can rename `inputs` to
`data_inputs`? Hopefully this makes it clearer.

tests/ttsim/interface_dag_elements/test_templates.py
```diff
@@ -0,0 +1,173 @@
...
+                    "z": z,
+                },
+            },
+            "rounding": True,
+            "date_str": "2025-01-01",
+            "backend": backend,
+        },
+        output_names=["templates__inputs"],
+    )["templates__inputs"]
+    assert actual == {"a": {"inp2": "float"}, "inp1": "int"}
```
MImmesberger, line 97: I should've mentioned this earlier but isn't it a bit confusing that these are scalar
types, but we expect numpy arrays?

PR #971, opened 2025-06-17
Annotations for `ColumnFunctions`

timmens:
- Postpone vectorisation until the last step of `specialized_environment` before creation of the DAG.
- Adjust signature s.t. `ColumnFunctions` output `(Float|Bool|Int)Columns` instead of scalars
- Make sure processed functions have access to other functions defined in the module by explicitly passing on globals to the wrapped function

### DIFF
--- a/src/ttsim/interface_dag_elements/labels.py
+++ b/src/ttsim/interface_dag_elements/labels.py
@@ -101,7 +101,7 @@ def processed_data_columns(processed_data: QNameData) -> UnorderedQNames:
 
 @interface_function()
 def input_columns(
-    processed_data_columns: UnorderedQNames | None,
+    processed_data_columns: UnorderedQNames,
     policy_environment: NestedPolicyEnvironment,
 ) -> UnorderedQNames:
     """The (qualified) column names in the processed data or policy environment.
@@ -109,17 +109,17 @@ def input_columns(
     Parameters
     ----------
     processed_data_columns:
-        The column names in the processed data or None. Will be returned if not None.
+        The column names in the processed data.
     policy_environment:
         The policy environment. The qualified names of the PolicyInput elements will
-        be returned if the processed_data_columns are None.
+        be returned if the processed_data_columns are empty.
 
     Returns
     -------
     input_columns:
         The (qualified) column names in the processed data or policy environment.
     """
-    if processed_data_columns is None:
+    if not processed_data_columns:
         return {
             k
             for k, v in dt.flatten_to_qnames(policy_environment).items()


src/ttsim/tt_dag_elements/vectorization.py
```diff
@@ -430,3 +439,83 @@ def _module_from_backend(backend: str) -> str:
...
+) -> str:
+    """Convert a scalar type to the corresponding array type."""
+    if scalar_type in ("int", "float", "bool"):
+        return _scalar_numerical_type_to_array_type(scalar_type, backend=backend)
+    return scalar_type
+
+
+def _scalar_numerical_type_to_array_type(
+    scalar_type: str,
+    backend: Literal["numpy", "jax"],
```
hmgaudecker, line 503:
```suggestion
```

src/ttsim/tt_dag_elements/vectorization.py
```diff
@@ -430,3 +439,83 @@ def _module_from_backend(backend: str) -> str:
...
+) -> str:
+    """Convert a scalar numerical type to the corresponding array type."""
+    scalar_type_to_array_type = {
+        "jax": {
```
hmgaudecker, line 507:
```suggestion
```
This branch should be the only thing we'll need.

tests/ttsim/tt_dag_elements/test_vectorization.py
```diff
@@ -681,3 +681,54 @@ def test_already_vectorized_func(xnp):
...
+        "y": "numpy.typing.NDArray[numpy.float64]",
+        "z": "numpy.typing.NDArray[numpy.bool_]",
+        "p1": "str",
+        "p2": "dict[str, float]",
+        "return": "numpy.typing.NDArray[numpy.float64]",
+    }
+    assert inspect.get_annotations(vectorized) == expected_annotations
+
+
+def test_vectorize_function_annotations_jax():
```
hmgaudecker, line 709:
```suggestion
def test_vectorize_function_annotations(backend, xnp):
```
See `conftest.py`. Will be automatically chosen based on flags. Sorry, this I should really have mentioned since it is undocumented so far...

Run jax tests via `pixi run tests-jax tests` now (the last `tests` just because otherwise the GETTSIM tests will run and mostly fail).

MImmesberger (approved):
Cool, thank you. Just fixed some type annotations in the TT part. Also, in [2072abe](https://github.com/iza-institute-of-labor-economics/gettsim/pull/971/commits/2072abe3559b28ea3da54f6d939890e8b78262c5) I added `PolicyInput`s as potential sources for aggregations. I don't remember perfectly what we discussed back then but I think it was not intentional to leave `PolicyInput`s out. Maybe you can double check @hmgaudecker

src/ttsim/interface_dag_elements/specialized_environment.py
```diff
@@ -152,10 +140,16 @@ def _add_derived_functions(
         }.items()
         if isinstance(v, ColumnFunction)
     }
+    policy_inputs = {
```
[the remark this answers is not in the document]
MImmesberger: Not sure I understand this completely, maybe we should discuss tomorrow as well. Just some thoughts:

> After all, policy_inputs are just a placeholder and if a user provides other data, this will end up a mess

That's precisely why we provide `input_columns` as potential aggregation sources too, I think?

> However, it may well be that this has stopped working now that we re-shuffled the order of vectorisations etc.

I am quite sure that this PR or the recent other PRs did not introduce this bug (we always excluded `PolicyInput`s here, going back to the introduction of `PolicyInput`s in March to the source code). The problem was that we didn't acknowledge `PolicyInput`s as potential source columns, which is why the root nodes of the DAG were, for instance, `vermögen_bg` and `vermögen_eg` instead of just `vermögen`. And I don't think thats something we want even if data is provided (regarding setting up processed data from policy inputs only if no data is provided).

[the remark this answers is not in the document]
hmgaudecker: Sure we want this behavior if no data is provided. But if data is there, policy inputs have no role to play whatsoever. They just hint at which inputs we expect in typical cases. If you want to provide those grouped columns, GETTSIM won't object. However, it should not suggest it, either. 

My thinking last week was that providing the column labels would be enough to get the desired behavior. Might well be that this never was the case! 

Happy to chat tomorrow afternoon, though.

src/ttsim/interface_dag_elements/labels.py
```diff
@@ -101,25 +101,25 @@ def processed_data_columns(processed_data: QNameData) -> UnorderedQNames:
...
-        be returned if the processed_data_columns are None.
+        be returned if the processed_data_columns are empty.
 
     Returns
     -------
     input_columns:
         The (qualified) column names in the processed data or policy environment.
     """
-    if processed_data_columns is None:
+    if not processed_data_columns:
```
MImmesberger, line 122: Works now. In the end, it was only this line.

hmgaudecker: Great work, Sherlock! Thanks!

hmgaudecker:
Would be happy to merge. However, we should eventually get some types from the `policy_input`s, e.g. now we see:

```python
{
    'einkommensteuer': {
        'abzüge': {
           'beitrag_private_rentenversicherung_m': 'FloatColumn | IntColumn | BoolColumn',
        }
    }
}
```

Would be pretty easy to look up in the `policy_environment` using optree's tree accessor (paths should be guaranteed to be the same in the (`templates`, input_structure`) tree and there) whenever a leaf has `|` in it, then just convert `float` to `FloatColumn`. Happy to postpone to a small PR, though.

PR #973, opened 2025-06-17
N types of data inputs

MImmesberger:

Make data inputs of different types possible.


MImmesberger:
Superseded by #975

PR #975, opened 2025-06-18
InputDependentInterfaceFunction

MImmesberger:

DAG depends on input data. Allows to specify input data in different ways (nested df, df with mapper, tree).

Supersedes #973.


MImmesberger:
@hmgaudecker You can have a first look now. Still need to add some tests but end-to-end tests pass, so I'm quite positive it works fine.

I actually had the same problem as you had: I need the `inputs` argument of `main` in a flattened version. I implemented an ugly preliminary solution in `main` just to make the tests run. Should discuss how we can get rid of that.

PR #976, opened 2025-06-19
Alternative interface: Top-level elements of interface DAG as arguments of main.

hmgaudecker:
As an alternative to #970, this has a very simple prototype to get a feel for how a more classic interface may look like. The interface DAG would purely be under the hood. 

In principle, we could use a hybrid version and accept, e.g., an `interface_dag_elements.InputData` object instead of dicts -- though in this case, they are probably simple enough?

Tests have not been adjusted except for `tests/ttsim/test_end_to_end.py`


hmgaudecker:
@MImmesberger, I am converging towards the view that we should just make this one the prototype and ditch #970. Just much more natural unless you are very deeply into thinking dags; something which is unlikely to be true for most users except for the two of us (the two sets comments on #855 certainly point into the opposite direction).

The idea is that you start like:

```python
from [ge]ttsim import main, input_data

main(
    date_str="2025-01-01",
    input_data=input_data.
)
```

and then you see the 5 mutually exclusive options appear in your IDE. `test_end_to_end` (the only one expected to work next to the templates test in gettsim) has the final result in action.

I'd probably rename `main` to `ttsim` (and `gettsim` for the derivative, which will just fix `orig_policy_objects__root` and for now also `backend`) and `oss` can go -- there really would be no difference left except for having less options that people never need to worry about.

hmgaudecker:
Perfect. Will merge so that we'll continue with this.

PR #987, opened 2025-06-27
Run fail_if functions first, closes #981

### LINKED ISSUE #981: BUG: `fail_if` functions aren't executed first when calling the concatenated function
### Bug description

`fail_if` function can catch errors that would occur downstream only if they are executed with priority (i.e. before the function that would break given the error). This is not ensured in the concatenated function we create via `dags`. This is why many `fail_if` functions won't throw an error.

hmgaudecker:
Attempt to use the fix outlined in #981.

@MImmesberger, can you please check whether this fixes the issue in the notebook you had somewhere? If possible, make a test case out of that. Thanks!


MImmesberger:
Doesn't seem to work, execution behaves exactly as before. Added test `test_fail_functions_are_executed_with_priority` such that we can check prototypes against it.

hmgaudecker:
Turns out the fix did actually work, but we were not including the fail-functions unless we specified the `include_...` conditions :roll_eyes:. Test passes now!

PR #990, opened 2025-07-02
Remove config elements from policy environment

MImmesberger:

Removes the following config elements from the policy environment by putting them in the interface DAG and partialling them into params and policy functions:
- backend
- xnp
- dnp
- num_segments
- evaluation_date (partial in evaluation_year, _month, _day)


hmgaudecker (approved):
Excellent, much clearer now!

PR #991, opened 2025-07-03
Implement remaining interface components

hmgaudecker:
Update the interface so that we end up with:

```python
def main(
    *,
    output: Output | None = None,
    date_str: DashedISOString | None = None,
    input_data: InputData | None = None,
    targets: Targets | None = None,
    backend: Literal["numpy", "jax"] | None = None,
    rounding: bool = True,
    fail_and_warn: bool = True,
    orig_policy_objects: OrigPolicyObjects | None = None,
    raw_results: RawResults | None = None,
    results: Results | None = None,
    specialized_environment: SpecializedEnvironment | None = None,
    policy_environment: NestedPolicyEnvironment | None = None,
    processed_data: QNameData | None = None,
    date: datetime.date | None = None,
    policy_date_str: DashedISOString | None = None,
    evaluation_date_str: DashedISOString | None = None,
    policy_date: datetime.date | None = None,
    evaluation_date: datetime.date | None = None,
    labels: Labels | None = None,
) -> dict[str, Any]:
```

Also get rid of the `_InterfaceDAGElements` class.


src/ttsim/main_args.py
```diff
@@ -0,0 +1,192 @@
...
+@dataclass(frozen=True)
+class Output(MainArg):
+    name: str | tuple[str, ...] | None
+    names: Iterable[str | tuple[str, ...]] | NestedTargetDict | None
+
+    def __init__(self, *args: Any, **kwargs: Any):  # noqa: ANN401, ARG002, ANN204
+        raise RuntimeError("Use any of the class methods to instantiate this class.")
+
+    @classmethod
+    def name(cls, name: str | tuple[str, ...]) -> Output:
```
MImmesberger, line 45: It's quite nice how we define the inputs of main, we should do something similar with
outputs here. Maybe a class `AllOutputNames` (naming can probably be improved)? E.g.

```python
from ttsim import AllOutputNames

main(
    ...
    output=Output.names(
        [
            AllOutputNames.fail_if.foreign_keys_are_invalid_in_data,
            AllOutputNames.result.tree
        ]
    )
)
```

src/ttsim/interface_dag.py
```diff
@@ -127,60 +136,54 @@ def lexsort_key(x: str) -> int:
...
-    accs, vals = optree.tree_flatten_with_accessor(  # type: ignore[var-annotated]
-        asdict(_InterfaceDAGElements()),  # type: ignore[arg-type]
-        none_is_leaf=True,
-    )[:2]
-    if "input_data" in inputs and isinstance(
-        inputs["input_data"], arg_templates.input_data.ABC
-    ):
-        inputs["input_data"] = inputs["input_data"].to_dict()
-    for acc, val in zip(accs, vals, strict=False):
+    # We only care about some function with a leaf name, not the precise content.
```
MImmesberger, line 146: Maybe rephrase the comment. Right now its probably only useful for yourself; I don't
understand what its supposed to say :smile:

src/gettsim/__init__.py
MImmesberger, line 1: Not sure what is common practice here. Should we provide the ttsim objects defined in `main_args` here as well?

MImmesberger (approved):

src/ttsim/main_args.py
hmgaudecker: ... and `AllOutputNames.result.tree` by now :innocent:

PR #993, opened 2025-07-03
Upserting of params

### LINKED ISSUE #980: ENH: Upserting objects in the policy environment

Upserting is complicated currently, mainly because the TTSIM objects take many input arguments that are mostly irrelevant for users doing policy experiments.

Updating a single parameter, for example, looks like this:

```python
old_beitragssatz = policy_environment["sozialversicherung"]["rente"]["beitrag"]["beitragssatz"]
new_beitragssatz = ScalarParam(
    leaf_name=old_beitragssatz.leaf_name,
    start_date=old_beitragssatz.start_date,
    end_date=old_beitragssatz.end_date,
    value=old_beitragssatz.value + 0.01,
    unit=old_beitragssatz.unit,
    description=old_beitragssatz.description,
    name=old_beitragssatz.name,
    reference_period=old_beitragssatz.reference_period,
)

modified_policy_environment = policy_environment.copy()
modified_policy_environment["sozialversicherung"]["rente"]["beitrag"]["beitragssatz"] = new_beitragssatz
```

Provide an upserting method for each policy object that return a new object with the new specification.

MImmesberger:

Closes #980 

As discussed, makes all arguments but 'value' of the params dataclasses optional.

### DIFF
--- a/src/ttsim/tt_dag_elements/param_objects.py
+++ b/src/ttsim/tt_dag_elements/param_objects.py
@@ -6,8 +6,7 @@ from typing import TYPE_CHECKING, Any, Literal
 
 import numpy
 
-_REQUIRED = object()
-
+PLACEHOLDER_FOR_REQUIRED_FIELDS = object()
 
 if TYPE_CHECKING:
     import datetime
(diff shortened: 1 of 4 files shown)


src/ttsim/tt_dag_elements/param_objects.py
```diff
@@ -47,10 +50,14 @@ class ScalarParam(ParamObject):
     A scalar parameter directly read from a YAML file.
     """
 
-    value: bool | int | float
+    value: bool | int | float = field(default_factory=lambda: _REQUIRED)  # type: ignore[assignment]
```
MImmesberger, line 53: Having no default here doesn't work because this dataclass inherits some default values from another dataclass (at least that's what I think is going on).

Below, we check explicitly, whether `value` was specified or the default was used.

One could also just set the default to `None` but I think the current way is a bit clearer.

hmgaudecker (approved):
Wonderful, thanks! I simplified a bit more.

PR #995, opened 2025-07-04
Small fixes related to #991

MImmesberger:

- Allow for dict output spec in main
- Return qnames when using AllOutputNames


MImmesberger:
Created with different base #996

PR #997, opened 2025-07-04
Improve errors when executing the interface DAG

### LINKED ISSUE #979: ENH: Things we need error messages for
### Missing root nodes

**Interface DAG**
- [ ] If we go with the tree interface: Unexpected/unknown inputs. Currently, doing `"orig_policy_objects": GETTSIM_ROOT` instead of `"orig_policy_objects": {"root": GETTSIM_ROOT}` results in an unreadable error message. (Fixed via arg dataclasses for `main`)

**TT DAG**
- Add a hint that if grouped root nodes are missing, users can insert individual-level variables as well.

### Data converters

- [ ] Throw error if input df mapper maps to a column that is not part of the df. Currently, it just uses the string as a policy input. Alternatively: Check that no column of the result has dtype `object`.

MImmesberger:

Closes #979

### DIFF
--- a/src/ttsim/interface_dag.py
+++ b/src/ttsim/interface_dag.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Literal
 
 import dags
 import dags.tree as dt
+import networkx as nx
 import optree
 
 from ttsim.interface_dag_elements import AllOutputNames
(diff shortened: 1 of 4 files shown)


hmgaudecker (approved):
Very nice, thanks! Just changed the wording here and there.

PR #1000, opened 2025-07-06
Use individual-specific Zusatzbeitrag

### LINKED ISSUE #929: ENH: Zusatzbeitrag KV

When computing social security contributions, we always use the mean Zusatzbeitrag. It is not possible to overwrite it because i) we don't have the infrastructure to overwrite `param_function`s and ii) we are inconsistent in how we access the Zusatzbeitrag (sometimes we use `mean_zusatzbeitrag` from the parameter yamls directly, sometimes we use a `param_function`. 

Make Zusatzbeitrag a `param_function` such that users can easily overwrite it. Similarly, the allgemeine Beitragssatz and the ermäßigte Beitragssatz should get their own `param_function` (this is necessary because they require the `_jahresanfang` functionality in some places).

Downstream functions ( e.g. `beitragssatz_arbeitnehmer`) should then be `policy_function`s.

**Importantly** in some places, we actually want to use the mean Zusatzbeitrag (calculcation of Midijobfaktor F). Don't replace this by the new `param_function` but create a new node `mean_zusatzbeitrag`.

Blocked until `param_function`s can be overwritten.

MImmesberger:

Closes #929


MImmesberger:
@hmgaudecker Tests with jax backend fail because the atol is not being applied; probably the usual problem as discussed [here](https://github.com/iza-institute-of-labor-economics/gettsim/pull/986#issuecomment-3022114115).

Do you have an idea what's going on? I thought your solution in #986 was general, so I didn't look into it.

´pixi run tests --backend=jax -k midijobfaktor.yaml`

hmgaudecker:
Darn, spent way too much time on this now. The fix in #986 was just for the `piecewise_polynomial`. We have the same issue cropping up here whenever we are vectorizing a policy function that has all-scalar inputs. In the years where this is failing, all the inputs to the `midijob_faktor_f_[...]`-functions are `param`s or `param_functions`, so `get_free_arguments()` will return an empty list.

I won't have time to fix this -- could you look into it,please, maybe together with @mj023 ?

hmgaudecker (approved):
Except for the test failures, looks great -- thanks!

MImmesberger:
I added a fix that is probably not optimal. Can you check and revert if you have a better idea @mj023?

@hmgaudecker From an implementation POV, the nicer approach would be to allow `param_function`s to depend on other `param_function`s and just change the decorator for Midijobfaktor F.
However, I'm not totally sure that we want to enforce the usage of `param_function`s in these cases if not absolutely necessary. In the case of the Midijob Faktor, for example, it's not super obvious that all arguments are params.

src/ttsim/interface_dag_elements/specialized_environment.py
```diff
@@ -307,15 +313,51 @@ def with_partialled_params_and_scalars(
...
+                hasattr(rounded_col_func, "__annotations__")
+                and "return" in rounded_col_func.__annotations__
+            ):
+                original_return_type = rounded_col_func.__annotations__["return"]
+
+            partial_func = functools.partial(
                 rounded_col_func, **partial_params_of_this_column_function
             )
+            # If the partialled policy function has no free arguments, we have to make
+            # sure it returns an array, not a scalar.
```
hmgaudecker, line 328: Please explain why, we will have forgotten before long.

hmgaudecker:
FYI -- hijacked this one for a quick fix I had forgotten in the morning.

hmgaudecker:
> Ah, it does. Apparently I had some intermediate state of development in mind.

Excellent. The wonders of using DAGs / `dags`. :laughing: 

> I switched to `param_function`s and removed the fix I implemented earlier. After all, the rounding issue only exists when users test against this specific column and choose the wrong decorator. As far as I understand your discussion above, this might get fixed anyhow on jax' end, if possible.

I don't think there is much of a rounding issue -- unless you mean that rounding apparently cannot be applied to arrays of 1-element-arrays. But then the same thing would crop up elsewhere. Unlikely enough to spend time on it now, hoewever.

> We could even strive for requiring all `policy_function`s to have at least one column input. However, we would have to rewrite the derived function logic then (they are all `policy_function`s even though some should be `param_function`s).

We'll never be able to guarantee that. We certainly don't want to force users to modify a bunch of GETTSIM functions when they replace a data-dependent rule by a scalar or parameter. After all, it is just a (rather minor) Jax issue.

Issue #750, opened 2024-05-03
BUG: Kinderwohngeld


MImmesberger:
### Bug description

We don't consider Kinderwohngeld, which can lead to children dropping out of the Bedarfsgemeinschaft. Because this may influence the Kindergeldübertrag (and changes the composition of the Bedarfsgemeinschaft), this also has effects on the Bürgergeld of the people that stay in the Bedarfsgemeinschaft.

[Here](https://harald-thome.de/files/pdf/2023/ASG-Merkblatt%20Kosa%202024.pdf) is a good example case.

MImmesberger:
[Here](https://www.verwaltungsvorschriften-im-internet.de/bsvwvbund_28062017_SWII4.htm) are some example cases that we may be able to reproduce as tests (search for "Kinderwohngeld" in document).

MImmesberger:
I haven't found the legal basis of Kinderwohngeld (I assume it is **implicitly** derived from regulations in SGBII and WoGG). But online resources strongly suggest that Kinderwohngeld is just a special case of Wohngeld, where a child can claim it even though it is part of a BG whose Eigenbedarf is not satisfied. That is, the child can drop out of the BG (e.g. see [here](https://www.wohngeld.net/besonderheit-beim-kinderwohngeld.html)).

Modelling the Kinderwohngeld should, thus, be straightforward once we have fixed #710 **and** endogenized the BGs via #622.

MImmesberger:
Switching to low-priority. The Kinderwohngeld is an optional mechanism, i.e. the Bedarfsgemeinschaft is not forced to apply for it.

PR #1007, opened 2025-07-09
Remove `kind` input

### LINKED ISSUE #704: ENH: Get rid of input variable `kind`

MImmesberger:

Closes #704 

This PR attempts to remove the `kind` input variable. This is not a quick and innocent change because `kind` tells GETTSIM much about the role of group members when it comes to transfer eligibility. 

I chose the following tricks to get rid of `kind`:

**Unterhalt**
Innocent changes.

**Elterngeld / Erziehungsgeld**
I didn't want to use the Familiengemeinschaft concept too much because it is not part of the Elterngeldgesetz. I added `ist_leistungsbegründendes_kind` which is True for children that may give rise to an Elterngeld/Erziehungsgeld claim.

I also added this concept to the Kindergeld, makes things much clearer IMO.

Alternative: As below, analyse the `fg` structure and look for children. 

**ALG2**
Use `ist_kind_in_bedarfsgemeinschaft` / `ist_kind_in_familiengemeinschaft` in order to tell which data raw refers to a child in the SGB II sense.

Note that children who live in their own BG are now considered to be "adults" from the POV of GETTSIM's SGB II rules. See #1009. This might be correct or not.
Before, users could specify this themselves via the `kind` input. But I doubt anyone did that and we didn't support it systematically, so I wouldn't worry about this too much.

**Wohngeld**

Innocent changes I believe. I removed a wrong if statement in `abzugsanteil_vom_einkommen_für_steuern_sozialversicherung` and an unnecessary check in `freibetrag_m_bis_2015`.

**Kinderzuschlag**

I replaced the `erwachsen` conditions with a condition derived from parent-child links (`hat_kind_in_gleicher_bedarfsgemeinschaft`).

Again, treatment of children might be wrong here as children who live in their own BG are considered to be adults by GETTSIM.

---
In general, we are very bad when it comes to children who claim transfers for themselves (KiZ, ALG2, Wohngeld). But this is a general problem of GETTSIM; this PR highlights it but doesn't make it worse. See issue #1009 and related ones (#750, ...)


src/_gettsim/familie/familie.py
hmgaudecker, line 1: In this file, I'd expect something like:

```py
def kind_in_fg(...) -> bool:
    [check conditions on parent-child-links and age]
```

Then we can just use the below functions. Note the name won't work because it has an `_fg` suffix while it's an individual-level variable, but you get the idea.

Same for `hh`, `bg`, possibly `wthh`, I suppose.

Issue #726, opened 2024-04-25
ENH: Improve on input variable `hat_kinder`


ChristianZimpelmann:
`hat_kinder` is described as `"Has kids (incl. not in hh)"` and is currently used for the calculation of contributions to Pflegeversicherung and for Erziehungsgeld.

- For Pflegeversicherung, we should improve the description to make sure it includes adult children and even deceased children ([source](https://www.haufe.de/personal/haufe-personal-office-platin/beitragszuschlag-fuer-kinderlose-in-der-pflegeversicherung-155-erfuellung-der-elterneigenschaft-von-2-elternteilen_idesk_PI42323_HI11444356.html#:~:text=Auch%20wenn%20das%20Kind%20zwischenzeitlich,Merkmale%20einer%20Lebendgeburt%20getragen%20hat.).
- For Erziehungsgeld, the use of `hat_kinder` seems incorrect. Probably rather check `not kind` or directly relate child-level benefits to parents.
- I was also wondering whether we can remove `hat_kinder` as input variable and make use of `p_id_elternteil_`. However, that would require to have children in other households and even deceased children in the input data set. Shouldn't go this route in the short-term.

Issue #513, opened 2023-01-30
ENH: rename `mietstufe` to `mietstufe_hh`


ChristianZimpelmann:
#### Is your feature request related to a problem?

Mietstufe is a hh-wide variable and should get the appropriate suffix.

Should be an easy change. Let's implement it when not too many PR's are open at the same time.

PR #1011, opened 2025-07-12
Simple renamings.

### LINKED ISSUE #451: Make `wohnort_ost` a household-wide variable.

`wohnort_ost` is defined on the individual level, but this should not differ within the household.

### Implementation

- rename input variable to `wohnort_ost_hh`
- adjust and check code where necessary

Potential difficulty: 
`wohnort_ost` is currently also used in `rentenwert`. But that function should actually not depend on the current place of living, but where the `entgeltpunkte` are earned (see #276)

MImmesberger:

Closes #451, #513, #726


hmgaudecker (approved):
Thank you!

PR #1021, opened 2025-07-15
Get JAX from PyPI and add Apple Metal support

timmens:

1. Fix pixi environment issues
2. Add support for Apple Metal with JAX

### How did you solve it

1. I download JAX completely via PyPI. The JAX wheels include a CPU jaxlib version for all operating systems out-of-the-box. Conda-forge has no compatible Windows jaxlib version for instance.
2. I remove MacOS-Intel from the JAX feature, as it is not available for jax >= 0.6
3. I add another feature (`metal`), which gives JAX the possibility to run more efficient on Apple Silicon

### Testing

This is tested on my Apple M4 MacBook and my Intel ThinkPad with CUDA support.


hmgaudecker (approved):
Thanks! Easier to see the changes relative to the "main" branch, so merging quickly -- let's do any additional changes in #1020

PR #1026, opened 2025-07-15
Fix #1023

### LINKED ISSUE #1023: BUG: `fail_if.input_df_mapper_columns_missing_in_df` is too greedy
### Bug description

Example in GEP-7 with `fail_and_warn=True` yields:

```
ValueError: All columns in the input mapper must be present in the input dataframe. The
following columns are missing: [-1, False, 0.0, 0.0, False]
```

### Solution

Discard non-string elements before checking.

MImmesberger:

Closes #1023

### DIFF
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -558,7 +558,9 @@ def input_df_mapper_columns_missing_in_df(
     """Fail if the input mapper has columns that are not in the input dataframe."""
     mapper_vals = dt.flatten_to_qnames(input_data__df_and_mapper__mapper).values()
     missing_columns = [
-        col for col in mapper_vals if col not in input_data__df_and_mapper__df.columns
+        col
+        for col in mapper_vals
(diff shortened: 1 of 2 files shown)


src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -558,7 +558,9 @@ def input_df_mapper_columns_missing_in_df(
     """Fail if the input mapper has columns that are not in the input dataframe."""
     mapper_vals = dt.flatten_to_qnames(input_data__df_and_mapper__mapper).values()
     missing_columns = [
-        col for col in mapper_vals if col not in input_data__df_and_mapper__df.columns
+        col
+        for col in mapper_vals
+        if col not in input_data__df_and_mapper__df.columns and isinstance(col, str)
```
hmgaudecker, line 563: Thanks! Apologies for the bad issue description. I think we should do the precise what we are doing in `df_with_mapped_columns_to_flat_data`. Maybe we can even refactor so we use the same code?

PR #1027, opened 2025-07-15
Better fail/warn tests

### LINKED ISSUE #1025: ENH: Improve fail/warn tests

We currently have many tests for our fail/warn functions, but we test them explicitly by selecting them as targets. As a consequence, we were missing a pretty big bug through which those fail/warn function were never executed if not specified as targets.

Add tests that emulate a more realistic environment in the sense that they test TTSIM end-to-end, but raise errors along the way. Possibly just change the target of the current tests to `results__tee` or similar.

MImmesberger:

Closes #1025 

### Todo

Add tests via `main` for

- [x] input_data_tree_is_invalid
- [x] environment_is_invalid
- [x] input_df_mapper_columns_missing_in_df
- [x] targets_tree_is_invalid


tests/ttsim/interface_dag_elements/test_failures.py
```diff
@@ -842,6 +844,7 @@ def test_fail_if_p_id_does_not_exist_via_main(backend):
             input_data={"tree": data},
             policy_environment={},
             tt_targets={"tree": {}},
+            date=datetime.date(2025, 1, 1),
```
hmgaudecker, line 847: Looks like the variants of `test_fail_if_p_id_does_not_exist` and `test_fail_if_p_id_is_missing` are exact duplicates?

hmgaudecker (approved):
Very nice, thanks!

PR #1031, opened 2025-07-17
Fix #1006

### LINKED ISSUE #1006: ENH: `("fail_if", "input_data_are_missing")`
@ChristianZimpelmann [writes](https://github.com/iza-institute-of-labor-economics/gettsim/pull/855#issuecomment-3052662548):

> ```python
> result = main(
>     date_str="2025-01-01",
>     main_target=MainTarget.specialized_environment.tax_transfer_dag,
> )
> ```
> 
> Leads to
> 
> ```
> ValueError: The following data columns are missing.
> 
> [ ....
> ```
> 
> Probably better to respond that the argument `input_data` is missing completely

Agreed. Related to #1005, here it will be a matter of checking whether `input_data` (as passed to main) is `None` **and** `processed_data == {}` while any of

- `("specialized_environment", "tax_transfer_dag")`
- `("raw_results", "columns")`

are required to compute the output.

MImmesberger:

Closes #1006

### DIFF
--- a/tests/ttsim/interface_dag_elements/test_failures.py
+++ b/tests/ttsim/interface_dag_elements/test_failures.py
@@ -930,19 +930,6 @@ def test_fail_if_root_nodes_are_missing_via_main(minimal_input_data, backend):
         )
 
 
-def test_fail_if_input_data_is_missing_via_main(backend):
-    with pytest.raises(
-        ValueError,
-        match=r"\('input_data', 'flat'\)",
-    ):
-        main(
-            main_target=MainTarget.specialized_environment.tax_transfer_dag,
-            date_str="2025-01-01",
-            orig_policy_objects={"root": METTSIM_ROOT},
-            backend=backend,
-        )
-
-
 def test_fail_if_root_nodes_are_missing_asks_for_individual_level_columns(
     minimal_input_data, backend
 ):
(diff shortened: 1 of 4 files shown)


tests/ttsim/interface_dag_elements/test_failures.py
```diff
@@ -930,6 +930,19 @@ def c(b):
         )
 
 
+def test_fail_if_input_data_is_missing_via_main(backend):
+    with pytest.raises(
+        ValueError,
+        match=r"\('input_data', 'flat'\)",
+    ):
+        main(
+            main_target=MainTarget.specialized_environment.tax_transfer_dag,
```
hmgaudecker, line 939:
```suggestion
            main_target=MainTarget.specialized_environment.tax_transfer_function,
```

Just to be sure, please add another test where we try to go end-to-end without input data. That is the most common case, I guess.

hmgaudecker (approved):
Very nice, thanks!

hmgaudecker (approved):
Apologies, I did not immerse myself enough into this when replying in the issue -- was thinking that the root nodes in `root_nodes_are_missing` were those of the interface DAG :see_no_evil: 

- Renamed to `fail_if.tt_root_nodes_are_missing`
- Changed the behaviour s.t. we send a different message when `processed_data` is empty
- Removed a test so that we continue to be a bit stricter (less code, less special cases)

Please double-check and merge as you see fit!

PR #1033, opened 2025-07-17
Tighten ruff checks

hmgaudecker:

Enable

- [ ] `INP001` (implicit namespace packages without init.)
- [ ] `PLR2004` (Magic values used in comparison)
- [ ] `PT006` (Allows only lists of tuples in parametrize, even if single argument)
- [ ] `PT007` (wrong type in parametrize)
- [ ] `S101` (use of asserts outside of tests)
- [ ] some more checks on individual files

### DIFF
--- a/src/_gettsim/sozialversicherung/rente/erwerbsminderung/anspruch.yaml
+++ b/src/_gettsim/sozialversicherung/rente/erwerbsminderung/anspruch.yaml
@@ -1,0 +1,15 @@
+---
+mindestpflichtbeitragsmonate:
+  name:
+    de: Mindestpflichtbeitragsmonate für Erwerbsminderungsrente
+    en: Minimum number of required contribution months for disability pension
+  description:
+    de: >-
+      ...
+    en: >-
+      ...
+  unit: Months
+  reference_period: null
+  type: scalar
+  2001-01-01:
+    value: 36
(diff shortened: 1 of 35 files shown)


src/_gettsim/sozialversicherung/rente/erwerbsminderung/anspruch.yaml
```diff
@@ -0,0 +1,15 @@
+---
+mindestpflichtbeitragsmonate:
```
hmgaudecker, line 2: Term correct? If you quickly find a reference, that would be great.

hmgaudecker (commented):
Autoreview.

src/_gettsim/sozialversicherung/rente/erwerbsminderung/anspruch.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name:
+    de: Mindestpflichtbeitragsmonate für Erwerbsminderungsrente
```
MImmesberger, line 4:
```suggestion
    de: Mindestpflichtbeitragszeiten für Erwerbsminderungsrente
```

src/_gettsim/sozialversicherung/rente/erwerbsminderung/anspruch.yaml
MImmesberger:
```suggestion
mindestpflichtbeitragszeiten_monate:
```
Just Pflichtbeitragszeiten in the law. Would be in line with what we do elsewhere.

PR #1039, opened 2025-07-21
Remove endogenous `wthh_id` and `bg_id`

### LINKED ISSUE #1010: ENH: Make BG and WTHH exogenous

We cannot deal with endogenous Vorrang-/Günstigerprüfungen with our static DAG. Rather, we should provide code to do these by calling GETTSIM repeatedly. 

We should remove the code that makes BG and WTHH endogenous and rather add them as a `policy_input`. From our existing code, we can create a preliminary version of the Vorrang-/Günstigerprüfungen in another repo (e.g. `gettsim-groupings`).

MImmesberger:

Closes #1010 as discussed in the comment.


MImmesberger:
Closing because of too many merge conflicts, not worth it to fix all of them.

PR #1041, opened 2025-07-22
Don't create DAG cycles via automatically added functions

### LINKED ISSUE #757: BUG: Error messages for variables that are disaggregated to the individual level
### Bug description

In #753 we introduced the variables `wohnfläche`, `heizkosten_m`, `bruttokaltmiete_m` that are computed from the basic input variables at `hh` level.

If the user forgets to add the individual level variable and the variable on `hh` level, the error message reports a cycle in the DAG. It would be more helpful for the user if this message reported the missing input column.

### Reproducable Example

Test input:

```yaml
---
info:
  note: ''
  source: ''
inputs:
  provided:
    p_id:
      - 0
    hh_id:
      - 0
    bruttokaltmiete_m_bg:
      - 600.0
    heizkosten_m_bg:
      - 0.0
  assumed: {}
outputs:
  _arbeitsl_geld_2_warmmiete_pro_qm_m_bg:
    - 600.0
```

Error message:

```text
E           ValueError: The DAG contains one or more cycles:
E
E           [
E               "['wohnfläche_hh', 'wohnfläche']",
E           ]
```

MImmesberger:

Closes #757

In fact, any occurences of automatically created cycles reported in #757 have already been solved via the namespaces structure.

In ALG2 we have

```python
@policy_function(start_date="2005-01-01")
def wohnfläche(
    wohnen__wohnfläche_hh: float,
    anzahl_personen_hh: int,
) -> float:
    """Share of household's dwelling size attributed to a single person."""
    return wohnen__wohnfläche_hh / anzahl_personen_hh
```

So we don't create a cycle anymore as we have `wohnen__wohnfläche -> wohnen__wohnfläche_hh` but `wohnen__wohnfläche_hh -> arbeitslosengeld_2__wohnfläche`.

Still, thanks to the work done by Lars in the past, a general TTSIM solution was easy to implement because it just copies the logic done for time-conversion functions.

### DIFF
--- a/src/ttsim/interface_dag_elements/automatically_added_functions.py
+++ b/src/ttsim/interface_dag_elements/automatically_added_functions.py
@@ -611,7 +611,7 @@ def create_agg_by_group_functions(
     potential_agg_by_group_sources = {
         qn: o for qn, o in all_functions_and_data.items() if not gp.match(qn)
     }
-    # Exclude objects that have been explicitly provided.u
+    # Exclude objects that have been explicitly provided.
 
     agg_by_group_function_names = {
         t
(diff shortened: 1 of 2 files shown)


hmgaudecker (approved):
Very nice, thank you!

Issue #1003, opened 2025-07-09
ENH: Expose more elements in GETTSIM interface


hmgaudecker:
We currently have to import user-relevant things like `ScalarParam` from `ttsim` internals, e.g. in the [playground notebook](https://github.com/iza-institute-of-labor-economics/gettsim/blob/gep-07/interface-playground.ipynb).

We should expose all things potentially relevant for users in `gettsim`, potentially in namespaces below that.

- both typing modules
- param objects
- column objects / param functions

*[pls expand via editing, if relevant]*

MImmesberger:
Relatedly, we should provide a wrapper for `plot_dag` that injects GETTSIM's root.

MImmesberger:
When doing this, we might want to change the docstrings of some objects. Mentioning the yaml file here, for example, might be confusing for users that just modify the `policy_environment`:

```python
@dataclass(frozen=True)
class ScalarParam(ParamObject):
    """
    A scalar parameter directly read from a YAML file.
    """
    ...
```

hmgaudecker:
> When doing this, we might want to change the docstrings of some objects. Mentioning the yaml file here, for example, might be confusing for users that just modify the `policy_environment`:

We'll have to postpone docs for now, I am afraid...

PR #1052, opened 2025-07-24
Extract TTSIM

hmgaudecker:
In sync with [TTSIM PR 1](https://github.com/ttsim-dev/ttsim/pull/1), this leaves just GETTSIM in here.

Will also tackle #1003 in the process.


MImmesberger (approved):
I'm not a huge fan of `gettsim.tt`, mainly because it's not immediately clear what the `tt` module is supposed to be. Maybe `gettsim.ttsim_objects`? Alternatively `gettsim.core` or `gettsim.policy_objects`?

hmgaudecker:
> I'm not a huge fan of `gettsim.tt`, mainly because it's not immediately clear what the `tt` module is supposed to be. Maybe `gettsim.ttsim_objects`? Alternatively `gettsim.core` or `gettsim.policy_objects`?

As discussed, let's leave it at that for now -- can always change before the new release, after all, ttsim is only in alpha stage (but would have to change there, too).

PR #1054, opened 2025-07-24
DOC: Simple example and comprehensive how-to notebook

### LINKED ISSUE #1034: DOC: example for seriously modifying the policy environment
~~Just stumbled across `_convert_plain_functions_to_policy_functions` in `stale_code_storage.py`. This is to remind ourselves to put that back somewhere. Probably best to start with an example notebook for the docs and then see what is a minimal interface to start with.~~

Based on discussion below: We'll just need an example of modifying the environment with different types of functions etc. -- GEP 7 example just has a tiny stab at modifying a `ScalarParam`.

MImmesberger:

Closes #1034

Tutorial on how to upsert
- parameters
- param_functions
- policy_functions


hmgaudecker (approved):
Wow! This is basically a tour de force of all GETTSIM features -- much more than what I had imagined here.

Maybe the title should be something like "Documentation of Taxes & Transfers Objects and how to modify a policy environment?". 

Please include in the docs.

I just modified the `ConsecutiveInt...` example to use the function for setting it up.

MImmesberger:
Thanks! Sorry for always forgetting about the changelog recently

PR #1056, opened 2025-07-25
Updates to plotting function

MImmesberger:

- Change plotting of dags to future-proof namespacing
- Following some [TTSIM changes](https://github.com/ttsim-dev/ttsim/pull/11/), this PR updates the exposed plotting function and adds `NodeSelector` as an importable element.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/gettsi
 
 ## Unpublished
 
+- {gh}`1056` Updates to plotting functions ({ghuser}`MImmesberger`)
+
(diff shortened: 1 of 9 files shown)


hmgaudecker (approved):
Perfect, thanks! 

The only question I'd have looking at this now would be naming now that we have it in a sub-namespace. That is, most characters appear twice here:

```python
plot_dag.plot_interface_dag(...)
plot_dag.plot_tt_dag(...)
```

I'd prefer:

```python
plot.dag.interface(...)
plot.dag.tt(...)
```

Would also be future-proof regarding plotting other things, like parameter values, function outcomes over time or over a grid.

Probably for a different PR though and would have to be mirrored in TTSIM, anyhow. Sorry for realising so late in the game.

PR #1058, opened 2025-07-26
Add dedicated namespace for Einnahmen

### LINKED ISSUE #1036: ENH: Add `einnahmen` namespace
Came up in the context of #1029. Following suggestion:

```
root
└── einnahmen
    ├── aus_nichtselbstständiger_arbeit
    |     └── bruttolohn_m
    ├── renten
    |   ├── gesetzliche_m(sozialversicherung...altersrente, sozialversicherung...erwerbsminderungsrente)
    |   ├── betriebliche_altersvorsorge_m
    |   ├── geförderte_private_vorsorge_m
    |   └── sonstige_private_vorsorge_m
    ⋮
└── einkommensteuer
    ├── einkünfte
    |   ├── aus_nichtselbstständiger_arbeit
    |       ├── betrag_y(einnahmen__aus_nichtselbstständiger_arbeit__bruttolohn_y)
    ⋮
    |       └── aus_forst_und_landwirtschaft_y
    ⋮
```

Some notes:
- Not necessary to include things where we only have the possibility of modelling `einkünfte` relevant for `einkommensteuer` (example: `aus_forst_und_landwirtschaft_y`), i.e., where we will only ever measure the net of proceeds and expenses.
- Conceptually, in most places `einnahmen__renten__gesetzliche_m` is relevant as opposed to the fact that it is in the sozialversicherung namespace. Would also make it clearer for people who want to override this.
- Similar to the `einnahmen__renten__gesetzliche_m` suggestion, we could think about copying over other sources of income. Though that might yield some room for confusion, so I'd do that only later.

MImmesberger:

Closes #1036


src/_gettsim/arbeitslosengeld_2/einkommen.py
```diff
@@ -98,24 +97,23 @@ def bruttoeinkommen_m(
...
-        + einkommensteuer__einkünfte__sonstige__rente__geförderte_private_vorsorge_m
-        + einkommensteuer__einkünfte__sonstige__rente__betriebliche_altersvorsorge_m
-        + sozialversicherung__rente__altersrente__betrag_m
-        + sozialversicherung__rente__erwerbsminderung__betrag_m
+        + einnahmen__aus_kapitalvermögen__kapitalerträge_m
+        + einnahmen__renten__sonstige_private_vorsorge_m
+        + einnahmen__renten__geförderte_private_vorsorge_m
+        + einnahmen__renten__betriebliche_altersvorsorge_m
+        + einnahmen__renten__gesetzliche_m
+        + einnahmen__sonstige__alle_weiteren_m
```
hmgaudecker, line 108:
```suggestion
        + einnahmen__sonstige_alle_weiteren_m
```
If we are getting rid of the `sonstige` namespace for `renten` (which is good, I think), we should do so all the way.

src/_gettsim/einkommensteuer/einkünfte/aus_kapitalvermögen/aus_kapitalvermögen.py
```diff
@@ -7,21 +7,25 @@
 
 @policy_function(end_date="2008-12-31", leaf_name="betrag_y")
 def betrag_y_mit_sparerfreibetrag_und_werbungskostenpauschbetrag(
-    kapitalerträge_y: float,
+    einnahmen__aus_kapitalvermögen__kapitalerträge_y: float,
```
hmgaudecker, line 10:
```suggestion
    einnahmen__kapitalerträge_y: float,
```
Would seem clear enough?

src/_gettsim/elterngeld/einkommen.py
```diff
@@ -7,7 +7,7 @@
 
 @policy_function(start_date="2007-01-01")
 def anzurechnendes_nettoeinkommen_m(
-    einkommensteuer__einkünfte__aus_nichtselbstständiger_arbeit__bruttolohn_m: float,
+    einnahmen__aus_nichtselbstständiger_arbeit__bruttolohn_m: float,
```
hmgaudecker, line 10:
```suggestion
    einnahmen__bruttolohn_m: float,
```
Also enough.

hmgaudecker (approved):
Looks great to me! Apologies for the scattered comments, hard to comment directly on the tree. At a high level, I think the only sub-namespace of `einnahmen` should be `renten`. Everything else is clear right away without the namespaces and mostly there is a single element.

As for `sonstige_alle_weiteren_m`, I do wonder whether we want / need that. We do not have any `einnahmen` when we cannot make a distinction between that and `einkünfte` elsewhere, so I wonder why would want to do it here?

MImmesberger:
I did not implement `einnahmen__sonstige_alle_weiter_m` or something similar. Instead, we just use `einkommensteuer__einkünfte__sonstige__alle_weiteren_m`.

Everything else is exactly as requested.

MImmesberger:
You can have another look, else feel free to merge!

PR #1071, opened 2025-08-08
Move to new plotting interface.

MImmesberger:

Twin PR to [TTSIM # 32](https://github.com/ttsim-dev/ttsim/pull/32). Move the plotting interface to the one defined over there.

Also, I removed some orphaned nodes I discovered when trying out the new mechanic.


MImmesberger:
It turns out that the jittability tests made use of the trick we added to create specialized environments without adding input data. This doesn't work anymore.

Three options:
- Cache the policy environment instead of the specialized policy environment. Then, create an input template to have fake input data, and then do the jittability test for all/many functions at once.
- Just as before, but test each TT function separately. This will increase runtime compared to the status quo.
- Cache the `specialized_environment_for_plotting_and_templates` (would have to rename it I think) and use the exact logic as before.

Preferences? @hmgaudecker

pyproject.toml
```diff
@@ -40,7 +40,7 @@ classifiers = [
 ]
 # Dependencies and requirements
 requires-python = ">=3.11"
-dependencies = ["ttsim-backend>=1.0a3"]
```
hmgaudecker, line 43: Just leaving this here as a reminder not to merge before releasing ttsim 1.0a4 (or 1.0b).

hmgaudecker (commented):
Looks great to me! 

Only reason not to approve is that we should wait for another TTSIM alpha (or beta) release.

MImmesberger:
Just a note on the failing tests: I'm 99% this is just a caching error because I fixed precisely this error via a commit in the TTSIM PR and tests run fine locally.

hmgaudecker (approved):
Publishing of ttsim done, so this is good to go!

PR #1074, opened 2025-08-10
DOC: General Update

MImmesberger:

- Update the docs for release
- Add installation instructions for uv, mamba, pip
- Update the Elterngeld notebook to new interface
- Add a general tutorial regarding history dependent TTs


MImmesberger:
@hmgaudecker Because TTSIM 1.0 is not released yet, I wasn't able to try out the installation via `pip`. So I mentioned only `graphviz` as an additional dependency (because you mentioned it). Might be missing something, not completely sure.

Issue #1073, opened 2025-08-10
DOC: explain colors in plotting interface


ChristianZimpelmann:
#### Location of the documentation

[https://gettsim.readthedocs.io/en/latest/how_to_guides/visualizing_the_system.html
](https://gettsim.readthedocs.io/en/latest/how_to_guides/visualizing_the_system.html)

#### Documentation problem

It is unclear whether the colors in the plots have any meaning. If so, it would be great to mention it. Otherwise, maybe mention this as well (and in this case think about whether different colors do make sense)

MImmesberger:
Thanks! We have one color for each top-level namespace. Will mention this in #1074.

hmgaudecker:
It would actually be good to fix the colors by namespace, rather than cycling through them (e.g., the `einkommensteuer` namespace is coloured in orange in the first graph [here](https://gettsim.readthedocs.io/en/latest/how_to_guides/visualizing_the_system.html) and red in the second. But I guess that would require changes to TTSIM?

hmgaudecker:
> A different solution would be to create a top-level-namespace-to-color map which is an input to `tt`, i.e. independent of which TT DAG to plot. We could either set the colors manually 

Yeah, this is what I meant. If staring at these things often enough, it will be important to keep the colors fixed to grasp things quickly in your mind. *(btw, while working on #1072, I was thinking whether we could get always-visible node labels back? they are useful for small graphs)*

> (but this might become a hassle once we add more namespace and try to accommodate different degrees of color blindness).

We should just expose the colormap argument and set a reasonable default. No way to accommodate all possible things you may want in different circumstances.

PR #1083, opened 2025-08-22
Plot with fixed colormap

hmgaudecker:
Make use of the possibilities that [TTSIM #44](https://github.com/ttsim-dev/ttsim/pull/44) opens up.

Add default colormap for GETTSIM:

- Top-level elements and background/general input variables: **blue**

- Taxes: **red**. 

  Exception: Einkünfte are mix of Einnahmen/Tax rules: **purple**.

- Social insurance: **yellow**

  Differentiate between programs and between pension contributions and pension benefits.

- Transfers: **green**

If not mentioned otherwise, differentiation within colors happens at the first level of the namespace hierarchy.

Addresses #1073, though that will be fixed in #1074.

### DIFF
--- a/src/gettsim/plot/dag/__init__.py
+++ b/src/gettsim/plot/dag/__init__.py
@@ -6,6 +6,42 @@ import ttsim
 
 from gettsim import germany
 
+GETTSIM_COLORMAP = {
+    # Top-level, background variables - blue.
+    ("top-level",): "lightskyblue",
+    ("einnahmen",): "mediumblue",
+    ("familie",): "skyblue",
+    ("wohnen",): "mediumturquoise",
+    ("hh_characteristics",): "skyblue",
+    ("ids",): "deepskyblue",
+    ("unterhalt",): "teal",
+    # Taxes - red. Exception: Einkünfte are mix of Einnahmen/Tax rules - purple.
+    ("einkommensteuer",): "crimson",
+    ("einkommensteuer", "einkünfte"): "purple",
+    ("lohnsteuer",): "red",
+    ("solidaritätszuschlag",): "darkred",
+    # Social insurance - differentiate between programs and between pension
+    # contributions and pension benefits.
+    ("sozialversicherung",): "gold",
+    ("sozialversicherung", "arbeitslosen"): "yellowgreen",
+    ("sozialversicherung", "kranken"): "yellow",
+    ("sozialversicherung", "pflege"): "khaki",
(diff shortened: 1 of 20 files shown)


src/gettsim/plot/dag/__init__.py
```diff
@@ -2,10 +2,46 @@
...
+    ("unterhalt",): "teal",
+    # Taxes - red. Exception: Einkünfte are mix of Einnahmen/Tax rules - purple.
+    ("einkommensteuer",): "crimson",
+    ("einkommensteuer", "einkünfte"): "purple",
+    ("lohnsteuer",): "red",
+    ("solidaritätszuschlag",): "darkred",
+    # Social insurance - differentiate between programs and between pension
+    # contributions and pension benefits.
+    ("sozialversicherung",): "gold",
+    ("sozialversicherung", "arbeitslosen"): "yellowgreen",
```
MImmesberger, line 26:
```suggestion
    ("sozialversicherung", "arbeitslosen"): "palegoldenrod",
```

To me 'yellowgreen' looks too greenish, seems like a transfer at first glance.

src/gettsim/plot/dag/__init__.py
```diff
@@ -2,10 +2,46 @@
...
+    ("sozialversicherung", "rente"): "goldenrod",
+    ("sozialversicherung", "rente", "beitrag"): "darkgoldenrod",
+    # Transfers - green
+    ("kindergeld",): "olive",
+    ("kinderbonus",): "darkolivegreen",
+    ("kinderzuschlag",): "olivedrab",
+    ("elterngeld",): "darkgreen",
+    ("erziehungsgeld",): "darkgreen",
+    ("unterhaltsvorschuss",): "seagreen",
+    ("wohngeld",): "turquoise",
```
MImmesberger, line 38:
```suggestion
    ("wohngeld",): "darkseagreen",
```

`turquoise` is very blue...

src/gettsim/plot/dag/__init__.py
JuergenWiemers: I agree, "palegoldenrod" would be better here.

src/gettsim/plot/dag/__init__.py
JuergenWiemers: Also agreed, and "Wohngeld" and "Kinderzuschlag" still look different enough with this change.

PR #1086, opened 2025-08-25
Tutorial for GETTSIM Personas

MImmesberger:

Adds a tutorial notebook for GETTSIM personas.

### Todo

A section about installation is missing. I guess we have to publish first?


docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+    "  [contributions are welcome!](https://github.com/ttsim-dev/gettsim-personas/issues/9))\n",
+    "- `policy_date`: The policy date of this persona. The policy date reflects the policy\n",
+    "  environment for which the persona was created.\n",
+    "- `evaluation_date`: The evaluation date of this persona, i.e. the date at which taxes\n",
+    "  and transfers should be computed. Used to calculate some of the persona's input data\n",
+    "  columns endogenously, e.g. the age at retirement.\n",
+    "- `input_data_tree`: The input data tree of this persona. This can be passed directly to\n",
+    "  GETTSIM's `main` function.\n",
+    "- `tt_targets_tree`: The targets that can be computed for this persona. This can also be\n",
+    "  passed to GETTSIM's `main` function.\n",
```
hmgaudecker, line 77:
```suggestion
```

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+    "\n",
+    "result = main(\n",
+    "    main_target=MainTarget.results.df_with_nested_columns,\n",
+    "    policy_date=example_persona.policy_date,\n",
+    "    input_data=InputData.tree(example_persona.input_data_tree),\n",
+    "    tt_targets=TTTargets.tree(example_persona.tt_targets_tree),\n",
+    "    include_warn_nodes=False,\n",
+    ")\n",
+    "\n",
+    "result"
```
hmgaudecker, line 108: Is it hard to change the display options so that empty columns are shown instead of `NaN`? Don't spend more than 5min on it, though.

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4",
+   "metadata": {},
+   "source": [
+    "### Optional Arguments When Instantiating a Persona\n",
+    "\n",
+    "You can provide a grid of earnings levels to compute taxes and transfers across\n",
+    "different earnings levels. For example, we may want to look at net income in our example\n",
```
hmgaudecker, line 119:
```suggestion
    "different earnings levels. For example, in our `example_persona`, the parents earn ... We may want to look at net income in our example\n",
```

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+    "    policy_date_str=\"2025-01-01\",\n",
+    "    bruttolohn_m_linspace_grid=einkommensteuer_sozialabgaben.Couple1Child.LinspaceGrid(\n",
+    "        p0=einkommensteuer_sozialabgaben.Couple1Child.LinspaceRange(\n",
+    "            bottom=4_000, top=4_000\n",
+    "        ),\n",
+    "        p1=einkommensteuer_sozialabgaben.Couple1Child.LinspaceRange(\n",
+    "            bottom=1_000, top=3_000\n",
+    "        ),\n",
+    "        p2=einkommensteuer_sozialabgaben.Couple1Child.LinspaceRange(bottom=0, top=0),\n",
+    "        n_points=100,\n",
```
hmgaudecker, line 147: Why only here? Use in all cases and explain that it defaults to 100, but needs to be the same across all inputs. Also explain why we can't just pass `4_000` for dad and kid.

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6",
+   "metadata": {},
+   "source": [
+    "Personas support different evaluation dates. Personas fix some input columns to constant\n",
+    "values (e.g. age at 30 years) and calculate the values of other input columns (e.g.\n",
+    "birth year as evaluation date minus 30 years). If no evaluation date is provided, the\n",
```
hmgaudecker, line 159:
```suggestion
    "birth year as evaluation date minus 30 years). In most cases, this will not matter much, but for many pension-related elements of taxes and transfers there often are birth year-based rules. If no evaluation date is provided, the\n",
```

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+    "    ),\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6",
+   "metadata": {},
+   "source": [
+    "Personas support different evaluation dates. Personas fix some input columns to constant\n",
```
hmgaudecker, line 157:
```suggestion
    "If you want to vary anything else than `bruttolohn_m`, use the upserting mechanism described below.\n",
    "\n",
    "Personas support different evaluation dates. Personas fix some input columns to constant\n",
```

docs/tutorials/personas.ipynb
```diff
@@ -0,0 +1,336 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "result = main(\n",
+    "    main_target=MainTarget.results.df_with_nested_columns,\n",
+    "    policy_date=persona_with_varying_rent.policy_date,\n",
+    "    input_data=InputData.tree(persona_with_varying_rent.input_data_tree),\n",
+    "    tt_targets=TTTargets.tree(persona_with_varying_rent.tt_targets_tree),\n",
+    "    include_warn_nodes=False,\n",
+    ")"
```
hmgaudecker, line 309:
```suggestion
    ")",
    "result"
```

docs/tutorials/personas.ipynb
MImmesberger: Something like this works:

```python
import pandas as pd
result_varying_rent.columns = pd.MultiIndex.from_tuples(
    [tuple("" if pd.isna(x) else x for x in col) for col in result_varying_rent.columns]
)

result_varying_rent.head()
```

But maybe we should implement this in TTSIM instead?

docs/tutorials/personas.ipynb
MImmesberger: Use what in all cases? `n_points` is an argument of the `LinspaceGrid`, not of the ranges defined for each person. There are no defaults for `n_points`.

docs/tutorials/personas.ipynb
hmgaudecker: No, we definitely do not want to mess with the data proper -- should leave there what pandas recognises as missing out-of-the box. If it does not work via display options, no problem at all!

docs/tutorials/personas.ipynb
```diff
@@ -222,7 +225,8 @@
    "metadata": {},
    "source": [
     "If you want to vary anything else than `bruttolohn_m`, use the upserting mechanism\n",
-    "described below.\n",
+    "described below. You can also specify LinSpaceRanges for multiple `p_id`s at once; they\n",
+    "just need to have the same length.\n",
```
MImmesberger, line 229: They will always have the same length because `n_points` is an attribute of the grid, not the range.

MImmesberger:
I added the required changes to `gettsim-personas` but this: `result_varying_income_of_secondary_earner.reset_index().set_index(["hh_id", "p_id"])` is not that easy because of the nested column names. 

The best I could come up with is this
Screenshot 2025-08-26 at 15 31 10

But I would just leave it as is tbh.

PR #1087, opened 2025-08-26
Remove wohnort_ost from simple_example notebook.

MImmesberger:

@felixschmitz made me aware that we have `wohnort_ost` as an input there. This PR removes it.

- The input is not necessary at all in 2025.
- It leads to errors in the years before because the actual input name should be `wohnort_ost_hh`.


hmgaudecker (approved):
Thanks!

Issue #738, opened 2024-04-29
BUG: Defintion of Einstandsgemeinschaften


MImmesberger:
Sorry to revive this discussion again: 

### Status Quo

In our current implementation, Einstandsgemeinschaft refers to couples who live together (specified via `p_id_einstandspartner`). This follows the definition in SGB II: [§7 Abs. 3a SGB II](https://www.gesetze-im-internet.de/sgb_2/__7.html) defines a Einstandsgemeinschaft as two people living together with a mutual willingness to take care for each other (so no children included here).

Here, the Einstandsgemeinschaft is part of the Bedarfsgemeinschaft. It seems like the purpose of the Bedarfsgemeinschaft in SGB II is to summarize couples who are not necessarily married under one term (afaik there is no connection to transfers beyond the membership in a Bedarfsgemeinschaft).

In SGB XII, the Einstandsgemeinschaft is the group that is entitled for some transfer and **includes underage children** who live in the same household and cannot cover their own needs. See, for example, this definition of [BMAS](https://www.bmas.de/DE/Soziales/Sozialhilfe/Leistungen-der-Sozialhilfe/leistungen-der-sozialhilfe.html).

>Bei in Privathaushalten lebenden Personen werden zusammenwohnende Partner sowie im Haushalt lebende minderjährige Kinder als sog. Einstandsgemeinschaft betrachtet. Eine Einstandsgemeinschaft ist dadurch gekennzeichnet, dass die Partner dieser Einstandsgemeinschaft füreinander mit ihrem Einkommen und Vermögen einstehen.

### Implications for GETTSIM

It seems to me that all occurrences of `_eg` in the code should actually refer to the SGB XII term, not the SGB II term (because they are used for SGB XII transfers). Hence, we would need to adjust the creation of `eg_id` accordingly.

I don't know yet how Einstandsgemeinschaft (SGB XII) relates to Bedarfsgemeinschaft (SGB II) other than the fact that in the latter, there needs to be at least one person that is able to work.

hmgaudecker:
Not sure. Seems to me like this is a definitional thing and we can just keep it as is -- we need that definition from SGB II in order to create `bg_id`s, right?

> I don't know yet how Einstandsgemeinschaft (SGB XII) relates to Bedarfsgemeinschaft (SGB II) other than the fact that in the latter, there needs to be at least one person that is able to work.

Sounds like it is fine to run with the `bg_id` for SGB XII as well.

[...]

PR #1092, opened 2025-09-05
Fix a couple of vectorization strategies.

hmgaudecker:
Just a couple of things I noticed while roaming the codebase related to [TTSIM #53](https://github.com/ttsim-dev/ttsim/pull/53).

Not sure about whether `ist_kind_in_einstandsgemeinschaft` is correct as is or it should have more conditions, though? We can leave that corner case for #738 though.

### DIFF
--- a/src/gettsim/germany/familie/familie.py
+++ b/src/gettsim/germany/familie/familie.py
@@ -143,7 +143,7 @@ def alter_monate_jüngstes_mitglied_fg(alter_monate: int, fg_id: int) -> float:
     pass
 
 
-@policy_function(start_date="2005-01-01")
+@policy_function(start_date="2005-01-01", vectorization_strategy="not_required")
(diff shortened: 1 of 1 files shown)


MImmesberger (approved):

hmgaudecker:
So we should leave it as a placeholder?

MImmesberger:
Not sure what you mean with placeholder. Recipients of Grundsicherung im Alter get additional money if they care for underage children (so we need it here, not a placeholder). The implementation silently assumes that no transfer recipient is married to a minor. But ofc, it would be better to do something similar as with `ist_kind_in_bedarfsgemeinschaft`.

We should note somewhere that children in GETTSIM are defined as individuals (below a certain age) that have a caretaker and are not caretakers themselves (as opposed to children just being minors).

PR #1113, opened 2026-01-07
Vorsorgepauschale Reform 2026

### LINKED ISSUE #1116: BUG: Lohnsteuer Steuerklasse 5/6 wrong since at least 2025
### Bug description

In #1113 we skipped the tests for Steuerklasse 5 and 6. They are off since at least 2025.

MImmesberger:

#1098 

Also closes #1116


MImmesberger:
Good news is that I'm able to replicate the results of the BMF Lohnsteuerrechner for Steuerklasse 1-4. Results are off for Steuerklasse 5 + 6, but they are off for 2025 as well. So I would suggest to postpone that one.

I also didn't manage to get the Lohnsteuer for privat Krankenversicherte right (see #1115), but I think it's not related to the calculation of Vorsorgepauschale (not 100% sure).

PR #1127, opened 2026-01-14
Remove `entgeltpunkte_updated`, rename `neue_entgeltpunkte` -> `neue_entgeltpunkte_y`

hmgaudecker:
- Adjust tests
- Document in example

ToDo:

- [x] Check for possibly other occurrences of adding flows to stocks
- [x] Document that aggregating stocks and flows needs to be done outside of GETTSIM. Can use discussion in [TTSIM #66](https://github.com/ttsim-dev/ttsim/issues/66) as a basis.

### DIFF
--- a/docs/how_to_guides/inputs_from_prior_periods.md
+++ b/docs/how_to_guides/inputs_from_prior_periods.md
@@ -12,6 +12,8 @@ therefore include both values taken from a fixed historical window (as in parent
 leave) and state variables that evolve over time and are carried forward into the
 current period.
 
+Also see {ref}`flows-and-stocks` for more background.
+
 ## General recipe
 
 In general, follow this recipe to calculate these taxes and transfers:
--- a/docs/tt_explanations/flows_and_stocks.md
+++ b/docs/tt_explanations/flows_and_stocks.md
@@ -1,4 +1,4 @@
-(flows_and_stocks)=
+(flows-and-stocks)=
 
 # Why stock transitions need to be calculated outside of GETTSIM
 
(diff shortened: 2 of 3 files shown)


docs/how_to_guides/inputs_from_prior_periods.md
```diff
@@ -1,15 +1,8 @@
 # How to calculate input variables from prior periods' states
 
 Many taxes and transfers require inputs from prior periods to serve as inputs for this
-year’s calculations. For example, parental leave benefits often rely on an estimate of
-the claimant’s net income during the 12 months preceding the child’s birth.
```
MImmesberger: I would have stopped reading here if I were interested in updating my inputs, so maybe add something like:

Another example is the calculation of public pension benefits, which relies on a measure of lifetime earnings that is accumulated and updated continuously each month. Prior-period inputs therefore include both values taken from a fixed historical window (as in parental leave) and state variables that evolve over time and are carried forward into the current period.

docs/tt_explanations/flows_and_stocks.md
```diff
@@ -0,0 +1,89 @@
...
+
+**This transition has to be done outside of GETTSIM.** That is, GETTSIM does not
+automatically accumulate flows into stocks because it would need to know the length of
+the period you are thinking about. Since
+
+1. these rules are typically very simple (as in the two examples above),
+1. adding this period length as a parameter to `main` would entail large costs, and
+1. it affects only a small subset of use cases (dynamic models),
+
+we decided against including it. Put differently, GETTSIM will only ever calculate
```
MImmesberger: "only ever" is a strong statement - perks of being BDFL 😄

PR #1132, opened 2026-01-30
Refactor piecewise polynomial (implement GEP-8)

hmgaudecker:
Refactor piecewise polynomial YAML and code for GEP-08, twin to [TTSIM #76](https://github.com/ttsim-dev/ttsim/pull/76)
    
Convert all piecewise polynomial YAML parameters from numbered dict format
to ordered lists with explicit interval strings:
- solidaritätszuschlag, einkommensteuertarif, vorsorge, behinderung,
  wohngeld, grundrente, bürgergeld/arbeitslosengeld_2 freibeträge,
  lohnsteuer vorsorge
- Rename keys: rate_linear->slope, rate_quadratic->quadratic,
  intercept_at_lower_threshold->intercept
- Update Python code to use new API (parameter_list instead of
  parameter_dict, new key names)
- Fix rates shape indexing: [coeff, interval] -> [interval, coeff]
  in lohnsteuer.py and solidaritätszuschlag.py
- Add extend_extend_intervals_to_real_line to fix stale adjacent intervals
  after upsert_tree merges in param_functions
- Re-export extend_extend_intervals_to_real_line from gettsim.tt

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -277,7 +277,7 @@ environments:
       - pypi: https://files.pythonhosted.org/packages/05/5b/83e1ff87eb60ca706972f7e02e15c0b33396e7bdbd080069a5d1b53cf0d8/simplejson-3.20.2-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/32/46/9cb0e58b2deb7f82b84065f37f3bffeb12413f947f9388e4cac22c4621ce/sortedcontainers-2.4.0-py2.py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/65/44/bb509c3d2c0b5a87e7a5af1d5917a402a32ff026f777a6d7cb6990746cbb/tabcompleter-1.4.0-py3-none-any.whl
-      - pypi: git+https://github.com/ttsim-dev/ttsim.git?branch=refactor-piecewise-polynomial#801627eed2fe85abc7d8d2ca18fb038d63449737
+      - pypi: git+https://github.com/ttsim-dev/ttsim.git?branch=main#9bbba34d26703f5f8febffaa2ff7d4d8397f44ed
       - pypi: https://files.pythonhosted.org/packages/8d/96/04e7b441807b26b794da5b11e59ed7f83b2cf8af202bd7eba8ad2fa6046e/wadler_lindig-0.1.7-py3-none-any.whl
       - pypi: ./
       osx-64:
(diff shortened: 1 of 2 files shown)


MImmesberger:
@hmgaudecker When discussion GEP 8, we forgot about `reference` and `note` keys as well as the `update_previous` mechanic. I propose we add a `intervals` which stores the list of intervals and their coefficients. At the level of the `intervals` key, we can put `reference`, `note` and `update_previous`.

Regarding the behavior of `update_previous` specs:
- No gaps between intervals before and after updating
- Overlapping intervals in new and old spec are fine; new spec takes precedence and the old intervals adjust accordingly.

Any comments?

hmgaudecker:
> Any comments?

Yes, I'll need more context...

> Overlapping intervals in new and old spec are fine; new spec takes precedence and the old intervals adjust accordingly.

Sounds scary. I think we should be very strict here. No need to use the update_previous mechanic; it's just a couple of extra lines after all.

MImmesberger:
> Sounds scary. I think we should be very strict here. No need to use the update_previous mechanic; it's just a couple of extra lines after all.

Yeah, needed to implement it (and fail to anticipate test results) to see how error-prone this is.

The only place where we use `update_previous` for piecewise functions is for the income tax and Soli. Copy+Paste of the parameters that don't change isn't too bad there. 

> Yes, I'll need more context...

So the only difference compared the GEP version will be that we do

```yaml
2026-01-01:
    reference: ...
    intervals:
        - interval: "[0, 100)"
           slope: 0.5
        - ...
```

instead of 

```yaml
2026-01-01:
    - interval: "[0, 100)"
       slope: 0.5
    - ...
```

hmgaudecker:
That looks good, but I still don't quite get what you mean by:

> Overlapping intervals in new and old spec are fine; new spec takes precedence and the old intervals adjust accordingly.

Are overlapping intervals

```yaml
2025-01-01:
    - interval: "[0, 100)"
       slope: 0.5
    - interval: "[100, 200)"
       slope: 0.4
2026-01-01:
    - updates_previous: true
    - interval: "[50, 150)"
       slope: 0.4
```

or 

```yaml
2025-01-01:
    - updates_previous: true
    - interval: "[0, 100)"
       slope: 0.5
    - interval: "[100, 200)"
       slope: 0.4
2026-01-01:
    - updates_previous: true
    - interval: "[0, 100)"
       slope: 0.4
```

? 

The first would be terrible, I don't see why we would not support the second.

MImmesberger:
I was talking about the first one and I agree we should not do it. We don't really need the second one as of now, but wouldn't hurt to cover that case.

hmgaudecker:
We could use it in Bürgergeld, but I am not sure whether this case would be covered:

```yaml
      - interval: '[100, 1000)'
        slope: 0.2
```

-> 

```yaml
      - interval: '[100, 520)'
        slope: 0.2
      - interval: '[520, 1000)'
        slope: 0.3
```

MImmesberger:
My hunch is that we should not allow for that. Splitting up intervals doesn't occur often and a strict "no changes in boundaries" rule is easier to keep track of (I noticed a week ago that I start to forget about all the features we implemented last summer 😄). But feel free to overrule.

hmgaudecker:
No, Just thought it might have worked out is the box!

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule ".ai-instructions"]
 	path = .ai-instructions
-	url = git@github.com:OpenSourceEconomics/ai-instructions.git
+	url = https://github.com/OpenSourceEconomics/ai-instructions.git
```
MImmesberger, line 3: Needed to run tests via GHA at `gettsim-personas`.

PR #1160, opened 2026-03-21
Add coverage support for auto-vectorized policy functions

MImmesberger:
## Summary

- When `--cov` is passed to pytest, switches all `PolicyFunction`s from `"vectorize"` (AST transformation) to `"loop"` (`numpy.vectorize`) strategy. This makes `coverage.py` see the original function bodies instead of reporting 0%.
- Only active during coverage runs — normal test runs are unaffected.

The conftest fixture monkeypatches `PolicyFunction.vectorize` to use `dataclasses.replace(self, vectorization_strategy="loop")` before delegating to the original method.

Depends on ttsim-dev/ttsim#92, which fixes three issues that previously prevented the `"loop"` strategy from working:
1. `look_up()` and `piecewise_polynomial()` returned arrays for scalar input
2. The `"loop"` path returned a `numpy.vectorize` object instead of a regular function

Related to https://github.com/ttsim-dev/ttsim/issues/91.

## Test plan

- [x] `pixi run ty` — type checker passes
- [x] `pixi run prek run --all-files` — linting passes
- [x] `pixi run -e py314-jax tests -n 7` — all 1277 tests pass
- [x] `pixi run -e py314-jax pytest ... --cov=src/gettsim/germany/kindergeld -k kindergeld` — 93% coverage reported (was 0% before)


hmgaudecker (approved):
Thanks, this is very nice! 

I added stuff to make the reports truly useful -- #1161 lists the additional tests we'll need.

PR #1163, opened 2026-03-24
Compute Bürgergeld via Bedarfsanteilmethode

### LINKED ISSUE #1162: Fix per-person Bedarfsanteilsmethode distribution in ALG2/Bürgergeld
Our calculation of Bürgergeld/ALG 2 on the Bedarfsgemeinschaft level is off. In principle the transfer is an individual level transfer that considers BG-level income for its calculation. The **Bedarfsanteilmethode** is currently not implemented but is the key mechanism to compute individual-level claims to Bürgergeld (and is therefore important to compute "crazy" grouping IDs).

Here is a DAG:

Image

The Bedarfsanteilsmethode (§ 9 Abs. 2 Satz 3 SGB II) distributes income proportionally by each person's share of total **ungedeckter** Bedarf. Per § 11 Abs. 1 Satz 5 SGB II, children's income must first be netted against the child's own Bedarf. The remaining adults' income is then distributed by shares of "ungedeckter Bedarf":

**The BG total is unchanged** — only the per-person distribution shifts (more to adults, less to children with Kindergeld). For BGs without children or where children have no income, the result is identical.

Also, see the example calculations here:

[Link](https://www.ingolstadt.de/Leben/Arbeit-Jobcenter/Arbeitsuchend/)
[Musterberechnungsbogen_Bürgergeld.PDF](https://github.com/user-attachments/files/26216471/Musterberechnungsbogen_Burgergeld.PDF)

MImmesberger:

Closes #1162


hmgaudecker (approved):
Very nice!

PR #1167, opened 2026-03-25
No Grundsicherung im Alter if a child's annual Gesamteinkommen is 100,000 € or more (§ 43 Abs. 5 SGB XII, 2005–2019)

### LINKED ISSUE #1148: Implement 100k children exclusion for Grundsicherung im Alter (§43 Abs. 5 SGB XII)
### Problem

[SGB XII §43 Abs. 5](https://www.gesetze-im-internet.de/sgb_12/__43.html) excludes applicants from Grundsicherung im Alter if their children (first-degree descendants) have annual gross income exceeding 100,000 EUR:

> "Unterhaltsansprüche der Leistungsberechtigten [...] sind nicht zu berücksichtigen, sofern deren jährliches Gesamteinkommen im Sinne des § 16 des Vierten Buches unter einem Betrag von 100 000 Euro liegt. [...] Wird dieses überschritten, ist Grundsicherung im Alter [...] nicht zu leisten."

**Not implemented.** Currently no check exists.

### References

- [SGB XII §43 Abs. 5](https://www.gesetze-im-internet.de/sgb_12/__43.html)
- [SGB IV §16](https://www.gesetze-im-internet.de/sgb_4/__16.html): Definition of Gesamteinkommen

### Related issues

- Related to #1146, #1147

MImmesberger:
## Summary

- Implement the 100k children exclusion per § 43 Abs. 5 SGB XII (originally Abs. 2, BGBl. I 2003 S. 3022) for policy dates **2005–2019**: applicants are excluded from Grundsicherung im Alter if any first-degree descendant has an annual Gesamteinkommen (§ 16 SGB IV) of 100,000 EUR or more. The threshold applies to each child individually.
- The Angehörigen-Entlastungsgesetz (BGBl. I 2019 S. 2135) repealed the exclusion as of 2020-01-01. Since then the 100,000 EUR limit only restricts the Unterhaltsrückgriff of the Sozialhilfeträger against the children (§ 94 Abs. 1a SGB XII), which does not affect the benefit — `betrag_m` therefore no longer depends on children's income from 2020 on.
- Add `einkommensgrenze_kinder` YAML parameter (100,000 EUR/year, 2005–2019, expired with a note on the repeal).
- New policy functions: `hat_gesamteinkommen_über_kindeseinkommensgrenze`, `hat_kind_mit_einkommen_über_einkommensgrenze_als_elternteil_1/_2` (`AggType.ANY` aggregations over both `p_id_elternteil_*` slots), and `hat_kind_mit_einkommen_über_einkommensgrenze`.
- Gate `betrag_m` on the new check until 2019; separate `betrag_m` variant from 2020 on.
- Add a documentation page explaining both regimes (`docs/tt_explanations/childrens_income_grundsicherung_im_alter.md`).
- Policy cases for the exclusion live at 2018-07-01 (pre-repeal).
- Bump ttsim-backend to current main for the `any_by_p_id` aggregation.

Closes #1148


src/gettsim/germany/grundsicherung/im_alter/im_alter.py
```diff
@@ -261,3 +271,48 @@ def vermögensfreibetrag_eg(
...
+    einkommensgrenze_kinder: float,
+) -> bool:
+    """Whether a person's Gesamteinkommen exceeds the children's income threshold.
+
+    Reference: § 43 SGB XII (BGBl. I 2003 S. 3022)
+    """
+    return einkommensteuer__gesamteinkommen_y >= einkommensgrenze_kinder
+
+
+@agg_by_p_id_function(agg_type=AggType.SUM)
```
hmgaudecker: Don't we have `AggType.ANY` now?

src/gettsim/germany/grundsicherung/im_alter/im_alter.py
```diff
@@ -261,3 +271,48 @@ def vermögensfreibetrag_eg(
...
+def anzahl_kinder_mit_einkommen_über_einkommensgrenze_über_elternteil_1(
```
hmgaudecker:
```suggestion
def anzahl_kinder_mit_einkommen_über_einkommensgrenze_als_elternteil_1(
```
Name is wrong in any case. But the way `elternteil_x` is used here confused me at first. Usually we look from the child to the two parents, here we need to check all children from the parents' side, regardless of whether the parent is recorded as `elternteil_1` or `elternteil_2`. Would be good to encode that in the name.

src/gettsim/tests_germany/policy_cases/grundsicherung/im_alter/2026-01-01/kind_über_einkommensgrenze.yaml
```diff
@@ -0,0 +1,298 @@
+---
+info:
+  note: |-
+    Retired couple (Person0 age 70, Person1 age 68), both past Regelaltersgrenze, no
+    income. Person2 is their adult child (age 40) living in a separate household with
+    annual gross income of 120,000 EUR (Gesamtbetrag der Einkünfte > 100,000 EUR).
```
hmgaudecker:
```suggestion
    monthly gross earnings of 10,000 EUR (annual Gesamtbetrag der Einkünfte > 100,000 EUR).
```
Or change `bruttolohn_m` -> `bruttolohn_y`, but I was skimming through the test looking for 120k... In any case, use earnings not income.

src/gettsim/tests_germany/policy_cases/grundsicherung/im_alter/2026-01-01/kind_unter_einkommensgrenze.yaml
```diff
@@ -0,0 +1,298 @@
+---
+info:
+  note: |-
+    Retired couple (Person0 age 70, Person1 age 68), both past Regelaltersgrenze, no
+    income. Person2 is their adult child (age 40) living in a separate household with
+    annual gross income of 72,000 EUR (Gesamtbetrag der Einkünfte < 100,000 EUR).
```
hmgaudecker: See comment in sibling file.

hmgaudecker (commented):
Just wondering whether the rule is really that simple cutoff? No exceptions, hardship, ... ?

MImmesberger:
> Just wondering whether the rule is really that simple cutoff? No exceptions, hardship, ... ?

Good that you push back. Apparently since 2020 parents can actually receive Grundsicherung im Alter if their children earn more than 100k annually (before 2020 it was indeed a restriction that excluded the parents from receiving the transfer; no hardship rules, at least in the SGB). Today, parents receive Grundsicherung im Alter, but the Sozialamt can reach out to the children to get that money back afterwards if they are more than 100k.

AFAICT the Sozialämter usually do this but since it is not an automatically applied rule, I'd say we don't model if past 2019.

I added an explanation to GETTSIM's documentation and the corresponding functions.

hmgaudecker:
Could you please fix the test-jax errors before I look into this? Thanks!

hmgaudecker:
> but the Sozialamt can reach out to the children to get that money back afterwards if they are more than 100k. AFAICT the Sozialämter usually do this

In this case, I think we should add an input variable `sozialamt_rückgriff_auf_kindereinkommen: bool` (or so). We don't want to have a hard policy break when in practice, there often is none.

hmgaudecker:
Actually, the more important reason: We want to be able to represent the clawback to children's income. The situation is not quite symmetric to take-up because we can always leave out a target and it won't be computed. If we left out the possibility here, we would not be able to represent it.

MImmesberger:
I'd prefer to postpone a proper treatment. It is complex because
- children are obliged to pay Unterhalt according to the Düsseldorfer Tabelle; in that sense the problem is similar to determining parental Unterhalt for children
- the Sozialhilfeträger's claim can never exceed the minimum of {individuelle Unterhaltsverpflichtung, geleistete Sozialhilfe, individuelle Leistungsfähigkeit}
- Especially in the presence of multiple children its not trivial how the Amt distributes their claim across children. To me it seems like the children below earnings of 100k are relevant too (their liability share is not lost but the Amt takes care of it).

We could ofc provide a target `maximaler_unterhaltsrückgriff_des_sozialamts` which equals the parental Grundsicherung im Alter. But I'd be afraid of giving users a false sense of security regarding the implementation. 

If you agree, I'll post the corresponding issue.

hmgaudecker:
Agree, but I think I prefer keeping the pre-2020 rule as the default then despite this going against our philosophy. Again, easier to turn that off than to turn Rückgriff on if we don't model it.

hmgaudecker (approved):
Thanks!

PR #1172, opened 2026-04-22
Add Wohngeld Basisanspruch to income considered for Wohngeld Mindesteinkommen check

### LINKED ISSUE #1171: Wohngeld: Mindesteinkommen-Check excludes Wohngeld itself (and possibly other income types per § 15 WoGG)
## Problem

The Mindesteinkommen check for Wohngeld (`einkommen_für_mindesteinkommen_m` in `src/gettsim/germany/wohngeld/voraussetzungen.py`) currently sums only:

- `arbeitslosengeld_2__nettoeinkommen_vor_abzug_freibetrag_m` (or `bürgergeld__…` from 2023)
- `unterhalt__tatsächlich_erhaltener_betrag_m`
- `unterhaltsvorschuss__betrag_m`
- `kindergeld__betrag_m`
- `kinderzuschlag__anspruchshöhe_m`

This sum is compared against the ALG2/Bürgergeld-Regelbedarf to determine whether the household reaches the Mindesteinkommen and thus qualifies for Wohngeld instead of being shunted to SGB II/XII.

**The would-be Wohngeld itself is not part of the sum.**

## Why this is wrong

Per VwV 15.22 to § 15 WoGG, the plausibility check asks whether the household's income *including the anticipated Wohngeld* ("voraussichtlich zu gewährende Wohngeld") covers the ALG2/Bürgergeld-Bedarf. Omitting Wohngeld flips borderline cases incorrectly: households that would only need SGB II top-up *without* Wohngeld, but would be self-sufficient *with* Wohngeld, are currently treated as ineligible.

Reference: <https://www.wohngeld.org/wohngeldgesetz-wogg/paragraph15/>

## Scope / follow-up

While auditing this, the full list of income types in § 15 WoGG / VwV 15.01 ff should be reconciled against what we currently include. The ALG2/Bürgergeld net-income aggregate already covers labor, self-employment, rental, capital, pensions, ALG I, and Elterngeld, so those appear fine — but the Wohngeld component is missing and should be added.

Tests under `src/gettsim/tests_germany/policy_cases/wohngeld/.../mindesteinkommen.yaml` should be extended with a case where Wohngeld itself is what tips the household over the Regelbedarf threshold.

MImmesberger:

Closes #1171


hmgaudecker (approved):
Very nice!

PR #1178, opened 2026-04-29
Implement § 17a WoGG Grundrentenfreibetrag in Wohngeld

MImmesberger:
Income for Wohngeld so far ignored the § 17a WoGG allowance for recipients of public pensions with at least 33 years of Grundrentenzeiten (introduced by Art. 4 Grundrentengesetz, effective 2021-01-01). The deduction is identical in formula to § 82a SGB XII: 100 € base plus 30 % of pension income above 100 €, capped at 50 % of Regelbedarfsstufe 1.

Adds the parameter `anrechnungsfreier_anteil_gesetzliche_rente` to wohngeld/einkommen.yaml and a new `freibetrag_m_ab_2021` policy function; the prior `freibetrag_m_ab_2016` is retained as `freibetrag_m_2016_2020`.


src/gettsim/tests_germany/policy_cases/wohngeld/2006-01-01/hh_id_15.yaml
```diff
@@ -96,16 +96,6 @@ inputs:
               - 0.0
               - 0.0
               - 0.0
-            grundrente:
```
MImmesberger, line 99: Some prior PR added these in the wrong namespace. Removing all of these

src/gettsim/germany/wohngeld/einkommen.py
```diff
@@ -231,6 +237,53 @@ def freibetrag_m_ab_2016(
...
+    anrechnungsfreier_anteil_gesetzliche_rente: PiecewisePolynomialParamValue,
+    grundsicherung__regelbedarfsstufen: Regelbedarfsstufen,
+    xnp: ModuleType,
+) -> float:
+    """Freibeträge on Einkommen relevant for Wohngeld calculation from 2021 onwards.
+
+    Wohngeld-Freibeträge inklusive § 17a WoGG (Grundrentenfreibetrag).
+    """
+    freibetrag_bei_behinderung = (
+        freibetrag_bei_behinderung_pauschal_m if behinderungsgrad > 0 else 0
```
hmgaudecker, line 267: does vectorisation allow for these guys nowadays? I distantly remember it did not at some point, but my memory might be off.

src/gettsim/germany/wohngeld/einkommen.py
```diff
@@ -231,6 +237,53 @@ def freibetrag_m_ab_2016(
...
+        freibetrag_kinder = min(
+            einnahmen__bruttolohn_m,
+            freibetrag_kinder_m["arbeitendes_kind"],
+        )
+    elif familie__alleinerziehend:
+        freibetrag_kinder = freibetrag_kinder_m["alleinerziehend"]
+    else:
+        freibetrag_kinder = 0.0
+
+    if sozialversicherung__rente__grundrente__grundsätzlich_anspruchsberechtigt:
```
hmgaudecker: Is that before or after taking it up?

src/gettsim/tests_germany/policy_cases/wohngeld/2006-01-01/hh_id_15.yaml
hmgaudecker: Oh, didn't we have some parameter `strict=True` or so for inputs in order to catch such cases? This shows that it would be useful to turn that on / get it back.

hmgaudecker (commented):
Very nice! Just a couple of smaller questions.

src/gettsim/germany/wohngeld/einkommen.py
MImmesberger: Yes I remember that as well, but apparently it works via `_ifexp_to_call`.

src/gettsim/germany/wohngeld/einkommen.py
MImmesberger: It's independent of actually taking up a DRV pension if that's what you're after.

hmgaudecker: Fine, but how would the WoG-Stelle check that? Difficult decision in any case.

src/gettsim/germany/wohngeld/einkommen.py
hmgaudecker: Excellent! Once the big gsia PR is gone (but before you have another pint), we should turn on the corresponding ruff rule and see what happens. Would make the code easier to read in many cases. And we should check whether we need to adjust docs.

src/gettsim/tests_germany/policy_cases/wohngeld/2006-01-01/hh_id_15.yaml
MImmesberger: Yes, see new TTSIM issue

src/gettsim/germany/wohngeld/einkommen.py
MImmesberger: #1195 + TTSIM PR

src/gettsim/germany/wohngeld/einkommen.py
MImmesberger: I'm sorry I remembered the wrong shortcut. The correct one is: It doesn't matter whether the person actually receives Grundrente. The Wohngeldstelle knows because the applicant needs to hand in their Rentenbescheid.

I added an `bezieht_rente` gate. This doesn't change the results (because with 0 pensions, the Freibetrag is 0 as well) but hopefully makes it clearer what is going on for future readers.

hmgaudecker:
Also here, jax tests fail.

hmgaudecker (approved):
Great, thanks a lot!

PR #1186, opened 2026-05-24
GEP 09 - Runtime checking via beartype

hmgaudecker:


docs/geps/gep-04.md
```diff
@@ -354,6 +354,145 @@ both ways without changing quantities. In case more complex conversions are need
...
+—and let `ty` check the body against the same scalar contract the author had in mind.
+The auto-vectorised wrapper that the DAG actually calls is generated separately; see
+{ref}`gep-9` for the layered inner/outer wrapper pattern that strips the user's scalar
+annotations on the inner runtime executor and stamps a synthesised column-typed
+signature on the outer beartype-checked forwarder.
+
+### Build-time annotation synthesis
+
+The auto-vectorised wrapper described in {ref}`gep-9` carries a synthesised column-typed
+signature stamped at DAG-build time, so the runtime claw checks columns against column
```
MImmesberger: `runtime claw`: We should specify what that is exactly

docs/geps/gep-04.md
```diff
@@ -354,6 +354,145 @@ both ways without changing quantities. In case more complex conversions are need
...
+types. Two kinds of DAG node are generated rather than written by hand, and both need a
+concrete type to take part in the DAG's annotation-consistency check:
+
+- **auto-vectorised wrappers** — the column-operating wrapper that the DAG calls in
+  place of a scalar policy function;
+- **aggregation wrappers** — the `grouped_*` / `*_by_p_id` primitive adapted to the
+  DAG's argument names for a group sum (`my_col_hh`) or a person-pointer aggregation.
+
+The aggregation primitives are the harder case. Each is declared with a stack of
```
MImmesberger, line 457: I think this is too much implementation detail. Very complex and distracts from the fundamental decision the community has to make

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+- * Created
+  * 2026-05-23
+- * Resolution
+  * Pending — see [#GEPs](https://gettsim.zulipchat.com/#narrow/stream/309998-GEPs/topic/GEP.2009)
+```
+
+## Abstract
+
+This GEP introduces runtime type checking across `ttsim`, `gettsim`, and
+`gettsim-personas` via [beartype](https://beartype.readthedocs.io)'s package claw, and
```
MImmesberger: I heard the term `claw` for the first time in a context unrelated to OpenClaw when you introduced beartype. This term feels ambiguous, but if it is a set term for type checkers we can keep it.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+  * Draft
+- * Type
+  * Standards Track
+- * Created
+  * 2026-05-23
+- * Resolution
+  * Pending — see [#GEPs](https://gettsim.zulipchat.com/#narrow/stream/309998-GEPs/topic/GEP.2009)
+```
+
+## Abstract
```
MImmesberger, line 18: The abstract doesn't really make clear what beartype buys us. 

I think it should mention these points:

- Currently, there is not a lot of type checking when using GETTSIM. This led to dangerous silent bugs in the past and unreadable errors.
- This GEP proposes to add beartype as a runtime type checker that automatically verifies user's input data against the expected input types.
- To do this, GETTSIM/TTSIM are becoming explicit about the data the packages expect and type errors are standardized. 
- When writing GETTSIM policy functions, type annotations will become mandatory

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+formalises the two-tier type vocabulary that the claw makes enforceable: a wide `UserX`
+family at the user boundary and a narrow canonical `X` family inside the package. Each
+user-facing entry point and decorator is wrapped with an explicit beartype configuration
+that re-raises type-violations as a documented subclass of `TTSIMError`. Auto-vectorized
+policy-function wrappers have their inherited scalar annotations stripped so that the
+claw checks the wrapper against its true (column) signature.
+
+## Motivation and Scope
+
+The ttsim DAG accepts a wide range of objects at its outer boundary — pandas Series,
```
MImmesberger:
```suggestion
The TTSIM DAG accepts a wide range of objects at its outer boundary — pandas Series,
```

Also everywhere else

MImmesberger:
```suggestion
The ttsim DAG accepts a wide range of objects as inputs — pandas Series,
```

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+numpy arrays, Python scalars, JAX arrays — and converts them internally into a narrower,
+performance-oriented representation: jaxtyping-shaped JAX or numpy arrays for columns
+and Python or numpy scalars for parameters. Today this distinction is implicit.
+`typing.py` exposes a single `Array`-based vocabulary, the canonical internal types are
+not named separately from their user-friendly supersets, and nothing enforces either
+contract at runtime. Three problems follow:
```
MImmesberger: I think contract is an informal term in that context (also in GEP 4). Maybe "Nothing enforces type compliance at runtime".

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+
+1. **Silent contract drift.** A policy function annotated `int` that is invoked with a
+   `jax.Array` works today, fails tomorrow on a backend swap, and has no guard at the
```
MImmesberger: The point being that swapping the backend means input types need to change? If so, we should write this more clearly.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+   boundary that would surface the mismatch with a useful error. Annotations are
+   documentation, not specification.
+
+1. **No single boundary for canonicalisation.** Every code path that takes user input
```
MImmesberger: We should avoid "boundary", too technical IMO.

Also, I don't see why this point here should be raised in a GEP when it sounds like lazy architecture.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+   re-implements the cast from "pandas Series or Python float or numpy scalar" to "JAX
+   float column / numpy float scalar". The conversion is scattered, inconsistent, and
+   impossible to type-check.
+
+1. **Indistinguishable bug classes.** When a TT DAG raises `TypeError`, the user cannot
+   tell whether they passed bad data, mis-declared a policy function, or hit an internal
+   ttsim bug. There is no exception vocabulary that maps to architectural layers.
+
```
MImmesberger, line 67: Add another point that points to [TTSIM #97](https://github.com/ttsim-dev/ttsim/issues/97). Missing type checkers can lead to dangerous bugs.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+### Same runtime, more discoverable failures
+
+The claw adds an O(n) container check on entry to every clawed function, but ttsim's
+entry points are called rarely (per-run, not per-row), so the cost is invisible at the
+boundary.
+
+## Backward Compatibility
+
+User-facing public API is unchanged. Anyone whose code raised `TypeError`, `ValueError`,
```
MImmesberger: Not true, users need to add type hints to the policy functions they have written themselves

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+or a bare `Exception` from inside ttsim will now see a `TTSIMError` subclass instead.
+Code that catches `Exception` keeps working; code that catches narrow built-in
+exceptions will have to broaden to `TTSIMError` (or the relevant subclass). Two
+pre-existing exception types are hoisted into the hierarchy without changing their
+definition site: `ConflictingActivePeriodsError` and `TranslateToVectorizableError`.
```
MImmesberger: Implementation detail I'd say. But no strong opinion.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+Both keep their original import path.
+
+Internal callers that relied on the wide-form types (passing a `pd.Series` into a
```
MImmesberger: I don't get what this paragraph is supposed to tell me.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+UserScalarFloat: TypeAlias = float | int | np.floating | np.integer
+UserScalarInt: TypeAlias = int | np.integer
+UserScalarBool: TypeAlias = bool | np.bool_
+```
+
+The column aliases use the `Array | np.ndarray` union so the same vocabulary covers both
+backends. This is the single source of truth for column shapes; callers do not branch on
+the backend.
+
+The aliases live at module top level, not under `if TYPE_CHECKING`. The claw needs them
```
MImmesberger, line 176: This can go

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+at runtime to rewrite call sites.
+
+The wide forms are restricted to the user boundary. Inside ttsim, the narrow forms are
```
MImmesberger: Maybe I'm missing something but this doesn't seem to be important for GEP readers

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+
+beartype_package("ttsim", conf=INTERNAL_CONF)
+
+# ...remaining imports
+```
+
+`beartype_package` installs an AST rewriter against the package's import hook. Every
+subsequent `import ttsim.*` produces a module whose annotated callables wrap themselves
+in a beartype check on load. There is no per-file decorator, no opt-in list, and no way
+to forget a function. `gettsim` and `gettsim-personas` do the same with their own root
```
MImmesberger: Didn't realise that when reviewing the TTSIM PR. Could this be an issue when users make isinstance checks or do some advanced things with `PolicyFunction` objects? Need to watch out for that.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+place across `ttsim`, `gettsim`, `gettsim-personas`, and `pylcm`. Every pixi test
+environment runs with the claw on; a missing or mistaken annotation is a build break.
+
+`.ai-instructions/modules/beartype.md` documents the conventions contributors follow:
+when to use `UserX` vs `X`, how to add a new boundary decorator, the wrapper-annotation
+rule, the claw-and-callable-instance gotcha, and the diagnostic workflow when a beartype
+violation surfaces. The module is included in the `tier-a` profile by default so every
+agent picks it up.
+
+## Alternatives
```
MImmesberger, line 469: I think all of these alternatives are rightfully not implemented. However, the real alternative is to write custom type checks as in the old days. 

Another alternative is not to enforce type annotations for user-written functions, forcing the type checker to be more lenient. I would emphasise that this is an option.

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+### The auto-vectorized-wrapper annotation problem
+
+Scalar policy functions are wrapped at DAG-build time by `ttsim.tt.vectorization` and
+`ttsim.tt.rounding`. The wrapper closes over the user function but is itself called on
+columns. If the wrapper inherits the user function's scalar annotations via
+`functools.wraps`, the claw checks column inputs against scalar annotations and rejects
+every legitimate call.
+
+The fix uses two layers. The **inner** runtime executor—`numpy.vectorize`, the
+AST-rewrite output, or the rounding callable—wraps the user function with
```
MImmesberger:
```suggestion
AST-rewrite output, or the rounding callable — wraps the user function with
```

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+
+### The auto-vectorized-wrapper annotation problem
+
+Scalar policy functions are wrapped at DAG-build time by `ttsim.tt.vectorization` and
+`ttsim.tt.rounding`. The wrapper closes over the user function but is itself called on
+columns. If the wrapper inherits the user function's scalar annotations via
+`functools.wraps`, the claw checks column inputs against scalar annotations and rejects
+every legitimate call.
+
+The fix uses two layers. The **inner** runtime executor—`numpy.vectorize`, the
```
MImmesberger:
```suggestion
The fix uses two layers. The **inner** runtime executor — `numpy.vectorize`, the
```

docs/geps/gep-09.md
```diff
@@ -0,0 +1,454 @@
...
+    return anzahl * satz
+```
+
+and accidentally calling it with `vectorization_strategy="not_required"` raises
+`PolicyFunctionDefinitionError` at decoration time — the scalar annotations are
+incompatible with column-direct execution.
+
+### Wider boundary types, narrower internal types
+
+Code that takes user input declares its parameters with `UserX` aliases. The same code's
```
MImmesberger: Rewrite to make clear what `UserX` is.
- We allow a broad selection of input types. The collection of these types is named `UserX` where X can be "FloatColumn", "IntColumn" etc. 
- We convert these types (via `_canonicalize_*` functions) to an internal representation with a much narrower type collection (e.g. pd.Series with uints is converted to...). The collection of these types is named `X`.
- This formalises how we think about types when using TTSIM, but does not change the API in any way.

docs/geps/gep-04.md
hmgaudecker: Replaced with "runtime type checker" — the deep mechanism stays in GEP 9 and `beartype.md` where the term `claw` is defined.

docs/geps/gep-04.md
hmgaudecker: I'd keep this section. The rule table for how each aggregation kind types its output (sum keeps float/int and promotes bool to int; mean is always float; min/max keep float/int; any/all produce bool; count is always int) **is** the spec for what derived nodes look like. Without it, the GEP says "derived nodes have concrete types" without committing to *which* concrete types — the alternative being the widened `FloatColumn | IntColumn` union from the `@overload` stack, which is a real and weaker option we are deliberately not picking. The rule belongs in the GEP because changing it would require changing the GEP; the `@overload` cross-check test is the part that's implementation, but the rule itself is architectural.

docs/geps/gep-09.md
hmgaudecker: Defined in the new Terminology subsection: "beartype's import-hook mechanism: a single `beartype_package()` call automatically applies `@beartype` to every annotated function in the package." Kept the term itself because it's `beartype`'s own vocabulary and the document refers to the mechanism often.

*ps - At least in the software ecosystem, bears with claws have been around for much longer than lobsters despite the recent gain in popularity of the latter...*

docs/geps/gep-09.md
hmgaudecker: Done. Rewrote the abstract along your four bullets (silent bugs today → beartype runtime checker → explicit type vocabulary + canonicalisation → annotations on user policy functions) and added a Terminology subsection right after defining `claw`, `runtime check`, and `boundary` so the rest of the document is readable without prior beartype exposure.

docs/geps/gep-09.md
hmgaudecker: Thx, applied everywhere.

docs/geps/gep-09.md
hmgaudecker: Replaced "nothing enforces either contract at runtime" with "no runtime check guarantees compliance." The term `contract` is still defined in the Terminology box for the few places where the document uses it later.

docs/geps/gep-09.md
hmgaudecker: Done.

docs/geps/gep-09.md
hmgaudecker: Rewrote the silent-drift example to drop the backend-swap phrasing: "A policy function annotated `int` that is invoked with a `jax.Array` runs silently today; the mismatch only surfaces if a downstream JAX-specific operation fails on the unexpected dtype, often far from the offending input."

docs/geps/gep-09.md
hmgaudecker: 1. `boundary` is now defined in the Terminology box ("a function the user calls directly: `main()`, `InputData.df_and_mapper`, `@policy_function`, …").
2. On the substantive concern: rewrote the motivation point to read as a factual observation rather than a complaint — "Every code path that takes user input re-implements the cast from `pd.Series` or Python scalar or numpy scalar to a canonical numpy / JAX form. The conversions are scattered, and nothing enforces that they agree." The point belongs because the GEP's proposal *is* to funnel those conversions through named `_canonicalize_*` helpers; naming the scattered pre-state is what motivates that.

docs/geps/gep-09.md
hmgaudecker: Added a fourth motivation bullet: "Past silent bugs. Missing or imprecise type checks have caused real,
hard-to-diagnose bugs (e.g. [TTSIM #97](https://github.com/ttsim-dev/ttsim/issues/97))."

docs/geps/gep-09.md
hmgaudecker: You're right, and I conflated two decorators in my first pass. The actual pre-GEP situation was uneven:

- `@policy_function` did **not** access annotations at decoration time. Missing annotations were silently tolerated — vectorization synthesises a wide default via `dags.get_annotations(func, default="IntColumn | FloatColumn | BoolColumn")` and the function compiles and runs with the imprecise default. The change here is "silent tolerance → friendly `PolicyFunctionDefinitionError`", which is sharper than I made it sound.
- `@policy_input` did access `func.__annotations__["return"]` directly and would `KeyError` if missing. That's the "obscure later error → friendly error" case.
- `@param_function`, `@agg_by_group_function`, `@agg_by_p_id_function`, `@group_creation_function` fall somewhere in between.

Rewrote Backward Compatibility to spell this out: "Pre-GEP, missing annotations were tolerated unevenly — `@policy_function` silently fell back to a wide default union via `dags`, while `@policy_input` raised a `KeyError` at decoration time. Post-GEP, all five `@*_function` decorators raise their matching `*DefinitionError` at decoration time, so the convention is enforced uniformly." Practical impact on existing `gettsim` code is still zero (every policy function is fully annotated) but the spec change is honest.

docs/geps/gep-09.md
hmgaudecker: I'd keep this. The two class names that users may have caught (`ConflictingActivePeriodsError`, `TranslateToVectorizableError`) are part of the public surface, and "they keep their import path" is exactly what someone with `try/except ConflictingActivePeriodsError:` in their pipeline needs to read. Happy to move it to a footnote if you'd prefer that instead.

docs/geps/gep-09.md
hmgaudecker: Collapsed to one sentence: "Internal code that passes wide types into narrow-typed functions surfaces as `BeartypeCallHintViolation` from the package-wide claw. These are pre-existing ttsim bugs to fix at the call site, not user-facing changes."

docs/geps/gep-09.md
hmgaudecker: I'd keep it. The reason aliases must live at module top level (not under `if TYPE_CHECKING`) is the kind of thing a future maintainer adding a new column alias needs to know upfront. Without it, the next person adds an alias under `TYPE_CHECKING`, breaks the claw at decoration time, and spends an afternoon figuring out why. GEPs being technical documents, this kind of forward-looking constraint feels in scope.

docs/geps/gep-09.md
hmgaudecker: I'd keep this too. After all, GEP's are our technical design documents. The rule "wide forms are restricted to the user boundary; internal code uses narrow" is the spec for the type split. Without it, the GEP only says "we have `UserX` and `X` aliases" without committing to where each is used. The follow-on "conversions are funnelled through `_canonicalize_*` helpers" is the spec that prevents wide types from leaking inward. That is precisely the property that the wide/narrow split is supposed to buy.

docs/geps/gep-09.md
hmgaudecker: Good catch. I didn't appreciate the user-visible angle. Added a Limitations subsection at the end of "Detailed Description": the binding does happen under normal `import`, but the standard policy-module loader (`orig_policy_objects.py` → `spec_from_file_location`) is claw-free, so isinstance checks survive for the standard load path. The narrow bite case (`from my_pkg.policy_module import my_fn` and then `isinstance(my_fn, PolicyFunction)`) is documented along with the workaround.

docs/geps/gep-09.md
hmgaudecker: Added two alternatives:

1. *Custom per-function type checks instead of runtime checking* — the pre-GEP `_fail_if_*` approach is what we had. Stays valid in principle, doesn't scale: each check requires editing a separate file and remembering to wire it in. The claw eliminates the wiring step.
  
2. *Leave annotations on user-written functions optional* — `_fail_if_missing_annotations` could downgrade to a warning and beartype would silently skip un-annotated parameters. Rejected because the goal degrades into "some functions are checked, some are not" with no in-band signal to the reader whether the missing check is deliberate or accidental. Since every existing GETTSIM policy function already carries full annotations, the strict policy enforces an existing convention, not a new requirement.

docs/geps/gep-09.md
hmgaudecker: Applied.

docs/geps/gep-09.md
hmgaudecker: Restructured the "Wider boundary types, narrower internal types" subsection with your three-bullet shape: `UserX` is the wide vocabulary; `_canonicalize_*` produces `X`; this formalises how we reason about types and pins each conversion to one named function per boundary, without changing the public API in shape.

docs/geps/gep-09.md
hmgaudecker: Applied.

MImmesberger (approved):

PR #1193, opened 2026-06-03
GEP 10: Units and Dimensionality

### LINKED ISSUE #1192: pint units: gettsim rollout (annotate everything, switch check ON)
## What to build

The full gettsim unit rollout in one PR, organized into per-namespace commits for reviewability. Annotate every policy function, policy input, and param with units (filling the params that currently have no unit), tag every bare literal in mixed-unit arithmetic, apply the [count] dimension where appropriate, and switch the always-on unit check on. Store historical values in their legal currency (e.g. DM) so the build-time conversion removes the manual DM-to-EUR conversions (gettsim #1174). After this PR every gettsim node is unit-checked.

## Acceptance criteria

- [ ] every function/input/param carries a unit; no null units remain
- [ ] historical DM values stored in DM with reference_period; build converts to the run currency
- [ ] all bare literals in mixed-unit arithmetic tagged
- [ ] always-on unit check enabled; full test suite green
- [ ] commits organized per namespace for review

## Blocked by

- https://github.com/ttsim-dev/gettsim/issues/1191

MImmesberger:
**GEP 10 — Units and Dimensionality**

Adopts [pint](https://pint.readthedocs.io) to add, in the spirit of GEP 9:

- **data-independent dimensionality checks** across the DAG: per-function bodies run on synthetic `Quantity`s plus an edge-consistency check at build time;
- **automatic currency resolution (DM↔EUR)**: historical values stored in their legal currency, converted to the run currency at build time;
- a **pint-backed reimplementation of the time-conversion arithmetic**, keeping GEP-1 suffix automation unchanged.

pint runs only at environment-build time and at the input boundary; it never wraps a live array, so the JAX/NumPy runtime and the GEP-9 type vocabulary are untouched. Parameters must name their legal currency; columns and functions use only agnostic `CURRENCY_*` units — that is what makes them provably currency-agnostic.

**Implementation**: ttsim-dev/ttsim#138 (infrastructure) + ttsim-dev/ttsim#141 (mettsim annotated end-to-end as the worked example); gettsim rollout is #1212 (closes #1192). No ttsim PR merges before the gettsim rollout is ready.


docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+enumeration — a `Unit` `StrEnum` shipped by `ttsim`, spelled identically in code (e.g.
+`unit=Unit.HOURS_FLOW`) and in YAML (e.g. `unit: HOURS_FLOW`):
+
+| token                            | resolves to                      | typical use              |
+| -------------------------------- | -------------------------------- | ------------------------ |
+| `CURRENCY_FLOW`                  | `CURRENCY / period`              | wages, claims, benefits  |
+| `CURRENCY`                       | `CURRENCY`                       | wealth, asset thresholds |
+| `DIMENSIONLESS`                  | `dimensionless`                  | shares, rates, counts    |
+| `DIMENSIONLESS_FLOW`             | `1 / period`                     | Zugangsfaktor per year   |
+| `YEARS`                          | `year`                           | ages, durations          |
```
hmgaudecker: **Calendar years and durations are different objects.** `YEARS` is used here for ages/durations *and* calendar years, but a calendar year is an affine point/ordinal, not a duration — they don't share the same arithmetic. Real pension code does `y_to_m(geburtsjahr) + (geburtsmonat-1)`: tag the years `YEARS` and the dry-run rejects `year + dimensionless`; tag them `DIMENSIONLESS` and the whole body goes dimensionless and falls back to the declaration unchecked. A distinct calendar-point/ordinal type plus a unit-aware date-difference helper, or explicitly scoping these bodies out, would resolve it.

Minor (same table): `HECTARES` exists in the `Unit` enum and the JSON schema but is missing from this vocabulary table.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+```yaml
+beitragssatz:
+  unit: DIMENSIONLESS   # a rate is dimensionless
+  reference_period: null
+  type: scalar
+  2024-01-01:
+    value: 0.013
+```
+
+**There are no exemptions** — every active node has a unit. Most nodes declare it
```
hmgaudecker: **Boolean-returning bodies are skipped, so the comparisons inside them go unchecked.** This says there are no exemptions and booleans are dimensionless nodes, but the environment checker skips boolean-output bodies entirely (`unit_checks.py:244`), so dimensional comparisons inside eligibility functions (e.g. `CURRENCY/month > CURRENCY`) are never checked — and most mettsim eligibility functions sit in that blind spot. The miss can change decisions, not just reported levels. Running boolean bodies through the explorer and accepting only a dimensionless final result would close it.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+
+**Parameters must be concrete; functions must be agnostic.** A parameter's numbers are
+written in *some* currency, so once a concrete currency is registered, an agnostic
+`CURRENCY_*` token on a parameter is a build error — the declaration must name the
+denomination (`EUR_FLOW`, not `CURRENCY_FLOW`). Columns and functions may *only* declare
+agnostic tokens.
+
+**The run currency.** The `currency` argument to `main()` defaults to the registered
+base currency; it is the currency the input data is taken to be in and that the outputs
+come out in. At environment build, every currency-denominated *parameter* is converted
```
hmgaudecker: **`require_converter` values are converted by a single uniform factor, which is wrong for polynomial coefficients.** 'every currency-denominated parameter is converted' holds, but `require_converter` is routed through the flat-dict path and scales every numeric leaf by the same factor (`policy_environment.py:313`). For a structured converter holding polynomial coefficients (the income-tax tariff is `require_converter`, not `piecewise_quadratic`), the order-`j` coefficient must scale by `f_out/f_in**j`; uniform scaling is dimensionally wrong (with `f_in=f_out=0.25`, `{0,1,0.5}`→`{0,0.25,0.125}` gives 84.375 instead of 1275 at x=25). EUR-only today (factor 1 hides it). The single `unit:` token also can't express a mapping's domain≠codomain, and the consumer opts out, so the tariff sits outside Layer-1.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+`1`, and `befreit` as a boolean stand-in. `einkommen_m * satz` is a flow times a
+dimensionless number, so it stays `EUR/month` — matching the declaration; the
+`mindestbetrag_m` arm matches too.
+
+**Every branch is covered, by re-running.** To evaluate `if befreit:` Python needs a
+yes/no, but a unit stand-in has no value to compare. So the stand-in intercepts the
+*truth test* itself (Python's `__bool__`) and hands it to a small driver — the **path
+explorer** — that decides which way to go, re-running the body and steering the open
+branches differently each time (a depth-first walk of the decision tree, in the style of
+*concolic* execution) until every reachable path is driven. The number of runs is the
```
hmgaudecker: **'Number of runs = number of reachable paths' is not what the explorer does.** It records booleans but no path constraints, so it visits infeasible combinations (`x>y` then `x<=y`) → possible false positive; and at the 1024-path cap (`unit_checks.py:931`) it only warns and returns success, so a wrong-unit branch first reached at path 1025 passes silently. Either hard-fail/opt-out on truncation, or describe this as syntactic branch-combination coverage.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+- **a result that comes out dimensionless** — a body inferring a dimensionless value (an
+  early `return 0.0`, or arithmetic that cancels) falls back to the declaration;
+
+**A body the dry-run cannot evaluate must opt out explicitly.** The dry-run executes a
+*scalar* body symbolically, so a body it cannot trace must opt out: vectorized functions
+(`vectorization_strategy="not_required"`, no scalar form for pint to walk), piecewise
+polynomials and lookup tables (evaluated by table machinery), and bodies calling `join`
+or a raw `xnp` op. Rather than silently trusting such a body, the check **rejects** it
+unless the author marks it `verify_units=False` on the decorator. The opt-out is of body
+*inference only*: the declared unit still stands and the body's edges are still checked.
```
hmgaudecker: **An opt-out does not keep the incoming edges checked.** 'the body's edges are still checked' overstates it: there is no independent consumer-side input contract, so a `verify_units=False` body's incoming edges are unchecked, and `ParamMappingObject` exposes only `output_unit` to the environment (`unit_checks.py:246,184`) — a `HECTARES` schedule can be evaluated at a currency input with no error. Narrowing this to 'output declaration + downstream checked uses', or adding per-argument input contracts, would match the implementation.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+derive their token from the source and the aggregation type, paralleling how
+{ref}`GEP 4 <gep-4>` resolves their types. `SUM`/`MEAN`/`MIN`/`MAX` preserve the source
+token; `COUNT` is a head count and is `DIMENSIONLESS`; `ANY`/`ALL` yield a boolean (a
+dimensionless quantity) and so auto-assign `DIMENSIONLESS` (as does a `SUM` over a
+boolean column — a head count). A `@group_creation_function` group id is auto-assigned
+`DIMENSIONLESS` (it is an identifier, and the decorator exposes no `unit=`). Where the
+source's token pins down a concrete currency (a parameter), the derived node inherits
+the **agnostic counterpart**.
+
+### Literals
```
hmgaudecker: **Bare non-zero literals in comparisons silently carry a dimension.** Ordering comparisons ignore every bare literal, not just `0` (`unit_checks.py:767`), so `wealth > 1_000_000` and `age < 33` are treated as unit-safe though the literals implicitly carry `CURRENCY`/`YEARS`. Permitting literal `0` as the polymorphic case but rejecting non-zero bare literals in ordering/min/max (unless tagged or promoted to a parameter) would catch these.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+
+- a body whose inferred unit disagrees with its declaration, on any reachable branch — a
+  stock times a per-year rate labelled as a stock, or a `_m` flow returned where `_y` is
+  declared;
+- an addition or subtraction of two non-equivalent quantities — a monthly flow plus a
+  yearly one (`betrag_m + freibetrag_y`), or a stock plus a flow. At run time the
+  assembled DAG computes on bare arrays with no pint, so such a combination is
+  unit-blind and silently wrong; the dry-run rejects it rather than letting pint's
+  build-time auto-conversion of same-dimension operands paper over it;
+- an ordering comparison (`<`, `<=`, `>`, `>=`) of two non-equivalent quantities;
```
hmgaudecker: Minor: this guarantee covers `< <= > >=`; `==`/`!=` are deliberately *not* screened (sentinel use). Worth stating explicitly so readers don't assume equality is dimension-checked.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+the {ref}`GEP 1 <gep-1>` conventions.
+
+(gep-10-periods)=
+
+### Flow tokens
+
+Flow units need a reference period source. For many functions and parameters, the period
+comes from the name suffix (`_y`, `_m`, …) as laid out in {ref}`GEP 1 <gep-1>`. Only
+parameters that cannot auto-convert because they are structured values (dicts,
+schedules, lookup tables) need to declare a `reference_period` to supply the period that
```
hmgaudecker: Minor: the implementation *forbids* `reference_period` on `scalar` parameters (a scalar's period comes only from its name suffix). Worth stating that prohibition here — the JSON schema doesn't reject it either.

hmgaudecker (commented):
External adversarial dimensional-safety audit of the GEP-10 stack (verdict: serious_gap; 5 serious, 4 moderate issues). Inline notes below flag places where the spec or implementation over-claims relative to the checks that actually run.

hmgaudecker (commented):
Two further minor specification points.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+| `HOURS_FLOW`                     | `hour / period` (dimensionless)  | working hours            |
+| `SQUARE_METERS`                  | `meter ** 2`                     | dwelling size            |
+| `CURRENCY_PER_SQUARE_METER_FLOW` | `CURRENCY / meter ** 2 / period` | rent caps                |
+
+A token ending in `…_FLOW` needs a period; every other token is complete as written and
+takes no period. Tokens are not pint syntax: each resolves internally to a pint unit
+(flow tokens after the period is filled in), but pint expressions never appear in a
+declaration.
+
+`HOURS_FLOW` is the one flow token that resolves to a *dimensionless* quantity: hours
```
hmgaudecker: **`HOURS_FLOW` is dimensionally identical to a bare dimensionless quantity.** Because hours/period is time/time, the checker cannot tell working hours from a plain share or count — so e.g. adding working hours to a ratio would not be caught. Worth stating next to the safety claim in the Abstract so the engine isn't read as complete dimensional verification (the dimensionless-result fallback has the same effect).

docs/geps/gep-10.md
```diff
@@ -0,0 +1,560 @@
...
+  boundary
+- ttsim [#121](https://github.com/ttsim-dev/ttsim/issues/121) — annotate mettsim, switch
+  check on, CI test
+- gettsim [#1191](https://github.com/ttsim-dev/gettsim/issues/1191) — register EUR/DM
+- gettsim [#1192](https://github.com/ttsim-dev/gettsim/issues/1192) — gettsim rollout
+
+Each package's params schema enumerates its own token vocabulary: the core tokens minus
+the agnostic currency tokens (the schema governs parameters, which must be concrete)
+plus the concrete variants of that package's registered currencies. It also enforces the
+`unit:` XOR `input_unit:`/`output_unit:` split per parameter `type:` and admits the
```
hmgaudecker: **The schema enforces the `unit:` XOR `input_unit:`/`output_unit:` split per `type:`, but does not constrain the *shape* of `unit:` by type.** A nested `unit:` mapping is accepted on `type: scalar`, which construction later rejects or mishandles. Conditioning the declaration shape on the parameter type in the schema would catch this at validation time.

docs/geps/gep-10.md
MImmesberger: Fixed — boolean-returning bodies are no longer skipped. Dropped the `boolean_nodes` clause from the env check, so a boolean body is now dry-run like any other (declares DIMENSIONLESS, full symmetry) and the comparisons inside eligibility functions are checked. Added `&`/`|`/`^`/`~` to `_DryRunQuantity` (screening both operands for dimensionless) so they run rather than forcing opt-outs; mettsim's one un-dry-runnable boolean (`in_same_household_as_recipient`, uses `join`) opts out explicitly. Pushed.

docs/geps/gep-10.md
MImmesberger: Fixed the silent pass: hitting the path cap now returns the opt-out-required error (symmetric with the per-run decisions cap) instead of warn-and-succeed. On the infeasible-path point — decided to keep syntactic branch-combination coverage (no path-constraint tracking) and make the wording honest: the GEP now says 'syntactically reachable branch combinations' and notes the cap opts out. Pushed.

docs/geps/gep-10.md
MImmesberger: Fixed — `_verify_one_body` now demands `verify_units=False` for any return that is neither a `pint.Quantity` nor a plain numeric/bool scalar (no recursion). `bounty_per_orc` opts out accordingly; the check also correctly caught a few `not_required` test fixtures returning columns/lists, now opted out. GEP opt-out paragraph extended. Pushed.

docs/geps/gep-10.md
MImmesberger: Fixed — the ordering comparisons reject a bare non-zero literal when the quantity is non-dimensionless (`0` stays the lone inline literal; dimensionless self stays lenient; `min`/`max` route through the same path). Scope is ordering only — `+`/`-` keep the lenient `0`-guard handling. mettsim's `age < 33` became an `age_of_majority` (YEARS) parameter. GEP catch-bullet amended. Pushed.

docs/geps/gep-10.md
MImmesberger: **S1 (calendar points vs durations):** Fixed. The GEP §vocabulary now documents a calendar-point sub-model — `CALENDAR_YEAR/MONTH/DAY` (affine points) alongside `YEARS/MONTHS/DAYS` (durations) — with the algebra table (`point − point → duration`, `point ± duration → point`; `point + point`, scaling a point, and mixing calendar axes are errors), the cyclic-ordinal rule (`geburtsmonat` stays `DIMENSIONLESS`), and a no-op migration note for existing `YEARS`. Implemented in ttsim as offset-unit engine + dry-run enforcement, with a mettsim worked example (`age = policy_year − geburtsjahr`).

**D1 (HECTARES missing from the table):** WONTFIX. `HECTARES` was deliberately dropped from the GETTSIM-facing vocabulary table — it lives in ttsim core for mettsim's worked example, but GETTSIM never declares it (German policy uses `SQUARE_METERS`), so it stays out of the table by design.

docs/geps/gep-10.md
MImmesberger: Reconciled in the GEP. A new **Converters (`require_converter`)** paragraph under §Currency spells out the two honest shapes:

- **homogeneous** — a single `unit:`, scaled uniformly; or
- **function-like** — `input_unit:` / `output_unit:` axes when the converter produces a schedule/lookup, whose typed output is converted per axis so an order-`j` coefficient scales by `f_out / f_in**j` rather than by one uniform factor.

It also states the rule for the mixed case (split into separate homogeneous parameters — e.g. a `satz` and a `YEARS` age bracket — since a blob has no surface to declare a unit per leaf) and the build-time rejection of a homogeneous converter that produces a function-like value. Commit `ad30820d`.

docs/geps/gep-10.md
MImmesberger: Schema enforcement done in ttsim `7470a9c8` (pint-120): on `type: scalar` the schema now forbids `reference_period` (a scalar takes its period from a time suffix on its name). The GEP-prose statement of the prohibition will land with the remaining GEP-prose clarifications (D2/D4). Note only the ttsim `params-schema.json` copy is tightened; the GEP's copy stays on the pre-pint form and migrates wholesale at gettsim #1192.

docs/geps/gep-10.md
MImmesberger: Schema enforcement done in ttsim `7470a9c8` (pint-120): on `type: scalar` the `unit:` declaration must be a single token, not a nested per-leaf mapping (only a dict parameter may declare per-leaf units, which construction otherwise mishandles). Combined with the D3 constraint in one conditional. Same caveat as D3 — ttsim copy only; the GEP copy migrates at gettsim #1192.

docs/geps/gep-10.md
MImmesberger: Resolved by documentation (commit `0422bb5a`). Replaced the overstated "the body's edges are still checked" with an honest split plus a **Known limitation** note:

- the declared *output* unit still holds, so downstream consumers are still checked;
- a body's *inputs* are the separately-verified declared outputs of their producers — so the incoming units are sound;
- what the opt-out drops is the *body-internal* check: a schedule's `input_unit` domain is not bound to the argument it's evaluated at, because the body is never dry-run.

Decided against adding a domain-binding check: standardized `piecewise_polynomial` evaluation makes the residual risk small, and it could escape end to end only if two opt-out bodies fed one another directly, which doesn't occur. The matching in-code comment is fixed in ttsim (`9fa15960`).

docs/geps/gep-10.md
MImmesberger: GEP prose added (commit `7c3b078a`). The Flow-tokens section now states that the two period sources are mutually exclusive: `reference_period` exists only for structured values, so declaring it on a single value — column, policy function, or scalar parameter — is an error, and the JSON schema rejects `reference_period` on `type: scalar`. (Schema tightening shipped earlier in ttsim's `params-schema.json`; the GEP's embedded copy migrates wholesale with the YAMLs in #1192.)

docs/geps/gep-10.md
MImmesberger: GEP prose added (commit `7c3b078a`). The schema paragraph now says the schema constrains, per `type:`, both the `unit:` XOR `input_unit:`/`output_unit:` split *and the shape of the declaration itself*: a `type: scalar` `unit:` must be a single token, the leaf-keys-to-tokens mapping form being admitted only for `type: dict`. (Matches the `allOf` tightening already in ttsim's `params-schema.json`; the GEP's embedded copy migrates wholesale with the YAMLs in #1192.)

docs/geps/gep-10.md
MImmesberger: Stated (commit `1c6fe8df`). Generalized the "What it cannot catch" bullet from the narrow "result comes out dimensionless" to the principle: the check is *dimensional*, not *semantic* — anything that collapses to the dimensionless dimension is indistinguishable to it. `HOURS_FLOW` is the worked example (`[time]/[time] = 1`, so adding working hours to a share or head count isn't caught), and the `return 0.0` / cancellation fallback is folded in as the same blind spot. Closes with "the engine guarantees *dimensional* soundness, not that every quantity is the intended *kind*."

docs/geps/gep-10.md
MImmesberger: Stated (commit `1c6fe8df`). Appended to the ordering-comparison bullet: equality (`==`, `!=`) is deliberately **not** screened, because it is the operator for sentinel and exact-marker tests — e.g. a person-pointer no-link marker (`p_id_empfänger == -1`) or an exact-zero guard (`kindersatz_m == 0.0`) — where the literal is a deliberate marker, not a hidden dimensioned bound. The note also states the trade-off explicitly: an equality between two genuinely non-equivalent quantities is not caught, so `==`/`!=` are reserved for marker tests, not for comparing computed amounts.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,969 @@
...
+policy using the currency in which the applicable law specifies its parameters. Users
+supply monetary data and request computed results in either Deutsche Mark or Euro, while
+statutory parameter values remain unchanged.
+
+## Motivation and scope
+
+Four problems motivate this GEP.
+
+1. **Arithmetic is not dimensionally validated.** GETTSIM currently treats values with
+   different meanings as ordinary numbers. A calculation can therefore add a monthly
```
hmgaudecker:
```suggestion
   different meanings as ordinary numbers. A calculation can therefore add a total
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,969 @@
...
+   amount to rent per square meter without failing immediately.
```
hmgaudecker:
```suggestion
   amount of rent to rent per square meter without failing immediately.
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+   mechanically converted Euro values with the original amount recorded only in
+   free-text metadata. However, storing DM values in their converted Euro form violates
+   GETTSIM's law-to-code principle.
+1. **Reference-period conversion factors are maintained separately.** GETTSIM already
+   converts between annual, quarterly, monthly, weekly, and daily values. The numerical
+   ratios for those conversions should come from the same unit system that validates
+   policy calculations.
+
+The GEP covers the unit infrastructure in TTSIM and the currency and policy declarations
+in GETTSIM. It treats Deutsche Mark as the earliest statutory currency modeled by this
```
hmgaudecker:
```suggestion
in GETTSIM. It treats Deutsche Mark and Euros as the currency space.
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+proposal.
```
hmgaudecker:
```suggestion
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Notably, the naming conventions defined in {ref}`GEP 1 <gep-1>` (group and time
+suffixes) and their automatic conversion are not superseded by this GEP, they remain in
```
hmgaudecker:
```suggestion
suffixes) and their automatic conversion remain in effect.
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+effect.
```
hmgaudecker:
```suggestion
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Most users continue to call `main()` with ordinary arrays or a DataFrame. Users may
+provide a `data_currency` argument to specify the currency of their monetary input and
+the desired currency of computed results. The default is Euro. Data currency may take up
+one of two registered currencies: `EUR` or `DM`.
+
+```python
+results = main(
+    policy_date_str="1999-01-01",
+    data_currency="EUR",
```
hmgaudecker, line 152: Would be a very strange thing for 1999 data.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Notably, the naming conventions defined in {ref}`GEP 1 <gep-1>` (group and time
+suffixes) and their automatic conversion are not superseded by this GEP, they remain in
+effect.
+
+(gep-10-usage)=
+
+## Usage and impact
+
+### Model users
```
hmgaudecker:
```suggestion
### Users of existing policy environments
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+   Deutsche Mark;
+1. evaluates the 1999 policy in Deutsche Mark using the statutory parameters; and
+1. converts computed currency-denominated results back to Euro.
+
+Input columns without a currency component, requested parameters, and requested input
+columns are not converted on output.
+
+(gep-10-trees)=
+
+### Unit-annotated data
```
hmgaudecker, line 174: This is a bit too standalone. I would expect a little explanation of what a `UnitAnnotatedColumn` is (underlying data model numpy/jax/configurable, ...), how one may go about annotating a DataFrame with units, ...

In the results section, it would be good to describe how one gets ordinary numbers back (and whether `UnitAnnotatedColumn`s work out of the box with plotting tools / numpy / jax).

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+The unit tag's physical dimension, period, and grouping level must agree with the
+declaration of the corresponding input in GETTSIM's DAG. Only the currency may differ,
+in which case the value is converted.
+
+For unit-annotated input, each leaf's concrete currency tag determines its source
+currency. `data_currency` determines the currency assumed for unannotated monetary input
+and the currency of computed results.
+
+The annotated result tree has the same structure. Computed currency values are tagged
```
hmgaudecker:
```suggestion
By default, the annotated result tree has the same structure. Computed currency values are tagged
```
Right? I could also always request that explicitly, I guess? Would be good to explain briefly.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+with the concrete data currency, while requested parameters retain their statutory
+currency.
+
+### Contributors
```
hmgaudecker:
```suggestion
### GETTSIM Contributors
```
Also mention "Users extending the policy environment"? Same thing, I guess, or can they turn pint off? If essentially the same, this should go directly below "Users of existing policy environments".

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+) -> float:
+    return hh_amount_m + cast_ttsim_unit(
+        bg_amount_m,
+        TTSIMUnit.CURRENCY.PER_MONTH.PER_HH
+    )
+```
+
+#### Parameters
+
+Parameter files record the currency and period in which the statute specifies a value.
```
hmgaudecker: Claudish -> English, please.

Also, why does currency get singled out here? Holds just as true for other values, right?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+    value: 1000.0
+```
+
+Currency-denominated parameters must use a concrete currency. A build-time check
+requires that currency to equal the statutory currency at the parameter's policy date.
+
+## Backward compatibility
+
+- Bare arrays and the DataFrame/mapper input interface remain supported.
+- `data_currency` defaults to Euro in GETTSIM, so current-policy Euro inputs and outputs
```
hmgaudecker:
```suggestion
- `data_currency` defaults to `EUR` in GETTSIM, so current-policy Euro inputs and outputs
```
I guess?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+  retain their existing denomination.
+- The former `reference_period` and `reference_level` fields are removed. Their
+  information becomes part of the compositional unit, for example `EUR_PER_YEAR_PER_FG`.
+- Existing policy functions, policy inputs, parameter files, and hand-written
+  aggregations require unit declarations.
+- There is no environment variable that disables all unit validation. Exceptions are
```
hmgaudecker: This will be too opaque to most people. Please translate to be comprehensible for occasional users.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| base               | The numerator of a compositional unit, such as `CURRENCY`, `DIMENSIONLESS`, or `YEARS`.                 |
+| period             | The denominator that distinguishes a flow from a stock, such as `MONTH` or `YEAR`.                      |
+| grouping level     | The entity to which a group property belongs, such as a household (`hh`) or Bedarfsgemeinschaft (`bg`). |
+| bare               | A quantity without a grouping-level denominator. Personal quantities are bare.                          |
+| stock              | A quantity without a period denominator, such as wealth or an age.                                      |
+| flow               | A quantity with a period denominator, such as income per month.                                         |
+| statutory currency | The currency in which the applicable statute specifies monetary amounts at a policy date.               |
+| data currency      | The currency of user-provided monetary columns and returned computed results.                           |
+| body validation    | Evaluation of a policy-function body with unit-carrying test values instead of data values.             |
```
hmgaudecker: While it is part of terminology, it really is out of place in this table. Just describe wherever it is relevant.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+(gep-10-vocabulary)=
+
+### Compositional unit vocabulary
+
+A unit consists of one base followed by at most one denominator of each category. The
+categories have a fixed order.
+
+```text
+base        := CURRENCY
+             | EUR | DM | ...
```
hmgaudecker:
```suggestion
             | EUR | DM
```
For all practical purposes, no? If we modeled Grundsteuer before 2024 we could add Reichsmark, if we calculated the price of Gebäudeversicherungen we could add Goldmark, but this seems far in the future if ever relevant.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+(gep-10-vocabulary)=
+
+### Compositional unit vocabulary
+
+A unit consists of one base followed by at most one denominator of each category. The
+categories have a fixed order.
+
+```text
+base        := CURRENCY
```
hmgaudecker, line 431: Explain below why we need both CURRENCY and specific incarnations thereof here.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+             | base _PER_ period
+             | base _PER_ level
+             | base _PER_ physical _PER_ period
+             | base _PER_ physical _PER_ level
+             | base _PER_ period _PER_ level
+             | base _PER_ physical _PER_ period _PER_ level
+```
+
+The parser rejects repeated denominators and non-canonical orderings such as
+`EUR_PER_BG_PER_MONTH`. The fluent Python builder enforces the same order through its
```
hmgaudecker: What is a fluent Python builder?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+```text
+base        := CURRENCY
+             | EUR | DM | ...
+             | DIMENSIONLESS
+             | HOURS
+             | SQUARE_METER | HECTARE
+             | YEARS | MONTHS | DAYS
+             | CALENDAR_YEAR | CALENDAR_MONTH | CALENDAR_DAY
+
+physical    := SQUARE_METER | HOURS
```
hmgaudecker, line 441: Explain below why hours is physical and not a period.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+             | base _PER_ physical
+             | base _PER_ period
+             | base _PER_ level
+             | base _PER_ physical _PER_ period
+             | base _PER_ physical _PER_ level
+             | base _PER_ period _PER_ level
+             | base _PER_ physical _PER_ period _PER_ level
+```
+
+The parser rejects repeated denominators and non-canonical orderings such as
```
hmgaudecker:
```suggestion
The parser rejects repeated denominators (like ...)  and denominators not ordered according to the above (like ```

File: docs/geps/gep-10.md [hunk 31]
```diff
@@ -0,0 +1,967 @@
...
+```python
+TTSIMUnit.CURRENCY.PER_MONTH.PER_BG
+TTSIMUnit.CURRENCY
+TTSIMUnit.DIMENSIONLESS
+TTSIMUnit.DIMENSIONLESS.PER_FG
+TTSIMUnit.HOURS.PER_WEEK
+```
+
+The Python representation and the YAML representation are two syntaxes for the same
+declaration. For example, `TTSIMUnit.CURRENCY.PER_MONTH.PER_BG` corresponds to
```
hmgaudecker:
```suggestion
declaration. They only differ by single `_` replacing single `.` as separators. For example, `TTSIMUnit.CURRENCY.PER_MONTH.PER_BG` corresponds to
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| `DIMENSIONLESS_PER_BG`                | `1 / [bg]`                      | persons in a Bedarfsgemeinschaft          |
+| `HOURS_PER_WEEK`                      | `[hours] / week` (bare)         | personal weekly working hours             |
+| `CURRENCY_PER_HOURS`                  | `CURRENCY / [hours]`            | hourly wage                               |
+| `CURRENCY_PER_SQUARE_METER_PER_MONTH` | `CURRENCY / meter**2 / month`   | monthly rent ceiling per square meter     |
+| `YEARS`                               | calendar-year duration          | age in years                              |
+| `CALENDAR_YEAR`                       | point on the calendar-year axis | birth year                                |
+
+`CURRENCY` is valid only in code-side declarations whose concrete currency depends on
+the policy date. Parameter YAML, rounding specifications, and input tags use a concrete
+registered currency.
```
hmgaudecker: Explain why.

Also explain best practices for describing policies that ended before 1999 / started in 2002. Would I use CURRENCY or DM/EUR and why?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Some policy expressions intentionally interpret a group property at another level. Such
+an expression requires `cast_ttsim_unit`, as described in
+{ref}`Explicit exceptions <gep-10-opt-out>`.
+
+(gep-10-booleans)=
+
+#### Booleans
+
+A boolean describes an entity. At the individual grain it is bare; at a group level it
```
hmgaudecker: Or corn? Claudish -> (economists') English, please.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Multiplication by a head count converts a bare per-person amount to a group total.
+
+```text
+familie__anzahl_personen_sn * sparerfreibetrag_y
+  = (1 / [sn]) * (CURRENCY / year)
+  = CURRENCY / year / [sn]
+```
+
+Some policy expressions intentionally interpret a group property at another level. Such
```
hmgaudecker: An example would help.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+| Boolean             | Declaration            | Resolved dimensionality |
+| ------------------- | ---------------------- | ----------------------- |
+| person indicator    | `DIMENSIONLESS`        | `dimensionless` (bare)  |
+| family indicator    | `DIMENSIONLESS_PER_FG` | `1 / [fg]`              |
+| household indicator | `DIMENSIONLESS_PER_HH` | `1 / [hh]`              |
+
+The logical operators `&`, `|`, and `^` preserve the level when both operands have the
+same level. When their levels differ, the result is bare — the individual "level" —
+because the expression is evaluated once per person.
```
hmgaudecker:
```suggestion
because the expression is evaluated bit-wise, i.e., per individual row.
```

docs/geps/gep-10.md
hmgaudecker: That is, just forward reference the section that I found ex post.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| ------------------------------ | -------- | ------------------------------ |
+| `P - P`                        | duration | `policy_year - geburtsjahr`    |
+| `P + D`, `P - D`               | point    | `geburtsjahr + statutory_age`  |
+| `P < P`                        | boolean  | `geburtsjahr <= policy_year`   |
+| `P + P`                        | error    | addition of two birth years    |
+| `P * n`, `P / n`               | error    | scaling a calendar point       |
+| point ordered against duration | error    | birth year compared with age   |
+| operation across calendar axes | error    | year point plus month duration |
+
+A month-of-year, day-of-week, or quarter is a cyclic ordinal rather than an absolute
```
hmgaudecker:
```suggestion
A month-of-year (January, February, ...), day-of-week (Monday, Tuesday, ...), or quarter is a cyclic ordinal rather than an absolute
```
I don't get the quarter one -- isn't that the same as `CALENDAR_MONTH`, just counting 1...4 instead of 1...12? Or is `CALENDAR_MONTH` actually `1999-11`? Then why don't we have `CALENDAR_QUARTER`?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+```yaml
+tarif:
+  input_unit: EUR_PER_YEAR
+  output_unit: EUR_PER_YEAR
+  type: piecewise_quadratic
+  # Values omitted.
+```
+
+The parameter schema rejects `unit:` on a mapping type that requires `input_unit:` and
+`output_unit:`. A time suffix in the parameter name describes the output axis and must
```
hmgaudecker: Can you give an example for the time suffix thing?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+agree with `output_unit`.
+
+`require_converter` parameters use one of three forms:
+
+- one `unit:` token for homogeneous content;
+- a per-leaf `unit:` mapping for heterogeneous content; or
+- `input_unit:` and `output_unit:` when the blob encodes a mapping between quantities.
```
hmgaudecker: blob?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+the corresponding `@param_function` as they may differ depending on the conversions in
+the param function.
+
+#### Parameter functions
+
+**Mapping parameters.**
+
+Parameter functions exposing mapping parameters (of type `PiecewisePolynomialParamValue`
+or `ConsecutiveIntLookupTableParamValue`) declare their input and output unit via
+`unit=InputOutputUnit(...)`:
```
hmgaudecker:
```suggestion
`unit=InputOutputUnits(...)`:
```
seems to be what is implemented in #1212, at least ?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+Units of `require_converter` parameters can be declared independently from the units of
+the corresponding `@param_function` as they may differ depending on the conversions in
+the param function.
+
+#### Parameter functions
+
+**Mapping parameters.**
+
+Parameter functions exposing mapping parameters (of type `PiecewisePolynomialParamValue`
```
hmgaudecker: Remind people that this is the function relating to `require_converter` and not something new.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+range. At a given policy date, validation requires declarations only for leaves present
+in the resolved value.
+
+#### Mapping parameters
+
+A schedule or lookup table is a mapping between quantities. It declares input and output
+axes rather than one scalar unit.
+
+```yaml
+tarif:
```
hmgaudecker:
```suggestion
raw_parameter_tarif:
```
That's the only example we have, but doesn't quite fit here. Alternatively, use a lookup table as example.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+For a multidimensional lookup table, `InputOutputUnit.input_unit` may be a tuple. Each
+tuple element declares the unit of the corresponding positional argument to `look_up`,
+and the number of declared axes must match the number of arguments. A
+`piecewise_polynomial` has exactly one input axis and therefore does not accept a tuple.
+Raw YAML mapping parameters use one `input_unit` token; a tuple is available only on a
+converter-produced schedule or an annotated schedule field.
+
+**Structured parameters.**
+
+Parameter functiuons exposing structured parameters in the form of generated dataclasses
```
hmgaudecker:
```suggestion
Parameter functions exposing structured parameters in the form of generated dataclasses
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+field is declared as `TTSIMUnit.CURRENCY`. Renamed or derived fields have no matching
+source path and cannot be compared automatically.
+
+(gep-10-auto)=
+
+### Generated nodes and aggregations
+
+#### Reference-period conversions
+
+A generated time-conversion node preserves its source base and grouping level and
```
hmgaudecker: source's base? maybe change the sentence more broadly

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+(CURRENCY / [hh]) / (1 / [hh])
+  = CURRENCY
+```
+
+A hand-written aggregation declares its unit. By default, that declaration must exactly
+match the unit derived from the source, aggregation type, and target level. An
+aggregation whose policy interpretation cannot be represented by these rules may set
+`verify_units=False`.
+
+`@agg_by_p_id_function` aggregates to the individual grain, which is bare, so its
```
hmgaudecker: Claudish...

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+bare because it is equivalent to a group sum divided by a head count.
+
+```text
+(CURRENCY / [hh]) / (1 / [hh])
+  = CURRENCY
+```
+
+A hand-written aggregation declares its unit. By default, that declaration must exactly
+match the unit derived from the source, aggregation type, and target level. An
+aggregation whose policy interpretation cannot be represented by these rules may set
```
hmgaudecker:
```suggestion
aggregation whose policy interpretation cannot be represented by these rules has to set
```
?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+(`1 / [target]`), and are bare at an individual target. `SUM`, `MIN`, and `MAX` over a
+non-boolean resolve to the target group level, so a bare source acquires it. A mean is
+bare because it is equivalent to a group sum divided by a head count.
+
+```text
+(CURRENCY / [hh]) / (1 / [hh])
+  = CURRENCY
+```
+
+A hand-written aggregation declares its unit. By default, that declaration must exactly
```
hmgaudecker: What is a default about this? If the qualifier is the opt-out, just remove "By default, "

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| generated time conversion            | assigned automatically                             | inherited agnostic base               | period derived from target suffix             |
+| generated aggregation                | assigned automatically                             | inherited agnostic base               | aggregation rule and target level             |
+| hand-written aggregation             | required `unit=`                                   | `CURRENCY` for monetary values        | exact match with derived unit unless disabled |
+| group-creation function              | required `unit=`                                   | agnostic `CURRENCY` only              | resolved like any other column                |
+| rounding specification               | unit required when attached to a monetary function | concrete currency                     | function unit and statutory currency          |
+| unit-annotated input column          | required on every leaf in that input mode          | concrete currency                     | declared node unit and data currency          |
+
+`UNSET_UNIT` has two uses that are distinguished by context:
```
hmgaudecker: Give examples. Where is this even used?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+(gep-10-literals)=
+
+### Numerical literals
+
+A dimensionless numerical literal cannot be added to, subtracted from, or ordered
+against a dimensioned quantity. Multiplication and division by a dimensionless factor
+are valid.
+
+```python
+betrag_m * 0.5  # Valid: the unit remains CURRENCY_PER_MONTH.
```
hmgaudecker: Order all examples according to the text.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+einkommen_m < 1000.0  # Invalid: the operands have different units.
+```
+
+A dimensioned threshold should normally be a parameter. If it must remain in the
+function body, `cast_ttsim_unit` assigns its intended unit.
+
+```python
+einkommen_m < cast_ttsim_unit(1000.0, TTSIMUnit.CURRENCY.PER_MONTH)
```
hmgaudecker: call by keyword

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+against the abstract `[currency]` reference. Every other currency states a `value`
+relative to a currency named before it. A currency's `statutory_from` is the date from
+which statutes denominate their numbers in it, until the next currency's date; at least
+one currency must become statutory.
+
+```python
+UNIT_SYSTEM = UnitSystem(
+    currencies={
+        "EUR": Currency(statutory_from="2002-01-01"),
+        "DM": Currency(value="EUR / 1.95583", statutory_from="0001-01-01"),
```
hmgaudecker:
```suggestion
        "DM": Currency(value="EUR / 1.95583", statutory_from="1948-06-20"),
```

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+#### Validation stages
+
+| Stage                         | When                                        | Input                                  | Validates                                                            |
+| ----------------------------- | ------------------------------------------- | -------------------------------------- | -------------------------------------------------------------------- |
+| declaration validation        | decoration or parameter loading             | declarations                           | required fields, grammar, suffixes, allowed currency bases           |
+| environment validation        | policy-environment assembly                 | unit-carrying test values              | function bodies, returns, branches, aggregations, statutory currency |
+| input-boundary validation     | processing unit-annotated input in `main()` | user tags and node declarations        | physical dimension, period, level, and source currency               |
+| numerical boundary conversion | before and after TT computation             | arrays, scalars, and conversion factor | input to statutory currency; computed results to data currency       |
```
hmgaudecker: why does function-body not show up here?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| ------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
+| Equality operators (`==`, `!=`) do not compare units.                    | Sentinel comparisons such as `p_id_empfänger == -1` remain possible, but an incompatible equality comparison is not detected. | Use ordering or an explicit policy condition where dimensional equivalence matters.            |
+| Branch exploration is limited to 1,024 paths and 64 decisions per run.   | Environment assembly fails when either limit is exceeded.                                                                     | Simplify the body or use `verify_units=False`.                                                 |
```
hmgaudecker:
```suggestion
| Branch exploration is limited to 1,024 paths and 64 decisions per run.   | Environment assembly fails when either limit is exceeded.                                                                     | Simplify the body.                                                 |
```

Limit sounds more than enough, no?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+- inconsistent schedule input and output axes; and
+- unsupported operations unless the function explicitly disables body validation.
+
+(gep-10-limitations)=
+
+#### Limitations
+
+| Limitation                                                               | Consequence                                                                                                                   | Required action                                                                                |
+| ------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
+| Equality operators (`==`, `!=`) do not compare units.                    | Sentinel comparisons such as `p_id_empfänger == -1` remain possible, but an incompatible equality comparison is not detected. | Use ordering or an explicit policy condition where dimensional equivalence matters.            |
```
hmgaudecker: Should this section be called Ttrade-offs / limitations?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| Branch exploration is limited to 1,024 paths and 64 decisions per run.   | Environment assembly fails when either limit is exceeded.                                                                     | Simplify the body or use `verify_units=False`.                                                 |
+| Not every NumPy/JAX or Python operation has a validation implementation. | The environment check rejects an unsupported body.                                                                            | Add validator support, use a local `cast_ttsim_unit`, or use `verify_units=False`.             |
```
hmgaudecker: Examples would be helpful, for the entire table.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+| Group nesting and membership are not represented in the unit registry.   | No automatic conversion exists between group levels.                                                                          | Use a head count, an explicit aggregation, or `cast_ttsim_unit` where substantively justified. |
+| A cast is an assertion, not a conversion or proof.                       | An incorrect cast can hide a unit error in that expression.                                                                   | Keep casts local and review them as policy assumptions.                                        |
+| `verify_units=False` skips the complete function body.                   | Consumers still use the declared return unit, but the producer body is not validated.                                         | Restrict the opt-out to functions that cannot be represented by the validator.                 |
```
hmgaudecker: Isn't this obvious?

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+
+(gep-10-opt-out)=
+
+#### Explicit exceptions
+
+`cast_ttsim_unit(value, unit)` changes the inferred unit of one expression during body
+validation. At numerical execution it returns `value` unchanged. It is appropriate for:
+
+- policy-defined arithmetic whose grouping interpretation differs from the general
```
hmgaudecker: Again, examples could be nice. E.g., I have no idea what "an explicit calendar-granularity conversion" is supposed to be and I lack imagination for a constant that cannot be represented as a parameter.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+- GETTSIM [#1193](https://github.com/ttsim-dev/gettsim/pull/1193) contains this draft
+  GEP.
+- GETTSIM [#1212](https://github.com/ttsim-dev/gettsim/pull/1212) contains the GETTISM
+  rollout.
+
+(gep-10-alternatives)=
+
+## Alternatives
+
+### Keep grouping levels outside Pint
```
hmgaudecker: Incomprehensible as is. I'd just drop it.

docs/geps/gep-10.md
```diff
@@ -0,0 +1,967 @@
...
+Rejected. Some statutory formulas contain coefficients whose numerical values depend on
+the currency convention even when the statute prints them as bare numbers. The Wohngeld
+basic formula is an example:
+
+```text
+1.15 * (M - (a + b*M + c*Y) * Y)
+```
+
+If `M` and `Y` have unit currency per month, then `a` is dimensionless and `b` and `c`
+have the implicit unit month per currency. Their numerical values must change when the
```
hmgaudecker:
```suggestion
have the implicit unit "month per currency". Their numerical values must change when the
```

hmgaudecker (approved):
Amazing work, congratulations! Just a bunch of textual changes, nothing about implementation.

docs/geps/gep-10.md
MImmesberger: Not necessarily, e.g. in a life-cycle model. Some users may also have pre-converted
values from the SOEP.

hmgaudecker: Ofc you can find examples, but why pick a contrived one?

MImmesberger: I don't see why it is contrived. Data coming out of SOEP preparation will be EUR only (e.g. [`plc0013_h`](https://paneldata.org/soep-core/datasets/pl/plc0013_h)), also for pre-2002. Or is your point that users will only ever have EUR data, almost never DM data?

hmgaudecker: 1999 was DM, default is EUR, anyways -- why put knots in people's heads?

MImmesberger:
Many thanks for the review! @hmgaudecker

Issue #1215, opened 2026-07-21
Midijob: childless LTC surcharge (§ 55(3) SGB XI) levied on the total reduced base instead of the employee base (§ 20(2a) S.6 SGB IV, § 58(1) S.3 SGB XI)


MaxGhenis:
**GETTSIM version**: 1.2.1 (PyPI distribution `gettsim==1.2`), policy date 2025-06-30.

**Summary**: in the Midijob transition band, the employee's long-term-care contribution levies the § 55 Abs. 3 SGB XI childless surcharge (0.6 pp) on the *total* reduced contribution base rather than on the *employee* reduced base — while the parity share (1.8 pp) correctly uses the employee base. The statutory text points to the employee base for both components.

**Statutory chain** (wording as consolidated):

- § 58 Abs. 1 S. 3 SGB XI: *"Den Beitragszuschlag für Kinderlose nach § 55 Absatz 3 Satz 1 **tragen die Beschäftigten**."* — the surcharge is part of the employee-borne share.
- § 20 Abs. 2a S. 6 SGB IV: *"**Zur Bestimmung des vom Arbeitnehmer zu tragenden Anteils** am Sozialversicherungsbeitrag wird als beitragspflichtige Einnahme der Betrag zu Grunde gelegt, der sich nach folgender Formel berechnet …"* — the employee-borne share computes on the reduced employee base (the `(AE − G) · 2000/(2000 − G)`-style formula), not on the S. 1 total base.
- § 55 Abs. 3 S. 1 SGB XI raises the *Beitragssatz* for childless members by 0.6 pp — a rate component of the member's contribution, which for Midijob employees is assessed on the § 20 Abs. 2a S. 6 base.

**Reproduction** (single childless employee age 35, monthly gross 1,307.14 € — inside the 2025 band 556.01–2,000 €):

```python
from gettsim import InputData, MainTarget, TTTargets, main  # 1.2.1
# single person, bruttolohn_m = 1307.142857, policy_date_str = "2025-06-30"
# target: sozialversicherung.pflege.beitrag.betrag_versicherter_m
```

- 2025 band bases: employee reduced base BE_ee = (AE − 556) · 2000/(2000 − 556) = 751.14 · 1.385042 = **1,040.35**; total reduced base (F = 0.6683) = F·556 + [(2000 − F·556)/(2000 − 556)]·(AE − 556) = **1,218.64**.
[...]

PR #1222, opened 2026-08-19
Assess the Midijob Pflege childless surcharge on the total base

MImmesberger:
Fixes a bug in the employee's long-term care contribution for Midijobs between
**October 2022 and June 2023**, found while investigating #1215.

This basically implements the same logic as for the 2023+ policy regime. Childless PV surcharge is applied to the total Bemessungsentgelt, not the AN share.


hmgaudecker (approved):
Very nice, thank you!

PR #1228, opened 2026-09-01
Fail the test suite on BeartypeClawDecorWarning

MImmesberger:
Companion to [ttsim#154](https://github.com/ttsim-dev/ttsim/pull/154), which fixes a `BeartypeClawDecorWarning` emitted on every `import gettsim` — it went unnoticed for months because it only ever showed up in the pytest warnings summary.

A claw warning means beartype could not decorate an object, so the package-wide runtime checking has a silent hole there. gettsim registers its own claw over `gettsim.*`, so the same failure mode applies to policy code here. This makes `BeartypeClawDecorWarning` an `error::` in the pytest config.

The `@interface_input` metadata dataclasses are the one known-benign case (a non-callable data holder has nothing to type-check) and stay exempt. The exemption is needed here even though `ttsim/__init__.py` installs it at import time, because pytest resets warning filters per test item.

Verified: `pytest src/gettsim/tests_germany` — 1325 passed, 559 skipped, no claw warnings.

https://claude.ai/code/session_01VCkEDhUCi4aJ6uLGTQeeNH


hmgaudecker (approved):
Thanks! Blocked on TTSIM 1.3.1 being released and the corresponding update to pyproject.toml, ofc.

PR #1233, opened 2026-09-16
Model Versicherungsfreiheit in der Renten- und Arbeitslosenversicherung

### LINKED ISSUE #1226: BUG: Versicherungsfreiheit past the Regelaltersgrenze is not modelled (§ 5 Abs. 4, § 172 Abs. 1 SGB VI)
## Summary

Employees who have reached the Regelaltersgrenze are versicherungsfrei in the statutory
pension insurance and owe no employee contribution; the employer alone pays half
(§ 5 Abs. 4 SGB VI, § 172 Abs. 1 SGB VI). GETTSIM charges them the employee share anyway.

## Reproduction

Person born 1959, one month past the Regelaltersgrenze, 4,000 €/month gross,
policy date 2026-01-01, no overrides:

| quantity | GETTSIM | expected |
|---|---|---|
| `sozialversicherung__rente__beitrag__betrag_versicherter_m` | **372.00** | **0.00** |
| `sozialversicherung__rente__beitrag__betrag_arbeitgeber_m` | 372.00 | 372.00 |

## Location in the code

`src/gettsim/germany/sozialversicherung/rente/beitrag/beitrag.py:72-77`

```python
def betrag_versicherter_regulärer_beitragssatz_m(
    einkommen_m: float,
    beitragssatz: float,
) -> float:
    """Public pension insurance contributions paid by the insured person."""
    return einkommen_m * beitragssatz / 2
```

No age-dependent branch anywhere upstream. `betrag_versicherter_m` is only zeroed for
marginal employment (`beitrag.py:26-40`, `48-68`).

## Legal basis

[§ 5 Abs. 4 SGB VI](https://www.gesetze-im-internet.de/sgb_6/__5.html) makes the group
versicherungsfrei; [§ 172 Abs. 1 SGB VI](https://www.gesetze-im-internet.de/sgb_6/__172.html)
then assigns the employer half of the contribution that would be due:

> Für Beschäftigte, die versicherungsfrei sind wegen 1. des Bezugs einer Vollrente wegen
> Alters nach Ablauf des Monats, in dem die Regelaltersgrenze erreicht wurde, … 3. des
> Erreichens der Regelaltersgrenze …, tragen die Arbeitgeber die Hälfte des Beitrags, der
> zu zahlen wäre, wenn die Beschäftigten versicherungspflichtig wären

A waiver is possible (§ 5 Abs. 4 Satz 2 SGB VI, employee then pays and accrues further
entitlements), which is why this needs a modelling decision rather than an unconditional
zero — see below.

## Open design question

Whether to model the waiver (Verzicht auf die Versicherungsfreiheit, § 5 Abs. 4 Satz 2
SGB VI) as an input column, or to assume versicherungsfreiheit unconditionally. The waiver
is a real and not uncommon choice, and it changes both the contribution and the accrual of
Entgeltpunkte. An input column defaulting to "no waiver" seems the honest option, but that
is a decision, not something the statute settles.

MImmesberger:
Closes #1226.

GETTSIM granted Entgeltpunkte and charged employee contributions to people who are versicherungsfrei — Altersvollrentner, and Minijobber without Versicherungspflicht. This models Versicherungsfreiheit in the Renten- and Arbeitslosenversicherung.

The reported bug is real but differently shaped than the issue describes: § 5 Abs. 4 SGB VI triggers on *drawing a Vollrente* past the Regelaltersgrenze, not on reaching it. Someone past the RAG who keeps working and draws no pension stays versicherungspflichtig, so the issue's repro expects `0.00` where GETTSIM's `372.00` was already correct.

## What changes

**Rentenversicherung** (§ 5 Abs. 4, § 172 SGB VI) — versicherungsfrei while drawing a Vollrente: at any age until 2016, and only past the RAG with the option to waive it from 2017 (Flexirentengesetz). No employee contribution; the employer keeps paying half.

**Arbeitslosenversicherung** (§ 28 Abs. 1 Nr. 1, § 346 Abs. 3 SGB III) — a different rule, not a copy: age alone, no pension needed. So a 68-year-old drawing nothing is versicherungsfrei in the ALV and versicherungspflichtig in the RV. Employer pays half until 2016, **nothing** from 2017 to 2021, half again from 2022.

**Geringfügige Beschäftigung** (§ 168 Abs. 1 Nr. 1b, § 163 Abs. 8 SGB VI) — with Versicherungspflicht the employer stays at the Pauschalbeitrag and the employee owes the residual rather than half, on at least the Mindestbeitragsbemessungsgrundlage (300 DM → 155 € → 175 €).

**Entgeltpunkte** (§ 76b SGB VI) — none while versicherungsfrei wegen Alters (Abs. 4); a Zuschlag reflecting the employer's Pauschalbeitrag for a Minijob without Versicherungspflicht, ≈ 15/18,6 of full points (Abs. 1–2); full points otherwise.

**Übergangsbereich** — for a versicherungsfreier Beschäftigter, half the contribution is assessed on the actual Bruttoarbeitsentgelt until 2022-09-30 and on the reduced beitragspflichtige Einnahme only from 2022-10-01 (§ 168 Abs. 1 Nr. 1d SGB VI a. F.; § 20 Abs. 2a S. 1 SGB IV).

**Parameter fix** — the Minijob-Pauschalbeitrag rises on 2006-07-01, not 2007-01-01 (Haushaltsbegleitgesetz 2006), in both Renten- and Krankenversicherung. The reduced 5 % rate for Privathaushalte is still not modelled; the parameters now say so.

## API impact

- **New required input** `sozialversicherung__rente__verzichtet_auf_versicherungsfreiheit`. `True` means mandatory coverage; the statutory instrument behind it differs by regime (Verzicht until 2012 and past the RAG, absence of a § 6 Abs. 1b Befreiung from 2013), which the docstring spells out.
- **Renamed** `bezieht_rente` → `bezieht_altersrente`. It only ever gated `altersrente__betrag_m`.
- **SV contributions now need `alter_monate` and `geburtsjahr`** — unavoidable, Versicherungsfreiheit is an age rule.
- **§ 17a WoGG** now gates on `altersrente__betrag_m > 0 or erwerbsminderung__betrag_m > 0`; an Erwerbsminderungsrentner with 33 Jahren Grundrentenzeiten was being denied the Freibetrag.

## Tests

New cases across the affected date ranges. Four of them carry official example calculations rather than hand-derived values:

| case | source |
|---|---|
| `2013-01-01/minijob_entgeltpunkte` | DRV, GRA zu § 76b SGB VI, Beispiel 1 |
| `2020-01-01/versicherungsfreiheit_im_übergangsbereich` | DRV, GRA zu § 172 SGB VI, Beispiel 1 |
| `2023-01-01/versicherungsfreiheit_im_übergangsbereich` | Übergangsbereich-Rundschreiben 2022-12-20, Beispiel 12 |
| `2025-01-01/minijob_mindestbeitrag` | Minijob-Zentrale, Mindestbeitrag |

The § 172 case is what caught the Übergangsbereich date split above.


MImmesberger:
Blocked on ttsim-dev/gettsim-personas#58, which adapts the personas to the renamed input and the new age/pension dependency of social insurance contributions. The Read the Docs build here installs `gettsim-personas` from PyPI, so `docs/tutorials/personas.ipynb` fails until that ships a release.

docs/how_to_guides/calculating_elterngeld.ipynb
```diff
@@ -128,7 +129,13 @@
     "    \"lohnsteuer\": {\"steuerklasse\": np.array([1])},\n",
     "    \"p_id\": np.array([0]),\n",
     "    \"hh_id\": np.array([0]),\n",
-    "    \"sozialversicherung\": {\"pflege\": {\"beitrag\": {\"hat_kinder\": np.array([False])}}},\n",
+    "    \"sozialversicherung\": {\n",
+    "        \"pflege\": {\"beitrag\": {\"hat_kinder\": np.array([False])}},\n",
+    "        \"rente\": {\n",
+    "            \"altersrente\": {\"betrag_m\": np.array([0.0])},\n",
+    "            \"verzichtet_auf_versicherungsfreiheit\": np.array([False]),\n",
```
hmgaudecker, line 136: This is annoying, but I guess we don't have a possibility that avoids having to specify it in cases where we know it does not apply? *(depends on age, which is only passed as part of the data, so can never cut it from the DAG)*

This said, I am sort of surprised it only shows up in relatively few cases here in the docs. But maybe we just don't have that many examples.

docs/how_to_guides/modifications_of_policy_environments.ipynb
```diff
@@ -107,6 +107,7 @@
    "source": [
     "INPUT_DATA_TREE = {\n",
     "    \"alter\": np.array([40, 40, 5]),\n",
+    "    \"alter_monate\": np.array([480, 480, 60]),\n",
```
hmgaudecker, line 110: We should not specify `alter` and `alter_monate` at the same time. With the new prominence of `alter_monate` (you basically need it as soon as you want a `betrag` from `sozialversicherung`, right?), we might even want to rename `alter` -> `alter_jahre`. Will only be relevant as an input variable in very few cases then.

docs/how_to_guides/calculating_elterngeld.ipynb
hmgaudecker: passing `"versicherungsfrei_wegen_alters": np.array([False])` would be nicer IMO, if that works. Didn't check the DAG.

src/gettsim/germany/sozialversicherung/rente/entgeltpunkte.py
```diff
@@ -4,38 +4,106 @@
...
+    beitrag__beitragsbemessungsgrenze_y: float,
+    beitrag__beitragssatz: float,
+    beitrag__minijob_arbeitgeberpauschale: float,
+    beitrag__beitragspflichtige_einnahmen_geringfügige_beschäftigung_y: float,
+    beitragspflichtiges_durchschnittsentgelt_y: float,
+    umrechnung_entgeltpunkte_beitrittsgebiet: float,
+) -> float:
+    """Earnings points for the wages earned in the current year.
+
+    Versicherungsfreiheit wegen Alters rules out earnings points, both from
```
hmgaudecker, line 65: Sentence does not work. "rules out any changes in earnings points" ?

src/gettsim/germany/sozialversicherung/rente/entgeltpunkte.py
```diff
@@ -45,21 +113,46 @@ def neue_entgeltpunkte_nach_wohnort(
...
+    beitrag__minijob_arbeitgeberpauschale: float,
+    beitrag__beitragspflichtige_einnahmen_geringfügige_beschäftigung_y: float,
     beitragspflichtiges_durchschnittsentgelt_y: float,
 ) -> float:
-    """Earning points for the wages earned in this year."""
-    if einnahmen__bruttolohn_y > beitrag__beitragsbemessungsgrenze_y:
-        versicherungspflichtiger_bruttolohn_y = beitrag__beitragsbemessungsgrenze_y
+    """Earning points for the wages earned in this year.
+
+    Versicherungsfreiheit wegen Alters rules out earnings points, both from
```
hmgaudecker, line 127: same same

hmgaudecker (approved):
Very nice! Just a few relatively minor comments!

docs/how_to_guides/calculating_elterngeld.ipynb
MImmesberger: `verzichtet_auf_versicherungsfreiheit` feeds into two kinds of Versicherungsfreiheit: i) wegen Alters (age >= NRA); ii) Minijob. So replacing with `versicherungsfrei_wegen_alters` won't work. 

Maybe at some point we could let users select a DAG namespace they want to use. All nodes from other namespaces become root nodes. This is closer to real world applications (if I have tax data and want to compute income taxes, I won't have the inputs for SGB II) and would get rid of these confusing cases where users need to think about Versicherungsfreiheit when they just want to calculate Elterngeld.

And yes, we don't have many examples in the tutorials.

docs/how_to_guides/modifications_of_policy_environments.ipynb
MImmesberger: > We should not specify alter and alter_monate at the same time.

Do you suggest making `alter` / `alter_jahre` a policy function?

This would be a temporary fix until we take care of #211 / computing age from birth date and evaluation date. Should be pretty high on our todo list anyhow.

docs/how_to_guides/calculating_elterngeld.ipynb
hmgaudecker: Ah, I was not aware of that. Are we 100% sure this is exactly the same decision?

docs/how_to_guides/modifications_of_policy_environments.ipynb
hmgaudecker: > Do you suggest making `alter` / `alter_jahre` a policy function?

I guess I am surprised here that it is not one... 

But yes, maybe we could give jax-datetime another, this time serious shot.

MImmesberger: `alter_monate` was relevant for young children only. But recently I started removing the faulty int-age-to-float-age comparisons in the pensions module.

docs/how_to_guides/calculating_elterngeld.ipynb
MImmesberger: Conditional on income yes, unconditional probably not (at least historically, defaults aren't necessarily the same). Alternative would be to add `verzichtet_auf_versicherungsfreiheit_wegen_minijob` and `verzichtet_auf_versicherungsfreiheit_wegen_alters`. I'm indifferent here

hmgaudecker: Would prefer the latter, separated, version. Quite different choice situations.

